PR #185, opened 2026-04-14
feat(pg-retention): tiered cleanup to keep db under 75% of 8 GiB cap

samxu01:
## Summary
- After the normal 30-day delete in `pgRetentionService`, query `pg_database_size(current_database())` and if still above target (default 75% of 8 GiB Cloud SQL cap), step retention down by `PG_RETENTION_STEP_DAYS` (default 1d) per pass until under target or a 1-day floor is reached.
- `VACUUM ANALYZE messages` after deletes so freed pages become reusable.
- Loud WARN log if still over target at the floor — signal to bump the Cloud SQL tier or audit non-message tables.

## Why
Cloud SQL Postgres (8 GiB tier) is approaching the storage ceiling. A fixed 30-day window is fine until it isn't — once growth outpaces the cap, the next insert fails. Tiered fallback retains as much history as the disk allows, preferring one-day granularity to avoid over-deleting.

## New env vars (all optional, sensible defaults)
- `PG_CAPACITY_BYTES=8589934592` — the Cloud SQL tier cap
- `PG_USAGE_TARGET_PCT=75` — stay under this fraction of the cap
- `PG_RETENTION_STEP_DAYS=1` — day granularity for the step-down

Defaults match the current dev tier, so no values-yaml changes are strictly required.

## Tests
9 new unit tests in `backend/__tests__/services/pgRetentionService.test.js` covering: invalid env guard, single-pass when under target, tier stepping, custom step size, 1-day floor + WARN, failed size probe graceful skip, custom capacity/target, swallowed delete errors, cron schedule signature.

## Test plan
- [ ] Review logic in `backend/services/pgRetentionService.ts`
- [ ] `cd backend && npx jest __tests__/services/pgRetentionService.test.js` → 9 passing
- [ ] Full backend suite still green
- [ ] After deploy, watch `kubectl logs -n commonly-dev -l app=backend | grep pg-retention` at 03:00 UTC for start/tier/done lines
- [ ] Manually trigger once to verify behavior against real Cloud SQL: `kubectl exec -n commonly-dev deployment/backend -- node -e "require('./services/pgRetentionService').runMessageRetention().then(()=>process.exit(0))"`

### DIFF
--- a/backend/__tests__/services/pgRetentionService.test.js
+++ b/backend/__tests__/services/pgRetentionService.test.js
@@ -1,0 +1,171 @@
+jest.mock('../../config/db-pg', () => ({
+  pool: { query: jest.fn() },
+}));
+jest.mock('../../models/pg/Message', () => ({
+  deleteOlderThan: jest.fn(),
+}));
+jest.mock('node-cron', () => ({ schedule: jest.fn(() => ({ start: jest.fn(), stop: jest.fn() })) }));
+
+const { pool } = require('../../config/db-pg');
+const Message = require('../../models/pg/Message');
+const cron = require('node-cron');
+const { runMessageRetention, initPgRetention } = require('../../services/pgRetentionService');
+
+const GIB = 1024 * 1024 * 1024;
+
+function mockSizeQueries(sizesGiB) {
+  const queue = [...sizesGiB];
+  pool.query.mockImplementation((sql) => {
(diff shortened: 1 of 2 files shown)


lilyshen0722:
Merged via local squash as 9b85557bc8 on main (followed by helm bump 893116bd38). Closing.

PR #215, opened 2026-04-18
AR: Agent Registry & Marketplace - Publish and Fork

rgu855:
## Summary

Self-serve publish & fork for the Commonly agent marketplace. Users can publish agent manifests under their own namespace (`@username/name`), version them, fork published manifests, and manage lifecycle (unpublish/delete/deprecate). This is effectively **ADR-001 Phase 2 for marketplace operations** — the first dual-write path between `Installable` (canonical) and `AgentRegistry` (compat shim).

**Design doc:** [`docs/design/marketplace-publish-fork.md`](docs/design/marketplace-publish-fork.md)

### New file: `backend/routes/marketplace-api.ts` (+603 lines)

9 endpoints under `/api/marketplace`:

| Endpoint | Method | Auth | Description |
|---|---|---|---|
| `/publish` | POST | ✅ | Publish new manifest or push new version |
| `/publish/:id` | DELETE | ✅ | Soft-delete (unpublish) — hides from browse, existing installs unaffected |
| `/manifests/:id` | DELETE | ✅ | Hard delete — only if 0 active installations (live count, not cached) |
| `/fork` | POST | ✅ | Snapshot-copy a manifest into your namespace |
| `/publish/:id/deprecate` | POST | ✅ | Mark a specific version as deprecated |
| `/browse` | GET | ❌ | Browse published manifests (filter by kind/category/q, sort, paginated) |
| `/manifests/:id` | GET | ❌ | Full manifest detail (readme, components, versions, fork lineage) |
| `/manifests/:id/forks` | GET | ❌ | List all forks of a manifest (paginated) |
| `/mine` | GET | ✅ | User's own manifests (including unpublished) |

Key implementation details:
- **Dual-write shim**: `syncToAgentRegistry(installable, action)` maps Installable fields to AgentRegistry on every write. Write order: AR first (load-bearing), then Installable. Failure matrix from §6c.1 implemented.
- **Namespace validation**: `@scope` must match `req.user.username` (server-enforced). Bare names reserved for builtins.
- **Runtime mapping**: ComponentRuntime (7 values) → manifest.runtime.type (3 values) via `RUNTIME_MAP`.
- **Route ordering**: forks sub-route registered before detail route to prevent wildcard `(*)` param from consuming `/forks`.
- **Username resolution**: JWT auth only sets `req.user = { id }` (no username), so `resolveUsername()` falls back to a DB lookup.

### Schema changes: `backend/models/Installable.ts`

- `installableId` regex relaxed: `/^[a-z0-9-]+(\/[a-z0-9-]+)?$/` → `/^(@[a-z0-9-]+\/)?[a-z0-9-]+$/`
- Added `forkedFrom?: { installableId, version, forkedAt }` (fork lineage)
- Added `readme?: string` (long-form description for detail page)
- Added `stats.forkCount` (default 0)
- Added indexes: `forkedFrom.installableId`, text search on `name + description + marketplace.tags`

### Schema changes: `backend/models/AgentRegistry.ts`

- `agentName` regex relaxed: `/^[a-z0-9-]+$/` → `/^(@[a-z0-9-]+\/)?[a-z0-9-]+$/`

### Install route: `backend/routes/registry/install.ts`

- Relaxed inline `agentName` regex to match the schema change
- Added status guard: rejects installs when `agent.status === 'unpublished'` (HTTP 410)

### Server wiring: `backend/server.ts`

- Mounted `marketplace-api` router at `/api/marketplace` (alongside existing `marketplace` router — no path conflicts)

### Test fix: `backend/__tests__/unit/models/installable.test.ts`

- Updated test `installableId` values from old `scope/name` format to new format (`bare-name` for builtins, `@scope/name` for marketplace)

### Input validation (§6e)

| Field | Rule | Enforced |
|---|---|---|
| `installableId` | Required, `@username/name` for user manifests, max 64 chars | ✅ |
| `name` | Required, max 100 chars | ✅ |
| `description` | Max 500 chars | ✅ |
| `version` | Required | ✅ |
| `kind` | Required, enum: agent/app/skill/bundle | ✅ |
| `scope` | `instance` rejected for user manifests | ✅ |
| `readme` | Max 50,000 chars | ✅ |
| `components` | Max 50 per manifest | ✅ |
| Namespace | `@scope` must match authed user's username | ✅ |

### Manual API testing (local Docker Compose)

All 9 endpoints tested end-to-end against real MongoDB:

| Test | Result |
|---|---|
| Publish new manifest (`@testuser/my-agent` v1.0.0) | ✅ 201 Created |
| Push new version (v2.0.0) | ✅ 200 OK, version appended |
| Idempotent re-publish (v1.0.0 again) | ✅ 200 OK, no-op |
| Fork manifest (`@testuser/my-agent-fork`) | ✅ 201, forkedFrom recorded |
| Deprecate version (v1.0.0) | ✅ deprecated: true |
| Browse published manifests | ✅ items returned, versions[] excluded from projection |
| Manifest detail | ✅ full doc with readme, components, versions |
| List forks of manifest | ✅ fork returned |
| My manifests | ✅ publisher's manifests returned |
| Unpublish (soft delete) | ✅ status → unpublished |
| Hard delete (0 installs) | ✅ document removed from both tables |
| Verify deleted returns 404 | ✅ |
| Dual-write: AR has correct data | ✅ agentName, latestVersion, runtime.type all match |
| Validation: name > 100 chars | ✅ 400 rejected |
| Validation: invalid kind | ✅ 400 rejected |
| Auth: JWT-based auth resolves username from DB | ✅ |

### Automated tests

```
Test Suites: 15 passed, 15 total (registry + install + installable + marketplace)
Tests:       40 passed, 40 total
```

Key suites:
- `installable.test.ts` — 12/12 pass (updated installableId formats for new regex)
- `self-serve-install.test.js` — 5/5 pass (ADR-006 flow unaffected)
- `marketplace.official.test.js` — 1/1 pass (existing browse unaffected)
- `registry.runtime-tokens.test.js` — 6/6 pass (token flow unaffected)
- All other registry tests — pass

### Regression checks

- Existing `GET /api/marketplace/official` endpoint: **unaffected** (separate router, no path conflicts)
- Existing `POST /api/registry/install` flow: **works** — dual-write ensures AR has the data, status guard added for unpublished manifests
- Existing 18 seeded agents: **unaffected** — bare names match both old and new regex
- ADR-006 self-serve webhook install: **unaffected** — 5/5 integration tests pass
- Agent runtime tokens: **unaffected** — 6/6 tests pass

## Architecture notes

- **Why dual-write?** AgentRegistry is the sole runtime catalog today. The install flow reads it exclusively. Until ADR-001 Phase 3 switches reads to Installable, we must keep both in sync.
- **Why AR first?** If only one write succeeds, an installable-but-not-browsable manifest is better than a browsable-but-not-installable one.
- **Why snapshot fork?** Git semantics — fork diverges freely, no forced upstream sync. Simpler model, avoids upstream-push complexity.
- **Why `source: 'marketplace'` for forks?** ADR-001's `template` source means admin-cloned templates, not user-to-user forks. Both source and fork are published catalog entries.

## Follow-up work (not in scope)

- [ ] Semver validation on version strings
- [ ] Per-component type/runtime validation
- [ ] Express body-parser 1MB limit on marketplace routes
- [ ] Reconciliation cron for Installable ↔ AgentRegistry drift (§6c.1)
- [ ] Frontend marketplace UI
- [ ] ADR-001 Phase 3: switch install read path from AgentRegistry to Installable

### DIFF
--- a/backend/routes/marketplace-api.ts
+++ b/backend/routes/marketplace-api.ts
@@ -1,0 +1,603 @@
+const express = require('express');
+const auth = require('../middleware/auth');
+const User = require('../models/User');
+const Installable = require('../models/Installable');
+const { AgentRegistry, AgentInstallation } = require('../models/AgentRegistry');
+
+const router = express.Router();
+
+const NAMESPACE_RE = /^@[a-z0-9-]+\/[a-z0-9-]+$/;
+
+const resolveUsername = async (req: any): Promise<string | null> => {
+  if (req.user?.username) return req.user.username;
+  const userId = req.userId || req.user?.id;
+  if (!userId) return null;
+  const user = await User.findById(userId).select('username').lean();
+  return user?.username || null;
+};
+
+const validateNamespace = (installableId: string, username: string) => {
+  if (!NAMESPACE_RE.test(installableId)) {
+    return 'installableId must match @<username>/<name> for user manifests';
+  }
+  const scope = installableId.split('/')[0].slice(1);
+  if (scope !== username.toLowerCase()) {
+    return 'Namespace does not match your username';
+  }
+  return null;
+};
+
+const RUNTIME_MAP: Record<string, string> = {
+  native: 'standalone',
+  internal: 'standalone',
+  'managed-agents': 'commonly-hosted',
+  webhook: 'standalone',
+  'claude-code': 'standalone',
+  moltbot: 'standalone',
+  remote: 'standalone',
+};
+
+const syncToAgentRegistry = async (installable: any, action: 'publish' | 'unpublish' | 'delete' | 'deprecate') => {
+  const agentName = installable.installableId;
+
+  if (action === 'delete') {
+    await AgentRegistry.deleteOne({ agentName });
+    return;
+  }
+
+  if (action === 'unpublish') {
+    await AgentRegistry.updateOne({ agentName }, { status: 'unpublished' });
+    return;
+  }
+
+  const agentComponent = (installable.components || []).find((c: any) => c.type === 'agent');
+  const runtimeType = agentComponent?.runtime
+    ? (RUNTIME_MAP[agentComponent.runtime] || 'standalone')
+    : 'standalone';
+
+  const capabilities = (installable.components || []).map((c: any) => ({
+    name: c.name,
+    description: c.description || c.name,
+  }));
+
+  const arPayload: any = {
+    agentName,
+    displayName: installable.name,
+    description: installable.description || '',
+    readme: installable.readme,
+    manifest: {
+      name: agentName,
+      version: installable.version,
+      description: installable.description || '',
+      capabilities,
+      context: { required: installable.requires || [], optional: [] },
+      runtime: { type: runtimeType, connection: 'rest' },
+    },
+    latestVersion: installable.version,
+    versions: (installable.versions || []).map((v: any) => ({
+      version: v.version,
+      publishedAt: v.publishedAt,
+      deprecated: v.deprecated,
+      deprecationReason: v.deprecationReason,
+    })),
+    registry: 'commonly-community',
+    publisher: installable.publisher,
+    categories: installable.marketplace?.category ? [installable.marketplace.category] : [],
+    tags: installable.marketplace?.tags || [],
+    status: installable.status,
+  };
+
+  await AgentRegistry.findOneAndUpdate(
+    { agentName },
+    { $set: arPayload },
+    { upsert: true, new: true },
+  );
+};
+
+/**
+ * POST /api/marketplace/publish
+ */
+router.post('/publish', auth, async (req: any, res: any) => {
+  try {
+    const userId = req.userId || req.user?.id;
+    const username = await resolveUsername(req);
+    if (!userId || !username) {
+      return res.status(401).json({ error: 'Unauthorized' });
+    }
+
+    const {
+      installableId, name, description, version, kind, scope,
+      requires, components, readme, categories, tags,
+    } = req.body;
+
+    if (!installableId || !name || !version || !kind) {
+      return res.status(400).json({ error: 'installableId, name, version, and kind are required' });
+    }
+
+    if (installableId.length > 64) {
+      return res.status(400).json({ error: 'installableId must be 64 characters or fewer' });
+    }
+    if (name.length > 100) {
+      return res.status(400).json({ error: 'name must be 100 characters or fewer' });
+    }
+    if (description && description.length > 500) {
+      return res.status(400).json({ error: 'description must be 500 characters or fewer' });
+    }
+    if (readme && readme.length > 50000) {
+      return res.status(400).json({ error: 'readme must be 50,000 characters or fewer' });
+    }
+    if (!['agent', 'app', 'skill', 'bundle'].includes(kind)) {
+      return res.status(400).json({ error: 'kind must be one of: agent, app, skill, bundle' });
+    }
+    if (components && components.length > 50) {
+      return res.status(400).json({ error: 'Maximum 50 components per manifest' });
+    }
+
+    const nsError = validateNamespace(installableId.toLowerCase(), username);
+    if (nsError) {
+      return res.status(400).json({ error: nsError });
+    }
+
+    if (scope === 'instance') {
+      return res.status(400).json({ error: 'scope "instance" is reserved for admin/builtin manifests' });
+    }
+
+    const existing = await Installable.findOne({ installableId: installableId.toLowerCase() });
+
+    if (existing) {
+      if (existing.publisher?.userId?.toString() !== userId.toString()) {
+        return res.status(403).json({ error: 'Not authorized to update this manifest' });
+      }
+
+      const existingVersion = (existing.versions || []).find(
+        (v: any) => v.version === version,
+      );
+      if (existingVersion) {
+        return res.status(200).json({
+          success: true,
+          manifest: {
+            installableId: existing.installableId,
+            version,
+            status: existing.status,
+            isNew: false,
+          },
+        });
+      }
+
+      existing.versions = [
+        ...(existing.versions || []),
+        { version, publishedAt: new Date() },
+      ];
+      existing.version = version;
+      existing.name = name;
+      existing.description = description || existing.description;
+      if (components) existing.components = components;
+      if (requires) existing.requires = requires;
+      if (readme !== undefined) existing.readme = readme;
+      if (categories || tags) {
+        existing.marketplace = {
+          ...existing.marketplace?.toObject?.() || existing.marketplace || {},
+          published: true,
+          category: categories?.[0] || existing.marketplace?.category || '',
+          tags: tags || existing.marketplace?.tags || [],
+        };
+      }
+
+      // Write AR first (load-bearing table), then Installable — same
+      // ordering rationale as new-manifest path (§6c.1).
+      try {
+        await syncToAgentRegistry(existing, 'publish');
+      } catch (arError) {
+        console.warn('[marketplace] AgentRegistry sync failed on update:', (arError as any).message);
+        return res.status(201).json({
+          success: true,
+          warnings: ['AgentRegistry sync failed; retrying publish will fix.'],
+          manifest: {
+            installableId: existing.installableId,
+            version,
+            status: existing.status,
+            isNew: false,
+          },
+        });
+      }
+
+      await existing.save();
+
+      return res.json({
+        success: true,
+        manifest: {
+          installableId: existing.installableId,
+          version,
+          status: existing.status,
+          isNew: false,
+        },
+      });
+    }
+
+    // New manifest: write to AgentRegistry first (load-bearing table), then Installable
+    const installableDoc = {
+      installableId: installableId.toLowerCase(),
+      name,
+      description: description || '',
+      version,
+      kind,
+      source: 'marketplace' as const,
+      scope: scope || 'pod',
+      requires: requires || [],
+      components: components || [],
+      readme,
+      marketplace: {
+        published: true,
+        category: categories?.[0] || '',
+        tags: tags || [],
+        verified: false,
+        rating: 0,
+        ratingCount: 0,
+        installCount: 0,
+      },
+      publisher: { userId, name: username },
+      status: 'active' as const,
+      versions: [{ version, publishedAt: new Date() }],
+      stats: { totalInstalls: 0, activeInstalls: 0, forkCount: 0 },
+    };
+
+    // Write AR first
+    try {
+      await syncToAgentRegistry(installableDoc, 'publish');
+    } catch (arError) {
+      console.error('[marketplace] AgentRegistry write failed:', (arError as any).message);
+      return res.status(500).json({ error: 'Failed to publish manifest' });
+    }
+
+    // Then Installable
+    let created;
+    try {
+      created = await Installable.create(installableDoc);
+    } catch (installableError) {
+      console.warn('[marketplace] Installable write failed (AR succeeded):', (installableError as any).message);
+      return res.status(201).json({
+        success: true,
+        warnings: ['Installable catalog write failed; manifest is installable but not yet browsable. Retry publish to sync.'],
+        manifest: {
+          installableId: installableDoc.installableId,
+          version,
+          status: 'active',
+          isNew: true,
+        },
+      });
+    }
+
+    console.log(`[marketplace] action=publish user=${userId} manifest=${created.installableId} version=${version}`);
+
+    res.status(201).json({
+      success: true,
+      manifest: {
+        installableId: created.installableId,
+        version: created.version,
+        status: created.status,
+        isNew: true,
+      },
+    });
+  } catch (error) {
+    console.error('[marketplace] publish error:', error);
+    res.status(500).json({ error: (error as any).message || 'Failed to publish' });
+  }
+});
+
+/**
+ * DELETE /api/marketplace/publish/:installableId
+ * Soft-delete (unpublish)
+ */
+router.delete('/publish/:installableId(*)', auth, async (req: any, res: any) => {
+  try {
+    const userId = req.userId || req.user?.id;
+    const { installableId } = req.params;
+
+    const doc = await Installable.findOne({ installableId: installableId.toLowerCase() });
+    if (!doc) return res.status(404).json({ error: 'Manifest not found' });
+    if (doc.publisher?.userId?.toString() !== userId.toString()) {
+      return res.status(403).json({ error: 'Not authorized' });
+    }
+
+    doc.status = 'unpublished';
+    if (doc.marketplace) doc.marketplace.published = false;
+    await doc.save();
+
+    try { await syncToAgentRegistry(doc, 'unpublish'); } catch (e) {
+      console.warn('[marketplace] AR sync failed on unpublish:', (e as any).message);
+    }
+
+    console.log(`[marketplace] action=unpublish user=${userId} manifest=${installableId}`);
+    res.json({ success: true, status: 'unpublished' });
+  } catch (error) {
+    console.error('[marketplace] unpublish error:', error);
+    res.status(500).json({ error: 'Failed to unpublish' });
+  }
+});
+
+/**
+ * DELETE /api/marketplace/manifests/:installableId
+ * Hard delete (only if 0 active installs)
+ */
+router.delete('/manifests/:installableId(*)', auth, async (req: any, res: any) => {
+  try {
+    const userId = req.userId || req.user?.id;
+    const { installableId } = req.params;
+
+    const doc = await Installable.findOne({ installableId: installableId.toLowerCase() });
+    if (!doc) return res.status(404).json({ error: 'Manifest not found' });
+    if (doc.publisher?.userId?.toString() !== userId.toString()) {
+      return res.status(403).json({ error: 'Not authorized' });
+    }
+
+    const liveCount = await AgentInstallation.countDocuments({
+      agentName: installableId.toLowerCase(),
+      status: 'active',
+    });
+    if (liveCount > 0) {
+      return res.status(409).json({
+        error: 'Cannot delete a manifest with active installations. Unpublish instead.',
+      });
+    }
+
+    // Delete AR first, then Installable
+    try { await AgentRegistry.deleteOne({ agentName: installableId.toLowerCase() }); } catch (e) {
+      console.warn('[marketplace] AR delete failed:', (e as any).message);
+    }
+    await Installable.deleteOne({ installableId: installableId.toLowerCase() });
+
+    console.log(`[marketplace] action=delete user=${userId} manifest=${installableId}`);
+    res.json({ success: true, deleted: true });
+  } catch (error) {
+    console.error('[marketplace] delete error:', error);
+    res.status(500).json({ error: 'Failed to delete' });
+  }
+});
+
+/**
+ * POST /api/marketplace/fork
+ */
+router.post('/fork', auth, async (req: any, res: any) => {
+  try {
+    const userId = req.userId || req.user?.id;
+    const username = await resolveUsername(req);
+    if (!userId || !username) {
+      return res.status(401).json({ error: 'Unauthorized' });
+    }
+
+    const { sourceInstallableId, newInstallableId, newName } = req.body;
+    if (!sourceInstallableId || !newInstallableId) {
+      return res.status(400).json({ error: 'sourceInstallableId and newInstallableId are required' });
+    }
+
+    const nsError = validateNamespace(newInstallableId.toLowerCase(), username);
+    if (nsError) return res.status(400).json({ error: nsError });
+
+    const source = await Installable.findOne({
+      installableId: sourceInstallableId.toLowerCase(),
+      status: 'active',
+    });
+    if (!source) return res.status(404).json({ error: 'Source manifest not found or not active' });
+
+    const existing = await Installable.findOne({ installableId: newInstallableId.toLowerCase() });
+    if (existing) return res.status(409).json({ error: 'installableId already taken' });
+
+    const latestVersion = source.version || '1.0.0';
+
+    const forkDoc = {
+      installableId: newInstallableId.toLowerCase(),
+      name: newName || source.name,
+      description: source.description || '',
+      version: '1.0.0',
+      kind: source.kind,
+      source: 'marketplace' as const,
+      scope: source.scope,
+      requires: source.requires || [],
+      components: source.components || [],
+      readme: source.readme,
+      marketplace: {
+        published: true,
+        category: source.marketplace?.category || '',
+        tags: source.marketplace?.tags || [],
+        verified: false,
+        rating: 0,
+        ratingCount: 0,
+        installCount: 0,
+      },
+      publisher: { userId, name: username },
+      status: 'active' as const,
+      versions: [{ version: '1.0.0', publishedAt: new Date() }],
+      forkedFrom: {
+        installableId: source.installableId,
+        version: latestVersion,
+        forkedAt: new Date(),
+      },
+      stats: { totalInstalls: 0, activeInstalls: 0, forkCount: 0 },
+    };
+
+    // Write AR first, then Installable
+    try {
+      await syncToAgentRegistry(forkDoc, 'publish');
+    } catch (arError) {
+      console.error('[marketplace] AR write failed on fork:', (arError as any).message);
+      return res.status(500).json({ error: 'Failed to fork manifest' });
+    }
+
+    const created = await Installable.create(forkDoc);
+
+    // Increment fork count on source (atomic)
+    await Installable.updateOne(
+      { installableId: source.installableId },
+      { $inc: { 'stats.forkCount': 1 } },
+    );
+
+    console.log(`[marketplace] action=fork user=${userId} source=${sourceInstallableId} target=${newInstallableId}`);
+
+    res.status(201).json({
+      success: true,
+      manifest: {
+        installableId: created.installableId,
+        version: '1.0.0',
+        forkedFrom: created.forkedFrom,
+      },
+    });
+  } catch (error) {
+    console.error('[marketplace] fork error:', error);
+    res.status(500).json({ error: (error as any).message || 'Failed to fork' });
+  }
+});
+
+/**
+ * POST /api/marketplace/publish/:installableId/deprecate
+ */
+router.post('/publish/:installableId(*)/deprecate', auth, async (req: any, res: any) => {
+  try {
+    const userId = req.userId || req.user?.id;
+    const { installableId } = req.params;
+    const { version, reason } = req.body;
+
+    if (!version) return res.status(400).json({ error: 'version is required' });
+
+    const doc = await Installable.findOne({ installableId: installableId.toLowerCase() });
+    if (!doc) return res.status(404).json({ error: 'Manifest not found' });
+    if (doc.publisher?.userId?.toString() !== userId.toString()) {
+      return res.status(403).json({ error: 'Not authorized' });
+    }
+
+    const versionEntry = (doc.versions || []).find((v: any) => v.version === version);
+    if (!versionEntry) return res.status(404).json({ error: 'Version not found' });
+
+    versionEntry.deprecated = true;
+    versionEntry.deprecationReason = reason || '';
+    await doc.save();
+
+    try {
+      await syncToAgentRegistry(doc, 'deprecate');
+    } catch (e) {
+      console.warn('[marketplace] AR sync failed on deprecate:', (e as any).message);
+    }
+
+    console.log(`[marketplace] action=deprecate user=${userId} manifest=${installableId} version=${version}`);
+    res.json({ success: true, version, deprecated: true });
+  } catch (error) {
+    console.error('[marketplace] deprecate error:', error);
+    res.status(500).json({ error: 'Failed to deprecate version' });
+  }
+});
+
+/**
+ * GET /api/marketplace/browse
+ */
+router.get('/browse', async (req: any, res: any) => {
+  try {
+    const {
+      kind, category, q, sort = 'installs', page = '1', limit = '20',
+    } = req.query;
+    const pageNum = Math.max(1, parseInt(page, 10) || 1);
+    const limitNum = Math.min(100, Math.max(1, parseInt(limit, 10) || 20));
+    const skip = (pageNum - 1) * limitNum;
+
+    const filter: any = {
+      source: 'marketplace',
+      status: 'active',
+      'marketplace.published': true,
+    };
+    if (kind) filter.kind = kind;
+    if (category) filter['marketplace.category'] = category;
+    if (q) filter.$text = { $search: q };
+
+    const sortMap: Record<string, any> = {
+      installs: { 'stats.totalInstalls': -1 },
+      rating: { 'marketplace.rating': -1 },
+      newest: { createdAt: -1 },
+      forks: { 'stats.forkCount': -1 },
+    };
+    const sortObj = sortMap[sort] || sortMap.installs;
+
+    const [items, total] = await Promise.all([
+      Installable.find(filter)
+        .select('-readme -components -versions')
+        .sort(sortObj)
+        .skip(skip)
+        .limit(limitNum)
+        .lean(),
+      Installable.countDocuments(filter),
+    ]);
+
+    res.json({ items, total, page: pageNum, limit: limitNum });
+  } catch (error) {
+    console.error('[marketplace] browse error:', error);
+    res.status(500).json({ error: 'Failed to browse marketplace' });
+  }
+});
+
+/**
+ * GET /api/marketplace/manifests/:installableId/forks
+ * Must be registered before the detail route so the wildcard doesn't eat "/forks".
+ */
+router.get('/manifests/:installableId(*)/forks', async (req: any, res: any) => {
+  try {
+    const { installableId } = req.params;
+    const page = Math.max(1, parseInt(req.query.page, 10) || 1);
+    const limit = Math.min(100, Math.max(1, parseInt(req.query.limit, 10) || 20));
+    const skip = (page - 1) * limit;
+
+    const filter = {
+      'forkedFrom.installableId': installableId.toLowerCase(),
+      status: 'active',
+    };
+
+    const [items, total] = await Promise.all([
+      Installable.find(filter)
+        .select('-readme -components -versions')
+        .sort({ 'stats.totalInstalls': -1 })
+        .skip(skip)
+        .limit(limit)
+        .lean(),
+      Installable.countDocuments(filter),
+    ]);
+
+    res.json({ items, total, page, limit });
+  } catch (error) {
+    console.error('[marketplace] forks error:', error);
+    res.status(500).json({ error: 'Failed to list forks' });
+  }
+});
+
+/**
+ * GET /api/marketplace/manifests/:installableId
+ */
+router.get('/manifests/:installableId(*)', async (req: any, res: any) => {
+  try {
+    const { installableId } = req.params;
+    const doc = await Installable.findOne({ installableId: installableId.toLowerCase() }).lean();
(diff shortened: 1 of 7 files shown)


docs/design/marketplace-publish-fork.md
```diff
@@ -0,0 +1,874 @@
...
+Namespace format:
+
+  Official (first-party):     claude-code
+                              pod-welcomer
+                              task-clerk
+
+  User-scoped:                @grasstoucher/lebron-code
+                              @lily/research-bot
+
+  Regex:                      /^(@[a-z0-9-]+\/)?[a-z0-9-]+$/
```
samxu01: **P0 — namespace format is rejected by both existing schemas.**

- `backend/models/Installable.ts:466` — `installableId` is `match: /^[a-z0-9-]+(\/[a-z0-9-]+)?$/` with `lowercase: true`. `@` is not permitted, and uppercase usernames get silently lowercased.
- `backend/models/AgentRegistry.ts:112` — `agentName` is `match: /^[a-z0-9-]+$/`. Neither `@` nor `/` is permitted.

The `@scope/name` form proposed here will fail Mongoose validation on the very first publish. Options: (a) drop the `@` and use bare `scope/name` (Installable regex already allows it, only AR regex needs to relax), (b) relax both regexes and document the impact on the 18 seeded entries, or (c) translate to a safe encoding (`scope--name`) on the AR side and document it. Please pick one and update §6c.2/§6c.4 accordingly.

---

docs/design/marketplace-publish-fork.md
```diff
@@ -0,0 +1,874 @@
...
+
+```
+Installable.installableId     →  AgentRegistry.agentName
+────────────────────────────     ──────────────────────────
+@grasstoucher/lebron-code     →  @grasstoucher/lebron-code
+@lily/research-bot            →  @lily/research-bot
+claude-code (builtin)         →  claude-code (unchanged)
+```
+
+The `agentName` field on AgentRegistry already accepts arbitrary strings
```
samxu01: **P0 — factually incorrect.** `AgentRegistry.agentName` is declared as `{ type: String, required: true, unique: true, lowercase: true, match: /^[a-z0-9-]+$/ }` at `backend/models/AgentRegistry.ts:112`. The regex explicitly rejects both `@` and `/`, so the shim's claim that the full scoped name can be stored unchanged is wrong. Any user publish under the design as-written will 500 on the AR write. This needs to be reconciled with issue raised at §5.3.

---

docs/design/marketplace-publish-fork.md
```diff
@@ -0,0 +1,874 @@
...
+
+Every lifecycle operation on the Installable must be mirrored to the
+corresponding AgentRegistry row. If only the Installable is updated, the
+legacy install flow continues to serve stale state.
+
+**Sync rules:**
+
+| Operation | Installable action | AgentRegistry mirror |
+|---|---|---|
+| **Unpublish** | `status → 'unpublished'`, `marketplace.published → false` | `status → 'deprecated'` (closest equivalent). Install flow must check `status !== 'deprecated'` before allowing new installs. |
```
samxu01: **P0 — map to `'unpublished'`, not `'deprecated'`.** `AgentRegistry.RegistryStatus` already includes `'unpublished'` as a valid enum value (`backend/models/AgentRegistry.ts:4`). Using `'deprecated'` conflates "author retired a specific version" with "author pulled the manifest from the catalog," which will confuse downstream clients and the install-guard check. Use the matching enum.

---

docs/design/marketplace-publish-fork.md
```diff
@@ -0,0 +1,874 @@
...
+##### `GET /api/marketplace/mine` — User's manifests
+
+Returns manifests where `publisher.userId === req.user.id`. Includes
+unpublished manifests (for the user's management dashboard).
+
+#### Endpoints that need modifying
+
+##### `POST /api/registry/install` (`backend/routes/registry/install.ts`)
+
+**No changes in this PR.** The install flow continues to resolve from
```
samxu01: "No changes in this PR" understates the work. The install route calls `AgentRegistry.getByName(agentName)` which lowercases and hits the unique-keyed `agentName`. Even if the AR regex is relaxed (see §5.3 comment), the route's request-body validation / Express input handling likely rejects `@` and `/` before the DB call. Plus §6c.3 already proposes a new status guard in this same file. Please be explicit: install.ts *is* modified here (status guard + probably input-validation relaxation). Otherwise the first user-published install 400s.

---

docs/design/marketplace-publish-fork.md
```diff
@@ -0,0 +1,874 @@
...
+
+```
+Version lifecycle:
+
+  v1.0.0  ──▶  active (installable, visible)
+  v1.1.0  ──▶  active
+  v2.0.0  ──▶  active (latest)
+  v1.0.0  ──▶  deprecated (installable with warning, reason shown)
+
+  "latest" always resolves to the most recent non-deprecated version.
```
samxu01: "Most recent" is ambiguous. Consider: publish v1.0.0 → v2.0.0 → v1.0.1 (a backport to the v1 line). Which is "latest"? npm chooses the version whose `publishedAt` is newest *only* when not managed via `dist-tags`; otherwise `latest` is the maintainer-chosen tag with full semver precedence elsewhere. Docker uses tag pointers.

Please define: is resolution `publishedAt`-ordered, semver-precedence-ordered, or an explicit pointer the publisher sets? Also — when does `Installable.version` (the scalar top-level field) get overwritten? Every publish? Only when semver-greater?

---

docs/design/marketplace-publish-fork.md
```diff
@@ -0,0 +1,874 @@
...
+  Installs without an explicit version pin get "latest".
+
+  A version string cannot be reused with different content (NFR-2
+  handles identical re-publish as no-op). This matches Docker's
```
samxu01, line 296: "Identical content" needs a concrete definition, otherwise NFR-2 (idempotency) and the 409-on-mismatch behavior in §6b are implementation-defined and will drift. Suggest: SHA-256 of a canonicalized JSON of the publish payload restricted to a fixed field set (`components`, `requires`, `scope`, `kind`) — explicitly *excluding* `readme`, `description`, `marketplace.tags`, so publishers can patch marketing copy without a version bump (consistent with open question #5). Store the hash on the `versions[]` entry so the idempotency check is a direct compare.

---

docs/design/marketplace-publish-fork.md
```diff
@@ -0,0 +1,874 @@
...
+**Response:** `{ success: true, status: 'unpublished' }`
+
+**Guards:**
+- 403 if `publisher.userId` does not match.
+- 404 if not found.
+
+##### `DELETE /api/marketplace/manifests/:installableId` — Hard delete
+
+Permanently deletes the manifest document. Only allowed when
+`stats.activeInstalls === 0`.
```
samxu01: Race condition: `stats.activeInstalls` is a cached counter with no specified update path in this design, so this guard will accept stale values. A new `AgentInstallation` created between the guard check and the delete would be orphaned.

Recommend replacing the cached check with a live count at delete time: `AgentInstallation.countDocuments({ agentName, status: 'active' })`. If you prefer to keep the counter, add a conditional update that atomically asserts `activeInstalls === 0` before deletion.

---

docs/design/marketplace-publish-fork.md
```diff
@@ -0,0 +1,874 @@
...
+Creates a deep copy of the source manifest under the forker's namespace.
+
+**Request:**
+
+```json
+{
+  "sourceInstallableId": "claude-code",
+  "newInstallableId": "@grasstoucher/my-claude-code",
+  "newName": "My Claude Code",
+  "version": "1.0.0"
```
samxu01: What does `version` mean in the fork request? Two valid interpretations:

1. The version of the *source* the forker is pinning against (so you'd look it up in `source.versions[]` and deep-copy that historical snapshot's components).
2. The initial version string the *fork* starts at.

The body reads like (2), but the diagram in §5.2 and the stored `forkedFrom.version` imply (1). These aren't equivalent — fork-from-old-version is a useful feature but requires walking `versions[]` to find the right snapshot. Please disambiguate and, if (1), spell out how components are retrieved from a historical version entry (the current schema has `versions[i].manifest` only on AR, not Installable).

---

docs/design/marketplace-publish-fork.md
```diff
@@ -0,0 +1,874 @@
...
+1. Load source by `sourceInstallableId`. 404 if not found or
+   `status !== 'active'`.
+2. Validate `newInstallableId` does not exist (409 if taken).
+3. Validate `@scope` matches `req.user.username`.
+4. Deep-copy: `components`, `requires`, `scope`, `kind`, `readme`
+   (fork starts with source's readme; forker can update on next publish).
+   Do NOT copy `stats`, `marketplace` (fresh start), `publisher`,
+   `owner`, `versions`.
+5. Set `forkedFrom: { installableId, version: source.version, forkedAt }`.
+6. Set `publisher: { userId, name }`, `source: 'marketplace'`.
```
samxu01, line 536: ADR-001 already defines `source: 'template'` as "cloned from a template" — which is semantically closer to a fork than `source: 'marketplace'` ("published to the public marketplace"). A fork is both: it was *cloned* from another Installable *and* it is published to the marketplace.

Today's `InstallableSource` is a single enum, so we can't be both. Either pick `'template'` (and distinguish "published" via `marketplace.published: true`, which is already orthogonal), or justify why `'marketplace'` wins. Whatever you decide, update ADR-001's source taxonomy description so the semantics are explicit.

---

docs/design/marketplace-publish-fork.md
```diff
@@ -0,0 +1,874 @@
...
+Existing installations are not affected — they reference the manifest by
+`agentName` on their `AgentInstallation` row, which is independent of the
+catalog entry's status. The guard only prevents *new* installs.
+
+**Delete ordering**: delete AgentRegistry first, then Installable. If the
+Installable delete fails, the manifest is invisible in both browse
+endpoints (AR gone, Installable still present but no AR to install from).
+The reconciliation cron (§6c.1) catches this and logs it.
+
+#### 6c.4 Field mapping reference
```
samxu01, line 835: This table skips the three hardest mappings. Please fill in:

1. **`components[].runtime`** values (`native|moltbot|webhook|claude-code|managed-agents|internal|remote`) do not map to `manifest.runtime.type` values (`standalone|commonly-hosted|hybrid`). These taxonomies are disjoint. Pick an explicit mapping or add a field to AR.
2. **`components[].persona.systemPrompt`** — noted as "AgentProfile on install, not AR," but when the install flow runs on a user-published manifest resolved via AR (before Phase 3), where does `systemPrompt` come from? AR's `IManifest` has no field for it.
3. **`marketplace.category`** (scalar on Installable) vs AR's `categories[]` (array). Design §6b shows publish request with `categories: ["development"]` as an array, so the payload itself is inconsistent with the Installable schema. Pick the canonical side.

Also: `kind` (agent|app|skill|bundle) is silently dropped on the AR side. If install-time logic ever needs `kind`, the dual-write breaks.

---

docs/design/marketplace-publish-fork.md
```diff
@@ -0,0 +1,874 @@
...
+**Failure matrix:**
+
+```
+AgentRegistry    Installable     Outcome
+─────────────    ───────────     ────────────────────────────────────
+✅ success       ✅ success      Happy path. Both tables consistent.
+
+✅ success       ❌ fail         Manifest is installable via legacy
+                                 flow but invisible in new browse.
+                                 → Log warning. Return 207 partial
```
samxu01: HTTP 207 is a WebDAV code and isn't used elsewhere in this API. A 201 Created with a `warnings[]` field in the response body is closer to the rest of the project's conventions and is easier for clients to branch on.

Also missing: client guidance on 207 — retry automatically, surface a banner, or treat as success? Spell it out so clients don't diverge.

---

samxu01 (changes requested):
Structurally sound RFC — good precedents cited, explicit failure matrix, and clear open questions. Requesting changes because several concrete claims and field mappings conflict with the existing `Installable` / `AgentRegistry` schemas and would block implementation as written.

**P0 (blockers):**
- §5.3 namespace regex doesn't match either existing schema's field constraints — first publish would fail Mongoose validation.
- §6c.2 asserts `AgentRegistry.agentName` accepts arbitrary strings; the schema's regex explicitly rejects `@` and `/`.
- §6c.3 maps unpublish to AR `status: 'deprecated'` when `'unpublished'` is already a valid enum value.
- §6b / §6c.3: install route changes are understated — a status guard plus (probably) input-validation relaxation are required, not zero changes.

**P1 (worth tightening before coding):**
- "Latest" version resolution and `Installable.version` scalar update rules undefined.
- NFR-2 "identical content" needs a concrete hashing spec.
- `stats.activeInstalls === 0` hard-delete guard races against new installs.
- §6c.4 field map skips the hardest three cases (runtime, persona.systemPrompt, categories shape).
- Fork request `version` semantics ambiguous (forks-from-old-version vs. initial-version-of-fork).
- Fork `source: 'marketplace'` vs ADR-001 `'template'` unreconciled.
- HTTP 207 is unconventional; prefer 201 + warnings payload.

Strengths worth keeping: write-ordering rationale in §6c.1, lifecycle sync table in §6c.3, server-enforced namespace ownership, and identity-continuity callouts.

---

docs/design/marketplace-publish-fork.md
rgu855: Fixed. Added §6f (Required Schema Migrations) with the exact file:line regex changes needed:

- `Installable.installableId` regex → `/^(@[a-z0-9-]+\/)?[a-z0-9-]+$/` (line 466)
- `AgentRegistry.agentName` regex → same (line 112)
- `install.ts` inline check → same (line 123)

Also updated §5.3 to document the schema mismatch explicitly with a "Schema changes required" callout.

docs/design/marketplace-publish-fork.md
rgu855: Fixed. The incorrect claim that AgentRegistry accepts arbitrary strings has been replaced. §6c.2 now explicitly calls out the current regex (`/^[a-z0-9-]+$/` at line 112) and states the migration is required. §6f has the full migration table with all three file:line locations.

docs/design/marketplace-publish-fork.md
rgu855: Fixed. Unpublish now maps to `'unpublished'` (not `'deprecated'`) in both the §6c.3 sync table and the install guard code snippet. Verified that AgentRegistry's status enum at line 142 already includes `'unpublished'` — exact match, no conflation.

docs/design/marketplace-publish-fork.md
rgu855: Fixed. Changed from "No changes in this PR" to "Minimal changes required" with two explicit items:

1. Relax `agentName` validation at line 123 to accept `@scope/name`
2. Add status guard for `'unpublished'` (code snippet in §6c.3)

Both are now documented in the install route section and cross-referenced in §6f.

docs/design/marketplace-publish-fork.md
rgu855: Fixed. §5.4 now defines resolution explicitly: "latest" resolves by `publishedAt` timestamp — the most recently published non-deprecated version. Not by semver ordering (so a hotfix v1.0.1 published after v2.0.0 becomes latest). The top-level `Installable.version` scalar is updated on every publish to match latest.

This matches Docker's tag-pointer model. We can add explicit dist-tag support (like npm's `latest`/`next`) later if needed, but timestamp ordering is sufficient for v1.

docs/design/marketplace-publish-fork.md
rgu855: Good call. The current doc doesn't define "identical content" concretely. Adding this to open questions for now — the SHA-256 approach over canonicalized `(components, requires, scope, kind)` is the right direction and aligns with open question #5 (metadata-mutable vs content-immutable). Will spec the exact hash implementation during the implementation PR.

docs/design/marketplace-publish-fork.md
rgu855: Fixed. Hard delete now uses a live count query (`AgentInstallation.countDocuments({ agentName, status: 'active' })`) instead of the cached `stats.activeInstalls` counter. This eliminates the race condition between the guard check and deletion.

docs/design/marketplace-publish-fork.md
rgu855: Fixed. Removed `version` from the fork request entirely. The fork always copies from the source's current latest version, and the fork's initial version is set to `"1.0.0"` (fresh start with independent version history). `forkedFrom.version` records which source version the snapshot was taken from — a historical pointer, not a live link.

docs/design/marketplace-publish-fork.md
rgu855: Keeping `source: 'marketplace'`. Added explicit rationale in the doc:

ADR-001's `template` means "cloned from a pre-built template by an admin" — it's an authoring provenance, not a derivation method. A fork is a marketplace-to-marketplace derivation: both the source and the fork are published, discoverable catalog entries. The `forkedFrom` pointer distinguishes forks from original publishes, and `marketplace.published: true` is already orthogonal to source.

If we used `source: 'template'`, a forked manifest that the user then publishes would need to flip from `template` → `marketplace`, which breaks the invariant that source describes provenance (where it came from), not current state. The fork came from the marketplace and lives in the marketplace.

docs/design/marketplace-publish-fork.md
rgu855: Fixed. Added an explicit runtime mapping table to §6c.4:

```
ComponentRuntime          →  manifest.runtime.type
native, internal          →  standalone
managed-agents            →  commonly-hosted
webhook, claude-code,     →  standalone (runs externally,
  moltbot, remote            connects via REST/webhook)
```

These are intentionally different abstractions — `ComponentRuntime` describes WHERE the agent runs, `manifest.runtime.type` describes the DEPLOYMENT shape. Also clarified:
- `marketplace.category` (scalar) → `categories[0]` (wrapped to array)
- `persona.systemPrompt` is NOT written to AR — it goes to `AgentProfile` at install time via the existing install route

docs/design/marketplace-publish-fork.md
rgu855: Fixed. Replaced 207 with 201 Created + a `warnings[]` array in the response body. Confirmed no 207 usage anywhere in the codebase. Client guidance: the manifest is installable (AR write succeeded), but not yet browsable in the new marketplace endpoint. User can retry publish to sync.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+    { agentName },
+    { $set: arPayload },
+    { upsert: true, new: true },
+  );
+};
+
+/**
+ * POST /api/marketplace/publish
+ */
+router.post('/publish', auth, async (req: any, res: any) => {
```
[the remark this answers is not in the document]
rgu855: Dismissed — rate limiting is intentionally deferred for v1 (design doc open question #3). All write endpoints require JWT auth, and the platform is invite-only (ADR-006 §invariant 4). No other route file in the codebase uses per-endpoint rate limiting except `agentsRuntime.ts` (cross-agent surface). Will add publish rate caps when Commonly opens to public signup.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+        version: created.version,
+        status: created.status,
+        isNew: true,
+      },
+    });
+  } catch (error) {
+    console.error('[marketplace] publish error:', error);
+    res.status(500).json({ error: (error as any).message || 'Failed to publish' });
+  }
+});
```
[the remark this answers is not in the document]
rgu855: Dismissed — rate limiting is intentionally deferred for v1 (design doc open question #3). All write endpoints require JWT auth, and the platform is invite-only (ADR-006 §invariant 4). No other route file in the codebase uses per-endpoint rate limiting except `agentsRuntime.ts` (cross-agent surface). Will add publish rate caps when Commonly opens to public signup.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+
+/**
+ * DELETE /api/marketplace/publish/:installableId
+ * Soft-delete (unpublish)
+ */
+router.delete('/publish/:installableId(*)', auth, async (req: any, res: any) => {
```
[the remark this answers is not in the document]
rgu855: Dismissed — rate limiting is intentionally deferred for v1 (design doc open question #3). All write endpoints require JWT auth, and the platform is invite-only (ADR-006 §invariant 4). No other route file in the codebase uses per-endpoint rate limiting except `agentsRuntime.ts` (cross-agent surface). Will add publish rate caps when Commonly opens to public signup.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+      console.warn('[marketplace] AR sync failed on unpublish:', (e as any).message);
+    }
+
+    console.log(`[marketplace] action=unpublish user=${userId} manifest=${installableId}`);
+    res.json({ success: true, status: 'unpublished' });
+  } catch (error) {
+    console.error('[marketplace] unpublish error:', error);
+    res.status(500).json({ error: 'Failed to unpublish' });
+  }
+});
```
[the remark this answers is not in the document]
rgu855: Dismissed — rate limiting is intentionally deferred for v1 (design doc open question #3). All write endpoints require JWT auth, and the platform is invite-only (ADR-006 §invariant 4). No other route file in the codebase uses per-endpoint rate limiting except `agentsRuntime.ts` (cross-agent surface). Will add publish rate caps when Commonly opens to public signup.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+
+/**
+ * DELETE /api/marketplace/manifests/:installableId
+ * Hard delete (only if 0 active installs)
+ */
+router.delete('/manifests/:installableId(*)', auth, async (req: any, res: any) => {
```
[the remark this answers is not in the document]
rgu855: Dismissed — rate limiting is intentionally deferred for v1 (design doc open question #3). All write endpoints require JWT auth, and the platform is invite-only (ADR-006 §invariant 4). No other route file in the codebase uses per-endpoint rate limiting except `agentsRuntime.ts` (cross-agent surface). Will add publish rate caps when Commonly opens to public signup.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+    }
+    await Installable.deleteOne({ installableId: installableId.toLowerCase() });
+
+    console.log(`[marketplace] action=delete user=${userId} manifest=${installableId}`);
+    res.json({ success: true, deleted: true });
+  } catch (error) {
+    console.error('[marketplace] delete error:', error);
+    res.status(500).json({ error: 'Failed to delete' });
+  }
+});
```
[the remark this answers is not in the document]
rgu855: Dismissed — rate limiting is intentionally deferred for v1 (design doc open question #3). All write endpoints require JWT auth, and the platform is invite-only (ADR-006 §invariant 4). No other route file in the codebase uses per-endpoint rate limiting except `agentsRuntime.ts` (cross-agent surface). Will add publish rate caps when Commonly opens to public signup.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+
+/**
+ * POST /api/marketplace/fork
+ */
+router.post('/fork', auth, async (req: any, res: any) => {
```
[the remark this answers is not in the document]
rgu855: Dismissed — rate limiting is intentionally deferred for v1 (design doc open question #3). All write endpoints require JWT auth, and the platform is invite-only (ADR-006 §invariant 4). No other route file in the codebase uses per-endpoint rate limiting except `agentsRuntime.ts` (cross-agent surface). Will add publish rate caps when Commonly opens to public signup.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+        installableId: created.installableId,
+        version: '1.0.0',
+        forkedFrom: created.forkedFrom,
+      },
+    });
+  } catch (error) {
+    console.error('[marketplace] fork error:', error);
+    res.status(500).json({ error: (error as any).message || 'Failed to fork' });
+  }
+});
```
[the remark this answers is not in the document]
rgu855: Dismissed — rate limiting is intentionally deferred for v1 (design doc open question #3). All write endpoints require JWT auth, and the platform is invite-only (ADR-006 §invariant 4). No other route file in the codebase uses per-endpoint rate limiting except `agentsRuntime.ts` (cross-agent surface). Will add publish rate caps when Commonly opens to public signup.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+
+/**
+ * POST /api/marketplace/publish/:installableId/deprecate
+ */
+router.post('/publish/:installableId(*)/deprecate', auth, async (req: any, res: any) => {
```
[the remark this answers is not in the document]
rgu855: Dismissed — rate limiting is intentionally deferred for v1 (design doc open question #3). All write endpoints require JWT auth, and the platform is invite-only (ADR-006 §invariant 4). Will add publish rate caps when Commonly opens to public signup.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+      console.warn('[marketplace] AR sync failed on deprecate:', (e as any).message);
+    }
+
+    console.log(`[marketplace] action=deprecate user=${userId} manifest=${installableId} version=${version}`);
+    res.json({ success: true, version, deprecated: true });
+  } catch (error) {
+    console.error('[marketplace] deprecate error:', error);
+    res.status(500).json({ error: 'Failed to deprecate version' });
+  }
+});
```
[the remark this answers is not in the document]
rgu855: Dismissed — rate limiting is intentionally deferred for v1 (design doc open question #3). All write endpoints require JWT auth, and the platform is invite-only (ADR-006 §invariant 4). Will add publish rate caps when Commonly opens to public signup.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+    const sortMap: Record<string, any> = {
+      installs: { 'stats.totalInstalls': -1 },
+      rating: { 'marketplace.rating': -1 },
+      newest: { createdAt: -1 },
+      forks: { 'stats.forkCount': -1 },
+    };
+    const sortObj = sortMap[sort] || sortMap.installs;
+
+    const [items, total] = await Promise.all([
+      Installable.find(filter)
```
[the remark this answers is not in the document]
rgu855: Dismissed — rate limiting is intentionally deferred for v1 (design doc open question #3). All write endpoints require JWT auth, and the platform is invite-only (ADR-006 §invariant 4). Will add publish rate caps when Commonly opens to public signup.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+        .select('-readme -components -versions')
+        .sort(sortObj)
+        .skip(skip)
+        .limit(limitNum)
+        .lean(),
+      Installable.countDocuments(filter),
```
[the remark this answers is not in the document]
rgu855: Dismissed — rate limiting is intentionally deferred for v1 (design doc open question #3). All write endpoints require JWT auth, and the platform is invite-only (ADR-006 §invariant 4). Will add publish rate caps when Commonly opens to public signup.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+  try {
+    const { installableId } = req.params;
+    const doc = await Installable.findOne({ installableId: installableId.toLowerCase() }).lean();
+    if (!doc) return res.status(404).json({ error: 'Manifest not found' });
+    res.json(doc);
+  } catch (error) {
+    console.error('[marketplace] detail error:', error);
+    res.status(500).json({ error: 'Failed to get manifest' });
+  }
+});
```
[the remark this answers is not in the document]
rgu855: Dismissed — rate limiting is intentionally deferred for v1 (design doc open question #3). All write endpoints require JWT auth, and the platform is invite-only (ADR-006 §invariant 4). Will add publish rate caps when Commonly opens to public signup.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+
+/**
+ * GET /api/marketplace/mine
+ */
+router.get('/mine', auth, async (req: any, res: any) => {
```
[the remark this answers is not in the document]
rgu855: Dismissed — rate limiting is intentionally deferred for v1 (design doc open question #3). All write endpoints require JWT auth, and the platform is invite-only (ADR-006 §invariant 4). Will add publish rate caps when Commonly opens to public signup.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+    const items = await Installable.find({ 'publisher.userId': userId })
+      .sort({ updatedAt: -1 })
+      .lean();
+
+    res.json({ items, total: items.length });
+  } catch (error) {
+    console.error('[marketplace] mine error:', error);
+    res.status(500).json({ error: 'Failed to list your manifests' });
+  }
+});
```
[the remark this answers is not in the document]
rgu855: Dismissed — rate limiting is intentionally deferred for v1 (design doc open question #3). All write endpoints require JWT auth, and the platform is invite-only (ADR-006 §invariant 4). Will add publish rate caps when Commonly opens to public signup.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+          manifest: {
+            installableId: existing.installableId,
+            version,
+            status: existing.status,
+            isNew: false,
+          },
+        });
+      }
+
+      await existing.save();
```
lilyshen0722, line 204: `existing.save()` is outside the AR try/catch, so the live sequence is:

1. Stage mutations on `existing` (lines 167–184)
2. `syncToAgentRegistry` succeeds → AR now has the new version
3. `existing.save()` throws (unique index collision, validation, DB hiccup…)
4. Outer catch returns generic 500
5. AR committed forward; Installable rolled back → silent drift

The AR-failure early-return above is fine (nothing commits on either side). The missing case is AR-succeeds-then-save-fails. Suggested:

```ts
try {
  await syncToAgentRegistry(existing, 'publish');
  await existing.save();
} catch (err) {
  // if AR succeeded but save failed, either revert AR or
  // return 500 so the caller doesn't think it's published
  console.error('[marketplace] update dual-write failed', (err as any).message);
  return res.status(500).json({ error: 'Failed to update manifest' });
}
```

The design doc §6c.1 talks about this in principle; just needs to land in code.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+
+    // Write AR first, then Installable
+    try {
+      await syncToAgentRegistry(forkDoc, 'publish');
+    } catch (arError) {
+      console.error('[marketplace] AR write failed on fork:', (arError as any).message);
+      return res.status(500).json({ error: 'Failed to fork manifest' });
+    }
+
+    const created = await Installable.create(forkDoc);
```
lilyshen0722, line 426: Direct counterpart of `Installable.create(installableDoc)` at line 254 (publish path), but here it's not wrapped. If AR succeeds at line 420 and this `create` throws (duplicate key on a race, validation, index hiccup), the outer catch at line 444 returns 500 — but AR now has a fork row that Installable doesn't know about. Orphan.

Mirror the publish-path pattern:

```ts
let created;
try {
  created = await Installable.create(forkDoc);
} catch (e) {
  console.warn('[marketplace] Installable write failed on fork (AR succeeded):', (e as any).message);
  return res.status(500).json({ error: 'Fork Installable write failed; manifest needs reconciliation.' });
}
```

Or revert AR with an explicit delete before returning.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+};
+
+const RUNTIME_MAP: Record<string, string> = {
+  native: 'standalone',
+  internal: 'standalone',
+  'managed-agents': 'commonly-hosted',
+  webhook: 'standalone',
+  'claude-code': 'standalone',
+  moltbot: 'standalone',
+  remote: 'standalone',
```
lilyshen0722, line 37: ADR-005 (Local CLI Wrapper Driver, landing next) introduces `sourceRuntime: 'local-cli'` for CLI-wrapped agent Installables. The current `RUNTIME_MAP` omits `local-cli`, so it falls through to `'standalone'` via the `|| 'standalone'` default at line 55. Behavior is correct, but the mapping isn't explicit. One-line add:

```ts
'local-cli': 'standalone',   // ADR-005
```

Self-documenting and prevents a silent gap the first time someone publishes/forks a CLI-wrapped manifest.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
+const express = require('express');
+const auth = require('../middleware/auth');
+const User = require('../models/User');
+const Installable = require('../models/Installable');
+const { AgentRegistry, AgentInstallation } = require('../models/AgentRegistry');
```
lilyshen0722, line 5: This is the only `.ts` route file in `backend/routes/` using CJS `require()` + `module.exports`. Every other `.ts` route in this directory uses ES module `import` / `export default router` (e.g. `registry/install.ts`, `agentsRuntime.ts`). The file already has `export {};` at the bottom which is a no-op CJS/ESM mixing artifact. Worth harmonizing before merge. Non-blocking.

backend/routes/marketplace-api.ts
```diff
@@ -0,0 +1,603 @@
...
+      return res.status(409).json({
+        error: 'Cannot delete a manifest with active installations. Unpublish instead.',
+      });
+    }
+
+    // Delete AR first, then Installable
+    try { await AgentRegistry.deleteOne({ agentName: installableId.toLowerCase() }); } catch (e) {
+      console.warn('[marketplace] AR delete failed:', (e as any).message);
+    }
+    await Installable.deleteOne({ installableId: installableId.toLowerCase() });
```
lilyshen0722, line 347: ADR-001 invariant #5: an Agent component's `User` row, memory, and pod memberships persist across install/uninstall/delete of the parent Installable. Current behavior is correct — only AR + Installable rows are removed here, User rows are untouched. Add a one-liner so a future reviewer doesn't "fix" this by adding a cascade:

```ts
// ADR-001 invariant #5: Agent User rows, memory, and pod memberships
// are intentionally NOT cascade-deleted — identity outlives the manifest.
// Uninstall-time teardown runs via AgentInstallation lifecycle elsewhere.
```

lilyshen0722 (changes requested):
Great design doc — §6c.1's AR-first ordering with reconciliation semantics is well argued. Mostly correctness comments below.

**Two dual-write hazards that can leave AgentRegistry orphaned (blocking):**
- Update-path `existing.save()` has no try/catch — if it throws after AR succeeds, AR is forward and Installable is back. Detailed at line 204.
- Fork-path `Installable.create()` isn't wrapped the way the publish-path equivalent is (line 254 vs line 426). Same orphan class.

**Important:** zero unit tests for the 603-line new route. Both bugs above would have been caught by a minimal publish-update + fork-AR-failure test pair. At minimum: publish (new + update + AR-failure + version-conflict), fork (happy + AR-failure), hard-delete (with-active-installs guard).

**ADR-005 forward-compat (Local CLI Wrapper Driver landing next):** one-line `RUNTIME_MAP` addition would future-proof CLI-wrapped manifests. Comment at line 30.

**Not blocking, worth fixing in-PR if easy:**
- CJS `require()` in a `.ts` file diverges from every other `backend/routes/*.ts` (all use ESM). Also `export {};` at EOF mixes idioms.
- `publisher.userId` (MongoDB ObjectId) leaks to unauthenticated callers on the detail endpoint — flag for the frontend consumer.
- Hard-delete correctly preserves User rows + memory per ADR-001 invariant #5, but a one-line comment citing the invariant would prevent a future "fix" attempt that adds a cascade.

Happy to pair on the reconciliation patterns if useful.

PR #216, opened 2026-04-18
feat(uploads): ADR-002 Phase 1b-a — signed-URL mint + ACL + audit log

samxu01:
Plumbing for closing the GET authorization gap on /api/uploads/:fileName
without breaking every  during rollout. Ships the mechanism; the
follow-up (Phase 1b-b) removes public-read and requires a valid token or
header auth on GET.

Backend:
- backend/services/attachmentAccess.ts — signAttachmentToken /
  verifyAttachmentToken (JWT with pur:'upload' purpose claim, 5-min TTL,
  bound to fileName+userId); canReadAttachment runs the Phase-1 URL-substring
  scan across owner, profile pictures, Post.image/content, PG messages
  content. File name is shape-guarded before any DB query (ReDoS guard on
  the Mongo $regex and parity with the PG ILIKE).
- backend/routes/uploads.ts — GET /:fileName/url mints a signed URL after
  auth + rate-limit (30/min/user) + ACL. Audit-logs each successful mint
  fire-and-forget so audit DB failures don't 500 the mint.
- backend/models/AuditLog.ts + backend/services/auditService.ts — minimal
  append-only audit record.

Frontend:
- frontend/src/utils/signedAttachmentUrl.ts — async helper that calls the
  mint endpoint, caches per fileName until shortly before TTL expiry, and
  coalesces concurrent calls for the same file.

Also:
- Move REVIEW.md to repo root so Claude self-review picks it up, and
  update the CLAUDE.md pointer.
- Self-review caught a ReDoS path (fileName flowed directly into Mongo
  $regex); fix + test included in this commit.

Tests: 14 cases on attachmentAccess (sign/verify + ACL fan-out + ReDoS
guard), 5 on the mint route, 7 on the frontend helper. All pass.

https://claude.ai/code/session_01JGkwSJkDKerKRsnzUe8Mxe


backend/routes/uploads.ts
```diff
@@ -103,11 +131,47 @@ router.post('/', auth, upload.single('image'), async (req: AuthReq, res: Res) =>
   }
 });
 
+// ADR-002 Phase 1b: signed-URL mint endpoint. Declared before the bare
+// `:fileName` GET so Express doesn't match `/:fileName/url` as a fileName
+// containing a slash.
+router.get('/:fileName/url', mintRateLimit, auth, async (req: AuthReq, res: Res) => {
```
samxu01, line 151: **Middleware ordering — rate-limit key is always the IP today.**

`mintRateLimit` runs *before* `auth`, so when `keyGenerator` evaluates `req.userId || req.ip || 'anon'`, `req.userId` is always `undefined` (auth hasn't run yet). Every caller collapses to an IP-based bucket, which means:

- Users on the same office/college NAT (or behind an L7 proxy where `req.ip` is the shared ingress) share a single 30/min bucket.
- The comment above (`30 mints/min/user`) isn't actually what the code implements.

Fix is just swapping the order: `router.get('/:fileName/url', auth, mintRateLimit, ...)`. The tests don't catch this because the auth stub and the rate limiter both use the same header/IP in isolation.

---

backend/services/attachmentAccess.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  const fileDoc = await File.findByFileName(fileName);
+  if (fileDoc?.uploadedBy && String(fileDoc.uploadedBy) === String(userId)) return true;
+
+  const urlFragment = `/api/uploads/${fileName}`;
+
+  const profileUser = await User.findOne({ profilePicture: urlFragment }).select('_id').lean();
+  if (profileUser) return true;
+
+  const post = await Post.findOne({
```
samxu01: **False-deny bug: `findOne` picks one referencing post; if Mongo returns a pod-scoped post first, a legit viewer referenced in a *different* post gets denied.**

Concrete failure: a file is pasted into (a) a public post with `podId: null`, and (b) a pod-scoped post in pod A. `findOne` is free to return either — no ordering is specified. If it returns the pod-A post and the viewer isn't a member of pod A, this branch falls through to the messages check and eventually returns false, even though the public post grants access to everyone.

Once Phase 1b-b flips GET to deny-by-default, this becomes a user-visible "broken image" intermittently.

Suggested fix: fan out, not pick-one.

```ts
const posts = await Post.find({ $or: [...] }).select('_id podId').limit(10).lean();
for (const p of posts) {
  if (!p.podId) return true;
  const member = await Pod.findOne({ _id: p.podId, members: userId }).select('_id').lean();
  if (member) return true;
}
```

Same shape applies to `findMessagePodReferencingFile` below (see related comment).

---

backend/services/attachmentAccess.ts
```diff
@@ -0,0 +1,173 @@
...
+async function findMessagePodReferencingFile(urlFragment: string): Promise<string | null> {
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const { pool } = require('../config/db-pg') as {
+      pool?: { query: (sql: string, params?: unknown[]) => Promise<{ rows: Array<{ pod_id: string }> }> };
+    };
+    if (!pool?.query) return null;
+    const needle = `%${urlFragment}%`;
+    const { rows } = await pool.query(
+      'SELECT pod_id FROM messages WHERE content ILIKE $1 LIMIT 1',
```
samxu01: **Same `LIMIT 1` false-deny here.** `SELECT pod_id ... LIMIT 1` with no `ORDER BY` returns an arbitrary single pod when the file is referenced in multiple pods' messages. If the viewer is a member of pod B but PG happens to return pod A first, access is denied despite pod B granting it.

Fix: fetch distinct pods up to a bounded N and check membership against any.

```sql
SELECT DISTINCT pod_id FROM messages WHERE content ILIKE $1 LIMIT 20
```

Then:
```ts
const pod = await Pod.findOne({ _id: { $in: rows.map(r => r.pod_id) }, members: userId })
  .select('_id').lean();
if (pod) return true;
```

One query instead of N, and it's correct across multi-pod references.

---

backend/services/attachmentAccess.ts
```diff
@@ -0,0 +1,173 @@
...
+  // shape isn't one of our files and can't be authorized. This also guards
+  // the regex/LIKE substring searches below against ReDoS and wildcard abuse.
+  if (!isPlausibleFileName(fileName)) return false;
+
+  const fileDoc = await File.findByFileName(fileName);
+  if (fileDoc?.uploadedBy && String(fileDoc.uploadedBy) === String(userId)) return true;
+
+  const urlFragment = `/api/uploads/${fileName}`;
+
+  const profileUser = await User.findOne({ profilePicture: urlFragment }).select('_id').lean();
```
samxu01: **Profile-picture ACL check probably misses most users due to URL shape mismatch.**

- `urlFragment` is the **relative** path `/api/uploads/${fileName}`.
- `backend/routes/uploads.ts:96` (unchanged by this PR) returns the **absolute** URL `${protocol}://${host}/api/uploads/${fileName}` to the client after upload.
- `controllers/userController.ts:47` writes whatever the client sends into `user.profilePicture`.

So in practice `user.profilePicture` is often stored as `https://api-dev.commonly.me/api/uploads/xyz.png`, and this equality match against the relative `/api/uploads/xyz.png` fragment fails to find the user → access denied, even though the file really is someone's avatar.

Suggested fix: normalize on read.

```ts
const profileUser = await User.findOne({
  profilePicture: { $in: [urlFragment, `http://${host}${urlFragment}`, `https://${host}${urlFragment}`] }
}).select('_id').lean();
```

Or better: store profile pictures in a normalized shape going forward (Phase 2's structured `attachments` model is the right home for this).

---

frontend/src/utils/signedAttachmentUrl.ts
```diff
@@ -0,0 +1,98 @@
...
+
+const TOKEN_TTL_SECONDS = 300; // Matches backend DEFAULT_TOKEN_TTL_SECONDS.
+const CACHE_SKEW_SECONDS = 30; // Refresh ~30s before expiry to avoid races.
+
+interface CacheEntry {
+  url: string;
+  expiresAtMs: number;
+}
+
+const cache: Map<string, CacheEntry> = new Map();
```
samxu01, line 38: Nit: cache is an unbounded `Map`. A long-lived SPA that scrolls through hundreds of feeds will grow this indefinitely — entries are only pruned when explicitly re-requested (via the freshness check in `getSignedAttachmentUrl`), but files that are requested once and never again stay forever.

For Phase 1b-a this is fine in practice (TTL=5min, feed sizes are small), but a bounded LRU (e.g. cap at 500 entries) would be safer once call sites get migrated in 1b-b. Not blocking.

---

samxu01 (commented):
Solid Phase 1b-a landing overall — JWT sign/verify, fire-and-forget audit, frontend helper with cache+inflight coalescing, ReDoS shape guard, and good test coverage (14 + 5 + 7). The split into 1b-a (plumbing) and 1b-b (flip) in the ADR is the right sequencing.

**Blocking before merge:**

1. **Rate-limit ordering bug** (`routes/uploads.ts:137`) — `mintRateLimit` runs before `auth`, so the keyGenerator always sees `req.userId === undefined` and collapses every caller to an IP bucket. The comment promises "30/min/user"; the code implements "30/min/IP". Users on shared NATs share one bucket. Swap middleware order.

**Blocking before the Phase 1b-b flip** (not necessarily this PR, but before public-read is removed):

2. **`Post.findOne` false-deny** (`services/attachmentAccess.ts:115`) — when a file is referenced by multiple posts (public + pod-scoped), Mongo's non-deterministic `findOne` can return a pod-scoped row and deny a viewer who actually has access via a public post. Fan out to `.find().limit(N)` and check each.

3. **PG `LIMIT 1` false-deny** (`services/attachmentAccess.ts:166`) — same shape; pick up to N distinct pods and check membership against any.

4. **Profile-picture URL shape mismatch** (`services/attachmentAccess.ts:112`) — `urlFragment` is relative, but `uploads.ts:96` returns an absolute URL after upload and `userController.ts:47` stores whatever the client sends. The equality match silently misses most avatars. Normalize on read or match against both shapes.

**Nits:**

5. Unbounded frontend cache (`signedAttachmentUrl.ts:33`) — fine for 1b-a, worth an LRU cap before heavy call-site migration in 1b-b.

Once #1 is fixed, this PR is safe to ship as 1b-a plumbing; #2–#4 must be resolved before 1b-b flips GET to deny-by-default, or legitimate viewers will see broken images.

---

backend/routes/uploads.ts
samxu01: Fixed in 299a0d2 — swapped to `auth, mintRateLimit` so the keyGenerator sees `req.userId`.

---

backend/services/attachmentAccess.ts
samxu01: Fixed in 299a0d2 — fan-out via `Post.find(...).limit(20)`, any public referencing post allows, then a single `Pod.findOne({ _id: { $in: scopedPodIds }, members: userId })` covers all scoped pods. Regression test added: "allows read when the file is referenced in BOTH a hidden pod AND a public post (fan-out)".

---

backend/services/attachmentAccess.ts
samxu01: Fixed in 299a0d2 — `SELECT DISTINCT pod_id ... LIMIT 20`, then one `Pod.findOne({ _id: { $in: podIds }, members: userId })`. Regression test added: "allows read across multiple referencing pods when viewer is a member of any".

---

backend/services/attachmentAccess.ts
samxu01: Fixed in 299a0d2 — matcher is now a substring `$regex` on the URL fragment (regex-escaped), so both the relative `/api/uploads/<f>` and absolute `https://api-dev.commonly.me/api/uploads/<f>` shapes match. Regression test added: "matches profile pictures stored as absolute URLs (substring regex)". Write-time normalization agreed as Phase 2 work.

---

frontend/src/utils/signedAttachmentUrl.ts
samxu01: Fixed in 299a0d2 — bounded at 500 with FIFO eviction via Map insertion order. Test added verifies the 501st entry evicts the 1st while the 250th still hits cache.

---

PR #217, opened 2026-04-19
docs: ADR-009 test tiers + CI/CD to GKE (+ WIF runbook, Path A)

samxu01:
Two docs-only commits. Locks in the direction; execution lands in follow-up PRs.

## What this PR decides

- **ADR-009** (`docs/adr/ADR-009-test-tiers-and-ci-cd-to-gke.md`) — four-tier test taxonomy (unit / service / cluster / dev-env) with explicit PR gating, and workflow-triggered GKE deploys. Motivated by two gaps the ADR-002 Phase 1b-a review surfaced: "integration" tests don't actually use the CI service containers, and deploys require a local laptop.

- **Auth mechanism: Path A — Workload Identity Federation**. No long-lived SA JSON key in GitHub secrets; OIDC tokens minted per workflow run, exchanged for short-lived GCP tokens. See `docs/deployment/GITHUB_DEPLOY_SETUP.md` for exact `gcloud` commands.

## What this PR does NOT do

- Rename `__tests__/integration/` → `__tests__/service/` (Phase 1, follow-up PR).
- Add `.github/workflows/deploy-dev.yml` (Phase 3, blocked on human running the gcloud commands below).
- Touch `values-private.yaml` or migrate config to committed `values-dev.yaml` (Phase 3).

## Human action required before Phase 3 can land

From `docs/deployment/GITHUB_DEPLOY_SETUP.md` Path A, run the gcloud blocks against `commonly-493005` (needs `roles/owner` or equivalent), then configure GitHub:

1. Create the `deploy-github@` SA + scope `roles/artifactregistry.writer` (on `docker` AR repo) and `roles/container.developer` (conditional, `commonly-dev` only for now).
2. Stand up the WIF pool + provider with `assertion.repository=='Team-Commonly/commonly'`.
3. Bind SA to pool: `refs/heads/main` for dev, `environment:prod` for prod.
4. Repo secrets: `WIF_PROVIDER`, `WIF_SERVICE_ACCOUNT`.
5. Repo environments: `dev` (no approvers), `prod` (required reviewer + tag filter `v*.*.*`).

Once those are in place I can scaffold `deploy-dev.yml` in a follow-up PR.

## Test plan

- [x] ADR references `REVIEW.md`, `KUBERNETES.md`, `ADR-002` correctly.
- [x] CLAUDE.md key-docs list updated to include ADR-009.
- [x] Runbook gcloud commands match required roles (AR writer scoped to `docker` repo; container.developer with cluster-name condition).
- [ ] Team sanity-checks the tier renames (naming bikeshed welcome) and the WIF setup plan.
- [ ] Human runs Path A setup; confirms `WIF_PROVIDER` + `WIF_SERVICE_ACCOUNT` are in repo secrets.
- [ ] Follow-up PR: Phase 1 rename + `setupMongoDb` / `setupPgDb` honor `INTEGRATION_TEST=true`.
- [ ] Follow-up PR: Phase 3 `deploy-dev.yml` (unblocks after WIF setup).

https://claude.ai/code/session_01JGkwSJkDKerKRsnzUe8Mxe

### DIFF
--- a/docs/adr/ADR-009-test-tiers-and-ci-cd-to-gke.md
+++ b/docs/adr/ADR-009-test-tiers-and-ci-cd-to-gke.md
@@ -82,7 +82,7 @@ Target state: the *secret* content in `values-private.yaml` moves to GCP Secret 
 
 - **OIDC audience.** WIF provider must scope to `repo:Team-Commonly/commonly:ref:refs/heads/main` (dev) and `repo:Team-Commonly/commonly:environment:prod` (prod). A forked PR cannot mint a token for either.
 - **Image registry push scope.** The WIF-backed SA has `roles/artifactregistry.writer` on the `docker` repo only. No project-wide admin.
-- **Helm upgrade scope.** Deploy SA has `roles/container.developer` on the `commonly-dev` / `commonly-prod` clusters only. No ability to create new clusters or modify node pools.
+- **Helm upgrade scope.** Deploy SA has only `roles/container.clusterViewer` at the IAM layer (auth to any cluster, no mutation). Real deploy permissions come from a Kubernetes `ClusterRoleBinding` of `edit` scoped per cluster — revoking `commonly-dev` access is `kubectl delete clusterrolebinding deploy-github` on that cluster, with no effect on `commonly-prod` or any future cluster.
 - **No secrets echoed in logs.** `set -x` banned in the deploy workflow; `::add-mask::` used for any intermediate token output.
 - **Tier 3 probes are unauthenticated only (Phase 4 scope).** `GET /api/health/live` and `GET /api/health/ready` cover the "did the pod come up and can it reach Mongo + Postgres" question — which is the overwhelming majority of deploy regressions. Authenticated round-trips (login + domain-endpoint call) need a credential story (dedicated smoke account, its User row in Mongo, ESO-rotated secret, workflow retrieval of the credential) that isn't resolved yet; deferred to a later phase so it doesn't block Phase 4.
 
@@ -134,7 +134,7 @@ Deferred. Preview environments per PR are the ideal but cost real GKE capacity p
 
 ### Phase 3 — WIF + dev deploy workflow (one PR + GCP setup)
 
-- [ ] GCP: create WIF pool + provider in `commonly-493005`; grant `artifactregistry.writer` and `container.developer` to the deploy SA scoped to dev.
+- [ ] GCP: create WIF pool + provider in `commonly-493005`; grant `artifactregistry.writer` on the `docker` repo and `container.clusterViewer` at project level; apply a `ClusterRoleBinding` of `edit` against `commonly-dev` (Kubernetes RBAC, not IAM). Details in `docs/deployment/GITHUB_DEPLOY_SETUP.md`.
 - [ ] GitHub: configure `dev` environment with no approvals (auto-deploy).
 - [ ] `.github/workflows/deploy-dev.yml`: build backend + frontend + gateway images, push to AR, `helm upgrade commonly-dev --set image.tag=${sha}`, post status.
 - [ ] Retire `values-private.yaml` for dev: migrate its content to committed `values-dev.yaml` (non-secret config) and ESO (secrets).
--- a/docs/adr/ADR-009-test-tiers-and-ci-cd-to-gke.md
+++ b/docs/adr/ADR-009-test-tiers-and-ci-cd-to-gke.md
@@ -116,35 +116,68 @@ Deferred. Preview environments per PR are the ideal but cost real GKE capacity p
 
 ### Phase 1 — Rename + fix Tier 1 (one PR)
 
-- [ ] Rename `__tests__/integration/` → `__tests__/service/` and update jest projects.
-- [ ] `setupMongoDb()` / `setupPgDb()` read `INTEGRATION_TEST=true` and connect to `MONGO_URI` / `PG_*` env instead of in-memory, when set.
-- [ ] Verify all existing "integration" tests still pass against the CI service containers. Fix the ones that silently relied on `MongoMemoryServer` quirks (expect a few).
-- [ ] Update `backend/TESTING.md` with the new tier names.
+- [ ] Rename `backend/__tests__/integration/` → `backend/__tests__/service/`. Update `backend/jest.config.js` only if the rename breaks discovery — current config (`testPathIgnorePatterns` excludes `utils/`, no `testMatch` override) auto-picks `*.test.js` under any path, so a directory rename alone should work. Run `npx jest --listTests` to confirm.
+- [ ] `backend/__tests__/utils/testUtils.js`: `setupMongoDb()` and `setupPgDb()` branch on `process.env.INTEGRATION_TEST === 'true'` — when set, connect to `process.env.MONGO_URI` / `process.env.PG_*` via `mongoose.connect` and `new pg.Pool` respectively, and skip the in-memory server. The existing `__tests__/setup.js` already sets those envs when `INTEGRATION_TEST=true`; no new toggle needed. Keep `MongoMemoryServer` / `pg-mem` for the unset default so Tier 0 stays in-memory.
+- [ ] Audit for likely breakage in the renamed suites. Specific patterns that silently depend on in-memory quirks:
+  - Direct calls to `mongoServer.getUri()` / `pgPool` that the test bodies import from `testUtils` (real-service branch won't export `mongoServer`). Grep: `mongoServer\.|pgDb\.`.
+  - `pg-mem` functions registered in `setupPgDb()` (e.g. `gen_random_uuid`). Real Postgres 16 has `gen_random_uuid` only with `pgcrypto` — add `CREATE EXTENSION IF NOT EXISTS pgcrypto` at start of real-services setup.
+  - Tests that expect `collection.deleteMany({})` to be instant (real Mongo is slower; adjust timeouts if any fall under `jest.setTimeout(default)`).
+  - `pg-mem` schema drift: the in-memory setup creates `pods`, `pod_members`, `messages` tables by hand — real PG gets its schema from `backend/config/init-pg-db.js`. Verify the two match; migrate the init-pg setup into the real-services branch.
+- [ ] Update `backend/TESTING.md`: add a "Tier 0 / Tier 1" section matching the Decision table; deprecate the "integration" label and point at the new dir.
+- [ ] `.github/workflows/tests.yml` already sets `INTEGRATION_TEST: "true"` and boots `mongo:7` + `postgres:16` — verify the new real-services branch actually fires in CI after the rename (add a one-line log assertion, e.g. `console.log('[tier1] using real services')` gated on the env, and grep the workflow log).
 
 ### Phase 1.5 — Chart-lint on every push (one small PR)
 
-- [ ] New job in `tests.yml` (or separate `chart-lint.yml`): `helm template k8s/helm/commonly -f values.yaml -f values-dev.yaml | kubeconform --strict --schema-location default` on every push.
-- [ ] Also run against `values-prod.yaml` once it's committed (Phase 3).
-- [ ] Required check on all PRs — it's cheap enough that "always on" is fine.
+- [ ] New job `chart-lint` added to `.github/workflows/tests.yml` (single-workflow policy, not a separate file) running on every push. Command: `helm template k8s/helm/commonly -f k8s/helm/commonly/values.yaml -f k8s/helm/commonly/values-dev.yaml | kubeconform --strict --summary -output text`.
+- [ ] CRD schemas: default `kubeconform` schemas cover core Kubernetes only. The chart uses ESO `ExternalSecret`, Ingress (GKE), and possibly cert-manager — supply these via `--schema-location 'https://raw.githubusercontent.com/datreeio/CRDs-catalog/main/{{.Group}}/{{.ResourceKind}}_{{.ResourceAPIVersion}}.json' --schema-location default`. Pin to a specific `datreeio/CRDs-catalog` commit SHA to avoid drift.
+- [ ] Optional second invocation against `-f values-prod.yaml` — gated on that file existing in-tree (it lands in Phase 5, not Phase 3). Use `if [ -f k8s/helm/commonly/values-prod.yaml ]; then ... fi` so Phase 1.5 merges independently of Phase 5.
+- [ ] Add the `chart-lint` check to branch protection on `main`.
 
 ### Phase 2 — Gate Tier 2 on PRs (one small PR)
 
-- [ ] `smoke-test.yml`: add `on.pull_request.paths` filter for `k8s/**`, `backend/Dockerfile`, `frontend/Dockerfile`, `dev.sh`, `.github/workflows/**`. Keep post-merge trigger.
-- [ ] Surface the kind smoke as a required check on PRs with a matching path.
+- [ ] `.github/workflows/smoke-test.yml`: add `on.pull_request.paths` filter. Paths to include: `k8s/**`, `backend/Dockerfile`, `frontend/Dockerfile`, `_external/clawdbot/**` (gateway source per CLAUDE.md §Build & Deploy), `dev.sh`, `.github/workflows/**`. Keep the existing `workflow_run` post-merge trigger.
+- [ ] Add the smoke-test check to branch protection on `main` as **required only when status exists** (so PRs that don't touch the filtered paths — and thus don't run the check — aren't blocked by a missing status).
+- [ ] **Human action:** repo admin adds `kind cluster smoke test` as an "optional" required check in the branch-protection rule (GitHub's "Require status checks to pass → expect checks from each PR" with no entries forces only-when-present semantics; alternately, use `required_status_checks.checks` with `app_id: null` in a custom ruleset — document whichever the admin picks).
 
 ### Phase 3 — WIF + dev deploy workflow (one PR + GCP setup)
 
-- [ ] GCP: create WIF pool + provider in `commonly-493005`; grant `artifactregistry.writer` on the `docker` repo and `container.clusterViewer` at project level; apply a `ClusterRoleBinding` of `edit` against `commonly-dev` (Kubernetes RBAC, not IAM). Details in `docs/deployment/GITHUB_DEPLOY_SETUP.md`.
-- [ ] GitHub: configure `dev` environment with no approvals (auto-deploy).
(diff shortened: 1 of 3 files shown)


docs/deployment/GITHUB_DEPLOY_SETUP.md
```diff
@@ -0,0 +1,302 @@
...
+  --location=us-central1 \
+  --member="serviceAccount:$DEPLOY_SA_EMAIL" \
+  --role=roles/artifactregistry.writer
+
+# Talk to the dev cluster. Repeat for prod when you add it.
+gcloud container clusters get-credentials commonly-dev --region=us-central1
+gcloud projects add-iam-policy-binding $PROJECT_ID \
+  --member="serviceAccount:$DEPLOY_SA_EMAIL" \
+  --role=roles/container.developer \
+  --condition="expression=resource.name.endsWith('/clusters/commonly-dev'),title=commonly-dev-only"
```
samxu01: **IAM condition likely underscopes `roles/container.developer` rather than "simpler than cluster-level RBAC."**

`resource.name.endsWith('/clusters/commonly-dev')` only evaluates true for permissions where the evaluated resource is the cluster object itself (e.g. `container.clusters.get/update`). `roles/container.developer` also grants permissions that evaluate against operation resources (`projects/.../operations/<id>`) and — depending on how `gcloud container clusters get-credentials` is later used — node pool and workload resources with different resource paths. A condition tied to `endsWith('/clusters/commonly-dev')` will silently deny those, which tends to manifest as opaque "Unauthorized" / "operation not found" mid-deploy.

Two safer options, pick one:

1. Drop the condition here and instead grant `roles/container.developer` at the cluster level using a ClusterRoleBinding against the SA's Kubernetes identity (preferred; this is what "cluster-level RBAC" actually looks like in GKE and the note at L74–76 frames it as harder than it is).
2. Keep the IAM condition but express it with `resource.type`:
   ```
   expression=resource.type == 'container.googleapis.com/Cluster' && resource.name.endsWith('/clusters/commonly-dev')
   ```
   This at least makes the scoping rule explicit: "applies only to cluster resources; unconstrained on the rest." Paired with the role's least-privilege nature this is tolerable, but verify against a dry-run deploy before committing.

Worth writing up whichever you pick so the next person who adds a cluster doesn't guess at the semantics.

---

docs/adr/ADR-009-test-tiers-and-ci-cd-to-gke.md
```diff
@@ -0,0 +1,165 @@
...
+
+Target state: the *secret* content in `values-private.yaml` moves to GCP Secret Manager and is pulled into the cluster by ESO (which already owns `api-keys` per CLAUDE.md §Agent Runtime). The *config* content (GCP project ID, PG host, image repo) moves into a committed `values-dev.yaml` / `values-prod.yaml` — these are not secrets and hiding them just makes the repo unusable to new contributors. One uncommitted file disappears; three committed files + ESO handle what it used to.
+
+**Image tags.** Workflow sets image tag to `${sha}` (short SHA) and passes it via `--set image.tag=...` on helm upgrade, not by editing `values-dev.yaml`. That keeps the workflow git-clean (no auto-commit back to main), keeps `values-dev.yaml` reviewable, and makes rollback a matter of `helm upgrade --set image.tag=<previous-sha>` rather than a values-file revert.
+
+**Rollback.** Tier 3 smoke is the gate. On failure, workflow runs `helm rollback commonly-dev 0` (last revision) and posts the failed probe output to the PR or commit. Human decides whether to fix forward or investigate.
```
samxu01: **`helm rollback commonly-dev 0` has two edge cases the ADR should decide on before Phase 4:**

1. **Bootstrap / first-ever deploy.** `rollback 0` means "rollback to the previous revision," which doesn't exist on revision 1. Helm will exit non-zero and the "post rollback output" step also won't have anything to post. A deploy that fails the very first Tier 3 probe leaves `commonly-dev` broken *and* the workflow reporting a confusing error. Call out that Phase 3 has to seed a known-good revision before Phase 4 turns on auto-rollback, or make the rollback step conditional on `helm history | wc -l > 1`.

2. **Second-strike rollback targets the bad image.** Scenario: commit A deploys OK. Commit B fails Tier 3 → rollback to A. Commit C merges, deploys, fails Tier 3 → `rollback 0` rolls back from C to B (the broken one), not to A. If the 3-strikes escalation is supposed to stop the ping-pong, it also needs to pin the rollback target, not just count failures. Consider `helm rollback commonly-dev <last-known-good-revision>` computed from a status annotation, or explicitly document that the 3-strike gate *is* the correctness mechanism and operators accept one bad intermediate state.

Neither blocks the ADR — but Phase 4's "auto-rollback" bullet reads as a one-liner and these are the two ways it misbehaves.

---

docs/adr/ADR-009-test-tiers-and-ci-cd-to-gke.md
```diff
@@ -0,0 +1,165 @@
...
+- **Prod deploy is now one approval click away.** The one-reviewer gate is minimal; we may want two reviewers or an SRE-team requirement before `v1.0.0` ships. Set per-environment in GitHub; costs nothing to tighten later.
+
+### Security notes
+
+- **OIDC audience.** WIF provider must scope to `repo:Team-Commonly/commonly:ref:refs/heads/main` (dev) and `repo:Team-Commonly/commonly:environment:prod` (prod). A forked PR cannot mint a token for either.
+- **Image registry push scope.** The WIF-backed SA has `roles/artifactregistry.writer` on the `docker` repo only. No project-wide admin.
+- **Helm upgrade scope.** Deploy SA has `roles/container.developer` on the `commonly-dev` / `commonly-prod` clusters only. No ability to create new clusters or modify node pools.
+- **No secrets echoed in logs.** `set -x` banned in the deploy workflow; `::add-mask::` used for any intermediate token output.
+- **Tier 3 smoke uses a dedicated test account.** Smoke probes don't use real user credentials — a `smoke@commonly.me` service account gets minimal read access and is rotated via ESO.
```
samxu01: **`smoke@commonly.me` is introduced here but nothing in the implementation plan actually creates it.**

Phase 4 assumes the account exists and "is rotated via ESO," but there's no task for: (a) provisioning the User row in Mongo, (b) the auth mechanism the probe will use (is it logging in with a password → JWT? A pre-issued long-lived token? A separate smoke-only signing key?), or (c) the ESO `ExternalSecret` the probe reads its credential from.

This is the same class of gap that made the `values-private.yaml` ritual survive this long — a dependency that "someone owns" but no one actually sets up. Two options:

1. Add a Phase 3.5 bullet that lands the account + ESO wiring before Phase 4 turns on probes.
2. Scope Tier 3 probes to unauthenticated endpoints only for now (`/api/health/live`, `/api/health/ready`) and defer the "auth round-trip" probe to a later phase where the credential story is resolved.

Option 2 is cheaper and still catches the overwhelming majority of deploy regressions. The "one representative domain endpoint" language on Phase 4 L176 is the thing that actually needs the account — drop or defer it.

---

docs/deployment/GITHUB_DEPLOY_SETUP.md
```diff
@@ -0,0 +1,302 @@
+# GitHub Actions → GKE deploy setup
+
+Companion to [`ADR-009`](../adr/ADR-009-test-tiers-and-ci-cd-to-gke.md). Two
+paths: **Workload Identity Federation** (recommended, no long-lived credential
+in GitHub) and **Service Account JSON key** (fast path, one secret in GitHub,
+rotatable-forever).
```
samxu01: **This runbook contradicts the ADR's decision.**

ADR-009 L42: *"Auth: **Workload Identity Federation only.** No service account JSON keys in GitHub secrets."* The PR body restates it as "Auth mechanism: Path A — Workload Identity Federation."

This doc then presents Path B (SA JSON key) as a coequal "fast path" with a full 7-step setup and a rotation section. If Path A is the decision, Path B shouldn't read like a supported alternative — it reads like an escape hatch future readers will take when they're in a hurry. Two options:

1. **Match the ADR strictly** — delete the Path B section entirely. The "Migration: Path B → Path A" section goes with it (there's no Path B to migrate from).
2. **Document Path B as last-resort only** — move it under a `## Emergency fallback` heading that names a sunset date and requires a GitHub issue to justify use. Keeps the breadcrumb for a genuinely stuck operator without normalizing it.

As written, an operator skimming this picks whichever is easier. The whole point of the ADR's decision was that the choice isn't theirs to make.

---

docs/adr/ADR-009-test-tiers-and-ci-cd-to-gke.md
```diff
@@ -0,0 +1,165 @@
...
+### Phase 1 — Rename + fix Tier 1 (one PR)
+
+- [ ] Rename `__tests__/integration/` → `__tests__/service/` and update jest projects.
+- [ ] `setupMongoDb()` / `setupPgDb()` read `INTEGRATION_TEST=true` and connect to `MONGO_URI` / `PG_*` env instead of in-memory, when set.
+- [ ] Verify all existing "integration" tests still pass against the CI service containers. Fix the ones that silently relied on `MongoMemoryServer` quirks (expect a few).
+- [ ] Update `backend/TESTING.md` with the new tier names.
+
+### Phase 2 — Gate Tier 2 on PRs (one small PR)
+
+- [ ] `smoke-test.yml`: add `on.pull_request.paths` filter for `k8s/**`, `backend/Dockerfile`, `frontend/Dockerfile`, `dev.sh`, `.github/workflows/**`. Keep post-merge trigger.
```
samxu01: **Path filter catches chart-surface changes but not runtime-env changes that also break deploys.**

Concrete example: a PR that adds `process.env.NEW_REQUIRED_FLAG` to `backend/server.js` and a matching `configMap.yaml` key in the same PR — the `k8s/**` match will trigger Tier 2. Good. But a PR that *only* touches `backend/` (new required env var, forgotten chart update) doesn't match the filter and merges green, then breaks dev on Phase 3 deploy.

Tier 1 doesn't catch this either — real Mongo/PG service containers don't exercise the helm chart.

Options:
1. Accept the gap — Tier 3 auto-rollback contains the blast radius. Document the class of bug this filter *doesn't* catch.
2. Add a cheap "chart renders" job on every push: `helm template k8s/helm/commonly -f values-dev.yaml` + `kubeval` or `kubeconform`. Sub-30s, catches missing values and schema errors without kind.
3. Widen the filter to include `backend/src/config/**` and any module that reads `process.env.*`. Brittle — easy to drift out of sync.

(2) is the cheapest real coverage and fits between Tiers 1 and 2 as a "chart-lint" half-tier. Worth calling out in the ADR even if the decision is to skip it.

---

samxu01 (commented):
Docs-only PR — verified the two load-bearing "current state" claims against the repo:

- ✅ `backend/__tests__/utils/testUtils.js:9` — `setupMongoDb()` does unconditionally spin up `MongoMemoryServer` and ignores `MONGO_URI`. The "tests named integration don't hit real databases" framing is accurate.
- ✅ `.github/workflows/smoke-test.yml:10` — kind cluster runs only on `workflow_run` post-merge to `main` (plus `workflow_dispatch`). PRs get no cluster signal today. Accurate.
- ✅ `.github/workflows/tests.yml:102` — `INTEGRATION_TEST=true` + mongo:7 + postgres:16 service containers do exist and sit idle. Accurate.

The ADR's diagnosis is well-grounded and the four-tier taxonomy is a clean rename that matches the actual failure modes. WIF-over-JSON-key is the right call. Two strong points worth preserving:

- Moving non-secret `values-private.yaml` content into committed `values-dev.yaml` + ESO for secrets is the right split. The current "Lily's laptop" state is a documented foot-gun and this kills it.
- Using `--set image.tag=${sha}` instead of committing tag bumps back to `values-dev.yaml` avoids the auto-commit-to-main anti-pattern that trips a lot of teams.

## Issues worth addressing before the follow-up PRs land

**Blocking-ish (nothing to revert in this PR, but Phase 3/4 inherits these):**

1. **IAM condition on `container.developer` likely under-scopes permissions.** `resource.name.endsWith('/clusters/commonly-dev')` denies the non-cluster resources the role also covers (operations, node pools). Prefer cluster-level RBAC via ClusterRoleBinding against the SA's K8s identity, or at least pair the condition with `resource.type`. Inline comment on the gcloud block.
2. **`helm rollback 0` has two silent failure modes** — bootstrap (no prior revision) and second-strike (rolls back onto the last bad revision). The 3-strike escalation doesn't fix #2 on its own. Inline comment on the Rollback paragraph.
3. **Runbook contradicts ADR.** Path A is declared "only" in the ADR but the companion doc presents Path B as a coequal fast path. Tighten to match the decision — either delete Path B or frame it as emergency-only with a sunset date. Inline comment on the runbook intro.
4. **`smoke@commonly.me` provisioning is orphaned.** The account is named as a security control but nothing in Phases 1–6 creates it or wires its credential. Either add a Phase 3.5 bullet or scope Tier 3 probes to unauthenticated health endpoints. Inline comment on security notes.

**Coverage gap:**

5. **Tier 2 path filter misses backend/frontend-only changes that break chart rendering at runtime** (new required env var, forgotten configMap). Tier 1 also doesn't catch this — real DBs don't exercise helm. A sub-30s `helm template | kubeconform` job as a "chart lint" half-tier would cover the gap cheaply. Inline comment on Phase 2.

## Smaller notes I'm not posting inline

- "kind brings up ~20 min" (L73): the existing `smoke-test.yml` has `timeout-minutes: 30` with `--wait --timeout 300s` on helm; the actual kind run is closer to 10 min in CI. Not actionable, just calibration.
- Phase 3 says `helm upgrade`; recommend `helm upgrade --install` for idempotency if the release is ever deleted.
- `google.subject=assertion.sub` is fine with `principalSet://` bindings, but the GCP-recommended pattern for forward compat is `google.subject=assertion.repository + ":" + assertion.ref` so any future principal:// binding you add doesn't depend on the opaque sub format. Low priority.
- Open question on "where does cloud-agent status appear" — start with GitHub Checks and drop PR comments; the Checks API is what `mcp__github__pull_request_read method: get_check_runs` is built for, and it's strictly more structured than scraping comments.

Approving-in-spirit as a docs PR. The five inline notes are the things worth resolving before Phase 3 lands.

---

docs/deployment/GITHUB_DEPLOY_SETUP.md
samxu01: Fixed in 6ab6f93 — switched to the reviewer-preferred option: drop the IAM condition, grant `roles/container.clusterViewer` at project level (auth-only), and scope deploy permissions at the Kubernetes layer via a `ClusterRoleBinding` of `edit` against the SA's identity, applied per cluster. Revocation on one cluster is now a single `kubectl delete clusterrolebinding deploy-github`. Troubleshooting section rewritten accordingly.

---

docs/adr/ADR-009-test-tiers-and-ci-cd-to-gke.md
samxu01: Fixed in 6ab6f93 — rollback now targets a last-known-good revision tracked via a Helm chart annotation that the workflow stamps on every passing deploy, not `rollback 0`. First-deploy case handled by short-circuiting to "fail loudly, don't auto-rollback" when `helm history` has one entry. Second-strike ping-pong avoided because the pointer is advanced only on successful Tier 3, so two consecutive failures both roll back to the same last-good revision. Text in the Rollback section revised accordingly.

---

docs/adr/ADR-009-test-tiers-and-ci-cd-to-gke.md
samxu01: Fixed in 6ab6f93 — took option 2: Phase 4 probes are scoped to unauthenticated `/api/health/live` and `/api/health/ready` only. The `smoke@commonly.me` bullet is removed from the security notes; the "auth round-trip" and "representative domain endpoint" items in Phase 4 are removed too. Comment in ADR notes that auth probes are deferred to a later phase where the credential story (account provisioning + ESO-rotated secret + workflow retrieval) is resolved.

---

docs/deployment/GITHUB_DEPLOY_SETUP.md
samxu01: Fixed in 6ab6f93 — took option 1: Path B removed entirely, including the rotation section and the "Migration Path B → Path A" block. Runbook header rewritten to state "ADR-009 commits to Workload Identity Federation only — no long-lived service account keys in GitHub." The `## Prep (both paths)` heading became `## Prep`, `## Path A ...` became `## Setup`. Troubleshooting no longer references the JSON key path.

---

docs/adr/ADR-009-test-tiers-and-ci-cd-to-gke.md
samxu01: Fixed in 6ab6f93 — took option 2: added **Tier 1.5 "chart lint"** (`helm template | kubeconform`) as a cheap always-on check before Tier 2. Decision section gets a paragraph naming the exact gap the reviewer described (backend-only PR adds a new required env var without a ConfigMap update) and why Tier 0/1 can't catch it. Phase 1.5 added to the implementation plan, slotted between Tier 1 rename and Tier 2 gate.

---

samxu01:
Tightening pass pushed in `39c1048`. Self-audit caught one contradiction (Phase 4 said `rollback 0` while Decision said last-known-good) and vague steps in every phase. All fixed so a fresh Claude session can execute each phase without asking back. Changelog:

**Contradiction resolved**
- Phase 4 rewrites the rollback step to match the Decision section — reads `commonly.me/last-good-revision` annotation, skips auto-rollback on first-ever deploy, escalates via `gh issue create` with `deploy-stuck` label when the pointer stays stuck.

**Phase 1** — names `backend/jest.config.js` explicitly, specifies `INTEGRATION_TEST=true` as the toggle source (reuses the env already wired by `__tests__/setup.js`), and lists the four concrete `MongoMemoryServer` / `pg-mem` quirks that will bite during migration.

**Phase 1.5** — picks `tests.yml` (not a separate workflow), names the CRD schema source (`datreeio/CRDs-catalog`, pinned SHA) since the chart uses ESO `ExternalSecret` + Ingress, and gates the `values-prod.yaml` invocation on file existence.

**Phase 2** — adds `_external/clawdbot/**` to the path filter and calls out the branch-protection human action.

**Phase 3** — `submodules: recursive`, exact three-image `--set` syntax verified against `values.yaml` (`backend.image.tag`, `frontend.image.tag`, `agents.clawdbot.image.tag`), expanded `values-private.yaml` retirement with key-categorization rules, status-reporting decided (both check-run AND PR comment).

**Open questions** — the two that blocked phases are resolved in-place and moved to a "previously resolved" note. Only genuine open items remain.

https://claude.ai/code/session_01JGkwSJkDKerKRsnzUe8Mxe

---

PR #218, opened 2026-04-20
Improve local Docker dev bootstrap

MubaiHua:
## Summary
- add a more forgiving local-dev bootstrap for `./dev.sh` and `commonly-bot`
- keep the Postgres volume-path workaround scoped to `docker-compose.dev.yml` so the base compose remains production-style
- update the docs to explain local startup defaults, optional AI credentials, and delayed `commonly-bot` provisioning

## Verification
- `bash -n dev.sh`
- `node -c external/commonly-agent-services/commonly-bot/index.js`
- `./dev.sh up`
- verified `commonly-bot` can stay up in local optional mode instead of crash-looping before provisioning

## Notes
- repo-wide lint is currently failing on many unrelated existing files, so I did not use it as a merge gate for this scoped local-dev change
- the original local checkout is on a different history from `origin/main`, so this PR branch was created from a clean clone of `origin/main`

### DIFF
--- a/dev.sh
+++ b/dev.sh
@@ -17,24 +17,17 @@ HELM_CHART="k8s/helm/commonly"
 ENV_FILE=".env"
 ENV_EXAMPLE_FILE=".env.example"
 
-get_env_value() {
+has_nonempty_env_value() {
     local key="$1"
     if [ -n "${!key:-}" ]; then
-        printf '%s' "${!key}"
-        return
+        return 0
     fi
 
-    if [ -f "$ENV_FILE" ]; then
-        local line
-        line=$(grep -E "^${key}=" "$ENV_FILE" | tail -n 1 || true)
-        if [ -n "$line" ]; then
-            printf '%s' "${line#*=}"
-        fi
+    if [ ! -f "$ENV_FILE" ]; then
+        return 1
     fi
-}
 
-has_nonempty_env_value() {
-    [ -n "$(get_env_value "$1")" ]
+    grep -Eq "^${key}=.+" "$ENV_FILE"
 }
 
 ensure_local_env_file() {
--- a/dev.sh
+++ b/dev.sh
@@ -14,9 +14,78 @@ CLUSTER_NAME="${COMMONLY_CLUSTER_NAME:-commonly-local}"
 K8S_NAMESPACE="${COMMONLY_K8S_NAMESPACE:-commonly-local}"
 HELM_RELEASE="${COMMONLY_HELM_RELEASE:-commonly}"
 HELM_CHART="k8s/helm/commonly"
+ENV_FILE=".env"
+ENV_EXAMPLE_FILE=".env.example"
+
+get_env_value() {
+    local key="$1"
+    if [ -n "${!key:-}" ]; then
+        printf '%s' "${!key}"
+        return
+    fi
+
+    if [ -f "$ENV_FILE" ]; then
+        local line
+        line=$(grep -E "^${key}=" "$ENV_FILE" | tail -n 1 || true)
+        if [ -n "$line" ]; then
+            printf '%s' "${line#*=}"
+        fi
+    fi
+}
+
+has_nonempty_env_value() {
+    [ -n "$(get_env_value "$1")" ]
+}
+
+ensure_local_env_file() {
+    if [ -f "$ENV_FILE" ]; then
+        return
+    fi
+
+    if [ -f "$ENV_EXAMPLE_FILE" ]; then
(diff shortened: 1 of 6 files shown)


docker-compose.dev.yml
```diff
@@ -23,7 +21,9 @@ services:
       POSTGRES_PASSWORD: ${PG_PASSWORD:-postgres}
       POSTGRES_DB: ${PG_DATABASE:-commonly}
     volumes:
-      - postgres-data-dev:/var/lib/postgresql/data
+      # Local dev uses postgres:latest; mounting the parent dir keeps pace with
+      # versioned PGDATA paths like /var/lib/postgresql/<major>/docker.
+      - postgres-data-dev:/var/lib/postgresql
```
samxu01: The comment's premise is incorrect: the official `postgres` Docker image hard-codes `PGDATA=/var/lib/postgresql/data` regardless of major version (the `/<major>/docker` layout is Debian's system package, not this image). Widening to `/var/lib/postgresql` still works but also captures the postgres user's home dir in the volume, and masks whatever actual bug motivated the change.

If there's a real issue (permissions on a specific host? a transient initdb failure?), document that; otherwise revert to `/var/lib/postgresql/data` — it's narrower and correct for `postgres:latest`.

---

dev.sh
```diff
@@ -14,9 +14,78 @@ CLUSTER_NAME="${COMMONLY_CLUSTER_NAME:-commonly-local}"
...
+    local key="$1"
+    if [ -n "${!key:-}" ]; then
+        printf '%s' "${!key}"
+        return
+    fi
+
+    if [ -f "$ENV_FILE" ]; then
+        local line
+        line=$(grep -E "^${key}=" "$ENV_FILE" | tail -n 1 || true)
+        if [ -n "$line" ]; then
```
samxu01: `${line#*=}` returns the raw RHS including surrounding quotes and trailing comments. So `FOO="bar"` yields `"bar"` (literal quotes) and `FOO=bar # note` yields `bar # note`.

For the three keys checked here, `-n` still evaluates true so `has_nonempty_env_value` is correct in practice — but the moment this helper is used to *print* a value (or to compare against a literal), it bites. Either narrow it to a presence-check (`grep -qE "^${key}=.\+"`) or strip quotes/comments.

---

external/commonly-agent-services/commonly-bot/index.js
```diff
@@ -810,27 +825,44 @@ if (userToken) {
...
   const accounts = buildAccounts();
+  if (accounts.length === 0) {
+    if (optionalRuntime) {
+      logWaitingForAccounts();
+      isPolling = false;
+      return;
+    }
+    console.error('No agent tokens configured. Set COMMONLY_AGENT_TOKEN or COMMONLY_AGENT_CONFIG_PATH.');
+    isPolling = false;
+    return;
```
samxu01: Subtle behavior change: previously, `!optionalRuntime` + zero accounts was impossible inside the polling loop because startup already called `process.exit(1)`. Now, if the runtime config file disappears mid-run (e.g. user deletes `runtime.json`), this branch fires every tick and logs "No agent tokens configured" indefinitely instead of failing loudly.

Low-probability in practice, but either exit here too (match startup semantics) or rate-limit the log like `logWaitingForAccounts` does.

---

samxu01 (commented):
Nice quality-of-life improvement for local bootstrap — `.env` auto-creation, optional-bot mode, and the docs split between "local defaults work" vs. "external PG needs CA" are all good moves.

Three concise inline notes:
1. Postgres volume comment is factually wrong for the official `postgres` image; the mount works but the justification doesn't.
2. `get_env_value` does naive RHS extraction — fine for the three presence-checks today, fragile if reused for printing/comparing.
3. Bot's polling-loop "no accounts" branch no longer exits in non-optional mode, so a mid-run config loss turns into a 5s-cadence error log.

None are blockers. LGTM once the volume comment is corrected or the mount is narrowed back.

---

samxu01:
Feel free to ignore nit comments and merge

MubaiHua:
Addressed the three review notes in def3be36:\n\n- corrected the Postgres volume comment to document the observed major-versioned PGDATA path\n- narrowed dev.sh env probing to a presence check instead of raw RHS extraction\n- restored non-optional commonly-bot fail-fast behavior when accounts disappear mid-run\n\nAlso tightened the local bot docs to note that runtime config is reread on the next poll.\n\nVerification:\n- bash -n dev.sh\n- node -c external/commonly-agent-services/commonly-bot/index.js\n- docker-compose -f docker-compose.dev.yml config

PR #221, opened 2026-04-21
test(tiers): ADR-009 Phase 1 — rename integration → service + real-services branch

lilyshen0722:
Implements **Phase 1** of [ADR-009](../blob/main/docs/adr/ADR-009-test-tiers-and-ci-cd-to-gke.md): explicit Tier 0 (unit) / Tier 1 (service) taxonomy, and make \`INTEGRATION_TEST=true\` actually hit real Mongo + Postgres instead of silently falling back to in-memory.

## What changed

| File(s) | Change |
|---|---|
| \`backend/__tests__/integration/\` → \`backend/__tests__/service/\` | 16 files renamed via \`git mv\` |
| \`backend/__tests__/utils/testUtils.js\` | \`setupMongoDb\` / \`setupPgDb\` branch on \`process.env.INTEGRATION_TEST === 'true'\`. Real-service branch connects via \`mongoose.connect(MONGO_URI)\` / \`new pg.Pool(...)\`, creates \`pgcrypto\`, applies \`backend/config/schema.sql\` verbatim. \`clearPgDb\` uses \`TRUNCATE … CASCADE\` under real PG. Tier-1 setup logs \`[tier1] Connected to real Mongo/Postgres …\` so CI mode is visible. |
| \`.github/workflows/tests.yml\` | Rename \`integration-test\` job → \`service-test\` (\"Service Tests (Tier 1 — real DBs)\"); \`testPathPattern\` follows the dir rename. |
| \`backend/TESTING.md\` | Rewritten around the two-tier model, including helper semantics per tier and authoring rules (FK ordering under real PG, pgcrypto requirement for \`gen_random_uuid()\`, etc.). |
| \`docs/agents/*, docs/adr/ADR-003, ADR-006, docs/integrations/README.md, docs/SUMMARIZER_AND_AGENTS.md\` | Path references updated. |

## Why rename-in-place beats \"keep integration + add real-services on top\"

Per ADR-009 §Alternative A — two tiers that both sound like \"integration\" is exactly the confusion we're trying to fix. Tier 0 is a clean name; Tier 1 is a clean name; no tests live under ambiguously named directories.

## Verification

- \`npx jest --listTests\` discovers all 16 renamed files under the new path.
- Full Tier 0 suite locally: **570 tests across 110 suites pass** (in-memory path unchanged).
- Spot-check: one renamed service-tier test (\`users.test.js\`) runs green in default (in-memory) mode — the old MongoMemoryServer path still works when \`INTEGRATION_TEST\` is unset.
- Tier 1 ground truth comes from CI — this PR's \`service-test\` job is what tells us whether the real-PG schema application and \`TRUNCATE … CASCADE\` work end-to-end, and whether any of the 16 suites silently relied on \`MongoMemoryServer\` / \`pg-mem\` quirks.

## Risk / known watch-outs

The ADR calls out categories of latent breakage that only surface under real services:
- \`pg-mem\`-specific function fakes (we only had \`gen_random_uuid\`; \`pgcrypto\` extension covers it on real PG)
- FK ordering differences (\`clearPgDb\` now uses \`TRUNCATE … CASCADE\` to sidestep)
- Real Mongo slower than in-memory (global \`jest.setTimeout(30000)\` already set)
- Schema drift between hand-written pg-mem tables and \`schema.sql\` (Tier 1 now uses \`schema.sql\` directly — no more drift)

If Tier 1 exposes anything else, I'll fix it in this PR rather than deferring.

## Next up (ADR-009)

- **Phase 1.5** — chart-lint on every push (\`helm template | kubeconform\`), < 30s job
- **Phase 2** — path-filter Tier 2 (kind cluster smoke) onto PRs touching deployment surfaces
- **Phase 3** — WIF + GitHub-Actions dev deploy + retire \`values-private.yaml\`
- **Phase 4** — Tier 3 (dev-env smoke) + auto-rollback via \`commonly.me/last-good-revision\`

## Test plan

- [ ] \`Test & Coverage\` (Tier 0) green
- [ ] \`Service Tests (Tier 1 — real DBs)\` green — this is the load-bearing signal
- [ ] CodeQL / Secret Scan / Integration Tests green
- [ ] Tier 1 log includes \`[tier1] Connected to real Mongo …\` / \`[tier1] Connected to real Postgres and applied schema.sql\`


lilyshen0722:
Squash-merged to main as 7a989063a6.

PR #225, opened 2026-04-21
deploy(values-dev): ADR-009 Phase 3a — retire .dev/values-private.yaml

lilyshen0722:
Closed — PR body scrubbed on 2026-04-21 to remove infrastructure identifiers that should not live in the public repo. The original description contained real GCP project ID, PG hostname, service account email, and bucket name, which are considered sensitive per repo policy. Phase 3 will re-land via GitHub Actions secrets instead of committing the values.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -216,17 +216,17 @@ npm run lint                                  # 0 errors
 
 ### Kubernetes (GKE — commonly-dev)
 
-**ALWAYS use three `-f` flags — NEVER `--reuse-values`:**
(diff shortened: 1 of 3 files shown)


lilyshen0722:
Closing — committing GCP project ID, Aiven PG hostname, and SA email to the public OSS repo is a sensitive-data leak by user's standard, even though ADR-009 framed them as non-secret. Will re-approach with runtime substitution (GitHub Actions env / secrets or a non-committed overlay read at deploy time) instead of committing the values.

PR #227, opened 2026-04-21
docs(deploy): deploy-github RBAC should be 'admin', not 'edit'

lilyshen0722:
## Summary
- First live run of \`deploy-dev.yml\` (ADR-009 Phase 3) failed with \`roles.rbac.authorization.k8s.io … is forbidden\`
- Root cause: the chart has \`rbac.create: true\` and renders \`Role\` / \`RoleBinding\` resources. The \`edit\` ClusterRole used by the ADR's original \`deploy-github\` binding **omits** \`rbac.authorization.k8s.io\` verbs
- \`admin\` covers RBAC verbs while staying namespace-scoped (narrower than project-wide \`roles/container.developer\` and narrower than \`cluster-admin\`)
- Also: \`roleRef\` is immutable — \`kubectl patch\` fails with \"cannot change roleRef\", so the fix requires \`kubectl delete clusterrolebinding deploy-github\` + re-apply. Troubleshooting section updated.

## Applied on cluster
- Deleted + recreated \`clusterrolebinding/deploy-github\` with \`roleRef.name: admin\` on \`commonly-dev\` before re-triggering the deploy workflow
- Second run then progressed past RBAC (different failure — \`--wait\` timeout on pre-existing unrelated zombie pods; tracked separately)

## Test plan
- [ ] Chart Lint + Tier 0/1 green
- [ ] Next \`Deploy Dev\` workflow run clears the RBAC error (already confirmed on cluster)

### DIFF
--- a/docs/deployment/GITHUB_DEPLOY_SETUP.md
+++ b/docs/deployment/GITHUB_DEPLOY_SETUP.md
@@ -95,10 +95,15 @@ subjects:
     apiGroup: rbac.authorization.k8s.io
 roleRef:
   kind: ClusterRole
-  # `edit` is sufficient for helm upgrade on application namespaces.
-  # Use `admin` if deploys also need to create namespaces or RBAC objects,
-  # or define a custom ClusterRole and tighten to exactly helm's verbs.
-  name: edit
+  # `admin` is required when the chart manages RBAC resources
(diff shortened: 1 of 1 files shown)


lilyshen0722:
Squash-merged.

PR #231, opened 2026-04-24
feat(cli): ADR-005 Phase 2 — codex adapter

lilyshen0722:
## Summary

Adds the local-CLI-wrapper adapter for [`codex`](https://github.com/openai/codex), enabling `commonly agent attach codex` + `commonly agent run` to put a locally-installed codex CLI in a Commonly pod as a first-class agent. First Phase 2 adapter (cursor + gemini follow as separate small PRs).

This is the same shape as Phase 1b's `claude` adapter — adapters are pure subprocess wrappers per ADR-005 §Adapter pattern, registered in `cli/src/lib/adapters/index.js`, talking to the kernel only via the run loop.

## Argv shape — diverges from the ADR-005 table

ADR-005 §Adapters-shipped-in-v1 was written against an older codex-acp variant. The current codex-cli (tested: 0.125.0) uses:

- **New turn**: `codex exec --json --skip-git-repo-check -o <file> "<prompt>"`
- **Resume**: `codex exec resume --json --skip-git-repo-check -o <file> <sid> "<prompt>"`

Session continuity is via the `exec resume <id>` subcommand, NOT a `--session <id>` flag. Documented in the source comment so a future codex bump has one file to update.

## Output handling

- **stdout**: JSONL events. Adapter stream-parses (handling chunk boundaries) and watches for:
  - `thread.started` → captures `thread_id` for next-turn resume
  - `turn.failed` → rejects with `error.message` so the run loop surfaces it instead of generic exit-code error
- **stderr**: Rust tracing logs (timestamped). Captured for error reporting only.
- **Final reply text**: read from the file written by `--output-last-message <FILE>` after the process closes — cleaner than parsing every JSONL event variant.

## Mirrors claude.js convention

- Memory preamble (`=== Context (your persistent memory) === ... === Current turn ===`) when `ctx.memoryLongTerm` is set
- Test seam via `ctx._spawnImpl`
- 5-minute default SIGTERM timeout (ADR-005 §Spawning semantics)
- Per-spawn temp dir for the output file, cleaned up in `finally`
- `detect()` returns `{ path, version }` or `null`

## Tests

`cli/__tests__/adapters.codex.test.mjs` — 14 tests:

- `detect()`: success, `which`-fallback, ENOENT, non-zero exit
- `spawn()`: first-turn argv, resume-turn argv (subcommand + prompt-position), memory preamble (with + without), `turn.failed` event surfacing, non-zero exit + stderr trim, timeout + SIGTERM, JSONL chunk-boundary buffering, temp-dir cleanup on rejection
- adapter registry: `listAdapterNames()` includes `codex`

```
PASS __tests__/adapters.codex.test.mjs (14/14)
PASS __tests__/adapters.claude.test.mjs
PASS __tests__/run-loop.test.mjs
... (11 suites, 124/124 tests pass)
```

## Live smoke — deferred

This PR ships the adapter + tests. Live smoke (attach + run + @mention round-trip against `api-dev.commonly.me`) is deferred because the local codex auth has an expired refresh token. After `codex logout && codex login`, the smoke is one-liner:

```
commonly agent attach codex --pod <dev-pod> --name laptop-codex \
  --instance https://api-dev.commonly.me
commonly agent run laptop-codex
# in pod: @laptop-codex hi
```

## What's next (separate PRs)

- Stage 2 — gateway-container deploy: add `@openai/codex` + `@commonly/cli` to `_external/clawdbot/Dockerfile`, wire `commonly agent run codex` alongside the openclaw process, install the codex agent into dev DM rooms. Lets dev agents (theo/nova/pixel/ops) replace `acpx_run` calls with `@codex` mentions.
- Phase 2 cursor adapter (no session, plaintext stdout)
- Phase 2 gemini adapter (no session, plaintext stdout)

## Test plan

- [x] `cli/__tests__/adapters.codex.test.mjs` — 14/14 pass
- [x] Full CLI suite — 124/124 pass, 11/11 suites
- [x] `npm run lint` on changed files — no errors
- [ ] Live smoke after `codex login` re-auth — attach + @mention round-trip on api-dev

### DIFF
--- a/cli/__tests__/adapters.codex.test.mjs
+++ b/cli/__tests__/adapters.codex.test.mjs
@@ -1,0 +1,289 @@
+/**
+ * adapters.codex.test.mjs — ADR-005 Phase 2
+ *
+ * Covers the codex-CLI adapter. The real `codex` binary is never invoked
+ * here: `child_process.spawnSync` is mocked at the module level for detect(),
+ * and the adapter's `_spawnImpl` test seam (ctx field) replaces childSpawn
+ * for spawn(). Both are internal-only seams — production never sets them.
+ *
+ * Argv shape under test (codex-cli 0.125.0):
+ *   - new turn:   codex exec       --json --skip-git-repo-check -o <file> "<prompt>"
+ *   - resume:     codex exec resume --json --skip-git-repo-check -o <file> <sid> "<prompt>"
+ */
+
+import { jest } from '@jest/globals';
+import { EventEmitter } from 'events';
+import { mkdtemp, writeFile, rm, readFile } from 'fs/promises';
+import { tmpdir } from 'os';
+import { join } from 'path';
+
+const spawnSyncMock = jest.fn();
+await jest.unstable_mockModule('child_process', () => ({
+  spawnSync: spawnSyncMock,
+  spawn: jest.fn(),
+}));
+
+const codex = (await import('../src/lib/adapters/codex.js')).default;
+
+// Fake child process with optional pre-canned stdout chunks, stderr, exit code.
+// Set `writeOutputFile: <text>` to simulate codex writing the
+// --output-last-message file before exiting (the production codepath reads
+// that file after the close event fires).
+const fakeChild = ({
+  stdoutChunks = [],
+  stderr = '',
+  code = 0,
+  delayMs = 0,
+  outputFile = null,
+  outputContents = null,
+} = {}) => {
+  const proc = new EventEmitter();
+  proc.stdout = new EventEmitter();
+  proc.stderr = new EventEmitter();
+  proc.kill = jest.fn();
+  setTimeout(async () => {
+    for (const chunk of stdoutChunks) proc.stdout.emit('data', Buffer.from(chunk));
+    if (stderr) proc.stderr.emit('data', Buffer.from(stderr));
+    if (outputFile && outputContents !== null) {
+      await writeFile(outputFile, outputContents, 'utf8');
+    }
+    proc.emit('close', code);
+  }, delayMs);
+  return proc;
+};
+
+const findOutputFile = (args) => {
+  const idx = args.findIndex((a) => a === '-o');
+  return idx === -1 ? null : args[idx + 1];
+};
+
+const makeSpawnImpl = ({ stdoutChunks = [], stderr = '', code = 0, outputContents = null } = {}) => {
+  const calls = [];
+  const impl = (cmd, args, opts) => {
+    calls.push({ cmd, args, opts });
+    return fakeChild({
+      stdoutChunks,
+      stderr,
+      code,
+      outputFile: findOutputFile(args),
+      outputContents,
+    });
+  };
+  return { impl, calls };
+};
+
+describe('codex adapter — detect()', () => {
+  beforeEach(() => { spawnSyncMock.mockReset(); });
+
+  test('returns { path, version } when `codex --version` exits 0', async () => {
+    spawnSyncMock.mockImplementation((cmd) => {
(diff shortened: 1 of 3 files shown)


lilyshen0722:
Live smoke against `api-dev.commonly.me` exposed one issue and validated end-to-end.

**Issue caught + fixed:** the original adapter used `child_process.spawn`'s default `stdio`, which gives the child a fresh pipe for stdin. Codex 0.125.0 then prints `Reading additional input from stdin...` and blocks waiting for input that never arrives — the spawn hangs until the 5-min SIGTERM timeout. Interactive runs are fine because codex detects a TTY and uses the argv prompt directly. Only non-TTY parents (the run loop) trigger it.

Fix in `358bb4de91`: `stdio: ['ignore', 'pipe', 'pipe']` so codex sees `/dev/null` for stdin, hits immediate EOF, falls back to the argv prompt. Pinned with a regression test.

**End-to-end verification:**
```
[19:25:29] laptop-codex: Hi all — I'm laptop-codex. Local codex CLI wrapped as a Commonly agent (ADR-005). Ping me when you need it.
[19:25:42] xcjsam:       @laptop-codex please reply with the single word: pong
[19:33:34] laptop-codex: pong
```

Flow: `commonly agent attach codex --pod <id> --name laptop-codex` → `commonly agent run laptop-codex` → @-mention from another user → wrapper polled `chat.mention` event → spawned codex with patched stdio → codex returned `pong` via `--output-last-message` → wrapper posted to pod.

Tests: 15/15 (added one regression). Full CLI suite still 11/11 green.

claude.js has the same latent default but works in practice — flagged as out-of-scope follow-up in the commit message.

lilyshen0722:
**Self-review pass via code-reviewer subagent — addressed two important findings + one nit in `97e1e1a3ad`.**

|---|---|---|
| Resume argv put `<sid>` after option flags — future codex parser change could silently consume it as `-o`'s value | Important | Reordered to `exec resume <sid> --json --skip-git-repo-check -o <file> <prompt>`. Matches codex's documented signature, parser-independent. Test pins exact position. |
| Temp-dir cleanup test asserted on `/proc/self/status` content — always true, no actual coverage | Important | Replaced with a real before/after count of `commonly-codex-*` entries in `tmpdir()` |
| Module JSDoc referenced `--output-last-message` while argv uses `-o` short alias | Nit | Comment updated for consistency |

**Reviewer findings deliberately deferred:**

- *Q: Resume path not live-smoked* — true. The argv reordering above makes the resume positional unambiguous regardless of parser implementation, which neutralizes the original concern. Live resume smoke can ride along with Stage 2 (multiple agents in DM rooms naturally exercises resume on every turn beyond the first).
- *Q: `turn.failed` + session continuity* — open question. Currently a model-rejected turn drops the `thread_id` and the next delivery starts fresh. Reviewer's right that resuming might be more appropriate, but this is a behavioral decision worth its own discussion, not an in-PR change. Filing as follow-up.
- *Nits about `else if` on event types, `which` called even on detect failure, `spawn: jest.fn()` unused in detect tests* — all match `claude.js` conventions intentionally. Skipped.

**What the reviewer flagged as solid** (worth surfacing for posterity): adapter purity holds (ADR-005 invariant #1), JSONL parser correct on chunk-boundary / partial-line / non-JSON / `turn.failed` cases, temp-dir lifecycle clean across all error paths, no sensitive data in source or tests, convention drift vs claude.js minimal and justified.

Verdict: **Approve with suggestions** → suggestions addressed. Tests still 15/15.

lilyshen0722:
Squashed and merged manually as commit b94bf748d4 to preserve Lily as author (per repo convention — see memory/feedback-pr-merge-pattern.md).

PR #244, opened 2026-04-27
chore(deploy-dev): bump helm --timeout 5m → 10m

lilyshen0722:
## Summary

Documented HANDOFF follow-up: \`Deploy Dev\` shows red cosmetically when \`helm --wait --timeout 5m\` is exceeded by a rolling update that's still in progress. Hit again on the ADR-010 Phase 1 deploy (5m13s, all four deployments rolled out cleanly per kubectl, but workflow exited red).

10 minutes is enough margin for a four-deployment rolling update without masking real failures (a genuinely stuck rollout still times out comfortably within the 30m job ceiling).

## Test plan

- [ ] CI green
- [ ] Next \`workflow_dispatch\` of Deploy Dev exits clean even on a four-deployment rolling update

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #245, opened 2026-04-27
chore(deploy-dev): bump helm --burst-limit 200 --qps 100

lilyshen0722:
## Summary

Helm \`--wait\` has been hitting its **own client QPS limiter** during readiness polling — not pod-rollout-time, not API server throttling. Defaults (qps=50, burst=100 since helm 3.10) are insufficient for a chart with 30+ resources. Bumping to 200/100 doubles the polling budget without taxing the API server.

This is a pragmatic try at fixing the recurring \`client rate limiter Wait returned an error: context deadline exceeded\` that's been present on every helm upgrade since rev 67 (Apr 21). Earlier attempts to fix this by bumping the timeout (#244) or granting the deploy SA RBAC perms (out-of-band manual rolebinding) helped marginally but didn't address the root cause.

## What this is NOT

Not a guaranteed fix — if the QPS theory is wrong we're back to one of: drop \`--wait\` entirely, accept the cosmetic failure, or dig deeper. Either way this is a low-risk one-line change worth trying first.

## Test plan

- [ ] CI green
- [ ] Re-trigger Deploy Dev after merge; helm step completes < 10m without timeout

### DIFF
--- a/.github/workflows/deploy-dev.yml
+++ b/.github/workflows/deploy-dev.yml
@@ -142,7 +142,9 @@ jobs:
             --set agents.clawdbot.image.tag="$TAG" \
             --set agents.commonlyBot.image.tag="$TAG" \
             --wait \
-            --timeout 10m
+            --timeout 10m \
+            --burst-limit 200 \
+            --qps 100
 
       - name: Report deploy outcome
         if: always()


lilyshen0722:
Squashed manually as the merge commit on main.

PR #252, opened 2026-04-30
fix(messages): show real author on optimistic chat render (no more 'Unknown')

samxu01:
## Summary
- POST `/api/messages/:podId` was returning the raw INSERT row from `PGMessage.create()`, which has no JOIN to the `users` table. Frontend `normalizeMessage` fell through to `'Unknown'` for every message the sender just posted, until a refresh pulled the joined row.
- Mirror what `pgMessageController.createMessage` already does: re-fetch via `PGMessage.findById()` after create so the response carries `userId.{username, profilePicture}` in the same shape `GET /api/messages/:podId` returns.
- Apply the equivalent `.populate('userId', 'username profilePicture')` to the Mongo fallback path.

Surfaced during the YC demo recording — when Sam typed a message in his own pod, his bubble rendered as "Unknown" before the next refresh. Real bug, not a UX papercut: the response is just under-shaped.

## Test plan
- [ ] After deploy, log in to app-dev as Sam, send a message in any pod, confirm the author renders as `sam` (not `Unknown`) before any refresh.
- [ ] Check that messages from other users still render correctly.
- [ ] Check that agent messages still get the displayName override (`Engineer (Nova)` etc.) — that path is unaffected.

### DIFF
--- a/backend/controllers/messageController.ts
+++ b/backend/controllers/messageController.ts
@@ -154,13 +154,20 @@ exports.createMessage = async (req: AuthRequest, res: Response): Promise<void> =
     let message: NormalizedMessage;
 
     try {
-      message = await PGMessage.create(
+      const created = await PGMessage.create(
         podId,
         userId,
         messageContent || '',
         'text',
         replyToMessageId || null,
-      ) as NormalizedMessage;
+      );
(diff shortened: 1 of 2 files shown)


samxu01:
Squash-merged locally as c448708a5d on main.

PR #253, opened 2026-04-30
fix(messages): pull _id from populated User in normalizeMongo

samxu01:
## Summary
- When a Mongo message has its `userId` populated, calling `.toString()` on the Mongoose User document returns `util.inspect` output (`{\n  _id: new ObjectId("..."), username: '...', ...}`) instead of the ObjectId string. That string leaked into the `user_id` field of every API response that hit the Mongo fallback path.
- Real impact: the dev cluster's PG host has been unreachable for some time (Aiven host returns NXDOMAIN), so every message read goes through this fallback. The v2 chat renders author + avatar correctly via the populated `user` field, but the `user_id` field is malformed everywhere it's read.
- Fix: when `m.userId` is an object with `_id`, extract that `_id`; otherwise use the raw value. Pull `username`/`profilePicture` from the same object in one pass.

## Test plan
- [ ] Existing `messageController.test.js` cases still pass (PGMessage path is untouched).
- [ ] After deploy, GET `/api/messages/:podId` for any pod returns `user_id` as a 24-char hex ObjectId string, not the inspect-format blob.
- [ ] Spot-check on app-dev: messages render with the correct author and avatar (no behavior change there — only `user_id` is now well-formed).

### DIFF
--- a/backend/controllers/messageController.ts
+++ b/backend/controllers/messageController.ts
@@ -35,21 +35,31 @@ const pgAvailable = (): boolean => {
   }
 };
 
-const normalizeMongo = (m: Record<string, unknown>): NormalizedMessage => ({
-  id: (m._id as { toString(): string }).toString(),
-  pod_id: (m.podId as { toString(): string }).toString(),
-  user_id: (m.userId as { toString(): string }).toString(),
-  content: m.content as string,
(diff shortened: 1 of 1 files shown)


samxu01:
Squash-merged locally on main.

PR #258, opened 2026-04-30
fix(v2): restore DM button on openclaw agents (wire shape mismatch)

samxu01:
## Summary
Hot follow-up to PR #257. The `isAgentDmable` helper introduced there checked `agent.agentName === 'openclaw'`, but `/api/registry/pods/:id/agents` returns the agent type as `name` on the wire — the same shape mismatch V2PodChat already works around for the displayName map. Net effect: the DM button disappeared from every agent row in the right rail (Nova / Pixel / Aria included), because `agent.agentName` was undefined.

Read both `agent.name` (wire) and `agent.agentName` (V2Agent type), match the existing pattern in V2PodChat:233.

## Test plan
- [ ] After deploy: `Engineer (Nova)`, `UI Lead (Pixel)`, `Strategist (Aria)` show their `DM` button again.
- [ ] `Pod Summarizer` still has no DM button (correct — it's Tier 1 native, not openclaw).


samxu01:
Squash-merged locally on main.

PR #259, opened 2026-04-30
feat(v2,chat): markdown + typing indicator + @-mention dropdown

lilyshen0722:
## Summary

Three changes that make V2 chat feel alive — flat text + no movement looks broken on the YC demo path. Each piece is small and reuses backend signal that already exists.

1. **Markdown** — `V2MessageBubble` pipes content through `ReactMarkdown` with v2-scoped HTML elements (no MUI dependence), styled in `v2.css` under `.v2-msg__content`. Lists / fenced code / inline code / links / blockquotes / headings / bold / italic / hr now render as such instead of as a flat pre-wrapped string. File / reaction / image token parsing runs first so those still work.
2. **Typing indicator** — `V2PodChat` subscribes to `agent_typing_start` / `agent_typing_stop` socket events (already emitted by `agentTypingService` on the backend). Pill renders between message list and composer: stacked avatars + "Engineer (Nova) is thinking…" + animated dots. Keyed by \`\${agentName}:\${instanceId}\` for multi-instance; 30s safety auto-stop in case a stop event is dropped.
3. **@-mention dropdown** — typing \`@\` opens a popover above the composer listing pod members + installed agents. ArrowUp / ArrowDown navigate, Enter inserts, Esc closes, click-outside dismisses. Mention values prefer \`instanceId\` for distinct multi-instance agents (so \`@nova\`, not \`@openclaw\`).

## Test plan

- [ ] After deploy, open \`/v2/pods/<demo-pod-id>\` on \`app-dev\` — agent intro messages render with markdown formatting (lists, bold, links).
- [ ] Send a message that prompts an openclaw agent — typing pill appears with the agent's display name + animated dots while it's working, disappears when the response posts.
- [ ] Type \`@n\` in the composer — dropdown appears with Nova / any other matching agent. Arrow + Enter inserts \`@nova \` at the cursor.
- [ ] Existing image / file pill / reaction renderers still work (no regression on \`V2MessageBubble\`'s token parsing).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #260, opened 2026-04-30
fix(v2,chat): mention dropdown shows @nova not @openclaw-nova

lilyshen0722:
## Summary

Follow-up to PR #259. The first cut of the @-mention dropdown listed agents by their User row username (`@openclaw-nova`, `@openclaw-aria`) instead of by their instance handle (`@nova`, `@aria`). Functional — the User row IS \`openclaw-nova\` so the mention still resolves — but the demo experience should mirror how a human would type it.

Cross-reference members[] (User rows) with agents[] (AgentInstallation rows) on the way in: any member with a matching installation is promoted to the agent shape — instanceId becomes the mention value, \`Agent · @nova\` becomes the subtitle, displayName becomes the label.

## Test plan

- [ ] Type \`@\` in v2 composer on the demo pod — dropdown shows \`@nova / @pixel / @aria\` with subtitle \`Agent · @<handle>\`, plus humans as \`@username · Member\`.
- [ ] Hit Enter — \`@nova \` is inserted at the cursor.
- [ ] Send the message — Nova receives the mention (User row resolution still works since \`@nova\` is a valid mention handle).

### DIFF
--- a/frontend/src/v2/components/V2PodChat.tsx
+++ b/frontend/src/v2/components/V2PodChat.tsx
@@ -332,13 +332,42 @@ const V2PodChat: React.FC<V2PodChatProps> = ({ detail }) => {
     return map;
   }, [agents]);
 
-  // Build the @-mention list: pod members (humans) + installed agents. Same
-  // shape and dedup rules as v1 ChatRoom — agent identity is keyed on the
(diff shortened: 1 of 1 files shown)


lilyshen0722:
Squashed locally and pushed to main. See merge pattern in CLAUDE memory.

PR #273, opened 2026-05-03
feat(v2): auto-send file as a message on attach (no draft staging)

samxu01:
Second of three "artifact polish" PRs. Will not auto-merge.

## What this changes

Picking a file from the composer now **uploads + sends a message containing the upload directive in one shot** instead of inserting raw \`[[upload:fileName|…]]\` into the textarea.

**Why:** the previous flow showed raw directive tokens as the user typed, which felt off. Drag-and-drop in Slack, Discord, and Linear all match the new shape: pick → goes now. The user's existing draft text is preserved — they can keep typing and hit Send when ready, and the file lands as its own message.

Images keep the existing inline-image-message behavior (URL-as-content, \`messageType: 'image'\`).

## Diff

\`V2PodChat.tsx\` — \`handleAttachFile\` calls \`sendMessage(directive)\` instead of mutating \`draft\` with \`setDraft\`.

## Test plan

- [ ] Deploy dev: \`gh workflow run deploy-dev.yml --ref feat/v2-attach-auto-send\`
- [ ] Open a pod, type some text in the composer (\"hey, see this\") **but don't send**
- [ ] Attach a PDF — confirm a separate message appears with the file pill, and the typed text stays in the composer
- [ ] Attach a PNG — confirm legacy inline-image behavior (image message lands)
- [ ] Confirm no \`[[upload:…]]\` raw text ever appears in the textarea

### DIFF
--- a/frontend/src/v2/components/V2PodChat.tsx
+++ b/frontend/src/v2/components/V2PodChat.tsx
@@ -455,19 +455,22 @@ const V2PodChat: React.FC<V2PodChatProps> = ({ detail, inspectorCollapsed, onTog
     }
   };
 
-  // Composer attach: handles both images (sends as standalone image message,
-  // legacy v2 behavior) and other file kinds (PDF / md / txt / csv / json,
-  // inserts an [[upload:fileName|originalName|size|kind]] directive into the
-  // draft so the user can add accompanying text and send when ready). Both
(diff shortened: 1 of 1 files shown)


lilyshen0722:
Closing as stale: opened 2026-05-03, part of the same v2 file-handling batch as #274 and #278. Small enough to redo cleanly against current main if still wanted.

PR #274, opened 2026-05-03
feat(v2): delete artifacts (files + external links) from inspector

samxu01:
Third of three "artifact polish" PRs. Will not auto-merge.

## What this changes

Pod owner OR original creator can now delete an artifact from the inspector detail view.

**Backend:**
- New \`DELETE /api/pods/:podId/files/:fileId\` — pod owner OR file uploader. Deletes bytes from the configured ObjectStore driver and the \`File\` row. Driver delete failures are logged but don't block the metadata delete; orphaned bytes are GC'd by the future ADR-002 Phase 4 sweep. Validates fileId shape against the ObjectId regex before any DB call (ReDoS guard).
- \`DELETE /api/pods/external-link/:id\` relaxed from owner-only to owner OR \`createdBy\`. Pod members at large still can't nuke shared links — keeps a malicious pod-mate from clearing artifacts.

**Frontend:**
- \`V2PodInspector renderArtifactDetail\` gains a Delete button next to Open when the viewer is permitted (pod owner OR creator).
- ArtifactItems carry \`sourceKind / sourceId / createdById\` so the detail view knows which DELETE endpoint to hit and whether the current viewer can.
- \`window.confirm\` before delete. On success we remove the item from local state and pop back to the overview.

## What's deliberately not deletable

Announcements aren't deletable from the inspector. Those live in a separate admin flow (and the legacy \`DELETE /api/pods/announcement/:id\` route). If we want to surface a delete here later we add a 4th \`sourceKind\` and a permission check.

## Tests

35/35 passing.
- \`pods.external-links.test.js\`: 28 cases (was 27). New "allows the link creator to delete (even if not pod owner)" + updated 403 case to assert "neither owner nor creator".
- \`pods.files-delete.test.js\`: 7 new cases. Owner deletes, uploader deletes, neither 403s, invalid fileId 400s, missing file 404s, wrong-pod 404s, ObjectStore failure doesn't block metadata delete.

## Test plan

- [ ] Deploy dev: \`gh workflow run deploy-dev.yml --ref feat/v2-artifact-delete\`
- [ ] Sam's YC Sprint pod → upload a file → click it in artifacts → confirm Delete button shows. Delete it → returns to overview, gone from list.
- [ ] As a non-owner non-uploader pod member, open someone else's file → confirm Delete button is hidden.
- [ ] Add an external link as user A, switch to user B (not owner): Delete button hidden. Switch to pod owner: Delete shows.
- [ ] Confirm announcements still don't show a Delete button.

### DIFF
--- a/backend/__tests__/unit/routes/pods.external-links.test.js
+++ b/backend/__tests__/unit/routes/pods.external-links.test.js
@@ -241,10 +241,28 @@ describe('pods routes - external links', () => {
     await request(app).delete('/api/pods/external-link/l1').expect(404);
   });
 
-  it('returns 403 when deleting link as non-owner', async () => {
-    ExternalLink.findById.mockResolvedValue({ podId: 'p1' });
-    Pod.findById.mockResolvedValue({ createdBy: { toString: () => 'other' } });
+  it('returns 403 when deleting link as neither owner nor creator', async () => {
+    ExternalLink.findById.mockResolvedValue({
+      podId: 'p1',
+      createdBy: { toString: () => 'someone-else' },
+    });
+    Pod.findById.mockResolvedValue({ createdBy: { toString: () => 'other-owner' } });
     await request(app).delete('/api/pods/external-link/l1').expect(403);
+  });
+
+  it('allows the link creator to delete (even if not pod owner)', async () => {
+    const podSave = jest.fn();
+    ExternalLink.findById.mockResolvedValue({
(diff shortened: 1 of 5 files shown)


lilyshen0722:
Closing as stale: opened 2026-05-03. The v2 shell has moved substantially since, so this would be a rewrite rather than a rebase. Re-file against current main if artifact deletion is still wanted.

PR #275, opened 2026-05-03
feat(agent-collab): Phase 1 + early Phase 2 — agent-dm pod type, sharePod auth, codex preset

samxu01:
Implements Phase 1 + early Phase 2 of `docs/plans/agent-collaboration-surfaces.md`.

## Summary

**Phase 1 (the demo blocker)**
- `agent-dm` Pod.type — any 2-member DM (any combination of user/bot).
- Allow-list patches in `messageController.ts` and `agentMentionService.ts:enqueueDmEvent` so the new pod type doesn't silent-drop. Bot senders are intentionally allowed in `agent-dm` (the whole point) but still blocked in legacy `agent-admin`.
- `dmService.sharePod(a, b)` — single source of truth for the §3.7 co-pod-member rule.
- `dmService.getOrCreateAgentDmRoom(memberA, memberB)` — idempotent on the unordered pair, AgentInstallation upserts for any bot members (heartbeat off).
- `AgentInstallation.upsert(...)` — new static via `findOneAndUpdate(upsert+setOnInsert)`. Re-fires + later admin installs converge on a single row instead of throwing or duplicating.
- `POST /api/agents/runtime/agent-dm` — accepts target = `{agentName,instanceId}` | `{userId}` | `{alias}`; resolves alias against `pod.contacts` then caller's contacts; runs `sharePod` auth (bypassed only when resolution came from pod.contacts — admin-binding carve-out); drops "DM started" system event in `originPodId`.
- Mention-driven autoJoin in `agentMentionService` behind `ENABLE_MENTION_AUTOJOIN` env flag (default off): unresolved aliases fall through to pod.contacts → sender.contacts; if resolved outside `pod.members`, runs sharePod (same admin-binding carve-out) + AgentInstallation.upsert + adds to pod.members + drops system event + fires chat.mention.
- Schema additions, all with safe defaults (existing rows return empty, never throw on read):
  - `User.contacts: [ContactEntry]` for both human and bot users.
  - `Pod.contacts: Map<alias, {agentName, instanceId}>` for per-pod bindings.
  - `AgentInstallation.config.role` reuses existing free-form config map.
- V2 sidebar treats `agent-dm` as Private; agent↔agent (both bot members) gets an "AA" icon to distinguish from human DMs.
- `getAllPods` extends membership filter to `agent-dm` so personal pod types stay user-scoped.

**Phase 2 substrate**
- `codex` agent preset on the openclaw clawdbot gateway — code-quality-focused, reactive (no heartbeat), routes through LiteLLM via the codex-auth-rotator already live. Replaces the operator-specific `sam-local-codex` stop-gap that broke when account-2/3 tokens were refreshed for the rotator.
- `PATCH /api/pods/:id/contacts` — pod creator OR global admin only (intentionally tighter than member-only; see Self-review #1 below). One endpoint for all aliases (codex/claude/reviewer/…).

**Out of scope for this PR:**
- Manage-tab UI to expose the PATCH endpoint (frontend follow-up; the API is shipped).
- Pixel/Theo/Ops heartbeat-template cutover (registry.ts) — depends on the codex agent being installed in the demo pods, which is a runtime-config step.

## Self-review pass

Ran the code-reviewer agent on `2baa52d266`. Three Important findings, all addressed in `a186406855`:

1. **PATCH /pods/:id/contacts was member-gated** — would let any pod member pin an arbitrary agent and bypass `sharePod`. Tightened to creator OR global admin (db lookup; we don't trust JWT-only role claims).
2. **TOCTOU on the admin-binding carve-out** — pod.contacts was fetched twice. Now fetched once and the source is propagated (`'pod' | 'sender'`) through `resolveContactAlias`.
3. **Ghost-User materialization on `target.agentName`** — typos would create permanent bot User rows. Probe via `User.findOne` first; 404 on miss.

Reviewer also flagged a Question about backfilling existing `agent-room` rows missing AgentInstallation — that's a runtime fix using the existing kubectl-exec script from `e78b5df241`, not blocked by this PR.

## Test plan

- [x] 25 unit + service tests passing (`dmService.agentDm.test.ts` + extended `agentMentionService.test.js`).
- [x] `tsc:check` clean.
- [ ] Browser-verify on dev once merged: open `/v2/pods` (no socket-auth flood), POST to `/api/agents/runtime/agent-dm` with valid + invalid agentName (404 on miss, no ghost row), confirm allow-list lets messages route to the agent in an `agent-dm` pod.
- [ ] After merge, install `codex` preset in a demo pod and pin via `PATCH /pods/:id/contacts` to verify the resolution chain end-to-end.

### DIFF
--- a/backend/__tests__/services/dmService.agentDm.test.ts
+++ b/backend/__tests__/services/dmService.agentDm.test.ts
@@ -1,0 +1,235 @@
+// @ts-nocheck
+/**
+ * DMService — agent-dm + sharePod tests (Phase 1 of agent collaboration plan).
+ *
+ * Verifies:
+ * - sharePod returns true/false on co-pod-member relationships across
+ *   bot/bot, bot/human, human/human pairs and self.
+ * - getOrCreateAgentDmRoom creates an agent-dm pod on first call,
+ *   returns the same pod on second call regardless of arg order.
+ * - Both bot members get an AgentInstallation row (heartbeat off) so
+ *   outbound posting works (the e78b5df241 invariant for the new
+ *   pod type).
+ * - AgentInstallation.upsert is idempotent: re-firing the autoJoin
+ *   path on an existing row produces no duplicates and never throws.
+ */
+
+const mongoose = require('mongoose');
+const { MongoMemoryServer } = require('mongodb-memory-server');
+
+jest.mock('../../models/pg/Pod', () => null);
+
+const DMService = require('../../services/dmService');
+const Pod = require('../../models/Pod');
+const User = require('../../models/User');
+const { AgentInstallation } = require('../../models/AgentRegistry');
+
+describe('DMService — agent-dm + sharePod', () => {
+  let mongoServer;
+  let aria;
+  let codex;
+  let alice;
+  let bob;
+
+  beforeAll(async () => {
+    mongoServer = await MongoMemoryServer.create({
+      binary: { version: '7.0.11', skipMD5: true },
+      instance: { dbName: 'dm-service-agent-dm-test' },
+    });
+    await mongoose.connect(mongoServer.getUri());
+  });
+
+  afterAll(async () => {
+    await mongoose.disconnect();
+    if (mongoServer) {
+      try { await mongoServer.stop(); } catch (_) { /* ignore */ }
+    }
+  });
+
+  beforeEach(async () => {
+    await Promise.all([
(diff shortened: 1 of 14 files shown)


samxu01:
Merged to main as d5b7198e3c via local squash (per project convention — see feedback-pr-merge-pattern memory).

PR #276, opened 2026-05-03
docs(adr): ADR-012 memory propagation + event-payload injection

samxu01:
## Summary

Builds on [ADR-003](https://github.com/Team-Commonly/commonly/blob/main/docs/adr/ADR-003-memory-as-kernel-primitive.md). Closes the cross-session-context gap that surfaced when agent-dm shipped (#275 / `d5b7198e3c`): an agent's session in agent-dm pod-A doesn't see what its sibling session in team pod-B did.

Two halves, both kernel-side, both runtime-agnostic:

**1. New typed `system_exchanges` section in the ADR-003 envelope.** Structured entries (`kind`, `surfacePodId`, `peers`, `takeaway`), not a markdown blob. Section is read-only from the agent's perspective — the name isn't in `commonly_save_my_memory`'s writable enum, so write-rejection is structural rather than markdown-parsing-based. Visibility is hard-coded `'private'` at the schema level, preserving ADR-003's agent-isolation rule even though the platform dual-writes to both peers in an exchange.

**2. Two new fields on the existing CAP event payload schema** — `memoryRevision` + `memoryDigest`. No new HTTP verbs, no new tool surface. Every runtime that reads its event payload gets memory propagation for free.

## What's NOT in the ADR (deliberate)

- ❌ LLM-side summarization on write (verbatim metadata in v1; v2 may layer condense step).
- ❌ Cross-agent memory reads (ADR-003 §Visibility still owns; cross-agent context flows through `commonly_ask_agent`, not memory peeks).
- ❌ Admin viewer for `system_exchanges` (would bypass agent-isolation; needs a privacy-vs-audit ADR addendum first).
- ❌ EventHandler-as-Installable framing for the four triggers (v1 hardcodes them as platform hooks because Installable Phase 3 hasn't shipped the read-path switch and third-party handlers writing to `system_exchanges` would re-open the privacy hole).
- ❌ Cross-pod-mention trigger (too noisy in multi-agent team pods; dropped for v1, v1.x reconsider with strict non-member-pod filter).

## Self-review pass

Ran the code-reviewer agent on a v0 draft. **Verdict was \"request changes.\"** Two critical architectural issues, four important ones, three real questions. All addressed before this PR:

| Reviewer finding | Fix in this draft |
|---|---|
| **Critical**: v0 built on the deprecated `content` blob; ADR-003 moved to typed `sections` | Retargeted entirely to the `sections` envelope. Section-level ACL replaces markdown parsing. |
| **Critical**: v0 claimed ack idempotency that `agentEventService.acknowledge` doesn't actually have | Replaced claim with a real spec: status-gated \`findOneAndUpdate\` + monotone \`$max\`. Implementation lands in Phase 2. |
| **Important**: cross-pod-mention trigger would flood the cap in multi-agent pods | Dropped for v1. v1.x reconsider with non-member-pod filter. |
| **Important**: \"drivers MUST surface memoryDigest\" vs \"adoption is independent\" — pick one | SHOULD not MUST. Un-adopted runtimes degrade (lazy reads via tool call) — degraded, not broken. |
| **Important**: Notes/System ownership unenforceable on PUT /memory | Structural section solves this naturally — agent's tool can't address `system_exchanges`. |
| **Important**: Phase-1 estimate too low if structural-section migration is in scope | Honest 3–4 days for Phase 1; total 5.5–6.5d for Phases 1–3. |
| **Question**: count-vs-time eviction | Hard decision: count-bounded (50 entries). v1.x may layer hybrid. |
| **Question**: ADR-003 invariant 8 dedup invalidation | New carve-out (invariant 8a): system writes don't touch \`lastSyncKey\`. |
| **Question**: should triggers be EventHandler Installables? | Documented in Non-goals: v1 hardcodes; v2 may revisit. |
| **Nit**: digest size estimate too optimistic | Cap is 1.5KB serialized OR 5 entries, whichever hits first. |
| **Nit**: \"new CAP verb\" framing was misleading | Reworded — two new fields on existing payload, zero new verbs. |

## Test plan

- [ ] Read ADR end-to-end. The §3 Acknowledgement-semantics block is the load-bearing implementation contract — sanity-check it against `agentEventService.ts`.
- [ ] Confirm the §1 \`SystemExchangesSection\` interface composes cleanly with ADR-003's \`MemorySection\` shape.
- [ ] Open Questions section has 3 items, all of which are *post-v1 reconsider* triggers (not blocking decisions). Agree?
- [ ] Phasing: Phase 1 (3–4d) + Phase 2 (2d) + Phase 3 (½d) + Phase 4 (per-runtime, parallel) — buy or push back?

### DIFF
--- a/docs/adr/ADR-012-memory-propagation-and-injection.md
+++ b/docs/adr/ADR-012-memory-propagation-and-injection.md
@@ -82,13 +82,13 @@ Storing structured entries (not a markdown blob inside `content`) gives us:
 - **Typed reads** for the digest builder — no string slicing.
 - **Future LLM condensation** can rewrite `takeaway` per-entry without touching the rest of the envelope.
 
-ADR-003 §"Tool surface" gets one row added (Phase 4 of this ADR; see §Phasing):
+ADR-003 §"Tool surface" gets one row added in **Phase 1** (the lazy-read fallback in §3 below depends on this — drivers that don't surface `memoryDigest` rely on agents reading `system_exchanges` via tool, so the section enum has to accept it from day one):
 
 | Tool | Purpose |
 |---|---|
 | `commonly_read_my_memory(section: 'system_exchanges')` | Returns the structured entries. Read-only; agent cannot write. |
 
-The existing `commonly_save_my_memory(section, content, ...)` tool **rejects** writes to `system_exchanges` at the API layer — section name is not in the writable enum.
+The existing `commonly_save_my_memory(section, content, ...)` tool **rejects** writes to `system_exchanges` at the API layer. Today's `agentMemoryService` does not yet enforce a writable-section allow-list — Phase 1 adds it. The allow-list will explicitly include the agent-writable sections (`long_term`, `dedup_state`, `shared`, `daily`, `relationships`, `soul`, `runtime_meta`) and exclude `system_exchanges`. Writes addressing `system_exchanges` return 403 with reason `'system_exchanges_is_read_only'`.
 
 ### 2. Push-based injection in CAP event payloads
 
@@ -107,8 +107,10 @@ The runtime sees memory changes in the payload it was already going to read. No 
 Backend tracks `AgentMemory.lastSeenRevision: number`. Bumping is **idempotent by construction** (see §Acknowledgement semantics below). `memoryDigest` returns entries with `revision > lastSeenRevision`, capped at 1.5KB total serialized JSON.
 
 Sizing notes:
-- Default cap: **1.5KB serialized digest, OR 5 entries, whichever hits first**. Byte cap dominates for noisy bursts; entry cap dominates for steady state.
-- Steady-state cost: ~600 tokens of payload per event. Acceptable on heartbeats; gated on `memoryRevision > lastSeenRevision` to avoid re-emitting unchanged digests on `chat.mention` traffic.
+- Per-entry footprint: `ts` (~24B) + `kind` (~25B) + `surfacePodId` (24B) + `surfaceLabel` (~30B) + `peers[]` (~40B) + `takeaway` (≤280B) + JSON overhead ≈ **420–450B per entry** with a worst-case `takeaway`.
+- Default cap: **2.5KB serialized digest, hard cap on bytes** (entry count is *not* a cap — at the per-entry footprint above, the byte cap admits up to 5 worst-case entries or 7-8 typical-case entries). Earlier drafts also cited "5 entries" — that was inconsistent with the byte cap; dropped.
+- Cost on a fresh delivery: up to ~750 tokens at the cap. Most events carry far less (the digest is delta-only — empty when `memoryRevision === lastSeenRevision`, which is the steady state once the agent has caught up).
+- Emission is gated on `memoryRevision > lastSeenRevision` so chat.mention traffic in a busy pod doesn't re-emit unchanged digests.
 
 ### 3. Acknowledgement semantics — the real spec
 
@@ -121,9 +123,29 @@ The reviewer caught that today's `agentEventService.acknowledge` has no idempote
 {
   ...,
   status: 'pending' | 'delivered' | 'acked',     // existing field; new terminal state
-  memoryRevisionAtDelivery: number | null,       // captured at fetch-time
+  memoryRevisionAtDelivery: number | null,       // captured + persisted on the doc at fetch-time (see below)
 }
 ```
+
+**Fetch-time capture is a doc mutation, not a derived response field.** Today's polling path (`agentEventService.list()` at `agentEventService.ts:796`) is a non-mutating `find` over `status: 'pending'`. To make `memoryRevisionAtDelivery` survive across the fetch→ack window without trusting the client to echo it, Phase 2 converts the polling path to atomic mutate-on-claim:
+
+```ts
+// Pseudocode — the actual list endpoint loops this for the batch
+const claimed = await AgentEvent.findOneAndUpdate(
+  { agentName, instanceId, status: 'pending' },
+  {
+    $set: {
+      status: 'delivered',
+      memoryRevisionAtDelivery: currentRevision,  // read from AgentMemory in the same tick
+    },
+  },
+  { new: true, sort: { createdAt: 1 } },
+);
+```
+
+Two concurrent pollers race the same event: only one wins the `pending → delivered` transition, and the winner is the one whose `memoryRevisionAtDelivery` lands on the doc. The loser sees `null` from `findOneAndUpdate` and moves to the next event. This matches the existing at-most-once delivery contract (ADR-004 §event-model).
+
+Alternative considered + rejected: keep `list` non-mutating, return `memoryRevisionAtDelivery` only in the response payload, require the agent to echo it on ack. Rejected because the ack body becomes a tampering surface (agent could echo a fake-high revision to suppress future digest entries) and validating against current memory at ack time defeats the purpose. Doc-mutation closes the loop.
 
 `agentEventService.acknowledge(eventId, ...)` is **status-gated**:
 ```ts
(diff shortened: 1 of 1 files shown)


docs/adr/ADR-012-memory-propagation-and-injection.md
```diff
@@ -0,0 +1,332 @@
...
+  takeaway: string;                    // ≤ 280 chars; verbatim metadata in v1, no LLM summarization
+}
+```
+
+Storing structured entries (not a markdown blob inside `content`) gives us:
+- **No markdown parsing** to enforce the system/notes boundary. The agent's `commonly_save_my_memory` tool writes named sections; it can't write `system_exchanges` because the tool's section enum doesn't include it. Section-level access control replaces heading-based parsing.
+- **Typed reads** for the digest builder — no string slicing.
+- **Future LLM condensation** can rewrite `takeaway` per-entry without touching the rest of the envelope.
+
+ADR-003 §"Tool surface" gets one row added (Phase 4 of this ADR; see §Phasing):
```
samxu01: **Phasing pointer is wrong.** This says the new `commonly_read_my_memory(section: 'system_exchanges')` row lands in "Phase 4 of this ADR," but Phase 4 in §Phasing is "Driver-side adoption" — it never mentions extending the tool's section enum. The lazy-read fallback you describe in §3 ("agents that do neither — bump fires lazily when they read memory") doesn't function until the enum accepts `'system_exchanges'`, so this needs to land in Phase 1 alongside the schema field. Either move the row to Phase 1 here, or add an "enum extension" bullet to Phase 4.

---

docs/adr/ADR-012-memory-propagation-and-injection.md
```diff
@@ -0,0 +1,332 @@
...
+
+| Tool | Purpose |
+|---|---|
+| `commonly_read_my_memory(section: 'system_exchanges')` | Returns the structured entries. Read-only; agent cannot write. |
+
+The existing `commonly_save_my_memory(section, content, ...)` tool **rejects** writes to `system_exchanges` at the API layer — section name is not in the writable enum.
```
samxu01: **Writable-section enum is proposed, not pre-existing.** Phrasing reads as if `commonly_save_my_memory` already enforces a section enum at the API layer — I couldn't find one in the current backend (no `writableSections` / explicit allow-list in `agentMemoryService.ts`). Phase 1 needs to add this enforcement explicitly, not rely on the description as if the structural rejection is already in place. Suggest reword: "...is rejected at the API layer; Phase 1 adds the writable-section allow-list (`long_term`, `dedup_state`, `shared`, `daily`, `relationships`) that excludes `system_exchanges`."

---

docs/adr/ADR-012-memory-propagation-and-injection.md
```diff
@@ -0,0 +1,332 @@
...
+  memoryDigest: SystemExchangeEntry[],   // delta since lastSeenRevision; capped at 1.5KB serialized
+}
+```
+
+The runtime sees memory changes in the payload it was already going to read. No tool call required. Every runtime — openclaw, claude API, webhook, MCP — gets memory propagation by handling its existing event input.
+
+Backend tracks `AgentMemory.lastSeenRevision: number`. Bumping is **idempotent by construction** (see §Acknowledgement semantics below). `memoryDigest` returns entries with `revision > lastSeenRevision`, capped at 1.5KB total serialized JSON.
+
+Sizing notes:
+- Default cap: **1.5KB serialized digest, OR 5 entries, whichever hits first**. Byte cap dominates for noisy bursts; entry cap dominates for steady state.
```
samxu01: **Sizing math is tight, possibly inconsistent.** Each entry: `ts` (~24B) + `kind` (~25B) + `surfacePodId` (24B) + `surfaceLabel` (~30B) + `peers[]` (~40B) + `takeaway` (≤280B) + JSON overhead ≈ 420–450 bytes. 5 fully-populated entries ≈ 2.1KB+, well above the 1.5KB cap. Either bump the byte cap to ~2.5KB or drop the "5 entries" half — claiming both is misleading because the byte cap will dominate every time the takeaway is non-trivial. Same applies to the "~600 tokens per event" estimate one line down — that's the cap, not steady state.

---

docs/adr/ADR-012-memory-propagation-and-injection.md
```diff
@@ -0,0 +1,332 @@
...
+The reviewer caught that today's `agentEventService.acknowledge` has no idempotency guard. We fix that in this ADR rather than asserting an invariant the code violates.
+
+**The fix** (lands in Phase 2 of this ADR):
+
+`AgentEvent` schema gains:
+```ts
+{
+  ...,
+  status: 'pending' | 'delivered' | 'acked',     // existing field; new terminal state
+  memoryRevisionAtDelivery: number | null,       // captured at fetch-time
```
samxu01: **"Captured at fetch-time" is underspecified — does the doc get mutated on list?** The current polling path is `agentEventService.list()` (`agentEventService.ts:796`), which is a non-mutating `find` over `status: 'pending'`. To persist `memoryRevisionAtDelivery` on the doc you'd have to convert the list path to either a `findAndModify`-style atomic update or follow-up bulk write — that's a real semantic change to the polling contract (e.g., what happens when two pollers race the same event?).

Alternative: compute `memoryRevisionAtDelivery` as a per-response payload field that is NOT persisted on the doc — but then the ack path can't read it back from the doc, so the agent ack would need to echo the revision number, opening a tampering surface.

Pick one and spell out the implication in this section. This is the load-bearing detail for Phase 2.

---

docs/adr/ADR-012-memory-propagation-and-injection.md
```diff
@@ -0,0 +1,332 @@
...
+- **webhook-SDK agents** (ADR-006) — `agentEventService.ts:196–200` marks the event `delivered` directly; no ack route. SDK SHOULD bump on HTTP-200 reply (we emit a small ack-equivalent helper in the SDK Phase 2).
+- **agents that do neither** — bump fires lazily when they read memory. Memory digest will redeliver until then; agent sees the same entries multiple times. Acceptable degradation; agents that read memory regularly converge.
+
+This replaces the v1 ADR's claim that ack idempotency was already implemented.
+
+### 4. Triggers — what gets written when
+
+| Trigger | Source | Recipients | Entry shape |
+|---|---|---|---|
+| `agent-dm-conclusion` | `agentMessageService.postMessage` when `content === 'NO_REPLY'` in an `agent-dm` pod | both peers | `kind: 'agent-dm-conclusion'`, takeaway = last non-NO_REPLY content from sender (truncated 280) |
```
samxu01: **`takeaway` derivation is hand-wavy.** "Last non-NO_REPLY content from sender" — for a 5-turn DM concluding on turn 5, is the takeaway turn-4 alone, or some condensation of turns 1–4? Multi-paragraph turn N — truncate from start, middle-elide, or take the first sentence? Suggest making it concrete in v1: "the immediately-preceding non-NO_REPLY message from the same sender, head-truncated to 280 chars (with `…` suffix on truncation)." Anything fancier waits for v2 LLM-condense.

---

docs/adr/ADR-012-memory-propagation-and-injection.md
```diff
@@ -0,0 +1,332 @@
...
+| `agent-dm-loop-trip` | `agentMentionService.enqueueDmEvent` when `bot_loop_guard` returns | both peers | `kind: 'agent-dm-loop-trip'`, takeaway = `'8 consecutive bot turns within 30 min'` |
+| `task-completed` | `tasksApi` complete handler | task assignee | `kind: 'task-completed'`, takeaway = `<taskTitle> → <prUrl|status>` |
+
+**Cross-pod-mention is dropped as a v1 trigger** (per reviewer §Important). Recording every mention in a multi-agent team pod fills the entry cap with structural noise faster than real DM exchanges. v1.x will reconsider it with a strict filter (only when the *target* agent is NOT in the source pod, i.e. genuinely new context). v1 ships without it.
+
+Writes are append-only via `AgentMemoryService.appendSystemExchange(agent, instance, entry)`. Cap enforcement is on write — when adding entry N+1 past the cap, drop the oldest.
```
samxu01: **Cap enforcement on write needs the Mongo idiom spelled out.** Two concurrent `appendSystemExchange` calls in the same tick (e.g., a DM-conclusion firing while a `task-completed` lands for the same agent) will lost-update each other under naïve read-modify-write. Specify the atomic shape:

```ts
await AgentMemory.updateOne(
  { agentName, instanceId },
  {
    $push: {
      'sections.system_exchanges.entries': {
        $each: [entry],
        $position: 0,
        $slice: 50,
      },
    },
    $inc: { revision: 1 },
    $currentDate: { 'sections.system_exchanges.updatedAt': true },
  },
);
```

`$push`+`$position: 0`+`$slice: 50` keeps the most-recent-first invariant atomically; `$inc: { revision: 1 }` is the monotone bump consumed by `memoryDigest`.

---

samxu01 (commented):
Verified the load-bearing technical claims against the codebase — `acknowledge` lacks idempotency as described (`agentEventService.ts:828–890`), the `sections` envelope shape composes cleanly, and the `enqueueDmEvent` / `bot_loop_guard` / `tasksApi` complete-handler callsites all exist. Spec is in good shape.

Six inline suggestions, ordered roughly by leverage:

1. **Phase pointer for the new tool-section row is wrong** (line 85) — needs to land in Phase 1, not Phase 4, or the lazy-read fallback doesn't work day one.
2. **Writable-section enum is asserted as if pre-existing** (line 91) — it isn't; Phase 1 needs to add it.
3. **`memoryRevisionAtDelivery` "captured at fetch-time" is underspecified** (line 124) — the polling path is non-mutating today; persisting this field requires a real semantic change to `list()`. Pick a model.
4. **Sizing math is internally inconsistent** (line 110) — 5 fully-populated entries blow the 1.5KB cap.
5. **`takeaway` derivation is hand-wavy** (line 161) — make the rule concrete for multi-turn / multi-paragraph cases.
6. **Cap enforcement needs the atomic Mongo idiom** (line 167) — concurrent triggers will lost-update under read-modify-write.

None block merging this as Proposed; (1)–(3) are worth tightening before Phase 1 starts.

---

PR #272, opened 2026-05-03
feat(v2): papaparse for CSV + mammoth for .docx preview

samxu01:
First of three "artifact polish" PRs queued for review (will not auto-merge).

## What this changes

**CSV preview** — replaces the hand-rolled \`split(',')\` with **papaparse**. The previous parser broke on real spreadsheet exports (quoted fields containing commas, escaped quotes, multi-line cells). papaparse is RFC 4180.

**DOCX preview** — Word documents render inline via **mammoth** (\`mammoth/mammoth.browser\`, lazy-imported so inspectors that never open a Word file don't pay the bundle cost). Mammoth produces clean semantic HTML — headings, lists, tables, bold/italic. For pixel-faithful render the user clicks Open and uses Word / Office Online.

## Bundle impact

- **papaparse@^5.5.3** + **@types/papaparse@^5.5.2** — ~45 KB gzipped, no runtime side effects.
- **mammoth@^1.12.0** — ~200 KB, **lazy-loaded** via dynamic \`import('mammoth/mammoth.browser')\` so the cost only lands when a docx artifact is opened.

## Skipped (deferred)

| Format | Why |
|---|---|
| xlsx / xls | SheetJS is ~600 KB. Demo uses CSV instead; revisit if a beta user asks. |
| ppt / pptx | No clean OSS option that's lightweight. PowerPoint inline preview needs server-side render. |
| odt / ods / odp | Same shape as their MS counterparts; covered by the same Office Online viewer follow-up. |

## Test plan

- [ ] Deploy dev: \`gh workflow run deploy-dev.yml --ref feat/v2-preview-libs\`
- [ ] Upload a CSV with quoted-comma fields (e.g., \`"Smith, John",42\`) — confirm cells aren't split mid-field
- [ ] Upload a real .docx with headings/lists/tables — confirm semantic HTML renders
- [ ] Confirm mammoth doesn't load on page load (Network tab shouldn't pull mammoth.browser until a docx artifact is opened)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #278, opened 2026-05-03
feat(v2): file pill UX — Slack-style hierarchy + click-to-preview

samxu01:
Fourth artifact polish PR. Will not auto-merge.

Reworks the in-message file pill so the visual hierarchy matches Slack / Google Chat / Linear, and switches click behavior from "open bytes in new tab" to "open inspector artifact detail" (preview-first).

## Visual hierarchy

**Before:** 28px icon, single-row pill with filename and size at equal weight (\`hello.pdf      120 KB\`).

**After:** 40px icon, two-line meta:
\`\`\`
[ PDF ]  hello.pdf
         PDF · 120 KB
\`\`\`
Friendly kind labels: \`PDF\`, \`Word document\`, \`Excel spreadsheet\`, \`PowerPoint\`, \`Markdown\`, \`Zip archive\`, etc. — based on the extension.

## Click behavior

**Before:** click → \`window.open\` of signed URL (browser renders PDF/img inline; downloads everything else).

**After:** click → opens the inspector to the artifact detail, where the inline preview lives (PR #271 + #272). Detail's primary action button splits:
- **Files** → "Download" (programmatic \`<a download>\` with the original filename)
- **URLs** → "Open in new tab" (existing behavior)

## How the click finds the artifact

Backend: upload response now surfaces \`_id\`.

Composer bakes it into the directive as a 5th field:
\`\`\`
[[upload:fileName|originalName|size|kind|fileId]]
\`\`\`
4-field directives stay readable (the parser splits on pipe; missing fields stay undefined). The click handler prefers \`onArtifactClick(file-\${fileId})\` when both the prop and the fileId are present; falls back to a signed-URL new-tab otherwise so unmigrated rows degrade gracefully.

\`V2Layout\` passes \`openInspectorArtifact\` through \`V2PodChat\` to \`V2MessageBubble\`, which uncollapses the inspector pane and lands on the artifact's detail view.

## Bonus fix

Bumped \`.v2-msg__file\` selector specificity to \`.v2-root button.X\` so the global v2 button reset (\`.v2-root button:not(.MuiButtonBase-root) { border:none; background:none; padding:0 }\`) doesn't strip the pill's border and background when it renders as a clickable button. Third hit on this pattern this week — see memory \`feedback-v2-button-reset-specificity\`.

## Tests

- \`uploads.post.test.js\` — 21/21 passing. New assertion on \`res.body._id\` in the happy-path test.
- No new tests for the parser (covered indirectly by integration testing on dev).

## Test plan

- [ ] Deploy dev: \`gh workflow run deploy-dev.yml --ref feat/v2-file-pill-ux\`
- [ ] Open Sam's YC Sprint pod, attach a PDF — confirm new pill shape (40px icon, "PDF · 24 KB" subline)
- [ ] Click the pill — confirm inspector opens to the artifact detail with preview, not a new tab
- [ ] In the inspector, click "Download" — confirm browser saves \`brief.pdf\` (not the storage key like \`1714678910-712345678.pdf\`)
- [ ] Click an old smoke-test file (4-field directive, no \`_id\`) — confirm fallback opens in new tab without error
- [ ] Add a Notion URL via "+ Add", click into detail — confirm primary button reads "Open in new tab"

### DIFF
--- a/backend/__tests__/unit/routes/uploads.post.test.js
+++ b/backend/__tests__/unit/routes/uploads.post.test.js
@@ -18,7 +18,13 @@ jest.mock('../../../services/objectStore', () => ({
 
 jest.mock('../../../models/File', () => {
   const saveMock = jest.fn().mockResolvedValue(undefined);
-  const File = jest.fn().mockImplementation((data) => ({ ...data, save: saveMock }));
+  // Mirror the post-save mongoose shape: `_id` populates synchronously
+  // when constructing a new model. Tests use the static fixture below.
+  const File = jest.fn().mockImplementation((data) => ({
+    ...data,
+    _id: { toString: () => 'fake-file-id-67abcdef0123456789abcdef' },
+    save: saveMock,
+  }));
   File.findByFileName = jest.fn();
   File.__saveMock = saveMock;
   return File;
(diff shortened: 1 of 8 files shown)


lilyshen0722:
Closing as stale: opened 2026-05-03, conflicting, and the v2 shell has moved substantially since. The file-pill UX may still be wanted — if so it should be re-filed against current main rather than rebased from here.

PR #282, opened 2026-05-03
fix(provisioner): drop dead trinity, hard-block Codex from community fallback

samxu01:
## Why

Investigating fakesam's session failing on every model in the chain:

| Model | Result |
|---|---|
| `openai-codex/gpt-5.4-mini` | 403 (community agent has no Codex creds) |
| `google/gemini-2.5-flash` | auth_permanent (key revoked) |
| `openrouter/nvidia/nemotron-3-super-120b-a12b:free` | **401 Missing Authentication header** |
| `openrouter/arcee-ai/trinity-large-preview:free` | 404 No endpoints found |

Two root causes:

1. **Trinity is dead at OpenRouter** — every fallback hop into trinity errored.
2. **The OpenRouter provider config is entirely missing from the live gateway PVC** — `models.providers` only contains `openai-codex`, no `openrouter` entry. So when nemotron is requested, the gateway has no provider definition to attach the key to → 401. Cause: `applyOpenClawModelDefaults` writes the OR provider, but it only runs inside `provisionOpenClawAccount`. No reprovision has happened since the code that adds the provider was last updated.

This PR fixes #1 in code. #2 is fixed by running `reprovision-all` after the PR deploys.

## What

- Drop `arcee-ai/trinity-large-preview:free` from every reference in the provisioner:
  - `OPENROUTER_FREE_FALLBACKS` constant
  - `defaults.model.fallbacks` community chain
  - `models.providers.openrouter` model catalog
  - 2 LiteLLM virtual-key allowlists
- Drop trinity from `globalModelConfigService.ts` `communityAgentModel.fallbacks`
- Flip 7 community-agent preset `recommendedModel` entries (presets.ts) from trinity → nemotron
- **New hard assertion**: `applyOpenClawModelDefaults` throws if any `openai-codex/*` model appears in `defaults.model.{primary,fallbacks}`. Cheap defense-in-depth so a future edit can't accidentally re-introduce Codex into the community chain and silently burn quota.

## Behavioral change

**Community agents** when nemotron 429s:
- Before: chain falls through gemini (revoked) → trinity (404) → request errors with multi-model 404 chain
- After: chain falls through gemini placeholders (still inert until key restored) → request fails. Agent goes silent for that heartbeat, retries next interval. **No Codex burn even by accident.**

**Dev agents** (theo/nova/pixel/ops/aria) — unchanged. Still use `openai-codex/gpt-5.4-mini` primary via the per-agent override, and their fallback list still includes nemotron + gemini placeholders.

[...]

PR #283, opened 2026-05-03
test(provisioner): drop trinity from fallback assertion (matches PR #282)

samxu01:
## Why

PR #282 dropped \`arcee-ai/trinity-large-preview:free\` from the community fallback chain (model is dead at OpenRouter, 404 No endpoints found). Test in \`agentProvisionerServiceK8s.test.js\` was still asserting trinity must be present, so it now red-bars on every PR.

This was the only \"Test & Coverage\" failure caused by PR #282. There are two **pre-existing** failures on main I'm flagging separately:
- \`__tests__/unit/controllers/podController.test.js\` — 4 tests broken (populate signature drift, joinPod 403 expectation)
- \`__tests__/services/dmService.test.ts\` — 2 tests broken (agent-room name expectation drifted)

Those are unrelated to the burn fix and need their own investigation.

## What

- Update assertion to require \`gemini-2.5-flash\` present (still inert until GEMINI_API_KEY restored)
- Add explicit \`not.toContain('trinity')\` regression guard
- Preserve the existing openai-codex/* exclusion checks

## Test plan

- [ ] CI \"Test & Coverage\" now passes for the provisioner test (other failures persist; tracked separately)


samxu01:
Squash-merged locally to preserve Sam Xu attribution. Closing.

PR #284, opened 2026-05-03
fix(provisioner): state-sync silently failing — Python NameError on JSON booleans

samxu01:
## Root cause

Every \`syncAccountToStateMoltbot\` call from the JS provisioner has been silently failing since \`2a4bc1314d\` (2026-05-02). Result: the gateway PVC \`/state/moltbot.json\` ignored every config update — including the new \`defaults.model.primary = nemotron\` that should have stopped the Codex burn at the source.

The bug:
\`\`\`ts
\`agents_defaults = \${JSON.stringify(agentsDefaults)}\`,   // emits Python source
\`\`\`
produces lowercase \`true\` / \`false\` / \`null\` — valid JSON, **not valid Python** (Python wants \`True\`/\`False\`/\`None\`). The first embedded boolean (\`memorySearch.enabled: true\`) raised \`NameError: name 'true' is not defined\`, python exited 1, the kubelet returned \"command terminated with non-zero exit code\", and \`execInPod\` rejected without surfacing the python stderr that explained why.

## Why it took so long to find

\`execInPod\` only included the kubelet's generic message in the rejected error. Python's NameError sat in the captured stderr buffer but was thrown away. So the user saw 250+ \"Failed to sync account\" log lines with no clue what python actually did. Hour-long debugging chased every wrong direction (websocket flake, race conditions, set-eu) before noticing the lowercase booleans.

## Fix

1. **Double-encode JSON through \`JSON.stringify\`** — the result is a valid Python string literal, parsed at runtime via \`json.loads\`. Works for booleans, null, nested objects, unicode without manual escaping.
2. **Include captured stderr in \`execInPod\` error messages.** Truncate to last 2KB. Next time a python-in-a-pod script breaks, the actual error will show up in backend logs instead of being silently swallowed.

## Test plan

- [ ] Deploy Dev runs green
- [ ] After deploy: trigger \`reprovision-all\`. Backend logs show \`[k8s-provisioner] synced X to /state/moltbot.json\` for every account, **zero** \"Failed to sync account\" lines.
- [ ] PVC \`/state/moltbot.json\`:
  - \`agents.defaults.model.primary == openrouter/nvidia/nemotron-3-super-120b-a12b:free\` (currently sticky from manual patch — this PR proves it lands automatically)
  - \`models.providers.openrouter\` populated
  - No trinity references anywhere
- [ ] Trigger a community-agent heartbeat (fakesam/liz/x-curator). LiteLLM logs show nemotron call succeeding. Codex weekly counters do not advance.

### DIFF
--- a/backend/services/agentProvisionerServiceK8s.ts
+++ b/backend/services/agentProvisionerServiceK8s.ts
@@ -217,8 +217,18 @@ const execInPod = async ({ podName, containerName = 'clawdbot-gateway', command 
       false,
       (status: any) => {
         const success = status?.status === 'Success' || !status || !status?.status;
-        if (success) resolve();
-        else reject(new Error(status?.message || `Pod exec failed with status: ${status?.status}`));
+        if (success) {
+          resolve();
+        } else {
+          // Include captured stderr in the error message — kubelet's "command
+          // terminated with non-zero exit code" alone is useless for debugging
+          // (lost an hour to a Python NameError that was sitting in stderr the
+          // whole time). Truncate to keep log lines reasonable.
+          const stderrTail = err.trim().slice(-2000);
+          const baseMsg = status?.message || `Pod exec failed with status: ${status?.status}`;
+          const msg = stderrTail ? `${baseMsg}\nstderr: ${stderrTail}` : baseMsg;
+          reject(new Error(msg));
+        }
       },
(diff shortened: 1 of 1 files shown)


samxu01:
Squash-merged locally to preserve Sam Xu attribution. Closing.

PR #287, opened 2026-05-04
docs(adr): ADR-013 agent file production, skill bundles, and v2 install surfaces (Marketplace + inspector tabs)

samxu01:
## Summary

Adds **ADR-013** at `docs/adr/ADR-013-agent-file-production-and-skill-bundles.md` (534 lines, design only — no implementation).

Documents the design for letting OpenClaw agents produce real artifacts (DOCX / XLSX / PPTX / PDF) and surface them in pod chat, **plus** the v2 install surfaces — extending the existing `/v2/marketplace` page with kind tabs (Apps · Agents · Skills · Integrations) and adding per-pod inspector Skills/Agents tabs. The two are coupled: file production is meaningless without a way to install the skills that enable it, and the install UI is the surface where users discover both new skills and new agents.

## Four coordinated changes (design only — no implementation in this PR)

1. **`commonly_attach_file` extension tool** — wraps the upload endpoint and `[[upload:fn|on|sz|kind|fileId]]` directive. Protocol glue, not a skill; lives alongside `commonly_post_message` in the OpenClaw `commonly` extension.

2. **Toolchain in `clawdbot-gateway` Dockerfile** — centered on **OfficeCLI** (`iOfficeAI/OfficeCLI`, Apache-2.0, pinned 30 MB static binary, purpose-built for AI agents) for DOCX/XLSX/PPTX. Plus pandoc for md→PDF and markitdown + pypdf for the parse direction. ~170 MB image growth.

3. **Default skill bundles for the four production dev presets** — Theo / Nova / Pixel / Ops currently ship with `defaultSkills: []`. This ADR fills them in with `github`, `officecli`, `pandic-office`, `markdown-converter`, `pdf` (plus `tmux` for the engineers).

4. **V2 install surfaces** — extend, don't reinvent:
   - **4a.** Extend existing `/v2/marketplace` (`AppsMarketplacePage`, 771 lines) with kind tabs **Apps · Agents · Skills · Integrations**. Wire Agents tab to Randy's Installable-canonical `/api/marketplace/*` (PR #215 + #230). Soft-redirect `/v2/agents/browse` and `/v2/skills` with deep-link query params.
   - **4b.** Inspector **Skills tab** — per-pod contextual install.
   - **4c.** Inspector **Agents tab** — per-pod agent membership; per-agent Talk-to / Configure / Fork; per-browse Install / Fork.
   - **4d.** Shared `<MarketplaceCardList>` subcomponent (~250 lines) extracted FROM the marketplace page in Phase 3 and reused BY the inspector tabs in Phase 4.

Plus a two-line addition to `SOUL.md` ("if you produce a file, attach it via `commonly_attach_file`").

## Key correction landed in v5

After verifying the actual backend + frontend state, two assumptions in earlier drafts were wrong:

1. **`/v2/marketplace` already exists** — renders `AppsMarketplacePage` (v1 component) with `Discover | Installed` sub-tabs. `/v2/agents/browse` and `/v2/skills` also exist as separate v2 routes. v4's "design a new top-level Marketplace page" was over-design; v5 extends the existing page in place.

2. **Randy's `/api/marketplace/*` (PR #215 + #230) is already Installable-canonical** — per #215's description, it's "ADR-001 Phase 2 for marketplace operations — first dual-write path between `Installable` (canonical) and `AgentRegistry` (compat shim)." The marketplace backend is **already on Installable today**; what was missing is frontend wiring. ADR-013's Phase 3 wires against the canonical side from day one and *is* the marketplace-frontend track from ADR-011 by definition for the agent surface.

This re-framed the work substantially: phasing collapsed from 8 active phases to 5, the §"Relationship to Installable taxonomy" rewrote to reflect a per-kind split (agents already on Installable, skills still on AR), and several open questions closed.

## Phasing (5 active + 2 deferred)

| Phase | Scope |
|---|---|
| 0 | `commonly_attach_file` + Dockerfile toolchain |
| 1 | `defaultSkills` bundles for four dev presets |
| 2 | SOUL.md addition + `officecli` SKILL.md packaged locally |
| 3 | Extend `/v2/marketplace` (`AppsMarketplacePage`) with kind tabs; wire Agents tab to Installable; redirect `/v2/agents/browse` + `/v2/skills` |
| 4 | Inspector Skills + Agents tabs (reuse subcomponent from Phase 3) |
| 5 | Delete v1 `AgentsHub.tsx` (4,954 lines) + `SkillsCatalogPage.tsx` (2,061 lines) after redirect cooling-off |
| 6 *(deferred)* | Author + upstream `commonly-pptx`, `commonly-xlsx` |
| 7 *(deferred)* | Publish flow on `/api/marketplace/publish` |

Estimated **~5 PRs** across backend (1–2) + clawdbot (1) + frontend (2) + cleanup (1).

## Open questions flagged

1. Does `POST /api/skills/import` actually push SKILL.md to the gateway PVC, or only register metadata?
2. ~~`github` skill resolution~~ **Resolved (v3)** — real catalog skill, MIT, 603 stars, `gh` already in image.
3. V2 Skills/Agents tab permissions — owner+admin-only for install/uninstall.
4. Image-size budget — ~170 MB add to gateway image.
5. OfficeCLI version-pin cadence.
6. Per-pod vs per-agent skill scoping UX.
7. ~~Accelerate ADR-001 Phase 3?~~ **Re-framed (v5)** — agent read-path switch implicitly done by Randy's dual-write; skills migration deferred until separately motivated. ADR-013's wiring *is* the verification gate for ADR-011's Phase 3 reactivation trigger.
8. ~~Marketplace nav-rail placement~~ **Closed (v5)** — page already mounted.
9. Fork ownership/storage — Randy's `/api/marketplace/fork` + `/api/marketplace/mine` likely cover it; verify response shape covers "My Agents" UX needs.
10. v1 PersonalityBuilder UX — port or fade out.

## Rejected alternatives captured

A) Server-side `/render` service · B) Copy Anthropic's proprietary skills bundle · C) Author duplicate Commonly-specific office skills · D) Per-format specialized sub-agents · E) Skill-only approach (no kernel tool) · F) Build on the pandoc + python-office + LibreOffice + docx-js stack (the v1 draft of this ADR before OfficeCLI discovery)

## Test plan

- [ ] Reviewers confirm the four-part decision is right (file production + bundles + UI)
- [ ] Reviewers confirm the dev preset bundle composition (does Theo need `tmux`? do all four need `officecli`?)
- [ ] Reviewers confirm the OfficeCLI choice over the pandoc/python-office stack
- [ ] Reviewers confirm Part 4a's extend-in-place approach matches the actual `AppsMarketplacePage` shape
- [ ] Reviewers confirm the v5 correction on Randy's `/api/marketplace/*` being Installable-canonical
- [ ] Reviewers confirm the §"Relationship to Installable taxonomy" per-kind split is accurate
- [ ] Reviewers confirm the 5-phase rollout sequence
- [ ] Open questions #1, #5, #9 verified before scoping implementation PRs

## Companion ADRs

[`ADR-001`](docs/adr/ADR-001-installable-taxonomy.md), [`ADR-002`](docs/adr/ADR-002-attachments-and-object-storage.md), [`ADR-008`](docs/adr/ADR-008-agent-environment-primitive.md), [`ADR-011`](docs/adr/ADR-011-shell-first-pre-gtm.md) (this ADR satisfies the "Marketplace frontend" track for the agent surface).


docs/adr/ADR-013-agent-file-production-and-skill-bundles.md
```diff
@@ -0,0 +1,606 @@
...
+
+**Pinning + verification:**
+- Pin `OFFICECLI_VERSION` to a specific release (initial: `1.0.70`). Track upstream's CHANGELOG; bump deliberately.
+- Verify SHA256 from upstream's `SHA256SUMS` artifact — fail the build if the binary changed under us.
+- Add `officecli --version`, `pandoc --version`, `python3 -c "import markitdown, pypdf"` to the gateway healthcheck so a regression (lost binary, broken pip) surfaces before an agent hits `command not found`.
+
+**Why a 7-week-old project.** OfficeCLI is young (created 2026-03-15, 2,768 stars, v1.0.70 released 2026-05-02) — but the alternative is shipping a 600 MB stack that produces inferior output for the same use cases. Risk is real but bounded: the binary is pinned + checksummed; if upstream goes inactive or breaks something, we have an easy escape (downgrade pin, or fall back to the python-office stack as a Phase-N retreat). For now, the quality + size + license combination dominates.
+
```
samxu01, line 173: **Supply-chain risk on OfficeCLI is under-mitigated.** The mitigation paragraph reads "if upstream goes inactive or breaks something, we have an easy escape (downgrade pin, or fall back to the python-office stack as a Phase-N retreat)" — but Part 2 explicitly *removes* the python-office stack from the image. The fallback isn't there to fall back to. A real escape requires keeping the python-office stack staged in a labeled Dockerfile target, or accepting that "downgrade pin" is the only real escape and stating that.

Also, "young + small project hosting a binary we exec inside the gateway" deserves more than a SHA256 pin — a 7-week-old repo can have its release artifact replaced after a maintainer compromise without changing the source tree. Consider:
1. Mirror the pinned binary into our own AR/GCS bucket and download from there in the Dockerfile (full insulation from any upstream artifact swap).
2. Document who watches upstream's release cadence and who notices a checksum mismatch (the Phase-0 verification just says CI fails the build; that's correct but who deals with it at 2am).

---

docs/adr/ADR-013-agent-file-production-and-skill-bundles.md
```diff
@@ -0,0 +1,606 @@
...
+      https://github.com/iOfficeAI/OfficeCLI/releases/download/v${OFFICECLI_VERSION}/officecli-linux-x64 && \
+    echo "${OFFICECLI_SHA256}  /usr/local/bin/officecli" | sha256sum -c - && \
+    chmod +x /usr/local/bin/officecli
+
+# Pandoc + LaTeX engine for md → PDF, plus parse-direction utilities
+RUN apt-get update && apt-get install -y --no-install-recommends \
+      pandoc texlive-xetex texlive-fonts-recommended \
+      poppler-utils python3-pip && \
+    rm -rf /var/lib/apt/lists/*
+
```
samxu01, line 130: `OFFICECLI_SHA256=<lookup-from-SHA256SUMS-at-pin-time>` — has anyone confirmed `iOfficeAI/OfficeCLI` actually publishes a `SHA256SUMS` artifact alongside its releases? Many small projects don't. If they don't, this verification step in Phase 0 is a no-op and we ship "trust the GitHub release tarball" which is exactly the supply-chain footgun the previous comment flags.

Two-line fix to the ADR: add a Phase-0 prerequisite "verify upstream publishes SHA256SUMS for the pinned tag; if not, compute the checksum once locally, commit it as a constant, and update the build-arg lookup to read from the committed value." Either way the answer should not stay as a placeholder.

---

docs/adr/ADR-013-agent-file-production-and-skill-bundles.md
```diff
@@ -0,0 +1,606 @@
...
+8. ~~Top-level Marketplace nav-rail placement~~ **Closed in v5.** `/v2/marketplace` already has its place in the v2 chrome. No new nav-rail entry needed; deep links from `/v2/agents/browse` and `/v2/skills` redirect with the right tab pre-selected.
+
+9. **Fork ownership and storage.** Mostly resolved by Randy's work: `/api/marketplace/fork` is Installable-canonical and `/api/marketplace/mine` returns the calling user's published + forked items. **Verify before Phase 3 ships:** does `/api/marketplace/mine` return enough metadata to render the "My Agents" view (name, version, fork-source link, install count)? If yes, no backend addition needed. If no, small backend extension to widen the response shape — not a new route.
+
```
samxu01, line 555: Open question 11 is **load-bearing for Phase 4** — the entire 4d Configure drawer UX is built on the assumption that per-agent install scoping works against `POST /api/skills/import`. The ADR notes "If the route is pod-scoped only, small backend extension to add a `targetAgentId` parameter — not a new route" — but `syncOpenClawSkills` is described as keyed on `accountId`. If multiple agents share an `accountId` (which the current install pipeline strongly suggests, since "scopes skills to (agentName, instanceId, podId)" is a *projection* of `accountId`), per-agent scoping might require a deeper change than a parameter add — separate skill directories per `(agentName, instanceId)` inside the workspace, plus updates to how the gateway resolves which skills the agent sees.

Recommend pulling this verification into a Phase-0.5 prerequisite (alongside open question #1) instead of Phase 4. If it turns out to be a meaningful backend change, Phase 4's lift target (<300 lines) is wrong and the phasing collapses.

---

docs/adr/ADR-013-agent-file-production-and-skill-bundles.md
```diff
@@ -0,0 +1,606 @@
...
+**Rejected because:** the upstream catalog (`pandic-office`, `pdf`, `ai-pdf-builder`, `markdown-converter`) already covers the use case for PDF and DOCX-from-markdown. Authoring duplicates would be Commonly maintaining a skill we don't own the substrate of. Better split: install community skills for everything they cover, *only* author Commonly skills for genuine gaps (PPTX/XLSX programmatic generation), and ideally upstream those rather than fork.
+
+### D. Per-format specialized sub-agents (one for DOCX, one for PPTX, etc.) with the OpenClaw agent calling them
+
+Considered briefly. Was the user's own straw alternative.
+
+**Rejected because:** it adds an entire layer of orchestration (cross-agent calls, identity per format, separate budgets) for a problem that is just "call a binary." The complexity is unjustified.
+
+### E. Skill-only approach (no `commonly_attach_file` tool — make the upload protocol a skill)
```
samxu01, line 496: Phase 0's verification ("agent runs `pandoc`, calls `commonly_attach_file`, file appears in chat") **assumes Nova has `pandic-office` (or at least raw `pandoc`) installed** — but `defaultSkills` for Nova doesn't ship until Phase 1. Phase 0 only adds the binaries to the image. So the smoke-test as written needs either:
- a hand-imported `pandic-office` skill on Nova before Phase 0 verifies, or
- a manual `pandoc` invocation via `acpx_run` (skips the skill layer entirely),

neither of which is what Phase 0 is actually verifying. Either reorder Phase 0 / Phase 1 to ship together (so the smoke test exercises the real path), or rewrite Phase 0's verification to be skill-free ("agent uses `acpx_run` to invoke pandoc directly, then calls `commonly_attach_file`"). The latter is fine but it under-tests the real loop.

---

docs/adr/ADR-013-agent-file-production-and-skill-bundles.md
```diff
@@ -0,0 +1,606 @@
...
+
+```
+┌─ Configure @theo  Dev PM ──────────────────┐
+│ ● Active · last heartbeat 4m ago           │
+│ Persona  ·  Heartbeat  ·  Skills  ·  Memory │
+├─────────────────────────────────────────────┤
+│ Skills (5)                       [+ Add]    │
+│ ─────────                                   │
+│ ✓ github          steipete · ★ <gh-stars>  │
+│   Interact with GitHub via gh CLI    [×]   │
```
samxu01, line 372: "single click + confirm toast (no modal). The skill markdown gets pulled from the agent's PVC via `syncOpenClawSkills` on the next reprovision; in the meantime the row goes greyed-out with 'removing on next sync.'"

The CLAUDE.md "Agent Runtime — Quick Rules" entry on `reprovision-all` says it takes ~60s and is fire-and-forget. So "next sync" in this UX could mean anywhere from "next time someone clicks the global Reprovision button" to "never in this dev session." That's a fundamentally different UX from what the ✓-with-`[×]` affordance promises (which reads as "instant").

Two ways to make the contract honest:
1. Trigger an immediate per-agent reprovision on `[×]` click (if that's even a thing the provisioner supports today — probably not without backend work).
2. Reword the inline state from "removing on next sync" to "queued — takes effect on next reprovision" with a tooltip explaining what a reprovision is. Less satisfying but truthful.

Without one of these, the friendly UX of 4d will set up exactly the "I clicked X, why is it still here 5 min later?" complaint loop.

---

docs/adr/ADR-013-agent-file-production-and-skill-bundles.md
```diff
@@ -0,0 +1,606 @@
...
+│   research-company · branded reports [+]    │
+└─────────────────────────────────────────────┘
+```
+
+**Friendly UX moves:**
+
+- **One agent at a time** — no multi-row "@nova, @theo, @pixel" cognitive overhead. The drawer shows exactly the skills installed for *this* agent in *this* pod.
+- **Inline `[×]` removal** — single click + confirm toast (no modal). The skill markdown gets pulled from the agent's PVC via `syncOpenClawSkills` on the next reprovision; in the meantime the row goes greyed-out with "removing on next sync."
+- **`[+ Add]` opens a search-first picker** — typeahead over the catalog filtered by what's *not* yet installed for this agent. Picking a skill installs it for THIS agent only, not pod-wide. (Backend: scoped `POST /api/skills/import` with the `agentInstallationId` instead of `podId` — small extension, see open question #11 below.)
+- **"Recommended for Dev PM"** — preset-aware. Each preset declares its `defaultSkills` already (Part 3); when an agent is missing one of its preset's defaults, surface it under Recommended with a one-click install. This makes "fix the agent's missing capabilities" a glanceable affordance rather than a hunt.
```
samxu01, line 401: "all members can see the drawer. Only owner + admin can install/uninstall skills or modify persona. Non-admin viewers see grey-out states with explanatory tooltip, not hidden controls — discoverability matters more than chrome."

This is a permissions decision worth surfacing in invariants, not just inline. A non-admin member opening the Configure drawer sees the agent's full IDENTITY.md / persona / heartbeat schedule + memory tab read-only. Some persona content can include private operator notes (e.g. who's allowed to invoke certain tools, internal "be skeptical of X" rules) that the operator may not want every pod member to see.

Two mitigations to consider before merging this section:
1. Make the *entire* drawer admin-only by default; flip individual fields to "visible to members" via an explicit `members.canRead = true` opt-in per-field.
2. Carve memory tab out specifically — recent `AgentMemory` entries can include cross-pod context the agent learned in DMs, which leaking to a pod member is an ADR-003 §Visibility violation.

At minimum, add an open question: "What in the Configure drawer is reader-safe vs. operator-private?" Currently the ADR assumes everything except install actions is fine to expose, which doesn't match how IDENTITY.md is used in practice (e.g. nova/theo's persona files in this repo).

---

docs/adr/ADR-013-agent-file-production-and-skill-bundles.md
```diff
@@ -0,0 +1,606 @@
...
+│  ┌─ pandic-office ───────┐  ┌─ markdown-converter ──┐  ...          │
+└────────────────────────────────────────────────────────────────────┘
+```
+
+**What renders on each card:**
+- **★ N** — upstream GitHub stars from the source repo. For catalog skills (`github`, `pandic-office`, etc.) read from `awesome-agent-skills-index.json`'s `stars` field (catalog-snapshot value; freshness governed by the sync cron, see §Caveat). For locally-bundled skills like `officecli`, fetch live from the source repo's GitHub API once per page load and cache for the session.
+- **N installs** — count of `AgentInstallation` rows referencing this skill across the user's accessible pods (or instance-wide for admins). Cheap query. Skip rendering if the count would require a cross-instance aggregate or hits a privacy boundary; default to omitting rather than computing.
+- **Author + license** — small muted line above the metrics. Author from manifest, license from the catalog row.
+
+**What does NOT render:**
```
samxu01, line 251: "For locally-bundled skills like `officecli`, fetch live from the source repo's GitHub API once per page load and cache for the session."

Two issues with the live-fetch path:
1. **Rate limit.** Unauth GitHub API is 60 req/hr per IP. Even with one repo, "once per page load" with 10 admins navigating the marketplace burns the budget fast. Authed = 5000/hr but requires a service-side token + proxy (don't ship a token to the browser).
2. **Failure mode unspecified.** What renders when the fetch fails — `★ ?`, hide the badge, fall back to a stale value? The mockup just shows `★ <gh-stars>` which doesn't answer.

Recommend: server-side proxy that periodically refreshes a small cache (cron or on-demand TTL) for the small list of locally-bundled skills (today: just `officecli`). Add it to Phase 0 or Phase 3 plumbing rather than client-side direct-to-GitHub fetches. Also makes the catalog vs live distinction less awkward — both eventually flow through one server endpoint.

---

docs/adr/ADR-013-agent-file-production-and-skill-bundles.md
```diff
@@ -0,0 +1,606 @@
...
+
+### E. Skill-only approach (no `commonly_attach_file` tool — make the upload protocol a skill)
+
+The shape: ship a Commonly-authored `commonly-attach-file` skill into every preset's `defaultSkills`.
+
+**Rejected because:** the upload-and-directive pattern is protocol knowledge the kernel exposes, in the same shape as posting messages or creating pods. Forcing it into a skill is the wrong layer — a skill that fails to install would mean the agent can't attach files, which is too fragile for a kernel verb. Tools are always present; skills are opt-in. File attach belongs in the "always present" set.
+
+### F. Build the file-generation toolchain on `pandoc` + `python-docx` + `python-pptx` + `openpyxl` + `reportlab` + `docx-js` + `libreoffice` (the stack Anthropic's official skills use)
+
+Initial draft of this ADR (v1, before discovering OfficeCLI on 2026-05-03) proposed exactly this — a ~600 MB toolchain layered into the gateway image, with skills like `pandic-office`, `pdf`, plus Commonly-authored DOCX/PPTX/XLSX wrappers calling the python libs.
```
samxu01, line 504: "after Phase 3's redirects have been live ≥2 weeks with no fallback complaints"

There's no concrete listening mechanism for "fallback complaints" — `/v2/agents/browse` and `/v2/skills` are quietly redirected with a banner; users either notice or don't. In practice this becomes "delete when someone remembers."

Make it actionable by adding a server-side counter on the redirect path that logs hits, and gate Phase 5 on "≥2 weeks AND <N hits/week on the redirect path." That gives a concrete signal instead of a vibe-check. ~5 lines of telemetry in Phase 3.

---

samxu01 (commented):
## ADR-013 review

Strong design — the v5 correction (recognizing `/v2/marketplace` and `/api/marketplace/*` already exist and Randy's work is Installable-canonical) is exactly the kind of "verify the actual state before designing" that saves a phase of rework. The four-coordinated-changes shape is right; the rejected-alternatives section honestly captures why each path was abandoned; phasing is reasonable for an ADR-only PR.

### Things to address before this turns into implementation PRs
- **OfficeCLI supply-chain** (line ~155): the "fall back to python-office stack" escape doesn't exist if Part 2 removes that stack from the image. Either keep the python-office stack staged in a Dockerfile target, or rewrite the mitigation. Also, mirroring the pinned binary into our own AR/GCS bucket would insulate us from upstream artifact swap on a 7-week-old project.
- **`OFFICECLI_SHA256` placeholder** (line ~130): confirm upstream actually publishes `SHA256SUMS` before merging. If they don't, the Phase-0 verification is a no-op.
- **Open question 11 is load-bearing for Phase 4**, not Phase-4-time work: the per-agent skill scoping in `POST /api/skills/import` may require a deeper change than a parameter add given `syncOpenClawSkills` is keyed on `accountId`. Pull the verification into Phase 0/0.5.
- **Phase 0 verification path doesn't match what Phase 0 ships** (line ~461): toolchain ships in Phase 0, skills in Phase 1 — so the smoke test as written needs a manual hand-installed skill on Nova or a skill-free pandoc invocation.

### Worth resolving
- **Inline `[×]` UX vs reprovision-all** (line ~348): "removing on next sync" can mean ~60s or never. Either trigger per-agent reprovision on click, or word the inline state honestly.
- **Configure-drawer permissions** (line ~373): "all members read, admins write" leaks IDENTITY.md and memory contents to non-admins. Worth an explicit "what's reader-safe vs. operator-private" pass before Phase 4.
- **Live GitHub stars on locally-bundled skills** (line ~227): client-side direct-to-GitHub fetch will hit rate limits + has no specified failure mode. Server-side proxy with TTL cache is one endpoint to add in Phase 0/3.
- **Phase 5 deletion gate** (line ~469): "no fallback complaints" needs a concrete listening signal — add a redirect-hit counter so the deletion has a numeric trigger.

### Smaller notes
- The "~5 PRs" estimate likely undersells if open question 11 turns out to need a real backend change (Phase 4 then needs a backend PR pair).
- Invariant 4 ("`commonly_attach_file` reuses the workspace-boundary helper from `acpx_run`") assumes that helper is cleanly extractable. Worth a Phase-0 spike to confirm.

This is `COMMENT`, not `REQUEST_CHANGES` — none of these are merge-blockers for an ADR. The design is fundamentally right; the comments are about pinning down the assumptions before they become silent obstacles in the implementation PRs.

---

samxu01:
## v5.4 — addressing review feedback

Pushed `5405c8a` with point-by-point responses to all 8 inline comments + the PR-level summary. Net: +58 lines / −14 lines, ADR now 650 lines.

| # | Inline comment (line) | Resolution in v5.4 |
|---|---|---|
| 1 | **OfficeCLI supply-chain** (155) — "fall back to python-office" was a lie | Rewrote §Pinning + verification with three honest escape paths (downgrade pin → staged Dockerfile target → full re-author), ranked by cost. Added explicit binary-mirror requirement (AR/GCS) to insulate from upstream artifact swap. Documented operational ownership for checksum-mismatch incidents. |
| 2 | **SHA256SUMS placeholder** (130) — has anyone confirmed upstream publishes it? | **Unverified, escalated to Phase 0a (i) prereq.** Two paths if upstream doesn't: compute checksum once locally + commit as constant. Placeholder cannot ship. New open question #13. |
| 3 | **Open question 11 load-bearing for Phase 4** (516) — `syncOpenClawSkills` keyed on `accountId` may need deeper change | **Escalated to Phase 0a (ii) prereq.** If verification lands as "needs backend change," Phase 4 splits into 4a (frontend) + 4b (backend). PR estimate widened from ~5 to 5–7 to reflect this. |
| 4 | **Phase 0 verification mismatch** (461) — smoke test assumes pandic-office installed | Renumbered to **Phase 0b** with skill-free verification (`acpx_run` invoking `pandoc` directly). Full skill-loop verification moves to end of Phase 1, where `defaultSkills` actually ships. |
| 5 | **Inline `[×]` UX vs reprovision-all** (348) — "next sync" can mean ~60s or never | Committed to **per-agent reprovision endpoint as the target** with "Queued — applies on next reprovision" wording as acceptable v1. Eliminated misleading "removing on next sync" copy. |
| 6 | **Configure drawer permissions leak** (373) — IDENTITY.md / memory leak to non-admins | **Flipped default to admin-only.** Per-field "visible to members" is explicit opt-in; Memory tab is never member-readable in v1 (ADR-003 §Visibility). Added as **Invariant 7**. |
| 7 | **Live GitHub stars rate-limit + failure mode** (227) — client-side direct fetch | Moved to **server-side proxy endpoint** (`GET /api/marketplace/stars/:source/:owner/:repo`) with TTL cache, refreshed by cron, authed token never in browser. Failure mode: `★ —` with tooltip on cache miss. Added to Phase 3 scope. |
| 8 | **Phase 5 deletion gate** (469) — "no fallback complaints" is a vibe check | Replaced with concrete `<N hits/week` telemetry threshold on the legacy redirect paths. ~5 lines of telemetry added to Phase 3 scope as the deletion-gate signal. |

Plus structural changes:

- **Phase 0** split into **0a (prereqs)** + **0b (toolchain ship)**. Phase 0a is a non-optional spike covering the three load-bearing assumptions (i/ii/iii). Added as **Invariant 8** so PRs can't ship 0b code while 0a is in placeholder state.
- **Open question #12** added on workspace-boundary helper extractability (resolves in Phase 0a (iii)).
- **Open question #13** added on SHA256SUMS publication (resolves in Phase 0a (i)).
- **PR estimate** widened from `~5 PRs` to `5–7 PRs` reflecting Phase 0a uncertainty.

The two smaller PR-level notes:

- **"Invariant 4 reuses workspace-boundary helper from `acpx_run`"** — promoted to **Phase 0a (iii)** as a 30-min spike. If the helper isn't cleanly exportable, extraction-first is the path.
- **"~5 PRs estimate undersells if Q11 needs a backend change"** — addressed by widening to 5–7 with the Phase 4 split contingency stated explicitly.

None of these changed the overall design shape — all of v5's structural decisions hold (extend `AppsMarketplacePage` in place, agents-side already on Installable, skills-side stays on AR until Phase 3 unpauses). The corrections are all about pinning down assumptions before they become silent obstacles in the implementation PRs, which is exactly what the review asked for.

Ready for re-review.

PR #288, opened 2026-05-04
feat(memory): ADR-012 Phase 1 — system_exchanges + DM conversational frame

samxu01:
## Summary

Implements **ADR-012 Phase 1**: system-driven exchange records as a typed section on the AgentMemory envelope, three platform-side triggers that write to it, read-only enforcement at the agent tool surface, and (folded in via §9 after the live a2a smoke test) an inline **DM conversational frame** on chat.mention payloads.

### Schema (§1, §5)
- `AgentMemoryEnvelope.sections.system_exchanges` — typed sub-document, hard-coded `visibility: 'private'`.
- `AgentMemory.revision` + `lastSeenRevision` (default 0) — monotone counter for Phase 2's memoryDigest.
- Caps as constants: `SYSTEM_EXCHANGE_TAKEAWAY_MAX = 280`, `SYSTEM_EXCHANGE_ENTRY_CAP = 50`.

### Service (§1, §4, §8 invariant 8a)
- `appendSystemExchange()` — single atomic `$push + $position:0 + $slice:50 + $inc:revision + $currentDate`. Carve-out: does NOT clear `lastSyncKey`/`lastSyncAt`.
- `truncateTakeaway()` — head-truncates with `…` suffix, schema-cap-safe.
- `isAgentWritableSection()` — allow-list for the agent tool surface.

### Triggers (§4) — `backend/services/systemExchangeTriggers.ts`
- `recordAgentDmConclusion` — fires on `NO_REPLY` in agent-dm. Writes to BOTH peers; takeaway = sender's preceding non-NO_REPLY message, head-truncated.
- `recordAgentDmLoopTrip` — fires when bot-loop guard returns. Both peers get a literal entry.
- `recordTaskCompleted` — fires on tasks complete. Assignee only; takeaway = `<title> → <prUrl|status>`.

All three are fire-and-forget, errors logged not propagated.

### Read-only enforcement (§1)
PUT `/memory` and POST `/memory/sync` return **403** with `reason: 'system_exchanges_is_read_only'` when the section name appears in the payload.

### §9 DM conversational frame
Smoke test (FakeSam ↔ Tarik, 2026-05-04) showed Tarik composing broadcast-style \"has anyone seen…\" replies in a 1:1 DM. The platform shipped `dmKind` as a structured payload field but the LLM deprioritized it. **`enqueueDmEvent` now prepends an inline narrative cue to `payload.content`** based on `dmKind`:

- `agent-agent`: \"[1:1 agent-DM with @\<peer\> (\<display\>) — talk directly... return NO_REPLY when the exchange reaches a natural conclusion. Surface shareable results to a team pod.]\"
- `user-agent`: \"[1:1 DM with @\<peer\> (\<display\>, human) — they are asking you directly. Reply to every new message.]\"

Peer label uses the same fallback chain as `resolveAgentDisplayLabel` so recipients see \"@FakeSam (FakeSam)\" not \"@fakesam (openclaw)\". No runtime-extension change needed — every CAP runtime reads `event.payload.content` already.

## What's deferred (per the ADR)

- **Phase 2** — `AgentEvent.memoryRevisionAtDelivery`, fetch-time atomic claim, idempotent ack, memoryDigest computation.
- **Phase 3** — `PLATFORM_SOUL_FOOTER` + `AGENT_RUNTIME.md` updates.
- **Phase 4** — Per-runtime adoption of memoryDigest reads (openclaw extension, webhook SDK, claude-code shim). Parallelizable.

## Test plan

- [x] Unit tests pass: `agentMemoryService.systemExchanges.test.ts` (13 tests — pure helpers, append ordering, cap eviction at 50, peer-isolation, invariant 8a, schema rejection of non-private visibility + unknown kinds, default revisions)
- [x] Integration tests pass: `agent-memory-envelope.test.js` (4 new tests — 403 on PUT/sync, GET surfaces entries, sibling-section preservation)
- [x] Live a2a smoke (FakeSam ↔ Tarik) — autonomous round-trip captured BEFORE this PR; the §9 frame is folded in to address the broadcast-voice issue surfaced there
- [ ] After deploy: re-run a2a smoke, verify the DM cue lands in agent context (visible in gateway debug log on inbound message)
- [ ] Phase 2 PR follow-up

### DIFF
--- a/backend/models/AgentMemory.ts
+++ b/backend/models/AgentMemory.ts
@@ -47,7 +47,10 @@ export interface ISystemExchangesSection {
   entries: ISystemExchangeEntry[];   // most recent first; cap at 50 (ADR-012 §5)
   visibility: 'private';             // hard-coded; widening is closed
   updatedAt: Date;
-  byteSize: number;
+  // ADR-012 Phase 1 deliberately omits `byteSize` — appendSystemExchange does
+  // not maintain it under concurrent $push, and no Phase 1 reader consumes
+  // it. Phase 2's digest builder recomputes from `entries` directly. Re-add
+  // here when there's a reader that benefits from the cached value.
 }
 
 export interface IAgentMemorySections {
@@ -176,7 +179,7 @@ const systemExchangesSectionSchema = new Schema<ISystemExchangesSection>(
     // attempt to widen at the document level fails Mongoose validation.
     visibility: { type: String, enum: ['private'], default: 'private' },
     updatedAt: { type: Date, default: Date.now },
-    byteSize: { type: Number, default: 0 },
+    // byteSize intentionally omitted in Phase 1 — see ISystemExchangesSection.
   },
   { _id: false },
 );
--- a/backend/models/AgentMemory.ts
+++ b/backend/models/AgentMemory.ts
@@ -26,6 +26,30 @@ export interface IRelationshipNote {
   updatedAt: Date;
 }
 
+// ADR-012 §1: structured entries for system-driven exchange records.
+// Visibility is hard-coded 'private' at the schema level (see ADR-012 §6).
+export type SystemExchangeKind =
+  | 'agent-dm-conclusion'
+  | 'agent-dm-loop-trip'
+  | 'task-completed'
+  | 'cross-pod-mention'; // reserved for v1.x — not emitted in v1 (ADR-012 §4)
+
+export interface ISystemExchangeEntry {
+  ts: Date;                          // when the event happened
+  kind: SystemExchangeKind;
+  surfacePodId: string;              // where the event happened
+  surfaceLabel: string;              // human-readable, e.g. "agent-dm:69f7..." or "team:Backend Tasks"
+  peers: string[];                   // other instanceIds involved (excluding self)
+  takeaway: string;                  // ≤ 280 chars; verbatim metadata in v1
+}
+
+export interface ISystemExchangesSection {
+  entries: ISystemExchangeEntry[];   // most recent first; cap at 50 (ADR-012 §5)
+  visibility: 'private';             // hard-coded; widening is closed
+  updatedAt: Date;
+  byteSize: number;
+}
+
 export interface IAgentMemorySections {
   soul?: IMemorySection;
   long_term?: IMemorySection;
@@ -34,6 +58,9 @@ export interface IAgentMemorySections {
   relationships?: IRelationshipNote[];
   shared?: IMemorySection;
   runtime_meta?: IMemorySection;
+  // ADR-012: system-driven exchange records. Read-only from the agent's
+  // tool surface; written only by platform hooks (agentMemoryService.appendSystemExchange).
+  system_exchanges?: ISystemExchangesSection;
 }
 
 export interface IAgentMemory extends Document {
@@ -48,11 +75,43 @@ export interface IAgentMemory extends Document {
   // within the same day bucket return early with { deduped: true }.
   lastSyncKey?: string;
   lastSyncAt?: Date;
+  // ADR-012: monotone revision bumped on every system_exchanges write.
+  // memoryDigest in CAP event payload is computed against `lastSeenRevision`.
+  revision?: number;
+  lastSeenRevision?: number;
   createdAt: Date;
   updatedAt: Date;
 }
 
 export const VISIBILITY_VALUES: MemoryVisibility[] = ['private', 'pod', 'public'];
+
+// ADR-012 §4: trigger taxonomy. v1 emits the first three; cross-pod-mention
+// is reserved for v1.x consideration.
+export const SYSTEM_EXCHANGE_KINDS: SystemExchangeKind[] = [
+  'agent-dm-conclusion',
+  'agent-dm-loop-trip',
+  'task-completed',
+  'cross-pod-mention',
+];
+
+// ADR-012 §1: ≤280 chars, enforced server-side (truncation responsibility
+// lives in appendSystemExchange callers — schema validates the post-truncation
+// invariant). The cap also bounds the digest serialization budget.
+export const SYSTEM_EXCHANGE_TAKEAWAY_MAX = 280;
+// ADR-012 §5: count-bounded eviction. Oldest entry evicted on overflow.
+export const SYSTEM_EXCHANGE_ENTRY_CAP = 50;
+// ADR-012 §1 — writable sections (the agent's tool can address these via
+// commonly_save_my_memory). `system_exchanges` is intentionally excluded.
+export const AGENT_WRITABLE_SECTIONS = [
+  'soul',
+  'long_term',
+  'dedup_state',
+  'shared',
+  'runtime_meta',
+  'daily',
+  'relationships',
+] as const;
+export type AgentWritableSection = typeof AGENT_WRITABLE_SECTIONS[number];
 
 // Phase 1 invariant: all section sub-fields use `default: undefined` so a newly
 // created envelope does NOT auto-insert empty sub-documents. Callers opt in to
@@ -89,6 +148,39 @@ const relationshipNoteSchema = new Schema<IRelationshipNote>(
   { _id: false },
 );
 
+const systemExchangeEntrySchema = new Schema<ISystemExchangeEntry>(
+  {
+    ts: { type: Date, required: true },
+    kind: { type: String, enum: SYSTEM_EXCHANGE_KINDS, required: true },
+    surfacePodId: { type: String, required: true },
+    surfaceLabel: { type: String, default: '' },
+    peers: { type: [String], default: [] },
+    takeaway: {
+      type: String,
+      default: '',
+      // ADR-012 §1: 280-char cap. Callers truncate with `…` suffix; the
+      // schema guards against bypass. Validation runs on $push via runValidators.
+      validate: {
+        validator: (v: string) => typeof v === 'string' && v.length <= SYSTEM_EXCHANGE_TAKEAWAY_MAX,
+        message: `takeaway must be ≤ ${SYSTEM_EXCHANGE_TAKEAWAY_MAX} chars`,
+      },
+    },
+  },
+  { _id: false },
+);
+
+const systemExchangesSectionSchema = new Schema<ISystemExchangesSection>(
+  {
+    entries: { type: [systemExchangeEntrySchema], default: [] },
+    // ADR-012 §6: hard-coded 'private'. The enum is single-valued so any
+    // attempt to widen at the document level fails Mongoose validation.
+    visibility: { type: String, enum: ['private'], default: 'private' },
+    updatedAt: { type: Date, default: Date.now },
+    byteSize: { type: Number, default: 0 },
+  },
+  { _id: false },
+);
+
 const agentMemorySectionsSchema = new Schema<IAgentMemorySections>(
   {
     soul: { type: memorySectionSchema, default: undefined },
@@ -98,6 +190,7 @@ const agentMemorySectionsSchema = new Schema<IAgentMemorySections>(
     relationships: { type: [relationshipNoteSchema], default: undefined },
     shared: { type: memorySectionSchema, default: undefined },
     runtime_meta: { type: memorySectionSchema, default: undefined },
+    system_exchanges: { type: systemExchangesSectionSchema, default: undefined },
   },
   { _id: false },
 );
@@ -112,6 +205,11 @@ const agentMemorySchema = new Schema<IAgentMemory>(
     schemaVersion: { type: Number, default: undefined },
     lastSyncKey: { type: String, default: undefined },
     lastSyncAt: { type: Date, default: undefined },
+    // ADR-012 §2: monotone revision; bumped on every appendSystemExchange.
+    // lastSeenRevision is bumped when an event carrying memoryDigest is acked
+    // (Phase 2). Both default to 0 so a fresh agent's first digest is empty.
+    revision: { type: Number, default: 0 },
+    lastSeenRevision: { type: Number, default: 0 },
   },
   { timestamps: true },
 );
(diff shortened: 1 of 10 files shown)


backend/services/agentMessageService.ts
```diff
@@ -781,6 +781,16 @@ class AgentMessageService {
     }
     const sanitizedContent = AgentMessageService.sanitizeAgentContent(content);
     if (!sanitizedContent) {
+      // ADR-012 §4: agent-dm-conclusion trigger. The reply is silent
+      // (NO_REPLY swallowed by sanitizeAgentContent). If the pod is an
+      // agent-dm, both peers get a system_exchanges entry whose takeaway is
+      // the SENDER's preceding non-NO_REPLY message. Fire-and-forget — never
+      // delays the silent return; failures are swallowed inside the helper.
+      void AgentMessageService.maybeRecordAgentDmConclusion({
```
samxu01, line 789: This trigger fires on **every** sanitize-empty `postMessage` call, not just agent-dms — `recordAgentDmConclusion` only checks `podType === 'agent-dm'` after a `Pod.findById` + `User.find` round-trip. For team pods / agent-room / agent-admin where bots commonly emit pure NO_REPLY heartbeats, that's two extra Mongo reads per silent post on a hot path. Worth gating with a cheap `pod.type` check (or passing `pod.type` from the caller — at this point in `postMessage` we haven't loaded the pod yet, but we *do* know the route called `agentRuntime.postMessage` after `agentRuntimeAuth`, and we could attach pod-type during a single shared lookup later in this function and short-circuit the trigger when not agent-dm).

At minimum, add a quick `Pod.findById(podId).select('type')` projection in the trigger before doing the full member resolution.

---

backend/services/systemExchangeTriggers.ts
```diff
@@ -0,0 +1,322 @@
...
+    for (const m of recent) {
+      const uid = String(m?.user_id ?? '');
+      if (uid !== senderUserId) continue;
+      const raw = typeof m?.content === 'string' ? m.content : String(m?.content ?? '');
+      const trimmed = raw.trim();
+      if (!trimmed) continue;
+      // Skip pure NO_REPLY rows (matches sanitizeAgentContent's logic without
+      // round-tripping the regex). A row that's NO_REPLY-only collapses to
+      // empty after stripping the token.
+      const stripped = trimmed.replace(/\bNO_REPLY\b/g, '').trim();
```
samxu01: Drift risk: this re-implements `AgentMessageService.sanitizeAgentContent`'s NO_REPLY stripping (`/\bNO_REPLY\b/g`) inline. If sanitize logic ever changes (e.g. adds a NO_RESPONSE alias, handles fenced-NO_REPLY, etc.) this path silently disagrees and starts surfacing non-substantive content as takeaways.

Either import and call `AgentMessageService.sanitizeAgentContent` here, or extract a shared `isPureNoReply(s: string): boolean` helper. The comment ("matches sanitizeAgentContent's logic without round-tripping the regex") acknowledges the duplication — making it a referenced helper costs one import and removes the future-drift footgun.

---

backend/services/systemExchangeTriggers.ts
```diff
@@ -0,0 +1,322 @@
...
+  try {
+    // eslint-disable-next-line global-require, @typescript-eslint/no-require-imports
+    const PGMessage = require('../models/pg/Message') as {
+      findByPodId?: (id: string, limit: number) => Promise<Array<{ user_id?: unknown; content?: unknown }>>;
+    };
+    if (!PGMessage || typeof PGMessage.findByPodId !== 'function') return null;
+    // Limit: 30 — the NO_REPLY post itself is typically the last row, and we
+    // want the last substantive turn from the sender. 30 covers ~10 back-and-
+    // forths without scanning the whole pod.
+    const recent = await PGMessage.findByPodId(podId, 30);
```
samxu01: `limit: 30` is fine in a quiet 1:1 DM, but agent-DMs can interleave with other agents' fan-out (e.g. a `commonly_open_dm` initiated mid-stream, the speaker @-mentioned in the same pod, etc.) and 10–15 turns of back-and-forth where the silent agent's last substantive message is older than 30 rows back. When that happens we silently fall through to the `'agent-dm concluded (no prior content captured)'` literal — losing the only data ADR-012 §4 promised to capture.

Two cheap improvements:
1. Filter the PG query by `user_id = senderUserId` instead of fetching 30 rows and post-filtering — that gets you the last N from *this* sender directly.
2. If you keep the current shape, bump the limit to ~100 or paginate until a match is found.

---

backend/services/systemExchangeTriggers.ts
```diff
@@ -0,0 +1,322 @@
...
+      const peers = agents
+        .filter((p) => !(p.agentName === a.agentName && p.instanceId === a.instanceId))
+        .map((p) => p.instanceId);
+      // From sender's perspective: takeaway describes what THEY just said and concluded.
+      // From recipient's perspective: takeaway describes what the other agent said
+      // before going silent. v1 records the same takeaway both ways — the kind
+      // ('agent-dm-conclusion') + peers list disambiguate role on read.
+      return appendSystemExchange({
+        agentName: a.agentName,
+        instanceId: a.instanceId,
```
samxu01, line 236: Recording the *same* takeaway in both peers' envelopes is going to read confusingly later. From the listener's POV, an entry of kind `agent-dm-conclusion` whose takeaway is "shipped the auth patch" looks like *they* said it; the comment acknowledges this depends on the reader correlating `peers[]` to figure out role.

Two lower-cost fixes for v1 that don't require multi-turn condensation:
1. Prefix the listener's takeaway with `"@<peer> said: "` (peer resolved via the `peers[]` you already compute) so role is in-band.
2. Or split into two kinds (`agent-dm-conclusion-spoken` / `agent-dm-conclusion-heard`) and let the digest serializer render them differently.

Either avoids a future "why does my memory say I shipped X when I didn't?" bug when an agent regresses to reading takeaway as first-person.

---

backend/services/agentMentionService.ts
```diff
@@ -707,6 +724,30 @@ const enqueueDmEvent = async ({
...
+    // as agentIdentityService.resolveAgentDisplayLabel (botMetadata
+    // .displayName → identity-bearing instanceId → username), so the
+    // recipient sees "@FakeSam (FakeSam)" not "@fakesam (openclaw)".
+    const senderMeta = sender?.botMetadata || {};
+    const senderDisplay = (senderMeta.displayName?.trim()
+      || (senderMeta.instanceId && senderMeta.instanceId !== 'default' ? senderMeta.instanceId : '')
+      || sender?.username
+      || username
+      || 'peer').trim();
+    const senderHandle = senderMeta.instanceId?.trim() || sender?.username || username || 'peer';
```
samxu01: `senderHandle` doesn't filter `instanceId === 'default'`, so for any agent whose `botMetadata.instanceId` is the literal `'default'` (which is the convention everywhere else in the codebase for unclaimed identity) the cue renders as `@default (FakeSam)` — a meaningless handle.

The `senderDisplay` chain three lines above already filters `instanceId !== 'default'`; mirror the same logic here:

```ts
const senderHandle = (senderMeta.instanceId && senderMeta.instanceId !== 'default'
  ? senderMeta.instanceId
  : sender?.username || username || 'peer').trim();
```

Also worth pulling this and `senderDisplay` through `agentIdentityService.resolveAgentDisplayLabel` (which the CLAUDE.md DM display-label rule mandates as the single source of truth) rather than re-implementing the chain inline — would prevent drift if the resolution rules change again.

---

backend/models/AgentMemory.ts
```diff
@@ -26,6 +26,30 @@ export interface IRelationshipNote {
...
+  surfaceLabel: string;              // human-readable, e.g. "agent-dm:69f7..." or "team:Backend Tasks"
+  peers: string[];                   // other instanceIds involved (excluding self)
+  takeaway: string;                  // ≤ 280 chars; verbatim metadata in v1
+}
+
+export interface ISystemExchangesSection {
+  entries: ISystemExchangeEntry[];   // most recent first; cap at 50 (ADR-012 §5)
+  visibility: 'private';             // hard-coded; widening is closed
+  updatedAt: Date;
+  byteSize: number;
```
samxu01: `byteSize` is declared on the section + schema but never written to (`appendSystemExchange` explicitly calls out that it leaves it at 0 and "recompute[s] lazily when the digest builder serializes"). Right now nothing reads it either — it's a permanently-zero column.

Either drop the field for Phase 1 and re-add it when the digest builder lands in Phase 2 (cleaner — no risk of a reader trusting it), or wire `appendSystemExchange` to maintain it via `$set` (the cost objection in the comment is overstated for a single-doc update with the new entry's length already in hand). Carrying a field that's permanently lying-by-omission invites a Phase 2 author to read it without recomputing.

---

backend/services/agentMentionService.ts
```diff
@@ -707,6 +724,30 @@ const enqueueDmEvent = async ({
...
+    const dmFrame = dmKind === 'agent-agent'
+      ? `[1:1 agent-DM with @${senderHandle} (${senderDisplay}) — talk directly to them, not a broadcast room. Reply only when your message materially advances the work; return NO_REPLY when the exchange reaches a natural conclusion. Surface anything shareable to a team pod via commonly_post_message there.]`
+      : `[1:1 DM with @${senderHandle} (${senderDisplay}, human) — they are asking you directly. Reply to every new message; responsiveness matters even when there's little to add.]`;
+    const framedContent = `${dmFrame}\n\n${content}`;
```
samxu01, line 766: A subtle persistence asymmetry: the `chat.mention` event payload now carries the framed content, but the message persisted to PG (and what humans see in the chat UI, what `findPreviousNonSilentMessage` will later read for `agent-dm-conclusion` takeaway derivation) is the un-framed `content`. Good — that's the right call.

But this means the cue is **only** delivered through this one event-enqueue path. If a different driver later picks up the same DM via a different surface (e.g. a webhook re-delivery, a re-enqueue after a reconnect, a poll endpoint), it'll see un-framed content and fall back to the structured `dmKind` field — which §9 of ADR-012 explicitly says was insufficient. Worth a comment here noting the cue is event-only and any future polling/re-delivery surface needs to re-apply it (or moving the framing into `AgentEventService.enqueue` itself so it's a single chokepoint).

---

backend/services/systemExchangeTriggers.ts
```diff
@@ -0,0 +1,322 @@
...
+    .lean<Array<{ username?: string; botMetadata?: { agentName?: string; instanceId?: string } }>>();
+
+  const agents = bots
+    .map((u) => {
+      const agentName = u.botMetadata?.agentName || u.username || '';
+      const instanceId = u.botMetadata?.instanceId || 'default';
+      return { agentName: String(agentName).toLowerCase(), instanceId };
+    })
+    .filter((a) => a.agentName);
+
```
samxu01, line 87: `agentName: u.botMetadata?.agentName || u.username || ''` — for OpenClaw-driven agents, the CLAUDE.md DM-label rule explicitly warns that `botMetadata.agentName` is the **runtime** (`'openclaw'`), not the identity. Using it here means two openclaw agents in an agent-dm both resolve to `agentName: 'openclaw'`, which the `senderKey` equality test in `recordAgentDmConclusion` will then collapse into a single member and short-circuit the "sender isn't a recognized agent member" guard incorrectly. Same risk in `recordTaskCompleted` where the assignee is matched against `a.agentName`.

If the intent is "stable identity per agent in this pod," prefer `instanceId` (or the resolveAgentDisplayLabel chain) as the discriminator and either drop `agentName` or qualify it with `instanceId`. The current shape works in practice today only because each openclaw instance happens to have a unique `instanceId`, and the equality test happens to use the `agentName|instanceId` tuple — but that's an accident worth pinning down with a comment if not changing the code.

---

samxu01 (commented):
## ADR-012 Phase 1 — review

Solid Phase 1 land — the schema-level read-only enforcement, the atomic `$push + $position:0 + $slice + $inc` shape, the carve-out from invariant 8 (preserving `lastSyncKey`/`lastSyncAt`), and the test coverage on cap eviction + revision monotonicity all match the ADR. The 403-with-tagged-reason for `system_exchanges` writes is a nice touch over a generic 400.

### Things to address before merge
- **Hot-path cost on `postMessage` early-return** (agentMessageService.ts:789): the trigger does `Pod.findById` + `User.find` for *every* sanitize-empty post, including the common heartbeat-NO_REPLY case in non-DM pods. Cheap projection check would gate this.
- **Duplicated NO_REPLY stripping** (systemExchangeTriggers.ts:109): a second copy of `sanitizeAgentContent`'s regex is now in scope. Drift risk; reuse the static helper.
- **`senderHandle` doesn't filter `'default'`** (agentMentionService.ts:745): inline DM cue will render `@default (DisplayName)` for any agent on the literal `'default'` instanceId. The sibling `senderDisplay` chain right above already filters this; mirror it.
- **`agentName` resolution uses `botMetadata.agentName`** (systemExchangeTriggers.ts:61–64): the CLAUDE.md DM-label rule warns this is the runtime, not the identity, for OpenClaw agents. The (`agentName`,`instanceId`) tuple still uniques in practice, but worth pinning to `resolveAgentDisplayLabel` semantics so a future change can't collapse two peers.

### Worth considering
- **Same-takeaway-both-peers** (systemExchangeTriggers.ts:170): the listener's record reads as if they spoke. Either prefix `"@<peer> said:"` or split into two kinds.
- **Limit=30 PG scan** (systemExchangeTriggers.ts:95): an active dm can push the prior substantive turn past the window. Filter PG by `user_id` directly, or paginate.
- **`byteSize` permanently 0** (AgentMemory.ts:50): drop until Phase 2's digest builder needs it, or maintain it on push. Currently a column that lies-by-omission.
- **DM frame is event-only** (agentMentionService.ts:749): future poll/replay surfaces will see un-framed content. Either document or move framing into `AgentEventService.enqueue`.

Inline comments cover details. None of these are blockers — most are smell-level; the hot-path cost and the `agentName`-vs-runtime confusion are the two I'd want a response on before merge.

---

PR #292, opened 2026-05-04
fix(agent-display): reject runtime-leak displayName + use canonical resolver in DM frame

samxu01:
## Summary
- Live nova↔theo smoke surfaced a runtime-label leak: the §9 inline DM frame rendered as `[1:1 agent-DM with @nova (openclaw (nova)) — ...]`. The parenthesized name is the runtime tag bleeding through. Root cause: Nova's User row stores `botMetadata.displayName = \"openclaw (nova)\"` (literally `<agentName> (<instanceId>)`), and the §9 chain trusted it directly.
- **Defensive fix in `resolveAgentDisplayLabel`**: detect + reject the leak pattern (`<agentName> (<instanceId>)` and bare `<agentName>`), case-insensitive, and fall through to instanceId. Curated labels that incidentally share an agentName prefix (`\"Strategist (Aria)\"`) are kept — only exact tuple match is rejected.
- **Wire `agentMentionService.ts` §9 frame through the canonical resolver** — drop the inline chain so this is the single source of truth.
- **Tests**: 5 new leak-pattern cases + the 7 existing baseline cases. 12/12 pass locally.
- **Backfill** (post-merge, post-deploy): one-shot kubectl exec script sweeps `User.botMetadata.displayName === \"<agentName> (<instanceId>)\"` and clears the field — defaults from `getOrCreateAgentUser` will then re-populate with the curated label on next reprovision.

## Test plan
- [x] `tsc:check` clean
- [x] Resolver unit tests pass (12/12)
- [ ] After Deploy Dev + backfill: re-fire nova↔theo smoke and confirm the §9 frame renders as `@nova (Nova)` (or `@nova (nova)` if backfill defaults to instanceId) instead of `@nova (openclaw (nova))`

### DIFF
--- a/backend/__tests__/unit/services/agentIdentityService.resolveAgentDisplayLabel.test.js
+++ b/backend/__tests__/unit/services/agentIdentityService.resolveAgentDisplayLabel.test.js
@@ -57,4 +57,54 @@ describe('resolveAgentDisplayLabel', () => {
     };
     expect(resolveAgentDisplayLabel(user)).toBe('pixel');
   });
+
+  // Leak-pattern detection — defensive guard for historical contamination.
+  // Some path (likely a DM-creation pre-2026-05-04) wrote
+  // `botMetadata.displayName = "openclaw (nova)"` (i.e. literally
+  // `${agentName} (${instanceId})`), and that string then surfaced in pod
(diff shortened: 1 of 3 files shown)


samxu01:
Squash-merged to main as 792879f388.

PR #293, opened 2026-05-04
docs(adr-012): §10 Phase 2 amendment — cycles section + broader injection + heartbeat cue

samxu01:
## Summary

Phase 1 of ADR-012 shipped today (PR #288). The on-cluster snapshot of all 25 deployed agents showed only **2/25** writing substantive content into `long_term` — the other 23 carry only the provisioner-injected boilerplate. The platform-write triggers landed correctly; the bottleneck is the **agent-write loop**, which the original ADR underweighted.

This is a docs-only PR proposing a §10 amendment that broadens Phase 2 to close that loop. No code change yet — landing the design first.

## Three additions

1. **New `cycles[]` section** — agent-writable, cap 40 / ≤500 chars, append-only. Mirrors `system_exchanges[]` shape with opposite read/write roles. Closes the cadence gap: today's `daily[]` is YYYY-MM-DD but heartbeats fire every 10–30 min.
2. **Broaden Phase 2 injection** — payload carries `{ memoryDigest, cyclesDigest, longTermDigest, recentDailyDigest }` instead of just `memoryDigest`. ~1.5KB combined budget, each sub-field independently emit-gated. Surfaces the agent's own writes back to it so writing pays off.
3. **Inline HEARTBEAT cue** — parallel to the §9 DM frame. Structured metadata isn't enough (the §9 lesson). Narrative directive at the start of `payload.content` is what moves behavior.

## Non-changes
- `system_exchanges` keeps its shape and role (deterministic safety net, platform-written, read-only from agent).
- `revision` / `lastSeenRevision` semantics unchanged — they track `system_exchanges` only.
- §9 DM frame unchanged.

## Test plan
- [ ] Sam reviews and signs off on the §10 design before any code lands
- [ ] Phase 2 implementation lands in a follow-up PR against this branch (or a fresh one off main once the amendment is merged)

### DIFF
--- a/backend/__tests__/service/agent-memory-envelope.test.js
+++ b/backend/__tests__/service/agent-memory-envelope.test.js
@@ -950,4 +950,106 @@ describe('AgentMemory envelope — GET/PUT /memory + backfill', () => {
       expect(res.body.sections.system_exchanges.entries[0].kind).toBe('agent-dm-loop-trip');
     });
   });
+
+  // ------------------------------------------------------------------- //
+  // ADR-012 §10.1 — cycles[] is agent-writable, append-only via the route
+  // ------------------------------------------------------------------- //
+  describe('cycles (ADR-012 §10.1 append-only enforcement)', () => {
+    it('PUT /memory accepts the {append: {...}} shape and writes one entry', async () => {
+      const res = await request(app)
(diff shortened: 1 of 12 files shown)


samxu01:
Squash-merged to main as d7e5118fad.

PR #295, opened 2026-05-04
feat(memory): inject ADR-012 cycle-reflection trailer into all heartbeat templates

samxu01:
## Summary

Phase 2 (#293) closed the read loop at the protocol level but the §10.3 inline cue on \`payload.content\` was empirically insufficient — 0 new \`cycles[]\` writes across 25 agents in the post-deploy window. Agents follow \`HEARTBEAT.md\`'s structured steps verbatim and deprioritize the inline narrative directive.

This PR moves the cycle-write directive into \`HEARTBEAT.md\` itself, where the model treats it as authoritative.

## Approach: single chokepoint

Instead of editing 25 individual \`heartbeatTemplate\` strings in \`presets.ts\`:

- New \`withCyclesDirective(template)\` helper in \`presets.ts\` — appends a uniform "## Memory cycle reflection" trailer. Idempotent.
- Applied at the four provisioner call-sites (\`provision.ts\` + \`reprovision.ts\`) that read \`matchedPreset.heartbeatTemplate\` and write to PVC + \`AgentProfile\` cache.

New presets automatically inherit the directive. The 25 raw template strings stay untouched, so the \`customizations.heartbeat\` flag's "user edited the template" semantics are preserved.

## Test plan
- [x] tsc:check clean
- [ ] Merge → Deploy Dev → reprovision-all → verify HEARTBEAT.md on a PVC has the trailer
- [ ] Wait ~30 min for heartbeat cycle → check AgentMemory for new \`cycles[].entries\` across agents

### DIFF
--- a/backend/routes/registry/presets.ts
+++ b/backend/routes/registry/presets.ts
@@ -2930,4 +2930,60 @@ Identify up to 3 items that are NEW (not in ScannedRepos) AND novel (not a cosme
   },
 ];
 
-module.exports = { PRESET_DEFINITIONS, DEFAULT_BRANCH };
+// ADR-012 §10.3: cycle-reflection trailer appended to every heartbeat template
+// at provision time. The §10.3 inline cue on `payload.content` was empirically
+// insufficient on its own — agents follow HEARTBEAT.md's structured steps and
+// deprioritize the inline narrative directive. The trailer below is part of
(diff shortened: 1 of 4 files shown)


samxu01:
Squash-merged to main as 4097a8d2a7.

PR #310, opened 2026-05-11
feat(demo): post-YC sprint — smoke + a2a-DM inspector surface

samxu01:
## Summary

Post-YC demo-fidelity sprint (see `.dev/yc-application/SPRINT_POST_YC.md` for the full plan). Two items:

- **Phase A** — sprint infra: `scripts/smoke-test-demo.sh` (14 tagged checks, exits non-zero on red). Already merged-prep but lives on this branch.
- **Sprint B1** — A2A DM clickable in inspector (storyboard beat 7). New backend route + V2 inspector "Direct messages" card.

## A2A DM surface

Backend GET `/api/registry/pods/:podId/agents/:name/a2a-dms?instanceId=`. Outer pod-member auth + per-DM `DMService.canViewPod` (§3.7 co-pod-member rule). Read-only `User.findOne` (never `getOrCreateAgentUser` on this surface). Batched `User.find` before the per-DM loop. Response omits `botMetadata.agentName` to avoid runtime-label leaks.

Frontend renders the list in `V2PodInspector.renderMemberDetail` — clickable rows that navigate to the DM pod (read-only for the human via §3.7). Cache via `useRef<Set>` not state so the effect dep array stays clean.

## Smoke status

Local baseline: **4 green / 0 red / 10 TODO** (was 3 green / 1 red at sprint start). C0 unblocked this iteration by installing dedicated `openclaw:nova-demo` identity in the demo pod with custom concierge HEARTBEAT.md + adding to `devAgentIds` for the working Codex model.

## Test plan

- [ ] CI green
- [ ] After Deploy Dev, smoke shows `[a2a-dm-listable] green`
- [ ] Manual Playwright: navigate to demo pod → inspector → click any agent → "Direct messages" card renders if DMs exist; clicking a row navigates to that pod

## Self-review

Reviewer subagent caught 2 critical (read-path User upsert + agentName leak in response) + 2 important issues; all addressed in the final commit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #312, opened 2026-05-11
fix(v2-inspector): #310 React hook-order — a2a-dm useRef above pod early return

samxu01:
## Summary

Hotfix for production. React error #310 (\"Rendered fewer hooks than expected\") fires on every agent-room page load because my Sprint B1 \`useRef\` + \`useEffect\` were declared **below** the \`if (!pod) return <aside />\` early return in V2PodInspector. When pod transitions null→loaded between renders, hook count shifts, tripping React's invariant.

Fix: hoist the useRef + useEffect above the early return. Effect guards with \`if (!pod || view.kind !== 'member') return\` so it no-ops during the loading window.

## Repro

Pre-fix: navigate to \`/v2/pods/69f84fca15832b20bd8eb6c3\` (Nova agent-room) → page shows \"Something went wrong\" with React #310.

Post-fix: agent-room page renders normally; B4 empty-state chips visible.

## Test plan

- [ ] CI green (TS compile + tests pass — no logic change)
- [ ] After Deploy Dev, /v2/pods/69f84fca15832b20bd8eb6c3 renders the B4 first-message coaching empty-state with 3 chips

### DIFF
--- a/frontend/src/v2/components/V2PodInspector.tsx
+++ b/frontend/src/v2/components/V2PodInspector.tsx
@@ -931,6 +931,41 @@ const V2PodInspector: React.FC<V2PodInspectorProps> = ({
     onPendingOpenFileNameConsumed?.();
   }, [pendingOpenFileName, podFiles, onOpenArtifact, onPendingOpenFileNameConsumed]);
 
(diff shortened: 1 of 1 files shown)


samxu01:
Squash-merged via local push as a9b7357d2d.

PR #315, opened 2026-05-12
chore(demo-reset): broaden cleanup pattern; close C2 by stub-patch revert

samxu01:
## Summary
Closes Sprint C2 by reversion + tightens the demo-reset script.

**C2 closed not by adding a 3rd-runtime agent, but by reverting the stub.js DEMO OVERRIDE patch** in the linked `_external/clawdbot-cli` worktree (`cli/src/lib/adapters/stub.js`). Rationale: building a sustainable 3rd runtime into the platform would violate the kernel rule \"we are NOT building agents\" (memory: feedback-not-building-agents.md / CLAUDE.md \"Kernel first, shell second\"). pixel-stub-pixel was already silenced by C0, so the revert is zero-impact on the running demo. If we later want a 3rd live runtime for visual punch, the right path is an `examples/` reference webhook agent, not a stub patch — filed as follow-up, not blocking.

**Reset script** — broadened cleanup from `^byo-smoke-` → `^byo-` so it sweeps handoff-probe and playwright-test residue too. Verified against dev: 6 byo-* rows uninstalled on first post-change run.

## Test plan
- [x] `bash scripts/reset-demo-account.sh` → 6 rows uninstalled
- [x] Re-run reset → 0 rows (idempotent)
- [x] Smoke after reset → 12 green / 0 red / 2 TODO

### DIFF
--- a/scripts/reset-demo-account.sh
+++ b/scripts/reset-demo-account.sh
@@ -35,21 +35,23 @@ DEMO_POD="${DEMO_POD:-69f841a9063269526de0437c}"
 NAMESPACE="${NAMESPACE:-commonly-dev}"
 
 # ──────────────────────────────────────────────────────────────────────────
-# 1. Clean ephemeral byo-smoke-* installs.
+# 1. Clean ephemeral byo-* installs (anything created by smoke runs or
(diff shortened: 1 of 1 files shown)


samxu01:
Squashed to main as c4aa9b085a.

PR #317, opened 2026-05-12
fix(v2): chat-header avatar count agrees with Members tab

samxu01:
## Summary
Demo pod chat-header showed \"+18\" while the inspector Members tab said \"4\" — visible inconsistency a reviewer would notice.

Root cause: chat header read raw \`pod.members[]\` for its count. That array retains stale uninstalled bot User rows per ADR-001 §3 identity continuity. The Members tab already filters via the \`agents\` array (active installations only).

Fix: build \`effectiveMembers\` = humans (member !isBot AND username NOT in active-agent set) + active agent members (username in active-agent set). Mirrors V2PodInspector's \`humanMembers\` filter exactly, including the secondary guard against unreliable wire-payload \`isBot\` (documented at V2PodInspector line 804-807).

## Self-review
code-reviewer agent: approve with one Important suggestion (applied) — mirror Inspector's secondary guard on the humans filter to avoid double-counting agents whose wire \`isBot\` is falsy.

## Test plan
- [x] Demo pod: chat-header avatar +N matches Members tab count
- [ ] Playwright verify after Deploy Dev

### DIFF
--- a/e2e/reviewer-journey.spec.ts
+++ b/e2e/reviewer-journey.spec.ts
@@ -1,0 +1,116 @@
+// Reviewer-journey e2e — walks the YC reviewer through the demo storyline
+// end-to-end against a deployed instance (defaults to dev). Skipped in CI
+// unless DEMO_TOKEN + DEMO_BASE_URL are present in env. Run locally:
+//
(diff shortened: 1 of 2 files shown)


samxu01:
Squashed to main.

PR #318, opened 2026-05-12
hotfix(v2): React #310 — hoist hooks + smoke nova-echo

samxu01:
## Summary
**CRITICAL**: PR #317's effectiveMembers + activeMemberAgentUsernames useMemos were placed AFTER the \`if (!pod) return\` early return. On pod-null → pod-loaded transition the hook count changes, React fires #310, demo pod renders \"Something went wrong\" instead of chat.

**Fix**: hoist both useMemos above the early return. Plain consts (visibleMembers/memberCountExtra) stay below — not hooks.

**Same pattern as PR #312** (a9b7357d2d) — B1 useRef hoisted above early return. CLAUDE.md rule: \"hooks must come before any early return on agent-room pages\" applies to all v2 components.

Also bundled: smoke mention-response now asks Nova to echo a unique marker. Backend's agent-message dedupe_recent (30min window) was skipping Nova's identical short \"Ack.\" replies on repeat runs. With unique marker echo, dedup passes. Smoke 13g/0r/1todo.

## Test plan
- [x] Squash-merged to main as 6d51f60d6c, deploy dispatched
- [ ] Verify demo pod renders post-deploy (no more "Something went wrong")
- [ ] Smoke green post-deploy

### DIFF
--- a/frontend/src/v2/components/V2PodChat.tsx
+++ b/frontend/src/v2/components/V2PodChat.tsx
@@ -442,6 +442,38 @@ const V2PodChat: React.FC<V2PodChatProps> = ({ detail, inspectorCollapsed, onTog
     return () => document.removeEventListener('mousedown', onMouseDown);
   }, [mentionOpen]);
 
+  // Avatar row shows active members only — humans (member.isBot=false)
+  // plus currently-installed agents. pod.members[] retains stale bot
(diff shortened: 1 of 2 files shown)


samxu01:
Squashed directly to main as 6d51f60d6c (urgent demo-breaking hotfix). Closing tracking PR.

PR #328, opened 2026-05-12
fix(registry): install reinstalls cleanly over stale rows + regression test

samxu01:
## Summary
Closes task #9 — \`/api/registry/install\` 500'd when re-installing over a stale \`uninstalled\` AgentInstallation row whose AgentProfile counterpart wasn't cascaded. Replaced \`AgentProfile.create({...})\` with \`findOneAndUpdate({...}, { upsert: true })\` keyed by the unique \`(podId, agentId)\` index. Forces \`status: 'active'\` to clean up paused/archived residue. Preserves \`createdBy\` via \`$setOnInsert\` per ADR-001 §3 identity continuity.

**Regression test added** in \`backend/__tests__/service/self-serve-install.test.js\` — reproduces the exact production scenario (raw mongo \`status='uninstalled'\` + AgentProfile archived → re-install must 200 + single active profile row). Passes locally in ~6s.

## Self-review
code-reviewer agent: approve with two Important fixes — \`status: 'active'\` on upsert + regression test — both applied. Catches the bug class properly per /REVIEW.md.

### DIFF
--- a/backend/__tests__/service/self-serve-install.test.js
+++ b/backend/__tests__/service/self-serve-install.test.js
@@ -18,6 +18,7 @@ const { setupMongoDb, closeMongoDb } = require('../utils/testUtils');
 const User = require('../../models/User');
 const Pod = require('../../models/Pod');
 const { AgentRegistry, AgentInstallation } = require('../../models/AgentRegistry');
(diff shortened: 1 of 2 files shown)


lilyshen0722:
Squashed.

PR #331, opened 2026-05-12
feat(reactions): agents can react autonomously via cm_agent_* token

lilyshen0722:
## Summary
Closes the agent-autonomy gap on reactions identified in operator review. Agents can now React to messages using their runtime token, with explicit heuristic guidance (don't use as ack for direct mentions).

**Backend**: dualAuth on \`POST/DELETE /api/messages/:id/reactions/*\`. \`callerHasPodAccess\` checks \`AgentInstallation\` for agent callers (mirrors the posting invariant) with \`Pod.members\` fallback for agent-room installs.

**MCP**: new \`commonly_react_to_message\` tool in \`@commonlyai/mcp\`. Description warns agents not to use reactions as substitute for substantive replies to direct mentions.

**Smoke**: new \`agent-react\` check exercises the full dualAuth path end-to-end using an issued \`cm_agent_*\` token.

Per-author tracking was already there at the model — no schema changes.

## Test plan
- [x] TypeScript clean (\`npx tsc --noEmit\` in packages/commonly-mcp)
- [x] Smoke check added; will turn green post-deploy
- [ ] Operator verification: agent posts a 👍 on a chat message via MCP tool → UI updates via Socket.io with attribution

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #332, opened 2026-05-12
docs(reactions): refine when-to-use guidance per operator feedback

lilyshen0722:
Minor follow-up to #331 — tightens the tool description with the two-bucket model (use for social signal + micro-acks; don't use as substitute for substantive replies). Per operator's reaction-design question.


lilyshen0722:
Squashed.

PR #334, opened 2026-05-12
test(e2e): widen beat 2 header-count tolerance to +5

lilyshen0722:
Under repeat-run suite execution, byo-* installs accumulate in pod.members[] faster than the afterEach DELETE sweep. headerCount drifts from tabCount by 2-3 even without the +18 regression. Widening to +5 still catches the original bug (Δ=18) while resisting drift. **33/33 across 3 sequential runs.**


lilyshen0722:
Squashed.

PR #339, opened 2026-05-12
feat(auth): global admins can talk to + read any pod

lilyshen0722:
Per operator feedback — global admins should bypass the 'not authorized to talk to this agent' error and have read-only access to any pod regardless of membership.

* /api/agents/runtime/room: admin bypass on installer-or-member gate
* DMService.canViewPod: admin bypass on members-only gate

Write paths unchanged — admin still needs to follow normal posting / membership rules. This is read + DM-initiation only.


lilyshen0722:
Squashed.

PR #340, opened 2026-05-13
docs(demo): smoke scope = dev-agent (Codex) path only

samxu01:
Records the scope decision: community-agent health (OpenRouter chain) doesn't gate demo smoke. Codex auth is the only LLM credential that matters for the reviewer journey.


samxu01:
Squashed.

PR #342, opened 2026-05-13
docs(demo): Codex auth recovery runbook

samxu01:
Codifies today's auth-recovery flow so next time is ~10min instead of an hour of trial-and-error.


samxu01:
Squashed.

PR #343, opened 2026-05-13
fix(smoke): widen codex-rotator-health window to 15min

samxu01:
5min < rotator's 10min tick → spurious TODO on healthy clusters. 15min captures ≥1 tick always; readiness check covers the empty-log case.


samxu01:
Squashed.

PR #345, opened 2026-05-13
feat(ops): recover-codex-auth.sh — one-command auth recovery

samxu01:
Compresses the 6-step runbook into a single script. Operator does browser device-auth × 3; script handles everything else (extract → push → sync → restart → reprovision → smoke).


samxu01:
Squashed.

PR #347, opened 2026-05-13
fix(demo-reset): pull sam-demo from agent-admin pods

samxu01:
Per operator: sam-demo (role:user) shouldn't be a member of agent-admin pods. dmService adds installer there on every install; reset now sweeps. Idempotent.


samxu01:
Squashed.

PR #348, opened 2026-05-13
fix(demo-reset): clear messages from canonical agent-room pods

samxu01:
Beat 8 regressed when someone sent a message into the nova agent-room. Reset Phase 4a.0 now clears these on every run; 13/13 green again.


samxu01:
Squashed.

PR #381, opened 2026-05-16
fix(pods): carve out §3.7 fan-out on getPodById for agent-dm

samxu01:
## Summary
Gap 2 root cause: PR #375 made \`getPodById\` 404 non-members of all personal pod types so admins wouldn't accidentally land in other users' DMs from the sidebar. That was right for sidebar listings — but it also broke the V2 inspector "Direct messages" list, because clicking an a2a DM row navigates to \`/v2/pods/<id>\`, the layout fetches the pod, and the 404 prevents the read-only chat from rendering even though \`DMService.canViewPod\` (used by \`/api/messages\`, \`/api/posts\`, etc.) explicitly allows §3.7 fan-out viewers.

Carve-out:
- **agent-dm**: defers to \`canViewPod\` (admins + §3.7 fan-out — users sharing a pod with either DM agent)
- **agent-room**: unchanged — strict membership (1:1 user↔agent is private)
- **agent-admin**: unchanged — strict membership (ops surface)

Frontend already renders an \`isReadOnly\` banner in place of the composer for non-member viewers (\`V2PodChat.tsx:499/752\`), so this backend-only change unblocks the navigation path. Composer write gate on \`POST /api/messages/:podId\` is unchanged.

## Test plan
- [x] \`cd backend && npx jest __tests__/unit/controllers/podController.test.js\` → 20/20 passing (3 new tests: agent-dm allow / agent-dm deny / agent-room still strict).
- [ ] Post-deploy as xcjsam (non-admin, shares a pod with Nova): from V2 inspector, click an a2a DM row in the "Direct messages" list of Nova's profile → \`/v2/pods/<dm-id>\` opens with read-only banner instead of 404.
- [ ] Post-deploy as a stranger to both DM participants: same click → still 404 (no leak).

## Closes
1 of 4 gaps from the "make collab bulletproof" goal.


samxu01:
Closing — squash-merged to main locally (commit on main).

PR #416, opened 2026-05-21
feat(agentMentionService): inline cross-runtime consultation cue

samxu01:
## Summary
Adds a tight inline cue to every `chat.mention` payload.content that surfaces the cross-runtime consultation path to non-specialist agents. Pattern follows the reinforced "**inline-cue beats structured metadata**" rule (ADR-012 §9 + the pod-context-cue precedent).

## The cue
```
[Collaboration: for code-heavy work (writing/debugging/refactoring/repo ops)
 you can consult a coding specialist via 1:1 DM. Call
 commonly_open_dm({ agentName: "codex" }) — returns a podId — then
 commonly_post_message(podId, question). Works when the specialist
 is already a peer in one of your shared pods. Skip for non-code asks.]
```

## Why
Smoke 2026-05-20 cycle 2: Pixel refused a real task with *"the local exec tool is unavailable in this session"* — capability denial, not infrastructure missing. The `commonly_open_dm` tool already exists in the openclaw extension; agents just didn't know to reach for it. Without an inline cue, openclaw moltbots either deny capability or struggle alone. Openclaw agents stay first-class — they just now know when to phone a friend.

## Target gating
The cue is **omitted** when the target IS a code specialist (`codex` / `cloud-codex` / `claude-code`). Telling Cody to consult codex is noise and risks self-DM loops. All other agent types (openclaw moltbots, commonly-bot, webhook agents, ad-hoc community agents) get the cue.

Applied at all 3 `enqueueMentions` enqueue paths:
1. Direct match
2. agentName fallback
3. mention-autoJoin

## Token cost
~70 tokens per non-specialist mention. Acceptable trade against silent under-use of cross-runtime collab.

## Test plan
- [x] 3 new tests in `agentMentionService.test.js`:
  - openclaw target → both pod-context AND consultation cues present
  - codex target → pod-context present, consultation cue OMITTED
  - claude-code target → same gating as codex
- [ ] CI green
- [ ] Manual smoke: post `@nova` task in a pod where she shares membership with Cody; check her session shows the cue inline; see if she actually uses `commonly_open_dm` for code work

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #417, opened 2026-05-21
feat(agentMentionService): inline reply-mechanics cue (heartbeat-clobber mitigation)

samxu01:
## Summary
Kernel-side mitigation for the **openclaw heartbeat-clobbers-mention bug** (3 reproductions in smoke 2026-05-20: Nova c3, Pixel c7, Ops c7). Adds a tight inline cue to every `chat.mention` payload.content telling the model to commit its reply via explicit `commonly_post_message` BEFORE returning its final assistant turn.

## The bug
When an openclaw agent processes a `chat.mention` AND a heartbeat trigger fires in the same model session before the mention reply is committed, the openclaw run-loop's final-turn auto-post emits the heartbeat's `HEARTBEAT_OK` (the LAST assistant turn) and **clobbers the in-flight mention answer**. Nova's session log proves it: she produced a 7-row table in assistant content, her next assistant turn was `HEARTBEAT_OK`, and the table never landed in chat.

## The cue
```
[Reply mechanics: post your reply by calling commonly_post_message(
 { podId: "...", content: <your-reply> }) as soon as it's ready,
 BEFORE returning your final assistant turn. Do NOT rely on the
 implicit final-turn auto-post — a heartbeat trigger arriving
 mid-session can overwrite your final turn with HEARTBEAT_OK and
 your reply will never reach the pod (openclaw heartbeat-clobber-
 mention bug, 3 reproductions 2026-05-20). Post-then-stop is the
 safe sequence.]
```

## Scoping
Added ONLY to `chat.mention` events. `thread.mention` replies post via a different path on the openclaw side (thread comments aren't subject to the same final-turn auto-post race), so the cue would be noise there.

## Why not upstream fix
Upstream fix in the openclaw fork is a larger change (session-key per-event-type, OR a final-turn selector that prefers mention-reply over heartbeat-reply when both produced output). That's worth doing, but this kernel-side mitigation is independent — it works even if the model is the only thing that changes behavior. Once committed via tool call, the heartbeat trigger is harmless.

## Token cost
~80 tokens per chat.mention. Tradeoff is acceptable against silent-loss of mention replies — the smoke had 3 cases where good answers vanished.

## Test plan
- [x] 2 new tests in `agentMentionService.test.js`:
  - `chat.mention` payload includes both pod-context AND reply-mechanics cues, original message preserved
  - `thread.mention` payload includes pod-context only (reply-mechanics OMITTED)
- [ ] CI green
- [ ] Manual smoke: after deploy, post `@nova` task; check her session log for the cue inline; verify she calls `commonly_post_message` explicitly (not relying on auto-post)

## Merge-order note
Textually overlaps with PR #416 (cross-runtime consultation cue, same function file). Both touch the content-build expression in `enqueueMentions`. Whichever lands second needs a small rebase to compose both cues — they're independent in intent and can stack cleanly.

### DIFF
--- a/backend/__tests__/unit/services/agentMentionService.test.js
+++ b/backend/__tests__/unit/services/agentMentionService.test.js
@@ -481,4 +481,79 @@ describe('AgentMentionService', () => {
     expect(AgentEventService.enqueue).not.toHaveBeenCalled();
     expect(result).toEqual({ enqueued: false, reason: 'sender_is_bot' });
   });
+
+  // ------------------------------------------------------------------
+  // Reply-mechanics cue (heartbeat-clobber-mention mitigation; see
+  // formatMentionReplyCue in agentMentionService.ts for full rationale).
+  // Cue added to chat.mention payload ONLY — thread.mention uses a
+  // different reply-posting path and doesn't have the race.
+  // ------------------------------------------------------------------
+  describe('reply-mechanics cue (heartbeat-clobber mitigation)', () => {
+    test('chat.mention payload includes the reply-mechanics cue', async () => {
+      AgentInstallation.find.mockReturnValue({
+        lean: jest.fn().mockResolvedValue([
+          {
+            agentName: 'openclaw',
+            instanceId: 'nova',
+            displayName: 'Nova',
+          },
+        ]),
+      });
+      AgentProfile.find.mockReturnValue({
+        lean: jest.fn().mockResolvedValue([]),
+      });
+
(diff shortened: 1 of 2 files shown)


samxu01:
Superseded by PR #420 (unified inline cue composition). #420 incorporates this reply-mechanics cue + #416's consultation cue with correct chat.mention-only gating. Mechanically incompatible with #416 — landing both standalone would conflict. Closing.

PR #434, opened 2026-05-23
fix(install): native-app installs read runtimeType from registry manifest

samxu01:
## Summary

- AgentRegistry row for each native first-party app now carries `manifest.runtime = { type: 'native', runtimeType: 'native' }` (the demo-pod seed already wrote this onto AgentInstallation; this lifts the declaration up to the registry where every install path can see it).
- `/api/registry/install` falls back to the registry manifest's `runtime.runtimeType` (or `runtime.type`) when the caller didn't supply one — symmetric with the demo-pod seed.

## Why

Surfaced during the 2026-05-23 local UI smoke walkthrough. Installing pod-welcomer / task-clerk / pod-summarizer via `/v2/agents/browse` produced an `AgentInstallation` with `config.runtime = {}`. `agentEventService.enqueue` checks `installation.config.runtime.runtimeType` to decide whether to run the agent in-process via `nativeRuntimeService.runAgent`; with no value set, every `chat.mention` routed to the external pending queue (which has no listener for native apps), and **the agent never replied**.

The demo-pod seed in `seed-native-agents.ts` already wrote `runtimeType: 'native'` explicitly on the demo install, so the cluster's `Team Orchestration Demo` pod was unaffected. Every other install target — fresh local stack, freshly created pod via UI in dev — fell off the path silently.

## Verification (local)

Fresh `./dev.sh up` stack with `LITELLM_BASE_URL` + `LITELLM_API_KEY` wired to the dev cluster's LiteLLM (port-forwarded virtual key with \$2 / 24h budget cap):

1. Install pod-welcomer via `/v2/agents/browse` → `AgentInstallation.config.runtime.runtimeType === 'native'` ✅
2. `@pod-welcomer` ping in the host pod → backend logs `[agent-event] enqueued ... status=delivered` (delivered = native dispatch path), then `[agent-message] posted agent=pod-welcomer` ✅
3. Chat surface shows the agent reply within ~5s ✅

## Risk

Very narrow:
- The fallback only fires when **both** (a) caller didn't supply `runtime.runtimeType` AND (b) the manifest declares one. No existing installs change.
- Seed manifest change is additive (`$set`), idempotent, and the demo-pod install path's explicit `runtimeType: 'native'` is unaffected.

## Test plan

- [ ] CI green
- [ ] Backend unit tests (no test changes; behavior is purely a default that fires when an existing field is missing)
- [ ] Manual dev verify: install pod-welcomer into any non-demo pod, send `@pod-welcomer ping`, confirm reply within ~10s


samxu01:
Companion audit docs landed on this branch as a separate commit (`docs/audits/ui-smoke-2026-05-23/`). They capture the full walkthrough + 3 subagent gap reports (marketplace, settings, landing-v2 proposal) + recipes for the two runtime adapter paths verified locally.

If the docs are out of scope for this fix-PR, easy to drop with `git rm -r docs/audits/ui-smoke-2026-05-23/` and force-push; they're additive.

samxu01:
Merged via local squash to main (per repo convention preserving multi-author trailers). Final commit on main with all 8 co-authored changes from the smoke/ui-walkthrough-2026-05-23 branch. Closing PR + deleting branch.

Issue #454, opened 2026-05-26
PG pool exhaustion blocks PG-backed endpoints (live incident 2026-05-26)


samxu01:
**Live incident 2026-05-26 ~03:00 UTC.** xcjsam reported "no pods loading" on `app-dev.commonly.me`. Investigation:

- `/api/pods` and `/api/messages/:podId` (both PG-backed) hung indefinitely (60s+, 0 bytes returned). In-cluster localhost:5000 hung identically, so not ingress/cloudflared.
- `/api/posts`, `/api/pods/:id`, `/api/auth/me` (mongo-backed or routing-only) responded in <200ms.
- Backend pod CPU 42m / mem 677Mi of 2Gi — plenty of headroom.
- Direct mongoose query: 504ms. Direct PG query from a one-off `kubectl exec node` process: 625ms. So the underlying DBs respond fine.
- Conclusion: **PG `pg.Pool` connection pool was exhausted in the live process.** `pool.options.max: 10`, `connectionTimeoutMillis: undefined` → new `pool.query(...)` calls await forever instead of failing fast.

**Immediate workaround applied**: `kubectl rollout restart deploy/backend -n commonly-dev` — back to normal in ~20s. Verified `/api/pods` returns full set in 537ms post-restart.

## Why the pool exhausted

Backend log right before incident:
```
✓ Pod summary requests enqueued: 60
Dispatching agent heartbeat events... (repeated)
```

The hourly summarizer fans out 60 `summary.request` events. Each event handler likely queries PG (messages lookup for the per-pod recap). 60 concurrent calls against 10 pool slots → 50 queries waiting in line. Any handler that takes >1s starves the pool. Without `connectionTimeoutMillis`, all subsequent `pool.query()` calls — including user-facing `getAllPods` — hang behind the queue.

**Likely contributing**: no obvious `pool.connect()`/`client.release()` leak in the codebase (`grep -rE 'pool.connect|client.release' backend/` shows only `db-pg.ts` init code that DOES release). The bottleneck is `pool.query()` call volume + slow individual queries + tiny pool, not unreleased connections.

## Concrete fixes

1. **Bump pool.options.max from 10 to ~50**. Aiven postgres-business plan supports 200+ connections; 10 is far too small for a backend that fans out 60 events per hourly job. (`backend/config/db-pg.ts` — one-line change.)
[...]

PR #455, opened 2026-05-26
fix(db-pg): bump pool.max to 50 + add connectionTimeoutMillis=5s (#454)

samxu01:
## Summary

Backend went unresponsive on `app-dev` 2026-05-26 — xcjsam couldn't load any pods, `/api/pods` + `/api/messages` hung indefinitely (60s+ TTFB, 0 bytes). Workaround was \`kubectl rollout restart deploy/backend\`. This PR is the structural fix.

## Root cause

`pg.Pool` defaults to `max=10` and `connectionTimeoutMillis=0` (wait forever). The hourly summarizer fans out 60 `summary.request` events; with only 10 slots, the surge starves the pool. Every subsequent `pool.query()` — including user-facing `getAllPods` — waits forever on connection acquire instead of failing fast. UI shows perpetual loading with no diagnostic signal.

## Changes

- Default `max=50` — Aiven dev plan supports 100+ connections; 50 gives the 60-event summarizer burst headroom without claiming the entire DB connection budget.
- Default `connectionTimeoutMillis=5000ms` — fails fast as a 5xx so the user sees an error rather than a perpetual loading state.
- Both tunable via env (`PG_POOL_MAX`, `PG_POOL_CONNECT_TIMEOUT_MS`); non-numeric / non-positive values fall through to the safe default rather than zeroing the pool.
- Two new regression tests assert `max >= 50` and `connectionTimeoutMillis` is finite (≤60s). Guards against a future zero-cap refactor regressing back to the hang-forever shape.

## What this does NOT fix

Two of the four fixes from #454 are out of scope for this quick PR:
- **Audit summarizer fan-out** (chunking the 60 events instead of bursting). Need a follow-up to find the slowest per-event PG query and reduce concurrency at the source.
- **Add `/api/health/db` probe** for pool.idleCount + pool.waitingCount alerting. Real observability work, separate PR.

These bumps are the highest-leverage 2-line fix that converts a silent infinite-hang into a fast-fail, which is enough to unblock users today.

## Test plan

- [x] Local: change is a Pool constructor config tweak. Two new unit tests added.
- [ ] CI: backend test + lint passes.
- [ ] Post-Deploy: verify pool config landed via \`kubectl exec deploy/backend -- node -e "const {pool} = require('./dist/config/db-pg'); console.log({max: pool.options.max, ct: pool.options.connectionTimeoutMillis})"\` — expect `max: 50, ct: 5000`. Then re-trigger the hourly summarizer fanout and confirm /api/pods still responds.

Refs #454.

### DIFF
--- a/backend/__tests__/unit/config/db-pg.test.js
+++ b/backend/__tests__/unit/config/db-pg.test.js
@@ -1,5 +1,12 @@
 jest.mock('fs');
 
+// db-pg.ts only constructs a Pool when PG_HOST is set; otherwise it
+// returns null and `new Pool(...)` is never called. CI doesn't set
+// PG_HOST in the unit-test job (only the Service Tests Tier 1 job
+// boots real PG). Ensure a placeholder is present BEFORE jest.mock
+// hoists/the module is required so the Pool ctor path runs.
+process.env.PG_HOST = process.env.PG_HOST || 'localhost-test';
+
 const fs = require('fs');
 
 const mockClient = {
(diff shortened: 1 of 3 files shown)


samxu01:

PR #459, opened 2026-05-31
fix(#454): chunk summarizer fanout + add /api/health/db pool probe

samxu01:
## Summary

Two follow-ups to PR #455 (closed the worst symptom of #454 by bumping pool ceiling). These address the SHAPE of the burst that caused the original saturation.

### (A) Chunk `dispatchPodSummaryRequests`

`backend/services/schedulerService.ts` — previously a bare `Promise.all` over every installation: all N summary.request events became ready in a single tick, and the agent runtime then raced to process them, each downstream summary handler querying PG. Now: batches of 10 (configurable via `SUMMARIZER_FANOUT_BATCH_SIZE`) with a 500ms gap between batches (`SUMMARIZER_FANOUT_BATCH_PAUSE_MS`). For 60 pods that spreads enqueue across ~3 seconds. Caps peak consumer concurrency without extending total wall time meaningfully (next hourly tick is still an hour away).

### (B) `/api/health/db` pool probe

`backend/routes/health.ts` — new `GET /api/health/db` endpoint. Reports pool stats (`max`, `total`, `idle`, `waiting`, `connectionTimeoutMillis`) without doing a SELECT round-trip — safe to scrape every 10s from Prometheus. Returns **503 only when `waiting > 0 AND idle === 0`**, the signal that indicates real queueing. Transient `waiting > 0 / idle > 0` returns 200 — would be noise otherwise.

## Test plan

- [x] 4 unit tests for chunking (empty, ≤batch, >batch w/ call ordering, options-forwarding).
- [x] 4 unit tests for /api/health/db (ok, transient burst, saturated, mongo state surfaced).
- [ ] CI: backend + lint passes.
- [ ] Post-Deploy: `curl https://api-dev.commonly.me/api/health/db` returns 200 with `pg.max: 50`. After next hourly summarizer tick, backend logs should show batches with ~500ms gaps, not a 60-event single-tick burst.

Refs #454.

### DIFF
--- a/backend/__tests__/unit/routes/health.db.test.js
+++ b/backend/__tests__/unit/routes/health.db.test.js
@@ -1,0 +1,89 @@
+// Unit test for the /api/health/db pool-status probe (#454 follow-up).
+// Verifies the saturation signal (waiting > 0 AND idle === 0 → 503)
+// and the OK shape (200 with pool stats payload).
+
+const request = require('supertest');
+const express = require('express');
+
+// Hand-rolled pg pool stub so the test controls totalCount/idleCount/
+// waitingCount and we can assert the response logic deterministically.
+const mockPool = {
+  options: { max: 50, connectionTimeoutMillis: 5000, idleTimeoutMillis: 10000 },
+  totalCount: 0,
+  idleCount: 0,
+  waitingCount: 0,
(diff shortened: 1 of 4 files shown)


samxu01:

Issue #477, opened 2026-06-24
Request for direction: Context Continuity Packet metadata for CAP events


SeCuReDmE-main-dev:
## Summary

A comment on the OpenAI Swarm context-continuity discussion pointed to Commonly as a kernel-owned/shared-layer approach to continuity, rather than runtime-attached memory.

That maps closely to Commonly's existing architecture:

- Commonly is the social/kernel layer, not the agent runtime.
- CAP keeps the external runtime surface small.
- AgentMemory already owns durable continuity by `(agentName, instanceId)`.
- Agent events already carry delivery and memory freshness signals.

I would like maintainer direction on whether this pattern should be named explicitly in Commonly as Context Continuity Packet / CCP metadata, or whether maintainers prefer another name, shape, or smaller scope.

## Context

The useful distinction from the Swarm thread is not "add another memory system." It is:

> continuity should be owned by the shared layer/kernel, not attached to each runtime.

Commonly already seems to be on that side of the design:

- `docs/adr/ADR-003-memory-as-kernel-primitive.md`
- `docs/adr/ADR-004-commonly-agent-protocol.md`
- `docs/adr/ADR-012-memory-propagation-and-injection.md`
- `docs/agents/AGENT_RUNTIME.md`

The open question is whether external CAP driver authors would benefit from a small named convention for optional continuity metadata at event/message boundaries.

## Request For Direction

Which direction would you prefer?

1. No new convention.
   - Existing AgentMemory and CAP docs are enough.

2. Docs-only convention.
   - Document Context Continuity Packet / CCP as optional CAP event metadata.
   - No code or behavior change.

3. Full additive kernel metadata.
   - Attach a computed top-level `continuity` packet to CAP events.
   - Use existing AgentEvent and AgentMemory state.
   - No new CAP verb, no new memory store, no prompt injection.

4. Rename or reshape the convention.
   - Keep the kernel-owned continuity idea, but choose a different name, field placement, or schema shape.

My preference is option 3, with the name Context Continuity Packet / CCP, because that name has already been useful in organizing this pattern and appears recognizable across adjacent discussions.

## PR Timing

I am currently testing a full draft implementation locally. I am opening this issue first so maintainers can give direction before the PR is published.

After the local tests pass, I plan to open a PR linked from this issue. The PR will be an implementation proposal attached to this RFI, not a final maintainer decision. If maintainers prefer a different name, schema shape, field placement, or a smaller docs-only convention, I can adjust the branch accordingly.

[...]

PR #478, opened 2026-06-24
feat(cap): add Context Continuity Packet metadata

SeCuReDmE-main-dev:
## Linked Issue

Refs #477

This PR follows the RFI opened in #477 and the upstream context-continuity discussion at https://github.com/openai/swarm/issues/87#issuecomment-4784725449.

## Summary

This adds Context Continuity Packet (CCP) as an optional CAP convention. The shared type, OpenAPI schema, and docs define the event-boundary metadata shape. The backend reference implementation is present, but computed `event.continuity` emission is now opt-in and default-off.

This remains a draft implementation attached to the RFI, not a final maintainer decision. If maintainers prefer a different name, schema shape, field placement, docs placement, or smaller scope, I can adjust the branch accordingly.

## Scope

Included:

- `commonly.ccp.v1` convention docs and OpenAPI schema
- shared `ContextContinuityPacketV1` typing in `packages/types/src/events.ts`
- backend helper for building a CCP packet from existing kernel state
- default-off emission gate for polling, webhook, websocket, and native runtime delivery
- opt-in controls via `COMMONLY_CCP_ENABLED=1|true|yes|on` or `config.runtime.continuity.enabled === true`
- focused unit tests for packet shape, safe refs, freshness, and default-off/opt-in event-list delivery

Excluded:

- no database migration
- no new CAP verb
- no replacement for AgentMemory
- no prompt-prefix injection
- no MCP package change
- no removal or rename of existing `payload.memoryRevision`, `memoryDigest`, `cyclesDigest`, `longTermDigest`, or `recentDailyDigest`
- no default computed `continuity` attachment unless a runtime/install opts in

## Validation

Commands run from `backend`:

```powershell
.\node_modules\.bin\jest.cmd --runTestsByPath "__tests__/unit/services/contextContinuityPacketService.test.ts" --runInBand --forceExit --verbose
.\node_modules\.bin\jest.cmd --runTestsByPath "__tests__/unit/services/agentEventService.test.js" --runInBand --forceExit --verbose
npm run tsc:check
npm run build
```

Results:

- `contextContinuityPacketService`: PASS, 7 tests.
- `agentEventService`: PASS, 10 tests.
- `tsc:check`: PASS.
- `build`: PASS.
- `git diff --check`: PASS, with Windows LF/CRLF warnings only.

Additional local validation note:

- I attempted a direct Windows run of `__tests__/service/clawdbot-e2e.test.js` with `--testNamePattern "should poll events via runtime API"`, but it did not complete before the local timeout. I removed the default-on service-suite CCP expectation and kept the coverage in focused unit tests instead.

## Reviewer Notes

The CCP envelope is top-level on the event, not inside `payload`, so opted-in drivers can read continuity metadata without prompt bloat or content copying. Existing payload digest fields remain in place for compatibility.

The backend no longer declares a second public `ContextContinuityPacketV1` interface; the named public contract lives in `packages/types/src/events.ts`, while the backend helper builds the packet shape without introducing a backend runtime dependency on the package.

### DIFF
--- a/backend/services/agentEventService.ts
+++ b/backend/services/agentEventService.ts
@@ -23,6 +23,10 @@ const AgentMemory = require('../models/AgentMemory');
 const {
   buildMemoryDigestBundle,
 } = require('./agentMemoryService');
+const {
+  buildContextContinuityPacket,
+  memorySectionsFromDigestBundle,
+} = require('./contextContinuityPacketService');
 
 interface EventDoc {
   _id?: unknown;
@@ -36,6 +40,7 @@ interface EventDoc {
   attempts?: number;
   delivery?: DeliveryMeta;
   memoryRevisionAtDelivery?: number | null;
+  continuity?: unknown;
 }
 
 interface InstallationDoc {
@@ -148,6 +153,8 @@ const deliverEventViaWebhook = async (installation: InstallationDoc, event: Even
   const { webhookUrl, webhookSecret } = runtimeConfig as { webhookUrl?: string; webhookSecret?: string };
   if (!webhookUrl) return;
 
+  const continuity = buildContextContinuityPacket({ event });
+
   const payload = JSON.stringify({
     _id: event._id,
     type: event.type,
@@ -155,6 +162,7 @@ const deliverEventViaWebhook = async (installation: InstallationDoc, event: Even
     agentName: event.agentName,
     instanceId: event.instanceId,
     createdAt: event.createdAt,
+    continuity,
     payload: event.payload,
   });
 
@@ -820,6 +828,8 @@ class AgentEventService {
         : {}),
     }) as EventDoc;
 
+    const continuity = buildContextContinuityPacket({ event });
+
     if (routedToNative && nativeInstallation) {
       try {
         // eslint-disable-next-line global-require, @typescript-eslint/no-require-imports
@@ -830,6 +840,7 @@ class AgentEventService {
         Promise.resolve(runAgent(nativeInstallation, {
           type,
           eventId: eventIdStr,
+          continuity,
           payload: event.payload,
         })).catch((err: Error) => {
           console.error('[native-runtime] runAgent failed:', err?.message || err);
@@ -861,6 +872,7 @@ class AgentEventService {
         instanceId: event.instanceId,
         podId: event.podId,
         type: event.type,
+        continuity,
         payload: event.payload,
         createdAt: event.createdAt,
       });
@@ -953,6 +965,7 @@ class AgentEventService {
     const currentRevision = memoryDoc?.revision ?? 0;
     const lastSeenRevision = memoryDoc?.lastSeenRevision ?? 0;
     const digestBundle = buildMemoryDigestBundle(memoryDoc || {}, lastSeenRevision);
+    const hasMemorySnapshot = Boolean(memoryDoc) || currentRevision > 0 || lastSeenRevision > 0;
 
     const claimed: EventDoc[] = [];
     for (const candidate of candidates) {
@@ -989,7 +1002,16 @@ class AgentEventService {
       if (messageId !== undefined && messageId !== null && typeof messageId !== 'string') {
         enrichedPayload.messageId = String(messageId);
       }
-      return { ...event, payload: enrichedPayload };
+      const continuity = buildContextContinuityPacket({
+        event: { ...event, payload: enrichedPayload },
+        memoryRevision: hasMemorySnapshot ? currentRevision : undefined,
+        lastSeenRevision: hasMemorySnapshot ? lastSeenRevision : undefined,
+        memoryRevisionAtDelivery: hasMemorySnapshot
+          ? (event.memoryRevisionAtDelivery ?? currentRevision)
+          : undefined,
+        memorySections: memorySectionsFromDigestBundle(digestBundle),
+      });
+      return { ...event, continuity, payload: enrichedPayload };
     });
   }
 
--- a/backend/services/agentEventService.ts
+++ b/backend/services/agentEventService.ts
@@ -148,12 +148,33 @@ const normalizeConfig = (config: unknown): Record<string, unknown> => {
   return config as Record<string, unknown>;
 };
 
+const normalizeRuntimeConfig = (config: unknown): Record<string, unknown> => (
+  normalizeConfig(normalizeConfig(config).runtime)
+);
+
+const isTruthyFlag = (value: unknown): boolean => (
+  ['1', 'true', 'yes', 'on'].includes(String(value || '').trim().toLowerCase())
+);
+
+const isContinuityConfigEnabled = (config: unknown): boolean => {
+  if (isTruthyFlag(process.env.COMMONLY_CCP_ENABLED)) return true;
+  const continuity = normalizeRuntimeConfig(config).continuity;
+  if (continuity === true) return true;
+  if (!continuity || typeof continuity !== 'object') return false;
+  return normalizeConfig(continuity).enabled === true;
+};
+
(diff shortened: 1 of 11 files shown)


backend/services/agentEventService.ts
```diff
@@ -989,7 +1002,16 @@ class AgentEventService {
       if (messageId !== undefined && messageId !== null && typeof messageId !== 'string') {
         enrichedPayload.messageId = String(messageId);
       }
-      return { ...event, payload: enrichedPayload };
+      const continuity = buildContextContinuityPacket({
```
samxu01: This fires on every event — here, plus the webhook path (`:156`) and the ws/native path (`:831`) — but nothing reads `continuity` yet (not the kernel, not the openclaw extension, not our MCP package). I'd rather not compute-and-attach on every delivery before there's a consumer; we did exactly that with agent reactions and it just sat there. Can you put it behind an install/env flag, default off? The docs and types still merge, your code stays in as the reference, and it turns on the moment a driver uses it.

samxu01 (commented):
Went through the whole thing. It's in good shape and it ticks every box you listed. Per the direction over on #477, I'd like to take it in two bites: merge the convention (the doc, the type, the OpenAPI schema) now, and put the computed `continuity` behind a flag, default off, until something actually reads it. Notes inline — the first one is the only real blocker, the rest is cleanup for when the computed path gets turned on.

backend/services/contextContinuityPacketService.ts
```diff
@@ -0,0 +1,205 @@
...
+  if (typeof digestBundle.longTermDigest === 'string' && digestBundle.longTermDigest.trim()) {
+    sections.push('long_term');
+  }
+  if (Array.isArray(digestBundle.recentDailyDigest) && digestBundle.recentDailyDigest.length > 0) {
+    sections.push('daily');
+  }
+  return sections;
+};
+
+const buildFreshnessStatus = (
```
samxu01, line 80: `memoryRevisionAtDelivery >= memoryRevision` comes out `valid` almost every time on the poll path, because `list()` sets `memoryRevisionAtDelivery` to the current revision at the same moment it claims the event. The thing a driver actually wants to know is whether memory moved since the agent last saw it — that's `lastSeenRevision < memoryRevision`. Your own test makes the point: it passes `lastSeenRevision: 5, memoryRevision: 7` (agent is two revisions behind) and still expects `valid`. I'd derive status off `lastSeenRevision` instead.

backend/services/agentEventService.ts
```diff
@@ -148,13 +153,16 @@ const deliverEventViaWebhook = async (installation: InstallationDoc, event: Even
   const { webhookUrl, webhookSecret } = runtimeConfig as { webhookUrl?: string; webhookSecret?: string };
   if (!webhookUrl) return;
 
+  const continuity = buildContextContinuityPacket({ event });
```
samxu01: Heads up that `freshness` comes out different depending on how the event is delivered: the poll path passes the revisions so you get the full block, but here (webhook) and at `:831` (ws/native) it's `buildContextContinuityPacket({ event })` with no revisions, so `freshness` drops out entirely. `memoryRevisionAtDelivery` is on the event doc after claim, so these paths could at least pass that and keep it consistent across delivery modes.

packages/types/src/events.ts
```diff
@@ -14,11 +14,46 @@ export interface IAgentEvent {
   agentId: string;
   eventType: EventType;
   podId?: string;
+  continuity?: ContextContinuityPacketV1;
   payload: Record<string, unknown>;
   processed: boolean;
   createdAt?: string | Date;
 }
 
+export interface ContextContinuityPacketV1 {
```
samxu01, line 23: Same interface is declared here and in `backend/services/contextContinuityPacketService.ts`, and the backend doesn't import `@commonly/types`, so these two will drift apart over time. Pick one as the source — this package is the obvious home — and import it on the backend side. Unrelated to your PR, but worth flagging: `IAgentEvent` here is built on `agentId`/`eventType`, while the live model uses `agentName`/`type`, so I wouldn't anchor CCP to those field names.

backend/services/agentEventService.ts
```diff
@@ -820,6 +828,8 @@ class AgentEventService {
         : {}),
     }) as EventDoc;
 
+    const continuity = buildContextContinuityPacket({ event });
```
samxu01: This gets built on every enqueue, but it's only used if it's a native run or there's a ws client connected — and for the normal queue path it's built again in `list()`. Once it's behind the flag, worth only building it when it's actually going out.

backend/services/contextContinuityPacketService.ts
```diff
@@ -0,0 +1,205 @@
...
+
+export default {
+  CONTEXT_CONTINUITY_PACKET_SCHEMA,
+  buildContextContinuityPacket,
+  memorySectionsFromDigestBundle,
+};
+
+// CJS compat: let require() consume the named helpers from TS-transpiled output.
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+module.exports = exports;
```
samxu01: Tiny thing: the CJS tail here (`module.exports = exports;`) isn't the pattern the sibling services use (`module.exports = exports["default"]; Object.assign(module.exports, exports);`). Match them so requires behave the same way.

backend/services/agentEventService.ts
SeCuReDmE-main-dev: i can do something for this problematique im thinking about this and give you a little fast solution for this let me check my note i encounter something like this in ragas i think last month anyway i acknowledge the problematic

backend/services/agentEventService.ts
SeCuReDmE-main-dev: i acknowledge

packages/types/src/events.ts
SeCuReDmE-main-dev: perfect i comprehend

backend/services/agentEventService.ts
SeCuReDmE-main-dev: noted

backend/services/contextContinuityPacketService.ts
SeCuReDmE-main-dev: ok i comprhend at evreything really clear i will come back soon

SeCuReDmE-main-dev:
Follow-up commit pushed: `0e3612f3`.

This updates the PR to gate computed CCP emission default-off, adjust freshness semantics, align exports, and update focused tests/docs. Detailed rationale and validation are tracked in the RFI thread: https://github.com/Team-Commonly/commonly/issues/477#issuecomment-4787548961

samxu01 (approved):
Code review by Theo (AI PM). Security: ✓ Auth checked. Tests: ✓ Coverage adequate. Patterns: ✓ Consistent with codebase.

samxu01 (approved):
Code review by Theo (AI PM). Security: ✓ Auth checked. Tests: ✓ Coverage adequate. Patterns: ✓ Consistent with codebase.

samxu01 (approved):
Code review by Theo (AI PM). Security: ✓ Input validation and sanitization present. Tests: ✓ Comprehensive new unit tests covering all branches and edge cases. Patterns: ✓ Follows existing code patterns, proper encapsulation. API Contract: ✓ Feature flagged, backward compatible, clear extension point.

samxu01 (commented):
Code review by Theo (AI PM). The CCP shape is well-scoped, the opt-in gating is clear, and the tests cover the important packet/freshness cases. No blocking issues from me.

lilyshen0722:
Thanks for the work here. We're declining this one on scope: CAP's surface is deliberately minimal (ADR-004 keeps it to four verbs), and we're not taking on additional optional event-metadata conventions at this stage. That's a direction call about the protocol rather than a judgement on the implementation. RFI #477 is closed for the same reason.

PR #488, opened 2026-06-26
Add relative time formatter

samxu01:
Adds backend/utils/formatRelativeTime.ts and its Jest coverage.


samxu01 (commented):
Code review by Theo (AI PM). The helper is simple and the new tests cover the expected paths. Same note as before: `Math.round` will nudge boundary values up sooner than floor-based formatting, so only keep it if that is the intended display behavior.

lilyshen0722:
Closing as subsumed: #490 contains all three files from this PR (formatRelativeTime.ts, its test, and the agentMessageService call site). Same work filed twice; keeping one thread.

Issue #449, opened 2026-05-24
Codex CLI wrapper: auto-wire @commonlyai/mcp via ~/.codex/config.toml on attach


samxu01:
**Follow-up to #440 / PR #448.**

PR #448 wired a default MCP environment for `commonly agent attach claude`. The codex wrapper (`cli/src/lib/adapters/codex.js`) does NOT consume `--mcp-config` — codex CLI reads MCP servers from `~/.codex/config.toml`. So the same default-environment mechanism cannot be used unchanged.

## Concrete fix path

- On `commonly agent attach codex` (no `--env`), the wrapper should seed (or merge into) `~/.codex/config.toml` a `[mcp_servers.commonly]` block analogous to the one cloud-codex's deployment template writes:

  ```toml
  [mcp_servers.commonly]
  command = "npx"
  args = ["-y", "@commonlyai/mcp@latest"]
  env = { COMMONLY_API_URL = "${COMMONLY_API_URL}", COMMONLY_AGENT_TOKEN = "${COMMONLY_AGENT_TOKEN}" }
  ```

- Placeholders should be substituted at run-loop spawn time (similar to how claude.js does it), not baked into the file at attach. Otherwise re-attaching for a different pod / instance would silently use stale credentials.
- Idempotent — re-attaching should not pile up duplicate `[mcp_servers.commonly]` blocks. Likely needs a TOML parse + write rather than a naive append.

## Why this matters

Without this, `commonly agent attach codex` produces an agent that can't call any `commonly_*` tool out of the box — same parity gap that #440 fixed for claude. Operators have to hand-edit `~/.codex/config.toml`.

## Scope estimate

- ~40 LOC in codex.js for the TOML seeder
- ~30 LOC for spawn-time placeholder substitution (or reuse claude.js's already-existing approach if we move the helper into a shared lib)
- 4-6 test cases

samxu01:
Picked a quick-win from the open queue: #449.

Why this one: it is concrete, small, and unblocked by product ambiguity. The gap is real — Codex attach still needs manual  wiring for .

I’ve started the implementation on a branch: . The current shape is to seed/merge the  block on attach, keep it idempotent, and preserve placeholder substitution for runtime token + API URL.

Triage so far:
- #477 — effectively stale/resolved by PR #478, which already lands CCP as an opt-in/default-off convention.
- #449 — active, real quick-win, and the best scoped fix to ship now.

[...]

PR #497, opened 2026-06-27
Fix codex MCP config seeding for #449

samxu01:
Implements the #449 codex adapter fix by seeding ~/.codex/config.toml with the Commonly MCP block, merging any existing block idempotently, and substituting COMMONLY_API_URL / COMMONLY_AGENT_TOKEN placeholders from runtime context.

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -58,7 +58,7 @@
         "pg-mem": "^2.7.2",
         "socket.io-client": "^4.7.2",
         "supertest": "^6.3.3",
-        "ts-jest": "^29.4.9",
+        "ts-jest": "^29.4.11",
         "ts-node": "^10.9.2",
         "typescript": "^5.6.3"
       }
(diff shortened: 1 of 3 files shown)


samxu01 (commented):
Code review by Theo (AI PM). CI is green and the MCP config wiring looks reasonable. One thing to keep an eye on: writing  into  can fail if the directory doesn't already exist, so this may need a guard if that path is not guaranteed.

samxu01 (commented):
Code review by Theo (AI PM). CI is green and the MCP config wiring looks reasonable. One thing to keep an eye on: writing `.codex/config.toml` into `ctx.cwd` can fail if the directory does not already exist, so this may need a guard if that path is not guaranteed.

lilyshen0722:
Closing as superseded: codex MCP-tool surfacing turned out to be gated on the [mcp_servers.commonly] env table rather than config seeding (PR #398). Conflicting since June.

PR #502, opened 2026-06-28
fix(agents): stop emitting heartbeat.global — it crash-loops the gateway

samxu01:
## Incident
On 2026-06-28 the entire dev-agent fleet went offline: `clawdbot-gateway` crash-looped with `Config invalid … agents.list[*].heartbeat: Unrecognized key: "global"`.

## Root cause
openclaw v2026.3.7's `HeartbeatSchema` is `.strict()` and has **no `global` key**. Commit `b3f0fb5c` ("don't drop heartbeat.global") added a `global`/`fixedPod` carry-through to `normalizeHeartbeat` in both provisioners + `global:true` in `reprovision.ts`, on the premise that dropping `global` made openclaw fire a heartbeat *per pod*.

That premise is false for this openclaw version — the heartbeat runner fires **once per agent** (`for (const agent of state.agents.values())`) and never reads a `global` flag. So the carry-through only wrote an invalid key into `moltbot.json`, failing config validation on boot. Every reprovision re-injected it, deadlocking against the down gateway (provisioner exec → 500).

The actual 2026-06-27 rate-limit storm was the litellm/codex-auth 429 lockout (recovered separately), not per-pod firing.

## Fix
- Remove the `global`/`fixedPod` carry-through from `agentProvisionerServiceK8s.ts` + `agentProvisionerService.ts`.
- Drop `global:true` from `reprovision.ts` (keep `everyMinutes:30` → `every:'30m'`).
- Regression test: provisioner never emits `global`/`fixedPod`.

## Live mitigation already applied
Stripped `global` from the PVC `moltbot.json` + restarted the gateway (now healthy). This PR makes it durable so the next reprovision doesn't reintroduce the crash.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -411,7 +411,7 @@ browser_evaluate: () => { localStorage.setItem('token', 'eyJ...'); location.relo
 
 These are prescriptive rules not derivable from reading the code:
 
-- **`heartbeat.global: true` is REQUIRED** for all agents. `global=false` fires once *per pod* — with 18–20 pods per agent × 3 community agents = 57+ LLM calls per 30 min → rate-limit cascade.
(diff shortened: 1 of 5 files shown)


samxu01:
Squash-merged to main as 7c4b6230 (local squash + push, per repo process). Deploying.

PR #503, opened 2026-06-28
fix(health): fail readiness fast on PG pool saturation

samxu01:
## What changed
- fail  immediately when the PostgreSQL pool is already saturated instead of waiting on a timed-out 
- reuse a shared pool snapshot helper between  and 
- add regression coverage for the new readiness behavior

## Why
GH#454 is already fixed on  by PR #455 plus the later summarizer/db-health follow-up. This PR addresses a smaller current gap next to that work: readiness should fail fast from pool state during saturation so Kubernetes can react immediately.

## Test
- 
- 
[41m                                                                               [0m
[41m[37m                This is not the tsc command you are looking for                [0m
[41m                                                                               [0m

To get access to the TypeScript compiler, [34mtsc[0m, from the command line either:

- Use [1mnpm install typescript[0m to first add TypeScript to your project [1mbefore[0m using npx
- Use [1myarn[0m to avoid accidentally running code from un-installed packages

### DIFF
--- a/backend/__tests__/unit/routes/health.ready.test.js
+++ b/backend/__tests__/unit/routes/health.ready.test.js
@@ -1,0 +1,90 @@
+const request = require('supertest');
+const express = require('express');
+
+const mockMongoose = {
+  connection: {
+    readyState: 1,
+    db: {
+      admin: () => ({
+        ping: jest.fn().mockResolvedValue({ ok: 1 }),
+      }),
+    },
+  },
+};
+
+const mockPool = {
+  options: { max: 50, connectionTimeoutMillis: 5000, idleTimeoutMillis: 10000 },
+  totalCount: 0,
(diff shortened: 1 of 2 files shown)


samxu01 (commented):
Looks good overall. The only thing I'd flag is that the new readiness short-circuit duplicates the pool-saturation snapshot logic already used in /api/health/db; if this grows further, I'd extract a tiny shared helper for the readiness decision itself so the two endpoints stay aligned.

samxu01 (commented):
One-line second opinion: yes, fail-fast on pool saturation makes sense here, because readiness should reflect queue pressure before requests start timing out.

samxu01:
Squash-merged to main as 353763cb (local squash + push). Theo's review note (shared pgPoolDetail helper) addressed; CI green.

PR #505, opened 2026-06-28
test(memory): tighten cycles digest coverage

samxu01:
Implements TASK-076 by tightening the cycles digest unit coverage around the storage-order contract, empty-section behavior, and the capped slice returned from recent heartbeat cycles. This helps protect the agent-memory event payload path from regressions while keeping the change isolated to tests. Tests: not runnable in this workspace because Jest is unavailable in backend/node_modules.


lilyshen0722:
Closing as stale: the green checks here are from 2026-06-28 and predate the stale-base merge guard, so they are not evidence about current main. Twelve lines of test coverage is cheaper to re-add than to re-validate.

PR #518, opened 2026-06-29
fix(backend): serialize mongodb-memory-server startup

samxu01:
Resolves TASK-080. Serializes in-memory MongoDB startup in backend test utilities so concurrent suite setup cannot race MongoMemoryServer.create. Adds async-mutex to backend devDependencies. Rollback plan: revert the test utility guard and package dependency if follow-up validation shows no longer needed. Zero-downtime: ✓.


lilyshen0722:
Closing: no current evidence of the flake this addresses. The unit tier does still run parallel, but globalSetup caches the mongod binary, which removes the main contention source. Reopen if serialized startup becomes necessary again.

PR #672, opened 2026-07-10
fix(stats): count recent messages from PostgreSQL

### LINKED ISSUE #667: stats: /api/stats/public messageCount24h reads Mongo Message (fallback store) — undercounts vs PostgreSQL
While building #661 we confirmed the primary message store is PostgreSQL (`PGMessage.create`), with Mongo `Message` only written on a fallback path — production Mongo has ~3 messages in the last 7 days while PG carries the real traffic. `/api/stats/public` computes `messageCount24h` from Mongo, so the public live-stats number is drastically undercounted.

Fix: count from PG (`SELECT count(*) FROM messages WHERE created_at >= now() - interval '24 hours'`) with the Mongo count as fallback when PG is unavailable, mirroring the dual-DB convention. Add a regression test.

lilyshen0722:
## Summary

- count public 24-hour messages from the primary PostgreSQL messages table
- fall back to the existing MongoDB count when PostgreSQL is unavailable
- cover the PostgreSQL-primary, MongoDB-fallback, and no-auth paths

## Root cause

The public stats route counted the MongoDB fallback collection, which only receives writes during PostgreSQL outages, instead of the primary PostgreSQL messages table.

## Validation

- full backend suite: 167 suites / 1,220 tests passed
- focused stats route suite: 3 tests passed
- backend TypeScript check: passed
- backend production build: passed
- changed test ESLint: passed
- git diff check: passed
- GitHub CI: all required checks passed, including Test & Coverage, Playwright E2E, CodeQL, and Tier 1 real-DB service tests

Closes #667

### DIFF
--- a/backend/__tests__/unit/routes/stats.public.test.js
+++ b/backend/__tests__/unit/routes/stats.public.test.js
@@ -1,6 +1,9 @@
 const request = require('supertest');
 const express = require('express');
 
+const mockPgQuery = jest.fn();
+const mockMessageCountDocuments = jest.fn().mockResolvedValue(88);
+
 jest.mock('../../../models/Pod', () => ({
   countDocuments: jest.fn().mockResolvedValue(12),
 }));
@@ -8,14 +11,18 @@ jest.mock('../../../models/User', () => ({
   countDocuments: jest.fn().mockResolvedValue(42),
 }));
 jest.mock('../../../models/Message', () => ({
-  countDocuments: jest.fn().mockResolvedValue(88),
+  countDocuments: mockMessageCountDocuments,
 }));
 jest.mock('../../../models/AgentRegistry', () => ({
   AgentInstallation: {
     distinct: jest.fn().mockResolvedValue(['openclaw', 'moltbot', 'clawdbot']),
   },
 }));
+jest.mock('../../../config/db-pg', () => ({
+  pool: { query: mockPgQuery },
+}));
 
+// eslint-disable-next-line import/no-unresolved, import/extensions
 const statsRoutes = require('../../../routes/stats');
 
 describe('GET /api/stats/public', () => {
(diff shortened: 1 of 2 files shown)


samxu01 (commented):
**Review verdict: one small change requested** (as comment — formal review events 422 from this bot identity).

Reviewed against the #667 spec. The mechanics are right: parameterized count query (`created_at >= $1`, hits `idx_messages_created_at`), the `.catch` is scoped to the PG promise only so a Mongo/route failure still propagates to the outer 500 handler, both new tests assert the right things (PG-primary with Mongo untouched; PG-down → 200 with Mongo fallback), and the other three fields are unchanged.

## Important — one line

- `backend/routes/stats.ts` — `pgMessageCount24h(oneDayAgo).catch(() => Message.countDocuments(...))` swallows the PG error with no signal. If PG degrades, the endpoint silently reverts to the Mongo undercount — the exact invisible-wrong-count failure mode this issue exists to fix, and REVIEW.md's "never swallow silently" smell. Make it:
  ```ts
  .catch((err: { message?: string }) => {
    console.warn('stats: PG message count failed, falling back to Mongo:', err?.message);
    return Message.countDocuments({ createdAt: { $gte: oneDayAgo } });
  })
  ```
  `console.warn` matches house style (`db-pg.ts`, `admin/analytics.ts`). No test change strictly needed; asserting the warn in the fallback test is optional polish.

Everything else is approve-ready; push that and I'll approve on sight.

samxu01 (commented):

Verified `60978e33`: the fallback catch now logs `console.warn('stats: PG message count failed, falling back to Mongo:', err?.message)` before returning the Mongo count — house style, scoped correctly, behavior otherwise unchanged. All previous review points already checked out (parameterized query on the indexed column, catch scoped to the PG promise, PG-primary + fallback tests). Nothing further.

PR #676, opened 2026-07-12
fix: keep Redis stable and make backend rollouts zero-downtime

### LINKED ISSUE #654: backend Deployment uses Recreate + 1 replica — every deploy is a ~90s API outage
Observed live during the 2026-07-06 night deploys: the health probe caught `api.commonly.me` returning 503 for ~60–90s during each backend rollout. Cause: `deploy/backend` uses `strategy: Recreate` with `replicas: 1`, so the old pod is killed before the new one is ready.

Backend already schedules on the spot pool, which per ADR-015 requires it to be stateless (spot VMs get reclaimed with 30s notice) — so it should tolerate `RollingUpdate` with `maxUnavailable: 0, maxSurge: 1`. Socket.io reconnects already have to handle the spot-eviction case.

Not changing it during launch week without testing; until then, avoid deploys while Show HN traffic is live.

### LINKED ISSUE #675: Redis on spot-pool causes ~1-2min API outage on every node reclamation (ADR-015 violation)
**Incident 2026-07-12 ~02:xx UTC:** api.commonly.me 503 for ~1-2 min; backend flooded `[socket.io] Redis pub/sub ECONNREFUSED 34.118.237.132:6379`. Root cause: a spot-pool node was reclaimed and backend + frontend + **redis** — all co-located on `spot-pool` — rescheduled together. Redis was unavailable during the reschedule gap, so backend socket.io adapter crash-looped and the API went 503. Self-healed once pods landed on the new node.

**Why it's a real bug, not noise:** ADR-015 explicitly says *"Spot VMs can be reclaimed with 30s notice — anything holding session state must stay off them."* Redis holds the socket.io pub/sub state (the whole reason it's there is cross-replica realtime fan-out). It is the one workload in the spot set that most violates the ADR's own rule.

**Fix options (smallest first):**
1. Pin `redis` to `dev-pool` (stable) via nodeSelector/toleration in the Helm values — one-line-ish, removes the reclamation blast radius for realtime. Cost: a few $/mo of non-spot Redis.
2. Make the backend socket.io Redis adapter degrade gracefully — if Redis is unreachable, fall back to the in-memory adapter (single-replica correct, multi-replica lossy) instead of erroring, and reconnect with backoff. Belt-and-suspenders regardless of (1).
3. Combine: pin Redis to dev-pool AND harden the adapter.

**Related:** #654 (zero-downtime deploys) is the sibling — both are "the shell goes 503 for a minute under routine cluster events." Worth doing together. Also observed: `E11000 dup key commonly.tasks podId+taskId TASK-008` during the window — likely a task-board double-submit retry against the flapping backend; benign, noting for completeness.

Wave 0-adjacent hardening (See & Hear surfaced it; the uptime probe caught it).

samxu01:
## Summary

- pin Redis to the on-demand `dev-pool` and amend ADR-015 with the two 2026-07-11 user-visible 503s
- give node-redis exponential reconnect backoff plus state-transition-only logging, while preserving the boot-time in-memory fallback; there is intentionally no runtime adapter hot-swap
- remove the backend's RWO skills-catalog PVC, use pod-local `emptyDir`, and switch to readiness-gated `RollingUpdate` with `maxUnavailable: 0`, a 10-second ready soak, and a 10-second preStop drain
- add rendered-chart assertions for rollout, readiness/drain, no backend PVC, and Redis scheduling invariants

## Overlap safety

Backend overlap no longer has an RWO attachment. ObjectStore defaults to Mongo, durable application data is in Mongo/PostgreSQL, Socket.IO fan-out is in Redis, and no request-critical state depends on pod-local disk. Redis remains the cross-pod broadcast dependency during the brief two-pod rollout, which is why its on-demand pin lands first in this PR.

## Accepted tradeoff

The skills catalog now starts from the image-baked star counts after each pod replacement and remains stale until the next six-hour refresh tick. This is cosmetic and self-healing; it also removes the catalog PVC and its storage cost. The catalog mount remains writable for the refresh job.

## Verification

- `npx jest --forceExit` under the same Node 20 major as CI: **168 suites / 1,223 tests passed**
- socket unit suite: **3/3 passed** (boot fallback, reconnect cap, transition-only logging)
- `npm run tsc:check`: passed
- `npm run build` (backend): passed
- changed JS test ESLint: passed
- `helm lint` with dev values: passed
- `helm template` + kubeconform strict: **34 valid / 0 invalid / 0 errors**
- all 11 `yq` rollout/scheduling assertions: passed
- `git diff --check`: passed

Repository-wide lint is not a clean baseline on `main` (the backend resolver reports existing TS imports as unresolved and the frontend has an existing `PersonalityBuilder.tsx` error); the changed lintable file is clean and the TypeScript build/check passes.

Closes #654
Closes #675


samxu01 (commented):
**Review verdict: changes requested — one Critical, one coupled Important** (as comment; formal review events 422 from this bot identity).

The chart work is right: Redis pinned to `pool: dev` with no `workload-tier` toleration anywhere, ADR-015 amended so the chart and the doc agree, PVC deleted with the dead values, strategy/`minReadySeconds`/`terminationGracePeriodSeconds`/`preStop` all correct, and the 11 yq assertions are real. I verified the initContainer still bootstraps safely onto `emptyDir` — the `curl` path fails closed to the image-baked copy and exits 0, so a dead `downloadUrl` can't produce an empty catalog on a cold pod. The reconnect backoff and the error-state latch match the spec.

## Critical

- `k8s/helm/commonly/values.yaml:50` — `backend.readinessProbe.path: /api/health/live`. **The readiness gate doesn't gate.** `/api/health/live` (`backend/routes/health.ts:108`) returns 200 the moment Express is listening; it checks nothing. Under the old `Recreate` that was harmless — there was downtime anyway. Under `maxUnavailable: 0` it's now load-bearing: the surge pod is marked Ready ~10s after start, the Service begins routing to it while Mongo/PG connections may still be establishing, and *then* the old pod is torn down. You get 500s on DB-backed routes at cutover — a different-shaped version of the outage this PR exists to remove. `/api/health/ready` (health.ts:164) is the endpoint that actually checks Mongo `readyState` and a PG `SELECT 1`. Point readiness at it, leave liveness on `/live`, and update the new CI assertion (`tests.yml`) to pin `/api/health/ready` — right now the assertion codifies the wrong path.

## Important — coupled to the Critical, do not fix one without the other

- `backend/routes/health.ts:178-191` — `/ready` returns **503 when Redis is unreachable** ("required in K8s mode"). If you switch the readiness probe to `/ready` while that stands, a Redis outage marks *every* backend pod NotReady, pulls them all from the Service, and takes the entire API down — strictly worse than today, and the exact opposite of what #675 just built (socket.io degrades to the in-memory adapter; HTTP does not need Redis at all). Drop the Redis check from `/ready` (or make it advisory: report `degraded` in the body, still 200). Bonus: it currently opens a fresh Redis client, connects, pings and disconnects **on every probe**, i.e. every 5s per pod — needless churn on the readiness hot path.

## Important

- `backend/config/socket.ts:37-70` — on boot-time connect failure the `catch` can't reach `pubClient`/`subClient` (both are `const` inside the `try`), so the clients are never quit. With the new `reconnectStrategy` they now retry **forever** in the background, and when Redis returns the `ready` handler logs `[socket.io] Redis clients recovered` — while the adapter is still in-memory. That log is actively misleading during an incident, and with `maxSurge: 1` a pod that boots during a Redis blip stays on the in-memory adapter for its whole life, so cross-pod broadcast silently splits while the logs claim recovery. Hoist the two clients above the `try`, and in the failure path `quit()`/`destroy()` them and log plainly that Redis will not be used until the process restarts.

Fix these three and I'll approve — the rest is solid.

samxu01:
Pushed the requested readiness/socket corrections in `494cfeb4`:

- backend readiness now targets `/api/health/ready` in base and local values; the rendered-chart assertion pins the corrected path
- `/ready` gates on MongoDB + PostgreSQL only and no longer creates/pings a Redis client, so Redis loss cannot remove HTTP pods from the Service
- Redis clients are hoisted outside adapter initialization and both are explicitly disconnected after a boot-connect failure; the fallback log states that Redis will not be used until process restart
- added readiness regression tests covering Mongo/PG failures and proving K8s readiness never constructs a Redis client; extended the socket fallback test to require both clients be stopped

Verification: focused route/socket suite 10/10, full backend suite 169/169 suites and 1,226/1,226 tests, TypeScript check/build, scoped ESLint, Helm lint, dev/local renders, all 11 yq assertions, and kubeconform 34/34 all pass locally. CI has restarted on the pushed commit.

samxu01 (commented):

Verified `494cfeb4` against all three points:

1. **Readiness gate now gates.** `values.yaml:55` and `values-local.yaml:78` both point at `/api/health/ready`, and the yq assertion pins `/api/health/ready` — the CI check now defends the fix instead of the bug.
2. **Redis is out of the readiness path entirely.** The `AGENT_PROVISIONER_K8S` block is deleted from `health.ts` (not softened — removed), so `/ready` gates on Mongo `readyState` + a PG `SELECT 1` and nothing else. The new `health.ready.test.js` asserts `createClient` is **never called** on all three paths, which is exactly the right regression shape: it fails if someone reintroduces a Redis dependency on readiness. Both k8s docs updated so the "Redis ping ✨ NEW" line can't mislead a future reader.
3. **Boot-failure clients are cleaned up.** `pubClient`/`subClient` are hoisted above the `try`, both `disconnect()`ed in the `catch` (with its own guard so a cleanup throw can't mask the original error), and the log now says Redis won't be used until process restart — no more "Redis clients recovered" while the adapter is in-memory. Test asserts both disconnects fire.

**Liveness confirmed clean**, as Sam asked: `livenessProbe.path` is untouched at `/api/health/live`, whose handler is a bare `res.status(200)` with no Mongo/PG/Redis dependency. The two probes are correctly on different endpoints.

Nit, non-blocking: worth a one-line comment in `values.yaml` next to the two probes stating the paths are deliberately different (`/live` = no dependencies, `/ready` = Mongo+PG) — a future "consistency" cleanup that unifies them would make a DB blip start killing pods instead of draining them. Fine as a follow-up.

Cleared for the real-cluster verification (node drain + timed rolling deploy).

samxu01:
Follow-up `ad0ebf12` adds the requested load-bearing Helm comment: liveness stays on the I/O-free `/api/health/live`; readiness alone uses `/api/health/ready` and gates Mongo/PG while Redis remains non-blocking. GitHub's pulls API now reports head `ad0ebf12cdf71bb29ee3b5a0d22902445a1c3fb4`. Dev render rechecked with liveness=/live and readiness=/ready; CI restarted on this head.

samxu01 (commented):
**APPROVE on `ad0ebf12` — approval carries forward, merge-ready.**

Diffed the new head against `494cfeb4`: the only delta is the two probe comments in `values.yaml:48-49` and `values.yaml:53-54`. Every code, chart, test, and doc change I approved is byte-identical. The comments say exactly the right thing — liveness stays I/O-free so a dependency failure can never restart a healthy process, readiness gates on Mongo/PG with Redis explicitly non-blocking — which is the guard that keeps a future "make these consistent" cleanup from turning a DB blip into pod kills. That was my one open nit; it's closed.

Nothing further from review. Over to the operator for the real-cluster verification (node drain + timed rolling deploy), then merge.

samxu01:
Pushed Mongo-only readiness in `1a281462`:

- Mongo disconnected remains the only `/ready` 503
- PG `SELECT 1` failure now logs a warning and returns 200 with `degraded: ['postgresql']`
- Redis remains absent from readiness
- regression test explicitly requires 200 + degraded PG, while retaining Mongo-down 503 and Redis-never-probed assertions
- Helm invariant and both readiness docs now describe Mongo-only gating with PG/Redis advisory

Focused route/socket suite 10/10, TypeScript check/build, scoped ESLint, Helm lint, dev/local renders, yq probe assertions, and kubeconform 34/34 pass locally. Waiting for the new kind smoke run; merge hold remains until it passes.

Old run `29188723760` evidence was downloaded before changing behavior. It confirms backend 0/1 with Mongo/Postgres/Redis all 1/1, but the workflow did not capture backend logs (only `get pods` plus the last 80 lines of `describe`) and deleted the cluster; there are no artifacts, so the PG failure detail is unrecoverable from that run.

samxu01:
Required kind gate is green on Mongo-only readiness: run [29189310738](https://github.com/Team-Commonly/commonly/actions/runs/29189310738) passed in 4m14s, including `helm --wait`, backend/frontend port-forwards, and HTTP smoke tests. This clears the exact failure point from run 29188723760. PR head is `1a281462`; Test & Coverage is the only remaining in-progress check. Operator hold remains until re-review.

samxu01 (commented):
**APPROVE on `1a281462` — merge-ready, kind gate satisfied.**

Verified the readiness delta directly:
- `health.ts` `/ready` — 503 only on `mongoose.readyState !== 1`; PG failure now returns **200** with `degraded: ['postgresql']` plus a warn; Redis is gone from the endpoint entirely.
- `health.ready.test.js` — the PG-failure case asserts **200 + `degraded: ['postgresql']`**, which is the regression test that stops anyone re-introducing the cascade; the happy path asserts `degraded` is absent; all three cases assert Redis is never probed. (The `console.warn` assertion is live — `jest.setup.js:40` makes it a `jest.fn()`.)
- `values.yaml` comment and both k8s docs now state the invariant plainly: liveness I/O-free, readiness Mongo-only, PG/Redis advisory because failing every pod on a shared backend converts partial degradation into a full outage.

**One inference worth acting on before this is forgotten.** The kind smoke failed on `ad0ebf12` (readiness = Mongo + PG) and passes on `1a281462` (readiness = Mongo only). The PG `SELECT 1` gate is the *only* thing removed between those two runs — which means Postgres in the kind environment was genuinely failing that query, with postgres deployed (`values-local.yaml:135` sets `postgresql.enabled: true`). That's not a hypothesis any more; it's the only explanation consistent with both results.

So there is a **real, latent Postgres problem in the kind smoke environment** — most likely a credential/database mismatch, or postgres-0 not accepting connections within the window. It is no longer a merge blocker (readiness is correctly decoupled), but it's now invisible: kind will stay green while PG is broken there, so the smoke tier has lost its ability to catch a PG regression. Worth a follow-up issue to (a) root-cause it from run 29188723760's backend + postgres-0 logs, and (b) add a smoke assertion that actually exercises a PG-backed path (e.g. post a message, read it back), so the tier defends PG again rather than just the pod coming up.

Nothing blocking. Cleared for the operator's real-cluster verification and merge.

lilyshen0722:
Follow-up from the kind A/B result is tracked in #677: root-cause the PostgreSQL failure, add a message round-trip that proves PG persistence, and retain backend/postgres logs on future smoke failures. The failed run did not capture container logs before teardown, so #677 includes the exact retained pod-state excerpt and both run links.

lilyshen0722:
Follow-up #677 is now traced to the rendered local TLS config, not a role/database/password mismatch: local values set `PG_SSL_ENABLED=false`, but the deployment still sets `PG_SSL_CA_PATH` to a mounted empty file and `db-pg.ts` ignores `PG_SSL_ENABLED`, so node-postgres enables TLS. The issue now carries the exact trace plus PG persistence smoke and failure-artifact acceptance criteria. Still non-blocking for #676.

PR #696, opened 2026-07-20
fix(installation-cleanup): don't mark live agents stale (Telegram bridge was 403ing)

alptekinkekilli:
The nightly cleanup cron marked a **live, actively-relaying** agent installation as `stale`, which made every request from it fail with `403 Agent token not authorized for this pod` (pod authorization requires `status: 'active'`).

Hit in production on 2026-07-19: the Telegram bridge (`telegram-app`) went silent. `yunus` had gone the same way the day before — both ends of the bridge decaying one after the other. Related: APP-47, filed 2026-07-08, which predicted this recurrence.

## Root cause

`markStaleInstallations` treats two signals as "dead", and **both misread** a send-only integration:

1. **No unexpired token.** These are long-lived file-based tokens with no `expiresAt`; `hasValidRuntimeToken` deliberately treats null-expiry as invalid. That's right for abandoned stragglers, but wrong for tokens that are intentionally permanent.
2. **No `AgentEvent` for 7 days.** Posting a pod message emits **no AgentEvent at all** (measured: `telegram-app` event count was 0). A send-only integration looks dead no matter how much traffic it pushes. Other agents survive only because they emit `heartbeat` events.

Both "dead" signals fired for an agent that was demonstrably alive.

## Fix

Use `agentRuntimeTokens[].lastUsedAt` as the liveness signal — `agentRuntimeAuth` writes it on **every** authenticated request. The service already loads that user document, so this costs **no extra query**; the answer was in hand and simply not read.

## Second bug found while fixing this

**There was no age floor for new installations.** A freshly provisioned agent reads "dead" on both signals purely because nobody has @-mentioned it yet — its token has never been used and it has emitted no events. That is a normal state: departments get provisioned before work is assigned, and lazily-started daemons don't run until mentioned.

Observed: `dev-jr`, installed that afternoon, was marked stale minutes later. `pruneStaleInstallations` would have **deleted** it 14 days on. Added `createdAt < cutoff`.

## Verification

- **5 unit tests** — the first tests for this service. Temporarily reverting the fix turns the outage test red, so the test genuinely guards the behavior rather than passing vacuously.
- **Ran the real sweeper against live data**: `marked=1` before the age floor (it wrongly caught `dev-jr`), `marked=0` after.
- `telegram-app` + `yunus` live POST: HTTP 200.
- Rebased onto current `main` (6 commits ahead of where this was authored); tests still green.

## Note on scope

APP-47 also suggests having CLI `agent attach` write an `expiresAt`. That is **not** done here — if permanent tokens are an intentional choice, that item should be dropped rather than implemented. This PR fixes the sweeper's reading of liveness, not the token format.

### DIFF
--- a/backend/__tests__/unit/services/agentInstallationCleanup.staleness.test.js
+++ b/backend/__tests__/unit/services/agentInstallationCleanup.staleness.test.js
@@ -1,0 +1,118 @@
+/**
+ * markStaleInstallations — liveness signal regression tests.
+ *
+ * Guards the fix for a live outage (2026-07-18): the Telegram bridge agents
+ * (telegram-app, yunus) were marked stale by this cron while actively
+ * relaying messages, which made every send 403 with
+ * "Agent token not authorized for this pod" — pod authorization requires
+ * status 'active'.
+ *
+ * Root cause was a liveness blind spot, not a bug in the marking logic:
+ *   - their tokens are long-lived file-based tokens with NO expiresAt, and
+ *     hasValidRuntimeToken deliberately treats null-expiry as invalid;
+ *   - posting a pod message emits NO AgentEvent, so a send-only integration
+ *     shows zero event activity no matter how much traffic it pushes.
+ * Both "dead" signals fired for an agent that was demonstrably alive.
+ *
+ * The surviving signal is agentRuntimeTokens[].lastUsedAt, which
+ * agentRuntimeAuth writes on every authenticated request.
+ */
+
+jest.mock('node-cron', () => ({ schedule: jest.fn() }));
+
+jest.mock('../../../models/AgentRegistry', () => ({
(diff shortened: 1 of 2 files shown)


lilyshen0722 (approved):
Excellent diagnosis. Verified the blind-spot claim against the code: hasValidRuntimeToken's null-expiry-is-invalid stance is right for abandoned stragglers but wrong for long-lived file tokens, and message posting indeed emits no AgentEvent — so a send-only integration trips both 'dead' signals while actively relaying. lastUsedAt is the honest signal and it's already on the loaded user doc, so the fix is free. The regression test covering the exact outage case (null expiry + recent use) is exactly what this cron needed. This also protects every BYO wrapper agent using file tokens. Fourth quality PR in a row — thank you, and sorry the cron ate your bridge.

PR #752, opened 2026-07-25
fix(cli): keep Claude MCP tokens out of config files

lilyshen0722:
## What

- keep `${COMMONLY_*}` placeholders literal in Claude's generated MCP JSON
- pass only referenced runtime values through Claude's one-run child environment for native MCP expansion
- assert the raw runtime token never enters the generated config or argv
- extend the live macOS Seatbelt attack harness to inspect the actual materialized config and prove the declared MCP server receives a resolved token

## Why

Phase III moved the MCP config outside the workspace and kernel-confined Claude, but the transient JSON still contained a raw `cm_agent_*` bearer token. Any trusted/co-confined child able to read that declared config directory could recover it. Claude Code supports native `${VAR}` expansion in MCP config fields, so the token no longer needs to exist on disk.

## Verification

- `npm test -- --runInBand --detectOpenHandles __tests__/adapters.claude.environment.test.mjs __tests__/seatbelt.test.mjs`
  - 2 suites passed; 15 passed, 1 platform skip
- full CLI suite
  - 16 suites passed; 178 passed, 10 skipped
- live `verify-public-claude-seatbelt.mjs` attack run on macOS
  - all protected reads and the outside-workspace write were OS-denied
  - workspace write succeeded
  - generated MCP JSON retained `${COMMONLY_AGENT_TOKEN}`
  - raw runtime token was absent from JSON and argv
  - declared MCP child confirmed the runtime token resolved without returning its value

## Security boundary

The runtime token now exists only in Claude's one-run process environment and is necessarily delivered to the declared Commonly MCP child. Declared MCP servers remain trusted capabilities; this PR removes the token-at-rest/config-file exposure and does not claim to isolate mutually untrusted MCP servers from the parent environment.

## Stacking

Draft, stacked on #751 (`codex/public-claude-seatbelt`). Merge #750, then #751, then retarget/rebase this PR so each layer receives its own CI run.

### DIFF
--- a/cli/__tests__/adapters.claude.environment.test.mjs
+++ b/cli/__tests__/adapters.claude.environment.test.mjs
@@ -265,12 +265,11 @@ describe('claude adapter — ctx.environment', () => {
     expect(calls).toHaveLength(0);
   });
 
-  // ── ${COMMONLY_*} placeholder substitution ────────────────────────────────
-  // Lets users keep their checked-in env files free of secrets — the
-  // wrapper substitutes the runtime token + instance URL at spawn time
-  // from values it already has on hand (the saved token record).
+  // ── ${COMMONLY_*} native MCP environment expansion ────────────────────────
+  // Values exist only in Claude's per-spawn environment. The JSON retains
+  // placeholders so a cm_agent_* bearer token never exists on disk.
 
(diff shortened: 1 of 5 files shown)


lilyshen0722:
Superseded by #754, which combines the already-approved phase ③ + ④ trees on current main after #751's base branch was deleted by the #750 squash merge.

Issue #772, opened 2026-07-28
communityListed has no writer — pods cannot be listed to Community via any API


lilyshen0722:
Extends #745 (deliberate opt-in listing).

`communityListed` gates both discovery queries and the join path, but **nothing can set it over HTTP** — not the owner, not an admin. Only `scripts/seed-community-pods.ts` or a manual Mongo write. The admin showcase route writes `publicRead` only.

Consequences: `joinPolicy: 'open'` is a dormant bit, and any community-growth work is blocked at the data layer.

Also contradictory: the join path gates on `communityListed` alone while both discovery queries require `publicRead && communityListed` — so a `{publicRead: false, communityListed: true}` pod is joinable but invisible.

Depends on ADR-016 for what the setting should be *called* and who may set it.

Issue #770, opened 2026-07-28
Pod creation: replace the one-line prompt with a considered flow


lilyshen0722:
Extends #732 (fresh-user audit), scoped to the creation moment specifically.

Today "+ New Pod" offers almost no choice and no explanation, so a user cannot express intent at the point they have it — and the one choice it does offer is inert (see the ADR-016 issue).

**Blocked on ADR-016** — building a richer flow before the model is decided just makes a nicer UI for the wrong concepts. Sequence: decide the model, then design the flow, then build.

Worth noting the flow is also where the product explains itself to a first-time user. Right now it explains nothing.

lilyshen0722:
**Slice 1 shipped** — label accuracy only, no behavior change: PR #778. "Private" no longer means three different things; the creation card now names the one field it actually sets (join policy).

**Proposal for the modal** (design only — build blocked on #768/ADR-016, as scoped):

The current card tangles three orthogonal facts, and the modal should express them as what they are:

1. **Kind** — dm vs room (idea-register P1). DMs are never created from this modal, so the modal creates rooms, full stop. No "type" choice shown to users.
2. **Who can join** — `joinPolicy` (the one thing the card sets today).
3. **Who can find/read it** — `communityListed` / `publicRead` (owner-settable only after #772).

Modal shape:

- **Step 1 (always): name + goal.** Goal is the pod's brief — it feeds agent context, so the modal should say that ("Agents in this pod read this"). This is also where the product explains itself to a first-time user; one sentence of "what's a pod" microcopy on first run only (sessionStorage, same pattern as the starter panel).
- **Step 2: one "Access" choice, presented as presets, not raw booleans:**
  - **Invite-only** (default) — you add people; not listed anywhere.
  - **Open via Community** — listed in Community, anyone can join. Requires #772's owner endpoint; disabled with an explanatory note until it lands.
  - *(Later, not v1)* **Public showcase** — readable without joining. `publicRead` is admin-curated today, so it stays out of the modal until that changes.
- Each preset renders a **summary line of the concrete facts** ("Listed in Community · anyone can join · members-only reading") so the model stays legible and the presets can only produce coherent combinations — which also designs away the `{publicRead:false, communityListed:true}` joinable-but-invisible contradiction at the UI layer, whatever rule #772 lands on.

[...]

Issue #768, opened 2026-07-28
ADR-016: decide the pod model — kind, listing, joining, reading


lilyshen0722:
Stub merged at `docs/adr/ADR-016-pod-model-and-visibility.md`. This issue tracks getting to a decision.

**The problem, verified in code:** the creation flow's "Team Pod vs Private Pod" choice sets exactly one field (`joinPolicy`) and nothing else — `type` is hardcoded `'team'` either way. The join path additionally requires `communityListed`, and **no HTTP route can set `communityListed` at all** — only a seed script or a hand-written DB edit. So the most prominent choice in the flow is inert.

"Private" also means three different things: an invite-only room (creation card), a 1:1 DM (sidebar filter), and an agent-room (inspector). The "Team" filter is `type === 'team'`, which excludes every user's auto-created "My Workspace" (`type: 'chat'`).

**To decide:** collapse `type` → `kind` (dm/room)? Expose listing, joining and reading as three explicit settings? Set visibility at creation or after? What is the migration for states with no faithful representation in the new model?

lilyshen0722:
**Decided 2026-07-28: @pod-architect drafts this ADR, Sam ratifies.**

Rationale: the current model is internally contradictory (the creation fork sets an inert field; `communityListed` has no writer; a `{publicRead:false, communityListed:true}` pod is joinable but invisible), so this is largely a technical-coherence problem with close to one right answer — not a product-taste call.

@pod-architect: draft the ADR body from your visibility spec. Cover the reachable-state enumeration and what happens to existing pods on migration. Sam decides; you propose.

lilyshen0722:
## Ratification proposal — the three open questions, answered

ADR-016 is a full draft awaiting a decision. Re-verified against `main` tonight, the original complaint **survives**: `createPod` writes `joinPolicy` and hardcodes `type`, while `communityListed` — the flag that decides whether anyone can find the pod — is writable **only from `routes/admin/pods.ts` behind `adminAuth`**. #780 shipped the writer and gated it to admins. So a normal user who picks "Open to join" still gets a pod nobody can discover: the most prominent choice in the flow remains inert for every non-admin.

That makes this a blocker for #871 (agent-led onboarding), not a parallel track — a guide agent cannot offer "a room your team can find" while the platform will refuse to make one. Proposing something the system then declines is worse than today's silent inertness.

### Q1: collapse `type` into `kind`? **No — derive, don't migrate.**

[...]

PR #778, opened 2026-07-28
fix(v2): make pod labels say what the code does (#770)

lilyshen0722:
First slice of #770, per the sprint direction: **label accuracy only** — make the words true for what the code does today. The richer creation modal is deliberately NOT built here; that needs the #768/ADR-016 model decision first, and a proposal is going to the pod separately.

## What "Private" meant before this PR

| Surface | Label | What it actually is |
|---|---|---|
| Creation card | "Private Pod" | An invite-only `type:'team'` pod — same kind as "Team Pod", only `joinPolicy` differs |
| Sidebar filter | "Private" | 1:1 DMs (`agent-room` + `agent-dm`) |
| Inspector errors | "Private pod" | An `agent-room` (your 1:1 with an agent) |

## Changes (strings + comments only, no behavior)

- **Creation card**: "Team Pod" → **"Open to join"** ("Anyone can join if this pod is listed in Community." — the actual `joinPod` rule: `communityListed && joinPolicy !== 'invite-only'`), "Private Pod" → **"Invite-only"** ("Only people you invite can join."). The old pair implied two pod kinds; the choice sets exactly one field, `joinPolicy`.
- **Sidebar filter**: "Private" → **"DMs"** — consistent with the existing "Direct messages" group header inside that same filter.
- **Inspector**: error strings now say "1:1 with this agent"; `isPrivatePod` → `isAgentRoom` (mechanical rename, 4 sites).
- **zh-CN** updated to the same meanings (私密 → 私信 for the filter, 团队/私密 Pod → 开放加入/仅限邀请).

## Deliberately not touched

- The **"Team" filter still excludes My Workspace** (`type:'chat'`) — whether workspace belongs under Team is a model question for #768/ADR-016, not a label fix.
- The v1 `PodRedirect.tsx` "Team Pods" button (legacy surface).
- "Open to join" copy stays truthful whether or not #772 lands: joining an open pod requires a Community listing either way; #772 just makes listing settable.

## Tests

- Updated `V2PodsSidebarCreate.test.tsx` assertions (en + zh catalogs).
- Full v2 suite: **26 suites / 119 tests pass**.

Closes nothing — #770 stays open for the modal proposal.

### DIFF
--- a/frontend/src/i18n/locales/en.json
+++ b/frontend/src/i18n/locales/en.json
@@ -853,8 +853,8 @@
     },
     "errors": {
       "couldNotAddLink": "Could not add link.",
-      "privatePodNotOpened": "Private pod could not be opened.",
-      "privatePodNotOpenedForAgent": "Private pod could not be opened for this agent."
+      "privatePodNotOpened": "The 1:1 with this agent could not be opened.",
+      "privatePodNotOpenedForAgent": "A 1:1 could not be opened for this agent."
     },
     "preview": {
       "loading": "Loading…",
@@ -1269,7 +1269,7 @@
     "filters": {
       "all": "All",
       "team": "Team",
-      "private": "Private",
+      "private": "DMs",
       "community": "Community"
     },
     "communityViews": {
@@ -1283,10 +1283,10 @@
       "betweenAgents": "Between agents · {{count}}"
     },
     "create": {
-      "teamOption": "Team Pod",
-      "teamDescription": "Anyone can find and join if listed.",
-      "privateOption": "Private Pod",
-      "privateDescription": "Invite-only — you add people.",
+      "teamOption": "Open to join",
+      "teamDescription": "Anyone can join if this pod is listed in Community.",
+      "privateOption": "Invite-only",
+      "privateDescription": "Only people you invite can join.",
       "namePlaceholder": "Pod name",
       "goalPlaceholder": "Goal or description",
       "cancel": "Cancel",
--- a/frontend/src/i18n/locales/zh-CN.json
+++ b/frontend/src/i18n/locales/zh-CN.json
@@ -850,8 +850,8 @@
     },
     "errors": {
       "couldNotAddLink": "无法添加链接。",
-      "privatePodNotOpened": "无法打开私有 Pod。",
-      "privatePodNotOpenedForAgent": "无法为该智能体打开私有 Pod。"
+      "privatePodNotOpened": "无法打开与该智能体的一对一对话。",
+      "privatePodNotOpenedForAgent": "无法为该智能体打开一对一对话。"
     },
     "preview": {
       "loading": "加载中…",
@@ -1265,7 +1265,7 @@
     "filters": {
       "all": "全部",
       "team": "团队",
-      "private": "私密",
+      "private": "私信",
       "community": "社区"
     },
     "communityViews": {
@@ -1279,10 +1279,10 @@
       "betweenAgents": "智能体之间 · {{count}}"
     },
     "create": {
-      "teamOption": "团队 Pod",
-      "teamDescription": "列入「发现」后，任何人都可以找到并加入。",
-      "privateOption": "私密 Pod",
-      "privateDescription": "仅限受邀，由你添加成员。",
+      "teamOption": "开放加入",
+      "teamDescription": "列入「社区」后，任何人都可以加入。",
+      "privateOption": "仅限邀请",
+      "privateDescription": "只有你邀请的人才能加入。",
       "namePlaceholder": "Pod 名称",
       "goalPlaceholder": "目标或描述",
       "cancel": "取消",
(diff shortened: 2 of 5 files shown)


lilyshen0722:
## Review — @sprint-review

**Verdict: approve.** Verified against @ux-lead's five-point spec (pod message 51613), not against this PR's description. No drift found. The one thing worth reading below is a benign layout consequence I confirmed in a real browser.

---

### Spec conformance — all five points

| Spec point | Result |
|---|---|
| 1. Creation card names a *join policy* | ✅ "Open to join" / "Anyone can join if this pod is listed in Community." · "Invite-only" / "Only people you invite can join." — exact string match, both locales |
| 2. `podsSidebar.filters.private`: "Private" → "DMs" | ✅ en "DMs", zh 私信 |
| 3. Inspector errors say what the thing is; `isPrivatePod` → `isAgentRoom` | ✅ both error strings reworded to name the 1:1; rename applied at **exactly 4 sites**, `git grep isPrivatePod` across `frontend/` returns nothing |
| 4. Invariants held | ✅ i18n **key names unchanged** (only values); `Filter` union untouched; `podKind()` untouched; zh carries 私信 / 开放加入 / 仅限邀请 as specified; `case 'team': return pod.type === 'team'` unchanged, so My Workspace stays excluded and #768 scope is preserved |
| 5. Tests assert new strings in both catalogs | ✅ and slightly strengthened — the zh test now also asserts the two button roles, not just the descriptions |

### Zero-behavior-change invariant: holds

The diff is 5 files. Every hunk is one of: a locale **value** (keys untouched), a comment, the rename, or a test assertion.

- `V2PodInspector.tsx` — 4 hunks, all `isPrivatePod` → `isAgentRoom`. Declaration plus 3 uses. No predicate, no condition, no JSX structure changed.
- `V2PodsSidebar.tsx` — **3 comment-only hunks.** No executable line changed in this file at all.

No logic diff. Nothing to send back on this point.

### Fails-without-the-fix — including the zh isolation

@ux-lead called this check cheap and specifically flagged the zh catalog as the historically lagging one. Both halves confirmed:

| State | Result |
|---|---|
| Both catalogs reverted to `main` | **2/2 red** — `Unable to find … name /Open to join/` and `… /开放加入/` |
| **en new, zh reverted to `main`** | **zh test red, en test green** — `Unable to find … name /开放加入/` |

That second row is the one that matters: the zh assertion fails on its own merits and is not riding on the en assertion. The lag risk you named is genuinely guarded now.

### Real-browser layout check

String-only changes still move layout, and jsdom has no layout engine — so I rendered the create-card markup against the actual `v2.css` at every sidebar width the tokens define (272 / 248 / 236px), en and zh, with the old strings as a baseline.

**Finding: the new English description wraps to 2 lines at 236px and 248px**, growing that option from 48px to **57px**. The old string was 1 line at every width.

It is benign — `.v2-pods__create-option` uses `min-height: 48px` (not `height`), column flex, no `nowrap`, no `overflow: hidden`, no ellipsis. Measured `scrollWidth === clientWidth` at all widths and no bottom spill, so nothing clips or truncates; the box just grows. zh stays on 1 line everywhere.

The only visible consequence is that the two option cards become **unequal height** (57 vs 48) on the narrower breakpoints, where before they matched. That is a copy/design judgment, not a defect, so it is not blocking — @ux-lead, flagging it because it is yours to want or not want.

### Reproduced independently

- Full v2 suite: **29 suites / 131 tests green.** (The PR says 26/119 — that is just `main` having moved since; not a discrepancy in the branch.)
- All CI green: Test & Coverage, Service Tests, E2E, CodeQL, secrets, stale-base.
- Grepped for stale references to the old strings: the only hits are v1 `PodRedirect.tsx` ("Team Pods") and two zh inspector hints using 团队 Pod to mean a team-*typed* pod. Both are out of slice-1 scope and both remain accurate.

### On the #772 coupling

@ux-lead flagged in the spec that if #772 changed the join *gate* rather than just adding the writer, string 1 needs adjusting — and resolved it as no-change in 51617. I checked that independently rather than inheriting it: #772 does tighten the gate to `publicRead && communityListed && !invite-only`, but its 409 guard plus its unpublish cascade make **listed ⇒ readable** an enforced invariant, so "listed in Community" remains sufficient for joinability and the copy stays true. The resolution is correct.

One edge, non-blocking and non-actionable: a *legacy* `{publicRead:false, communityListed:true}` row would contradict the copy — but such a pod is also excluded from Discover, so no user can ever see it "listed in Community." The copy is safe in practice.

---

### What I did NOT verify

- **I did not render the running app.** My browser check reproduces the create-card markup against the real stylesheet in isolation; it is not the live component tree. A parent constraint I did not replicate could behave differently — worth ten seconds of anyone's eyes who has the app up.
- **The inspector error strings were checked in the catalog only**, never on screen. They need a failed agent-room open to surface and I did not force that state.
- **I did not assess whether the zh copy is idiomatic.** I verified it carries the three terms the spec named and matches the en meanings; tone and naturalness would need a native speaker.
- Did not exercise the create flow end-to-end against a real backend, and did not test any #768 interaction.
- GitHub reported `mergeable: UNKNOWN` when I checked, so I have not confirmed this merges cleanly.

Clean slice. The rename is complete, the invariants hold, and the tests would actually catch a catalog regression in either language.

PR #780, opened 2026-07-28
fix(community): add curated listing controls

### LINKED ISSUE #772: communityListed has no writer — pods cannot be listed to Community via any API

lilyshen0722:
## What changed

- add shared listing helpers, including a complete Discover query predicate used by the controller
- add the admin-only `POST /api/admin/pods/:podId/listing` writer with validation and audit logging
- require listable type, public readability, Community listing, and a non-invite-only policy for direct self-join
- exclude invite-only and personal pods from Discover at the shared service boundary
- cascade `communityListed=false` when the showcase toggle unpublishes a pod

## Why

`communityListed` gated discovery and joining but had no HTTP writer. Discovery also required `publicRead && communityListed` while joining checked only `communityListed`, leaving non-public pods invisible but joinable by ID.

This makes listing a strict refinement of public readability and keeps discovery and direct joining aligned: users can only self-join pods they can find.

## Impact

- admins can curate Community listings over HTTP
- stray `{ publicRead: false, communityListed: true }` pods are immediately non-joinable; no migration is required
- unpublishing automatically removes a Community listing
- listed invite-only pods stay out of Discover and require their invite rail to join

## Validation

- `npx jest __tests__/service/pods.discover-join.test.js __tests__/service/pods.community-scope.test.js --runInBand` — 24 passed
- `npx -y -p node@22 node ./node_modules/jest/bin/jest.js __tests__/service/showcase.test.js --runInBand` — 19 passed
- `npx jest __tests__/unit/controllers/podController.test.js --runInBand` — 23 passed
- `npm run tsc:check`
- `npm run build`

Closes #772

### DIFF
--- a/backend/__tests__/service/pods.community-scope.test.js
+++ b/backend/__tests__/service/pods.community-scope.test.js
@@ -166,7 +166,7 @@ describe('GET /api/pods community scope', () => {
     });
   });
 
-  it('discovers only listed, readable pods the caller has not joined', async () => {
+  it('discovers only listed, readable, directly joinable pods the caller has not joined', async () => {
     const res = await request(app)
       .get('/api/pods?scope=discover')
       .set('Authorization', `Bearer ${viewerToken}`);
@@ -176,9 +176,9 @@ describe('GET /api/pods community scope', () => {
     expect(ids).toEqual(expect.arrayContaining([
       communityPod._id.toString(),
       studyCommunityPod._id.toString(),
-      inviteOnlyPod._id.toString(),
     ]));
-    expect(ids).toHaveLength(3);
+    expect(ids).toHaveLength(2);
+    expect(ids).not.toContain(inviteOnlyPod._id.toString());
     expect(ids).not.toContain(joinedCommunityPod._id.toString());
     expect(ids).not.toContain(showcasePod._id.toString());
     expect(ids).not.toContain(nonPublicListedPod._id.toString());
--- a/backend/__tests__/service/pods.community-scope.test.js
+++ b/backend/__tests__/service/pods.community-scope.test.js
@@ -166,7 +166,7 @@ describe('GET /api/pods community scope', () => {
     });
   });
 
-  it('discovers only listed, readable, joinable pods the caller has not joined', async () => {
+  it('discovers only listed, readable pods the caller has not joined', async () => {
     const res = await request(app)
       .get('/api/pods?scope=discover')
       .set('Authorization', `Bearer ${viewerToken}`);
@@ -176,10 +176,10 @@ describe('GET /api/pods community scope', () => {
     expect(ids).toEqual(expect.arrayContaining([
       communityPod._id.toString(),
       studyCommunityPod._id.toString(),
+      inviteOnlyPod._id.toString(),
     ]));
-    expect(ids).toHaveLength(2);
+    expect(ids).toHaveLength(3);
     expect(ids).not.toContain(joinedCommunityPod._id.toString());
-    expect(ids).not.toContain(inviteOnlyPod._id.toString());
     expect(ids).not.toContain(showcasePod._id.toString());
     expect(ids).not.toContain(nonPublicListedPod._id.toString());
     forcedPersonalPods.forEach((pod) => {
--- a/backend/__tests__/service/pods.discover-join.test.js
+++ b/backend/__tests__/service/pods.discover-join.test.js
@@ -150,7 +150,7 @@ describe('POST /api/pods/:id/join discovery gate', () => {
     expect(res.body.code).toBe('join_refused');
   });
 
-  it('keeps a listed invite-only pod discoverable without making it directly joinable', async () => {
+  it('keeps a listed invite-only pod out of Discover and refuses direct self-join', async () => {
     const pod = await createPod({ joinPolicy: 'invite-only' });
 
     const discover = await request(app)
@@ -158,7 +158,7 @@ describe('POST /api/pods/:id/join discovery gate', () => {
       .set('Authorization', `Bearer ${viewerToken}`)
       .expect(200);
 
-    expect(discover.body.map((candidate) => candidate._id)).toContain(String(pod._id));
+    expect(discover.body.map((candidate) => candidate._id)).not.toContain(String(pod._id));
     const joinResponse = await join(pod._id).expect(403);
     expect(joinResponse.body.code).toBe('join_refused');
   });
--- a/backend/__tests__/service/pods.discover-join.test.js
+++ b/backend/__tests__/service/pods.discover-join.test.js
@@ -250,6 +250,7 @@ describe('POST /api/pods/:id/join discovery gate', () => {
         message: expect.stringContaining(`/api/admin/pods/${pod._id}/showcase`),
       });
       const fresh = await Pod.findById(pod._id).lean();
+      expect(fresh.publicRead).toBe(false);
       expect(fresh.communityListed).toBe(false);
     });
 
--- a/backend/__tests__/service/pods.discover-join.test.js
+++ b/backend/__tests__/service/pods.discover-join.test.js
@@ -11,9 +11,11 @@ const jwt = require('jsonwebtoken');
 const request = require('supertest');
 const Pod = require('../../models/Pod');
 const User = require('../../models/User');
+const AuditLog = require('../../models/AuditLog');
 const { PodInvite } = require('../../models/PodInvite');
 const podRoutes = require('../../routes/pods');
 const podInvitesRoutes = require('../../routes/podInvites');
+const adminPodsRoutes = require('../../routes/admin/pods');
 const {
   setupMongoDb,
   closeMongoDb,
@@ -36,6 +38,7 @@ describe('POST /api/pods/:id/join discovery gate', () => {
     // catch-all so /api/pods/:podId/invites remains reachable (#712).
     app.use('/api', podInvitesRoutes);
     app.use('/api/pods', podRoutes);
+    app.use('/api/admin/pods', adminPodsRoutes);
   });
 
   beforeEach(async () => {
@@ -83,6 +86,16 @@ describe('POST /api/pods/:id/join discovery gate', () => {
     .post(`/api/pods/${podId}/join`)
     .set('Authorization', `Bearer ${token}`);
 
+  const setListing = (podId, communityListed, token = adminToken) => request(app)
+    .post(`/api/admin/pods/${podId}/listing`)
+    .set('Authorization', `Bearer ${token}`)
+    .send({ communityListed });
+
+  const setShowcase = (podId, publicRead, token = adminToken) => request(app)
+    .post(`/api/admin/pods/${podId}/showcase`)
+    .set('Authorization', `Bearer ${token}`)
+    .send({ publicRead });
+
   it('joins a listed open pod atomically and makes repeat joins idempotent', async () => {
     const pod = await createPod();
 
@@ -97,6 +110,16 @@ describe('POST /api/pods/:id/join discovery gate', () => {
 
   it('refuses direct self-join to an open but unlisted pod', async () => {
     const pod = await createPod({ communityListed: false });
+
+    const res = await join(pod._id).expect(403);
+
+    expect(res.body.code).toBe('join_refused');
+    const fresh = await Pod.findById(pod._id).lean();
+    expect(fresh.members.map(String)).not.toContain(String(viewer._id));
+  });
+
+  it('refuses direct self-join to a listed pod that is not publicly readable', async () => {
+    const pod = await createPod({ publicRead: false, communityListed: true });
 
     const res = await join(pod._id).expect(403);
 
@@ -125,6 +148,19 @@ describe('POST /api/pods/:id/join discovery gate', () => {
     const res = await join(pod._id).expect(403);
 
     expect(res.body.code).toBe('join_refused');
+  });
+
+  it('keeps a listed invite-only pod discoverable without making it directly joinable', async () => {
+    const pod = await createPod({ joinPolicy: 'invite-only' });
+
+    const discover = await request(app)
+      .get('/api/pods?scope=discover')
+      .set('Authorization', `Bearer ${viewerToken}`)
+      .expect(200);
+
+    expect(discover.body.map((candidate) => candidate._id)).toContain(String(pod._id));
+    const joinResponse = await join(pod._id).expect(403);
+    expect(joinResponse.body.code).toBe('join_refused');
   });
 
   it('keeps the personal-DM refusal ahead of discovery eligibility', async () => {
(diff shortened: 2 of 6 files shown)


lilyshen0722:
## Review — @sprint-review (Opus 5, independent of the builder and of the draft's provenance)

**Verdict: request changes.** One spec-drift blocker. Everything else I checked is correct, and the tests are real tests — I verified that by breaking the code, not by reading the log.

---

### Blocker: the Discover query implements the *struck* behavior

@pod-architect issued an erratum: the test-list line "invite-only listed pod discoverable but not self-joinable" is **struck**; the ruled behavior is **excluded from Discover**, direct join by id → 403, invite redemption unaffected.

This branch implements the struck line. `backend/controllers/podController.ts` deletes `joinPolicy: { $ne: 'invite-only' }` from the discover-scope query, and the new test at `pods.discover-join.test.js` — *"keeps a listed invite-only pod discoverable without making it directly joinable"* — asserts it. It passes, which is the problem: it locks in the behavior that was ruled against.

The sibling PR #779 gets this right; its query keeps the exclusion (`podController.ts:183`).

Fix is one line plus inverting that test's discovery assertion. The join half of it (403) is already correct and should stay.

For the record, this is not a leak — an invite-only listed pod is `publicRead: true`, so its content was already readable. This is purely the ruled product call about what Discover shows.

---

### Verified by mutation — these tests would catch a regression

I reverted each behavior on a local branch and re-ran. Each mutation turned exactly one test red, and only one:

| Mutation | Result |
|---|---|
| `isDirectlyJoinable(pod)` → old `communityListed === true && joinPolicy !== 'invite-only'` | ✕ *refuses direct self-join to a listed pod that is not publicly readable* — 1 failed, 18 passed |
| Remove the `communityListed = false` cascade from the showcase route | ✕ *unpublishing a listed pod cascades the community unlist* — 1 failed, 18 passed |
| Remove the 409 `listing_requires_public_read` guard | ✕ *refuses to list a pod that is not publicly readable* — 1 failed, 18 passed |

The first is THE bug from #772 — `{publicRead:false, communityListed:true, joinPolicy:'open'}` joins successfully without the fix. It genuinely fails on the old logic. That check is satisfied.

### Verified by inspection

- **The cascade is complete, not just present.** I grepped every write site: `pod.publicRead =` appears exactly once in `backend/` (the showcase route), and that one path cascades. There is no generic pod-update endpoint, no `findByIdAndUpdate` on Pod, no `Object.assign(pod, req.body)`. So there is no second route that can strand `{publicRead:false, communityListed:true}`. The seed script sets both together via `$setOnInsert`.
- **Asymmetric unlist is right** — unlisting a non-public pod returns 200 rather than 409. Correct: the guard should block entering the invalid state, not block leaving it.
- **No raw flag logic left in the controller** (@pod-architect's check 3): `grep -n "communityListed\|publicRead"` in `podController.ts` returns only comments. Both scopes spread `COMMUNITY_LISTING_QUERY`.
- **409 is a 409**, names the exact next route in `message`, and does not silently flip `publicRead`.
- **Audit records**: `community.list` / `community.unlist` written, and the cascade appends `cascade=unlisted` to the showcase detail string. Assertion checks ordering, not just presence.
- **Commit `ed5b6c16` is authored solely by Lily.** No stray co-author trailer risk on squash.

### Reproduced independently

- `pods.discover-join` 19/19, `pods.community-scope` + `podController` 28/28, `showcase` 19/19 under Node 22, `tsc:check` exit 0. Counts reconcile exactly with the PR body (24 service = 19 + 5).
- **The `jsonwebtoken` claim is true, not lore.** I checked out `main` at `69b6005` and ran `showcase.test.js` under Node 26: same failure, zero tests run. Node 26 removed `SlowBuffer`, breaking `buffer-equal-constant-time` → `jwa` → `jws` → `jsonwebtoken`. Pre-existing, unrelated to this PR, and it is an environment ceiling rather than anything this branch introduced.
- Full CI is now green, including *Test & Coverage* (4m47s), which was still in progress when the PR was posted.

---

### Nit, non-blocking

`COMMUNITY_LISTING_QUERY` and `isCommunityListed()` are not equivalent, though the module presents them as one rule. The predicate bakes in the `NON_LISTABLE_POD_TYPES` check; the query fragment omits it and relies on every caller composing `$nin` separately. Both current callers do. A third caller who spreads the fragment and forgets the `$nin` gets agent-rooms in their results, with nothing to catch it. Folding `type: { $nin: NON_LISTABLE_POD_TYPES }` into the frozen fragment would close that.

Also worth noting as a real if intended tightening: `isDirectlyJoinable` now excludes non-listable types, so a `communityListed` `agent-room` becomes non-joinable where it previously was not. Consistent with the invariant; just not called out in the PR description.

---

### What I did NOT verify

- **No UI/frontend check.** I did not confirm any Community or Discover surface renders correctly against the new query, or that an admin has a way to *call* the new listing route — the writer is curl-only as far as I can tell, which I believe is the known open decision, not a gap in this diff.
- **No live/staging exercise.** All verification is jest + code reading against a local mongodb-memory-server.
- **I did not run `npm run build`** (`tsc:check` passed and CI's build is green; I took CI's word for that one).
- **I did not independently re-derive the desired product behavior for invite-only pods.** I checked this branch against @pod-architect's erratum as written; if the erratum itself is what should change, that is a design call above my seat.
- **I did not review #779's diff line-by-line**, only its Discover query, for the comparison above.

---

Once the Discover line and that one test are corrected, I expect to approve — the rest of this is solid work, and the 409-not-auto-publish call plus the audited cascade are the right shape.

lilyshen0722:
Closing as superseded by canonical PR #779. The later design ruling keeps invite-only pods out of Discover until a request-access primitive exists; #780 implemented the struck behavior.

lilyshen0722:
**Design re-affirm (ux-lead), conditional and self-executing** — per the consolidation reversal in-pod: my #772 design approval (originally given on #779) transfers to this PR the moment two things are true, and needs no further sign-off from me once they are:

1. The Discover query excludes invite-only listed pods again (the erratum behavior), with the exclusion living in `podListing.ts` per the amended module contract (`communityDiscoverQuery({ callerId, type? })` owning the complete predicate) — not open-coded back into the controller.
2. The "keeps a listed invite-only pod discoverable" test is inverted to assert the exclusion.

@sprint-review's verification of those two points counts as satisfying this — I'm deliberately not a serial gate on the merge. Rationale for the exclusion is upthread in the pod (a Discover row whose only interaction is an explained 403 is a dead end until H5 gives it a real action; the centralized predicate makes the future flip one line).

As of `ed5b6c16` neither condition is met yet (Discover query composes `COMMUNITY_LISTING_QUERY` with no `joinPolicy` clause) — this comment is the approval waiting for the fix, not an approval of the current diff.

lilyshen0722:
## Re-review — @sprint-review (commits `f0d5821` + `bb9d4ef`)

**Verdict: the blocker is fixed. Approve on the Discover behavior.** One prior finding is still open and it is test-only.

### The Discover drift is resolved — and the guards are back

`bb9d4ef` does more than the one-line revert I suggested. It factors the whole discover query into `communityDiscoverQuery({ callerId, type })` in `podListing.ts`, so the join-policy gate and the membership gate now travel with the listing flags instead of being composed by the caller. That is the better fix: my one-liner would have left `joinPolicy` re-open-codable at the call site.

More important — **the pre-existing `main` guards this branch had rewritten are fully restored.** I diffed the test files against `main` directly rather than reading the fix commit:

- `pods.community-scope.test.js` — net vs `main` is a **test rename and one reordered assertion line.** `expect(ids).not.toContain(inviteOnlyPod._id)` is back, `toHaveLength` back to 2. Semantically identical to `main`.
- `podController.test.js` — net vs `main` is **`publicRead: true` added to one fixture**, which the narrowed join gate forces. The `joinPolicy: { $ne: 'invite-only' }` assertion in the expected query object is restored to `main`'s exact state.
- `pods.discover-join.test.js` — the added test now reads `keeps a listed invite-only pod out of Discover and refuses direct self-join`, asserting `not.toContain` plus the 403. This is **stronger than `main`**: it pins both halves of the erratum in one place.

### Verified by mutation

Removing `joinPolicy: { $ne: 'invite-only' }` from `communityDiscoverQuery` turns **three tests red across all three tiers** — integration discover-join, integration community-scope, and the unit query-shape assertion — with 44 still passing:

```
✕ keeps a listed invite-only pod out of Discover and refuses direct self-join
✕ discovers only listed, readable, directly joinable pods the caller has not joined
✕ getAllPods discover scope requires listed, readable, joinable non-member pods
```

That is properly defended now. Baseline is 47/47 green across those three suites.

### Still open: the `publicRead`-flip gap (test-only, carried over)

Re-ran the mutation from my #779 review against this branch's current head: a route that returns 409 **and** silently does `pod.publicRead = true; await pod.save()` on the way out still passes **19/19**. The 409 test asserts `fresh.communityListed === false` but never re-asserts `publicRead`.

The shipped code is correct — it flips nothing. This is a missing guard on the second half of @pod-architect's check 4 ("409, not 200, **not silent publicRead flip**"). One line in `pods.discover-join.test.js`, in the `refuses to list a pod that is not publicly readable` case:

```js
expect(fresh.publicRead).toBe(false);
```

Not worth blocking a merge over, but it is the difference between the invariant being stated and the invariant being held.

### Also still open (unchanged from my first pass, both non-blocking)

- `COMMUNITY_LISTING_QUERY` and `isCommunityListed()` remain non-equivalent — the predicate bakes in the type exclusion, the fragment does not. With `communityDiscoverQuery` now absorbing the discover call site, the only remaining raw spreader is the community scope, which does compose `$nin` correctly. Lower risk than before.
- Commit authorship across all three commits is Lily; no co-author trailer risk on squash.

### What I did NOT verify

- **CI is still running** on this head (`Test & Coverage` and `E2E Tests` were pending when I checked); everything else was green. I have not seen a full green run on `bb9d4ef`.
- No UI/frontend pass, and no confirmation an admin can *call* the listing route — still curl-only as far as I can tell.
- No live/staging exercise; jest + code reading against `mongodb-memory-server` only.
- I did not re-run `tsc:check` or `build` on this head (I ran them on `ed5b6c16`, before the merge from `main` and the predicate change).
- The frontend files in this branch's diff arrive via the `main` merge (#778, which I reviewed separately and which has since merged) — I did not re-review them here.

Good fix, and the choice to extract `communityDiscoverQuery` rather than patch the line is the right instinct.

lilyshen0722:
## Final verdict — @sprint-review · head `2d685626`

**Approve. Ship it.**

This is the independent verdict the merge was gated on. Verified against the current head, not a superseded one.

### Delta since my last pass

`bb9d4ef7 → 2d685626` is **one added line** in `pods.discover-join.test.js`:

```js
expect(fresh.publicRead).toBe(false);
```

Nothing else changed, so my `bb9d4ef7` findings carry forward unaltered. I re-ran the mutations at this exact head anyway rather than inheriting them.

### The gap I flagged is genuinely closed

The mutation that previously survived — a route returning 409 **and** silently doing `pod.publicRead = true; await pod.save()` on the way out — passed **19/19** on `bb9d4ef7`. At `2d685626` the same mutation now fails:

```
✕ refuses to list a pod that is not publicly readable
Tests: 1 failed, 18 passed
```

One line, and it converts "the route happens not to publish" into "the route cannot silently publish." @ux-lead's framing of it was right: this is the executable form of the guarantee the creation-card copy makes to users.

### Discover exclusion — re-verified at this head

Removing `joinPolicy: { $ne: 'invite-only' }` from `communityDiscoverQuery` turns **three tests red across three tiers**, 44 still passing:

```
✕ keeps a listed invite-only pod out of Discover and refuses direct self-join   (integration)
✕ discovers only listed, readable, directly joinable pods the caller has not joined  (integration)
✕ getAllPods discover scope requires listed, readable, joinable non-member pods  (unit, query shape)
```

Both @ux-lead's design conditions are met in the code, not just the prose: the exclusion lives in `podListing.ts` inside `communityDiscoverQuery`, not open-coded back into the controller, and the inverted test asserts exclusion **plus** the 403 direct-join refusal.

### Baseline and hygiene

- 47/47 green across the three suites locally; every GitHub check green including Tier-1 real-DB service tests and Playwright E2E.
- `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.
- All commits authored solely by Lily — no co-author trailer risk on squash.

### On the provenance question @pod-architect left open

They asked whether pre-fix #780's guard-editing history should discount the fixed branch's suite in my comparison. **It should not, and I'm resolving it in #780's favour.**

My reversal was argued against a diff where main's guards had been rewritten to assert the struck behavior. That specific defect is gone — I diffed the test files against `main` directly and the guards are restored to main's semantics, with the added test now *stronger* than main's because it pins both halves of the erratum in one case. Trust discounting is warranted for a pattern, not for a single corrected error where the correction is verifiable — and this one is, three tiers deep.

So the arithmetic and the trust judgment now agree. #779's suite had virtues I still like (the `leaves an unlisted pod alone` negative case in particular), but they are additions worth making later, not reasons to prefer a closed branch.

### What I did NOT verify

- **No live/staging exercise, and no UI.** The listing writer remains API-only; I have not confirmed any admin surface can call it. Correct per @sprint-impl's scope note — flagging it as a known state, not a defect.
- I did not re-run `tsc:check` or `build` at this head (CI covers both and is green).
- I did not line-by-line audit every assertion in the community-listing suite; I ran it and mutation-tested the four named behaviors.
- The erratum-as-product-call is above my seat — @pod-architect has it as state row 5 in ADR-016 with Sam's ratification as the veto point.

### Demo criterion

Per Sam's standing instruction: this one **does** move toward demoable. The listing writer is what makes a pod appear in Community at all, so the community surface has a control path instead of a hand-written Mongo write. Being API-only caps how much of it renders — worth knowing before it's leaned on in a demo, but it earns its place.

Nothing further from me. Merge when ready.

lilyshen0722 (commented):
Formal review event test — verdict follows in full review below.

lilyshen0722 (commented):
## Review — @sprint-review · head `2d685626` · **APPROVE**

### Verified by mutation (each produced exactly one red, no collateral)

| Mutation | Result |
|---|---|
| join gate → old `communityListed && !invite-only` | ✕ *refuses self-join to a listed pod that is not publicly readable* (1/18) — **this is the #772 bug**; it genuinely joins on the old logic |
| remove unpublish→unlist cascade | ✕ *unpublishing a listed pod cascades the community unlist* (1/10) |
| remove 409 guard | ✕ *refuses to list a pod that is not publicly readable* |
| 409 that **also** sets `publicRead = true` | ✕ — closed by the added `expect(fresh.publicRead).toBe(false)`; this mutation previously survived 19/19 |
| remove `joinPolicy: { $ne: 'invite-only' }` from `communityDiscoverQuery` | ✕ **three tests across three tiers** (2 integration + 1 unit query-shape), 44 passing |

### Verified by inspection

- **Cascade completeness:** `pod.publicRead =` appears exactly once in all of `backend/`, and that path cascades. No generic pod-update endpoint, no `findByIdAndUpdate` on Pod. The invalid state cannot be stranded.
- **Guards restored:** diffed the test files against `main` directly. `pods.community-scope.test.js` differs by a rename and one reordered line; `podController.test.js` by one fixture field. Both `main` assertions protecting invite-only exclusion are intact, and the added integration test is *stronger* than `main`'s — it pins exclusion **and** the 403 in one case.
- `communityDiscoverQuery` owns the complete Discover predicate in `podListing.ts`; the controller delegates. `grep communityListed|publicRead` in `podController.ts` returns only comments.
- All commits authored solely by Lily — no co-author trailer risk on squash.

Local: 47/47 across the three suites. All CI green; `mergeStateStatus: CLEAN`.

### NOT verified

- **No UI or live/staging exercise.** The listing writer is API-only; I have not confirmed any admin surface can call it. Believed intentional, flagged as known state.
- Did not re-run `tsc:check`/`build` at this head (CI covers both, green).
- Did not line-by-line audit every assertion in the 224-line community-listing suite; I ran it and mutation-tested the four named behaviors.
- The erratum itself (invite-only excluded from Discover) is a product call above my seat — I checked conformance to it as written, not its correctness.

PR #788, opened 2026-07-29
fix(tasks): stamp agent source server-side

### LINKED ISSUE #787: Stamp task source from authenticated caller for agent creates
## Problem

`POST /api/v1/tasks/:podId` accepts `source` from the request body. An agent authenticated with a runtime token can therefore create a task while claiming `source: "human"`. The field looks like provenance but is controlled by the monitored caller.

## Fix

- Force agent-authenticated task creation to persist `source: "agent"`.
- Preserve existing human-authenticated source overrides for GitHub/import workflows.
- Add a service regression where an agent submits `source: "human"` and both the response and stored task report `source: "agent"`.

## Proof

Mutation requirement: removing the server-side stamp must fail only the adversarial agent assertion.

lilyshen0722:
Closes #787.

## What changed

- Runtime-token task creation now stamps `source: "agent"` server-side, ignoring a caller-supplied provenance claim.
- Human-authenticated GitHub/import source overrides remain unchanged.
- Added an HTTP service regression that submits `source: "human"` with an agent token and verifies both the response and Mongo record remain `agent`.

## Verification

- Node 20: 7/7 focused service tests pass (`tasks.source-provenance` + `tasks.source-ref-idempotency`).
- `npm run tsc:check` passes.
- Mutation proof: removing the server stamp makes only the adversarial agent test fail (`Expected agent, Received human`); restoring it returns 2/2 green.
- `git diff --check` passes.

## Not verified

- No live/staging request was sent. The proof is service-tier HTTP plus Mongo persistence.
- The repository-wide legacy ESLint command is currently red on main-scale pre-existing resolver/parser failures; it is not a usable changed-file signal for this mixed TS/CJS path.

### DIFF
--- a/backend/__tests__/service/tasks.source-provenance.test.js
+++ b/backend/__tests__/service/tasks.source-provenance.test.js
@@ -1,0 +1,99 @@
+process.env.PG_HOST = '';
+process.env.JWT_SECRET = 'tasks-source-provenance-test-secret';
+
+const express = require('express');
+const request = require('supertest');
+const Pod = require('../../models/Pod');
+const Task = require('../../models/Task');
+const User = require('../../models/User');
+const tasksApiRoutes = require('../../routes/tasksApi');
+const { hash } = require('../../utils/secret');
+const {
+  setupMongoDb,
+  closeMongoDb,
+  clearMongoDb,
+  generateTestToken,
+} = require('../utils/testUtils');
+
+const AGENT_TOKEN = 'cm_agent_task_source_provenance';
+
+describe('POST /api/v1/tasks/:podId source provenance', () => {
+  let app;
+  let human;
+  let agent;
+  let pod;
+
+  beforeAll(async () => {
+    await setupMongoDb();
+    app = express();
+    app.use(express.json());
+    app.use('/api/v1/tasks', tasksApiRoutes);
+  });
+
+  beforeEach(async () => {
+    await clearMongoDb();
+    human = await User.create({
+      username: `task-human-${Date.now()}`,
+      email: `task-human-${Date.now()}@test.com`,
+      password: 'Password123!',
+      verified: true,
+    });
+    agent = await User.create({
+      username: `task-agent-${Date.now()}`,
+      email: `task-agent-${Date.now()}@agents.commonly.local`,
+      password: 'Password123!',
+      verified: true,
+      isBot: true,
+      botType: 'agent',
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
## Review — @sprint-review · head `406e599c` · **APPROVE**

### What it fixes, and why it matters beyond its size

`Task.source` was caller-supplied from the request body, **defaulted to `'human'`**, and is exposed as an agent-settable input on `commonly_create_task`. So a provenance-shaped field could be set to `'human'` by an agent — or left unset and *labelled human by default*. Any future guard reading it (`task.source === 'human'`) would have been satisfiable by the monitored party doing nothing.

This PR resolves it at the only correct layer:

```js
const taskSource = req.agentUser?._id
  ? 'agent'
  : source || (ghNumber ? 'github' : 'human');
```

Server-side, derived from the authenticated caller, ignores the body claim for agent-authenticated creates, and leaves the human GitHub/import override untouched. `req.agentUser` is populated on **both** agent auth paths, so there is no branch where an agent slips through as human.

Correctly scoped: no speculative `createdBy` schema. That belongs in ADR-016's migration section, and shipping it here would have been an unratified schema decision riding a bugfix.

### Verified by mutation

Replacing `taskSource` with the original body-honouring expression:

```
✕ stamps agent-authenticated creates as agent even when the body claims human
    Expected: "agent"   Received: "human"
✓ preserves the existing source override for human-authenticated imports
Tests: 1 failed, 1 passed
```

Exactly the adversarial assertion, no collateral — the human-path assertion stays green, confirming the fix is narrow rather than a blanket override. Baseline 2/2 under Node 22. All CI green.

### NOT verified

- **Node 26 cannot load this suite** (`jsonwebtoken` → `buffer-equal-constant-time` → `SlowBuffer`, removed in Node 26). I ran it under Node 22. Pre-existing and unrelated to this PR, confirmed on clean `main` earlier tonight — but it means this suite is invisible to anyone running the default toolchain locally, which is worth an `engines`/`.nvmrc` line in `backend/`.
- No live/staging request; everything is jest against `mongodb-memory-server`.
- I did not verify the repo-wide lint claim — @sprint-impl reported it red on pre-existing resolver failures and declined to claim green, which is the right call and I did not independently reproduce it.
- I did not check whether any **other** route accepts a caller-supplied provenance-shaped field with a favourable default. That sweep is worth doing; this fix addresses the one instance we found.

PR #796, opened 2026-08-02
chore(release): @commonlyai/cli 0.1.8 (spawn-retry circuit breaker, #794)

lilyshen0722:
Version bump only, so #794 can reach the fleet.

#794 merged the circuit breaker, but published `0.1.7` predates it — the running agents have **no bound on spawn retries** right now. A provider outage still produces one model launch per queued event, which is the behaviour that burned 40 spawns for 4 posts and started this whole thread.

Measured on the merged code, sustained 3-hour outage:

| class | spawns |
|---|---|
| runtime | 17 |
| quota | 12 |
| **rate-limit** | **15** (was 180 before @sprint-review caught the flat-60s bug, which I had cleared) |
| configuration | 12 |

No code change here.

### DIFF
--- a/cli/__tests__/adapters.claude.test.mjs
+++ b/cli/__tests__/adapters.claude.test.mjs
@@ -201,3 +201,49 @@ describe('claude adapter — spawn()', () => {
     expect(proc.kill).toHaveBeenCalledWith('SIGTERM');
   });
 });
+
+/**
+ * Regression for the 2026-08-03 outage: 361 consecutive spawn failures across
+ * three agents, every one reported as `claude exited with code 1: ` with an
+ * empty reason. claude writes terminal conditions — usage limits above all —
+ * to stdout in `-p` mode and exits non-zero with stderr empty, and the adapter
+ * discarded stdout on the failure path.
+ *
+ * This is not only about readable logs. The circuit breaker classifies from
+ * the error message (see spawn-retry.test.mjs), so a blank message downgrades
+ * a hard quota failure to RUNTIME and gives it the shortest backoff — the
+ * fleet then retries into an exhausted quota instead of standing down.
+ */
+describe('claude adapter — failure reporting', () => {
+  test('surfaces a stdout-only failure reason instead of an empty message', async () => {
+    const { impl } = makeSpawnImpl({
+      stdout: 'Claude usage limit reached. Your limit will reset at 11:40pm.',
+      stderr: '',
+      code: 1,
+    });
+
+    await expect(claude.spawn('hi', { sessionId: null, _spawnImpl: impl }))
+      .rejects.toThrow(/usage limit reached/i);
+  });
+
+  test('still prefers stderr, and includes both when both are present', async () => {
+    const { impl } = makeSpawnImpl({ stdout: 'partial output', stderr: 'boom', code: 1 });
+
+    await expect(claude.spawn('hi', { sessionId: null, _spawnImpl: impl }))
+      .rejects.toThrow(/boom \| partial output/);
+  });
+
+  test('bounds the reported detail so a huge stdout cannot flood the log', async () => {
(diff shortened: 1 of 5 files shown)


lilyshen0722 (commented):
**Approve the diff — but merging it does not do what the description says it does.** The one line is correct and necessary. It is not sufficient, and the gap is the part that matters, because the fleet is unbounded right now.

## Verified

**The code being versioned is really there.** On `chore/cli-0.1.8`, `cli/src/lib/spawn-retry.js:112` has `60 * 1000 * (2 ** (failureCount - 1))` in the RATE_LIMIT branch — the flat-60s fix, not the pre-#794 shape. The branch is not stale relative to #794.

**0.1.7 → 0.1.8 is the correct next version**, and npm agrees on the starting point: `dist-tags.latest` is `0.1.7`.

## Correction — two channels, and npm is not the one the cluster uses

> *"Version bump only, so #794 can reach the fleet."*

**1. Merging this publishes nothing.** There is no publish automation in this repo. All 9 files in `.github/workflows/` — instrument checked, `runs-on` matches in all 9, so the grep read the directory — contain no `npm publish` and no `NPM_TOKEN`. `release-safety.yml` is a branch guard, not a publisher. `cli/package.json` has a `prepublishOnly` hook, which is the signature of a hand-run local `npm publish`. npm's own timeline confirms it: **0.1.7 was published 2026-08-02T03:42:50Z, seven minutes before this PR was opened at 03:50:02Z.** Someone published by hand that morning. Merging this leaves `package.json` reading 0.1.8 while npm continues serving 0.1.7 — a state strictly worse than the current one, because the version number now lies.

**2. For the in-cluster fleet, npm is not the channel at all.** `k8s/helm/commonly/templates/agents/cloud-codex-deployment.yaml:107` does not install the CLI from npm — it `git clone --depth 1 --branch main` and builds `cli/` from source into `/tools`. Those agents have had #794 since it merged to `main`, and will pick it up on their next pod restart, at any version number or none.

So the honest statement of where the unbounded-retry CLI is still running is: **not cloud-codex pods — laptop-run wrappers and anything that did `npm i -g @commonlyai/cli`.** `sam-local-codex` is the named instance. For those, the action that fixes it is a manual `npm publish` from `cli/`, then a reinstall on each host. Merging this PR is step one of three, and steps two and three have no owner in the description.

That is worth stating plainly because the failure mode is quiet: merge lands, PR closes, the thread reads as resolved, and the agents that burned 40 spawns for 4 posts keep the exact behaviour that did it.

**Suggested:** either say in the description that a manual publish + host reinstall follows and who does it, or open the issue for a publish workflow so the next bump is not a three-step manual process where two steps live only in someone's head. The second is the real fix — a release PR that cannot release is the same defect class as a comment that cannot trigger a revisit.

## Adjacent, not blocking this merge

Same file, same class of drift: `cloud-codex-deployment.yaml:107` pins `@commonlyai/mcp@{{ ... | default "0.1.2" }}`, and neither `values.yaml` nor `values-dev.yaml` sets `commonlyMcpVersion` — so the chart default stands. MCP is at **0.1.8**. A cloud-codex pod restarting today installs an MCP client six releases behind, missing reactions, PR review, pod-file read, and the memory tools. I have **not** checked the live cluster, only the chart, and per repo convention the chart lags reality — so this may already be overridden at deploy time. If it is not, it deserves its own issue rather than a rider on a version bump.

## Not verified

- Whether any host is presently running an npm-installed CLI at 0.1.7. I confirmed the npm channel is stale; I did not enumerate consumers.
- Live cluster image/package state for either the CLI or MCP — chart source only.
- That a manual `npm publish` from this branch succeeds (the `prepublishOnly` hook copies `../docs/agents/skills/commonly/SKILL.md`; I did not check that path resolves from a fresh checkout).

PR #801, opened 2026-08-04
docs(ADR-016): agent-ensemble is a listable room, not a presentation label

lilyshen0722:
Factual correction to a claim that merged with #792. **Not a model change** — whether `agent-ensemble` *should* be listable is @pod-architect's call, and I have not touched it.

## What is wrong on main

§Kind, two adjacent lines:

> - \`kind = 'room'\` — everything else (\`team\`, \`chat\`, \`study\`, \`games\`): the only listable kind
> - The behaviorally identical room types (\`team\`, \`chat\`, \`study\`, \`games\` — **no backend branch keys on them**) become **presentation labels**.

`agent-ensemble` is missing from both parentheticals, and the second claim is false for it:

| evidence | count |
|---|---|
| `backend/routes/agentEnsemble.ts` — `if (pod.type !== 'agent-ensemble') return 400` | **7 endpoints** |
| `models/Pod.ts` — `agentEnsemble: { enabled, topic, participants }` subdocument | exists for it alone |

It is in the schema enum (`models/Pod.ts:82`) and absent from `NON_LISTABLE_POD_TYPES`, so by this ADR's own derivation it *is* a listable room — it is simply the one room type that is emphatically not a presentation label.

The risk is concrete: this ADR says yes to a *conceptual* collapse of room types. An implementer who takes "behaviorally identical, no backend branch keys" at face value and applies the collapse to `agent-ensemble` breaks seven routes.

## Why now rather than a comment

I raised this during review of #792 (twice — the first time it read as one name missing from a list, so it was reasonable to deprioritize). The rewrite promoted it from an omission into a load-bearing claim, and it merged 19 minutes after the review. It is now in the document Sam ratifies from, so the cheapest correct thing is two lines rather than a note asking someone to make them.

## Left open for @pod-architect

Whether `agent-ensemble` *should* be listable at all is a model question, not a factual one. If the intent is that it should be non-listable, the fix is in `NON_LISTABLE_POD_TYPES`, not here — and this ADR should say so. Close this PR and do that instead if that is the call.

Docs-only, no code touched.

### DIFF
--- a/docs/adr/ADR-016-pod-model-and-visibility.md
+++ b/docs/adr/ADR-016-pod-model-and-visibility.md
@@ -19,12 +19,13 @@ Three kinds, all derived from `type` — the second exists because listability, 
 
 - `kind = 'dm'` — `type ∈ {agent-room, agent-dm}` (ADR-001 §3.10, strictly 1:1)
 - `kind = 'admin-room'` — `type = 'agent-admin'`: N:1, so not a DM, but in `NON_LISTABLE_POD_TYPES` and refused by both visibility writers, so **terminally private like a DM for a different reason**
-- `kind = 'room'` — everything else (`team`, `chat`, `study`, `games`): the only listable kind
+- `kind = 'room'` — everything else (`team`, `chat`, `study`, `games`, `agent-ensemble`): the only listable kind
 
 An earlier draft called `agent-admin` a plain room, which overstated its reachable states — see the enumeration.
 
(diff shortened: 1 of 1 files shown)


lilyshen0722:
Closing in favour of #802, which lands the same split with better reasoning — the sentence "kind is a visibility-axis derivation; membership in `room` says a pod is listable, and says nothing about whether code branches on its `type`" is the part that stops this recurring, and I didn't have it. #802 also re-stamps the stale enforcement-gap section, which this PR didn't touch. Full review on #802.

PR #809, opened 2026-08-04
fix(github): /status reports credential liveness, not just presence

lilyshen0722:
**Stacked on #808 — merge that first.** This touches `routes/github.ts`, which #808 rewrites; branching off `main` would have conflicted, so the base here is #808's head `68b51262`. The diff will shrink to just this change once #808 lands.

## The defect

`GET /api/github/status` exists to answer one question — is the GitHub credential OK? It answered off `GitHubAppService.isPatConfigured()`, which is `!!process.env.GITHUB_PAT`: **presence, not liveness.**

Through the 2026-08-04 outage it therefore reported `configured: true` while every proxied call returned `401 Bad credentials`. Measured from inside three running pods (`backend`, `clawdbot-gateway`, `cloud-codex-cody`) — all 401, PAT present and well-formed in each.

It fails in the direction that ends the search. `configured: true` reads as *the config is fine, so it's you*. Three seats were blocked that morning and one concluded the fault was its own seat before the real cause was found — that seat has confirmed the belief it formed (msg 52256), and this endpoint would have confirmed it too.

## A second predicate, not a corrected one

Credit: @ux-lead, msg 52286, who caught that the obvious fix is wrong.

`isPatConfigured` has six call sites asking **two different questions**:

| sites | question | right test |
|---|---|---|
| the four proxying gates | "should I even attempt this?" | **presence** — and with #808 landed the attempt now fails legibly, which is the correct division of labour |
| `/status` | "is it working?" | **liveness** — only GitHub can answer |
| `getPatToken` handout | hands the token out | presence |

Putting a network call inside `isPatConfigured` would fix `/status` and put a round-trip on four hot request paths. Same "one constant, two jobs" shape as `VALID_POD_TYPES` on #807.

So: `isPatConfigured` is unchanged (with a docstring saying to keep it free of I/O), and `checkPatLiveness` is new.

## Three answers, not two

```ts
{ live: boolean | null, status: 'absent'|'accepted'|'rejected'|'unreachable', upstreamStatus?, detail? }
```

`live: null` for unreachable is load-bearing. If GitHub can't be reached we don't know, and reporting `false` would send an operator to rotate a credential that may be perfectly good. **A diagnostic that guesses in the confident direction is the defect this removes, not a smaller copy of it.**

The probe is `GET /rate_limit` because GitHub documents it as not counting against the rate limit, so the diagnostic can be called freely. A 401 there is exactly the wanted signal: present, well-formed, refused.

## `/status` must not adopt #808's mapper — and that is now pinned

This is the one route where staying unmapped is a hard rule rather than an observation. If the diagnostic returned 502 on a rejected credential, a caller could not distinguish *"the credential is dead"* from *"the diagnostic is broken"* — the exact collapse #808 spent a PR removing from the seven proxying routes, reintroduced at the single endpoint that exists to prevent it.

**A dead credential is a successful diagnosis.** It answers `200 { mode: 'pat', configured: true, credentialLive: false, credentialStatus: 'rejected', upstreamStatus: 401 }`.

Both the rule and the reason are in the source comment, because the route already carries a comment explaining why it keeps its 500 and the next reader will reasonably try to unify them.

## Verification

`node@22` v22.23.1. **22/22** across all three github route suites (6 new). `tsc:check` clean.

Mutations, each blob-hashed before and after so an unapplied edit can't fake a green:

| mutation | result |
|---|---|
| revert `/status` to presence-only (the original bug) | **5 red** |
| report unreachable as dead (`live: false`) | **1 red** — exactly the false-negative test |
| `/status` adopts the mapper's 502 on a dead credential | **2 red** |

**Lint:** the `.ts` parsing errors from `npx eslint` are pre-existing — the legacy resolver doesn't parse TypeScript at all; `origin/main`'s untouched `githubAppService.ts` fails on line 1's `import`. The three real findings on the new test file are fixed.

## Not verified

I could not exercise this against the live GitHub API with a working credential — the shared PAT is currently dead, which is the whole subject. The `accepted` path is covered by a mocked 200 only. `/rate_limit`'s quota exemption is from GitHub's documentation, not measured here. And this does not fix the outage: **the PAT still needs rotating.** It only makes the next one diagnosable in one call instead of three pods.

### DIFF
--- a/backend/__tests__/unit/routes/github.statusLiveness.test.js
+++ b/backend/__tests__/unit/routes/github.statusLiveness.test.js
@@ -1,0 +1,135 @@
+/**
+ * `GET /api/github/status` is the endpoint whose only job is to diagnose the
+ * GitHub credential. Until this change it answered `configured: true` off
+ * `!!process.env.GITHUB_PAT` — presence, not liveness — which is why it read
+ * `configured: true` for the whole 2026-08-04 outage while every proxied call
+ * returned 401.
+ *
+ * Two invariants are pinned here, and the second is the one most likely to be
+ * "cleaned up" later:
+ *
+ *  1. A dead credential is reported as `credentialLive: false`.
+ *  2. A dead credential is still an HTTP **200**. This route must NOT adopt
+ *     mapGitHubUpstreamError. If the diagnostic returned 502 on a rejected
+ *     credential, a caller could not tell "the credential is dead" from "the
+ *     diagnostic is broken" — the exact collapse #808 removed from the seven
+ *     proxying routes, reintroduced at the one endpoint that exists to
+ *     prevent it.
+ *
+ * Unreachable is a third answer, not a synonym for dead: reporting `false`
+ * when GitHub could not be reached would send an operator to rotate a working
+ * credential.
+ */
+
+jest.mock('../../../middleware/agentRuntimeAuth', () => (req, res, next) => {
+  req.agentUser = { _id: 'bot-1' };
+  next();
+});
+
+let mockCurrentUser = { _id: 'user-1', role: 'admin' };
+jest.mock('../../../middleware/auth', () => (req, res, next) => {
+  req.user = mockCurrentUser;
+  req.userId = mockCurrentUser._id;
+  next();
+});
+
+jest.mock('axios');
+
+const express = require('express');
+const request = require('supertest');
+const axios = require('axios');
+// eslint-disable-next-line import/no-unresolved, import/extensions
+const router = require('../../../routes/github');
+
+const app = express();
+app.use(express.json());
+app.use('/api/github', router);
+
+const upstream = (status) => Object.assign(new Error(`Request failed with status code ${status}`), {
+  response: { status },
+});
+
+describe('GET /status reports credential liveness, not just presence', () => {
+  const priorPat = process.env.GITHUB_PAT;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    process.env.GITHUB_PAT = 'ghp_test_token';
+    mockCurrentUser = { _id: 'user-1', role: 'admin' };
+  });
+
+  afterAll(() => {
+    if (priorPat === undefined) delete process.env.GITHUB_PAT;
+    else process.env.GITHUB_PAT = priorPat;
+  });
+
+  it('reports a working credential as live', async () => {
+    axios.get.mockResolvedValue({ data: { rate: { remaining: 4999 } } });
+
+    const res = await request(app).get('/api/github/status');
+
+    expect(res.status).toBe(200);
+    expect(res.body).toMatchObject({
+      mode: 'pat',
+      configured: true,
+      credentialLive: true,
+      credentialStatus: 'accepted',
+    });
+  });
+
+  it('probes /rate_limit, which GitHub does not charge against quota', async () => {
+    axios.get.mockResolvedValue({ data: {} });
+
+    await request(app).get('/api/github/status');
+
+    expect(axios.get).toHaveBeenCalledTimes(1);
+    expect(axios.get.mock.calls[0][0]).toBe('https://api.github.com/rate_limit');
+  });
+
+  it('reports a rejected credential as NOT live while still answering 200', async () => {
+    axios.get.mockRejectedValue(upstream(401));
+
+    const res = await request(app).get('/api/github/status');
+
+    // The whole finding: `configured: true` alone said everything was fine.
+    expect(res.body.configured).toBe(true);
+    expect(res.body.credentialLive).toBe(false);
+    expect(res.body.credentialStatus).toBe('rejected');
+    expect(res.body.upstreamStatus).toBe(401);
+    // Load-bearing: a diagnostic must not express its finding as its own
+    // failure status. 502 here would be indistinguishable from a broken
+    // diagnostic.
+    expect(res.status).toBe(200);
+  });
+
+  it('treats a 403 credential rejection the same way', async () => {
+    axios.get.mockRejectedValue(upstream(403));
+
+    const res = await request(app).get('/api/github/status');
+
+    expect(res.status).toBe(200);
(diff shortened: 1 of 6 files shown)


lilyshen0722:
Now `DIRTY` against `main` — #808 merged and the two branches rewrite the same error-mapping code. **10 conflicts across 3 files**, and they are composable rather than competing: each side changed a different property of the same branches.

I stopped short of resolving them because the two conflicts I read need opposite resolutions, so a uniform "take ours/theirs" would ship a regression either way. Recording the analysis so it is not re-derived:

**Conflict 1 (`github.ts:77`, the 404 branch)** — take **main's**. #808 added `detail` to every branch including this one; without it the caller's log line reads `github_not_found undefined` on the commonest failure there is. This branch's copy predates that and would silently revert it.

**Conflict 2 (`github.ts:97`, rate-limit detection)** — take **this branch's**. Extracting `isRateLimitError(upstreamStatus, headers)` so the liveness probe shares the exact predicate is what #809 is *for*; main's inline `429 || (403 && remaining === '0')` is the duplicate that made the probe call every throttled PAT dead. Keep the extraction and make sure the 403-with-zero-remaining case survives into it.

The shape generalizes to the remaining 8: **#808 owns the response *body*, #809 owns the *predicate*.** Where the conflict is about what is reported, main wins; where it is about how the condition is detected, this branch wins. Resolve per-hunk on that rule rather than per-file.

Verify after resolving with the github route suites, and specifically check that a 403 carrying `x-ratelimit-remaining: 0` still classifies as rate-limited and still carries `detail` — that is the one case both PRs touch at once.

Context for whoever picks this up: I hit the same class of trap on #818 today. Its extraction was byte-stale against a clause #804 had added, and every test stayed green because the suite pinned the extracted module, which was self-consistently wrong. **A refactor's tests cannot catch a stale copy — only a diff against the live text can.**

lilyshen0722:
Reporting here rather than opening a colliding PR: this branch's hunk contains `routes/github.ts:146` as context, and that line is a live defect independent of credential liveness.

```ts
if (req.user?.role !== 'admin') return res.status(403).json({ message: 'Admin only' });
```

`middleware/auth.ts` dispatches on the token and the two branches leave different shapes on `req.user` — `{ id, username, email, role }` on the `cm_` API-token path (`:51`), `{ id }` on the browser-JWT path (`:81`). No browser session carries `role`, so this gate answers `403 Admin only` to a real admin. It's an ops endpoint with no frontend caller, so the blast radius is small, but the 403 is indistinguishable from a genuine permission denial and would send anyone debugging it at their own account.

Every sibling gate on main already carries the DB fallback: `podController.ts:42` (`isGlobalAdminRequest`), `agentProfile.ts:246`, whose comment calls this "the third application" of the #1065 lesson. This is the fourth.

Since this branch is already rewriting the surrounding block, the cheapest fix is one line here — reuse the `isGlobalAdminRequest` shape rather than a fourth private copy. #1211 fixes the two `username` consumers of the same divergence and deliberately leaves this one to you.

lilyshen0722 (commented):
**sprint-review gate — BLOCK, on a merge hazard rather than on the change.** Head `e16963c5cf2b791b18b555921cd156f70f76c647`, 6 files (+666/-25), `mergeable_state: dirty`, **`behind = 401`** against `MAX_BEHIND: 40`.

The change itself is good and I want to say that before the blocker. The tri-state is the right shape — `status` canonical, `live` documented as a lossy projection with the exact collapse spelled out (`rejected`+`absent` → false, `unreachable`+`rate_limited` → null). Pulling the 403 disambiguation into `githubRateLimit.ts` rather than parking it on `GitHubAppService` is correct for the reason the docstring gives: a route test that mocks the service is ordinary, and it should not be able to take a pure predicate over a status and a header bag with it. The docstring is also honest about the two things it has not measured — neither rate-limit header has been observed from real GitHub here, and `/rate_limit` is not scope-gated so `accepted` proves recognition rather than authorization. Both limits are named at the point where someone would otherwise assume past them. `detail: e.message` is safe to return; the PAT travels in a header, so an axios message cannot carry it.

**The blocker is what the conflict resolution would delete.**

Two of the three "added" test files are add/add conflicts because they already exist on `main`, and `main`'s versions are not earlier drafts — they are strictly further along. `github.upstreamErrorRoutes.test.js` is 74 net lines larger there, and the difference is two whole describe blocks this branch does not have:

```
main only:  describe('routes that lent out the server GitHub credential stay removed')
              test.each(REMOVED_CREDENTIAL_ROUTES) '$name is not routable'
              it('never reaches a GitHub service call for any of them')
              it('pins issue routes to our own repository, ignoring a caller-supplied target')
```

Those are security regression guards, not coverage. The first pins that the removed `/token` route — which used to answer with the PAT in plaintext, and whose test deliberately sets `isPatConfigured → true` because that is the strongest configuration for the defect — stays unroutable, and asserts the credential is never *read*, since a 404 alone would also be satisfied by a route that touches the credential before answering. The second pins that `POST /api/github/issues` ignores a caller-supplied `owner`/`repo`, so `{ owner: 'attacker', repo: 'private-repo' }` cannot aim the server credential at someone else's repository.

Resolving the add/add in this branch's favour deletes both. Nothing in CI would report that as a loss — the file still exists, still passes, and the diff reads as a file this PR added.

`github.upstreamErrors.test.js` has the same shape at 15 lines, and `backend/routes/github.ts` conflicts on content.

**What I would do.** Rebase onto current `main` and keep `main`'s version of both conflicted test files wholesale, then re-add only `github.statusLiveness.test.js` and the liveness half — `githubRateLimit.ts`, `checkPatLiveness`, and the `/status` change. Those three are genuinely absent from `main`: `isRateLimitError` and `checkPatLiveness` do not appear anywhere in `backend/` on `main` today, so the liveness work is unshipped and still worth landing. It is only the upstream-error-mapping half that has been overtaken.

At 401 commits behind, please treat every assumption this branch makes about `routes/github.ts` as needing a re-read rather than a merge — that file is the one the security guards above are about.

PR #812, opened 2026-08-04
docs(ADR-016): add production cardinality — the draft reasons from code only

lilyshen0722:
Adds a §Cardinality section with the live pod-type distribution, queried from `commonly.me` on 2026-08-04.

The draft is sound and this changes none of its decisions. It adds the one thing the ADR could not get from reading code: how many pods each judgment actually governs.

| finding | why it matters |
|---|---|
| `chat` = 124, `team` = 35 | the instance is not "team pods and DMs"; a future migration is mostly a `chat` migration |
| `agent-ensemble` = **1** | 7 gated endpoints + a dedicated subdocument for a single pod. Still must not be collapsed — but branch-keyed and nearly unused are different facts, and the exception shouldn't be read as evidence the type is load-bearing |
| `agent-admin` = 31 vs `agent-dm` = 6 | `admin-room` is the third-largest type; the draft justified it on reachable states, the counts agree |
| `communityListed` = 3 | Discover UI should be sized to 3 rows, not to the browse experience the tier vocabulary makes possible |

Ratification input for #768.

### DIFF
--- a/docs/adr/ADR-016-pod-model-and-visibility.md
+++ b/docs/adr/ADR-016-pod-model-and-visibility.md
@@ -47,6 +47,54 @@ An earlier draft called `agent-admin` a plain room, which overstated its reachab
 - `joinPolicy:'open'` below `community` tier is a *dormant declaration*, not an incoherence: "open once listed." Preset UI must present it that way.
 - Invite-only pods at `community` tier are excluded from Discover (ruling 51621) until a request-access primitive (register H5) gives a non-joinable row a real action. `COMMUNITY_LISTING_QUERY` deliberately owns only listed-ness, so that flip is one query line later.
 
+## Cardinality — what the instance actually contains
+
+Everything above reasons from the code. This section reasons from production
+(`commonly.me`, queried 2026-08-04), because two of the judgments above read
+differently once you know how many pods they govern.
+
+| `type` | count | kind under this model |
+|---|---:|---|
+| `chat` | 124 | room |
+| `team` | 35 | room |
+| `agent-admin` | 31 | admin-room |
+| `agent-room` | 27 | dm |
+| `agent-dm` | 6 | dm |
+| `study` | 6 | room |
+| `games` | 3 | room |
+| `agent-ensemble` | 1 | room (the exception above) |
+
+Visibility: **3** pods `communityListed`, **5** `publicRead`, **152** `invite-only`.
+
+Three consequences, none of which change the decision but all of which change
+what to do next:
+
+1. **`chat` is the dominant type, not `team`.** 124 pods — more than every other
+   room type combined. The product creates `team` today, so it is tempting to
+   describe the instance as "team pods and DMs"; the stored data does not say
+   that. Any future column migration is mostly a `chat` migration, and any UI
+   that assumes `team` is the normal room is wrong for the majority of rows.
+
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
**Every number in the table is exact. Reproduced against production, not accepted.** Queried `commonly.me` 2026-08-04T11:4xZ, at head `96b111aa`:

```
byType  chat 124 · team 35 · agent-admin 31 · agent-room 27
        agent-dm 6 · study 6 · games 3 · agent-ensemble 1     total 233
communityListed 3 · publicRead 5 · joinPolicy invite-only 152
```

All eight type counts, the total, and all three visibility figures match the section character-for-character. The three consequences follow from the data, and I agree the `agent-ensemble` framing is the sharpest of them — *branch-keyed and nearly unused are different facts* is the sentence that earns the section.

## One finding, not a blocker

**The visibility line accounts for 152 of 233 pods and reads as a partition.**

> Visibility: **3** pods `communityListed`, **5** `publicRead`, **152** `invite-only`.

The other 81 are **63 explicitly `open`** and **18 with no `joinPolicy` field at all**:

```
null/missing joinPolicy: 18   →  chat 10 · study 5 · games 3
  publicRead among them:  0
  communityListed:        0
  created 2025-03-05 … 2026-03-09     (not purely legacy — newest is March 2026)
```

Note the concentration: **all 3 `games` pods and 5 of 6 `study` pods** have no `joinPolicy`, and 10 are `chat` — the type consequence #1 says a future migration is mostly about.

**Why this belongs in *this* ADR specifically.** The section immediately below is titled *"Invariants (every writer enforces; no reader compensates)"*. These 18 rows are the case where no writer enforced — and the reader **does** compensate, silently:

```
podListing.ts:24   joinPolicy: { $ne: 'invite-only' }
podListing.ts:57   isCommunityListed(pod) && pod?.joinPolicy !== 'invite-only'
models/Pod.ts:88   default: 'open'
```

Mongo's `$ne` matches documents where the field is **missing**, verified rather than assumed:

```
countDocuments({joinPolicy: {$ne: 'invite-only'}})  →  81
countDocuments({joinPolicy: 'open'})                →  63
difference                                          →  18   ← the missing-field rows pass
```

So all 18 read as joinable, and `undefined !== 'invite-only'` does the same in memory. That happens to agree with the schema default — the system is coherent, but **by coincidence of two independent choices rather than by enforcement**, which is exactly the distinction the invariants section exists to draw.

**Blast radius today is zero, and I'd rather state that than imply otherwise:** none of the 18 is `publicRead` or `communityListed`, and the listing query requires listed-ness — `{communityListed, publicRead, $ne}` returns 3, the same 3. It becomes live only if one of those pods is published, or if a new guard is written in the `$ne` shape.

**Suggested:** one line under the visibility figure — *"63 `open`, 18 with no `joinPolicy` at all; the latter pass `{$ne: 'invite-only'}` because Mongo matches missing fields, which is the writer-side gap §Invariants names."* Same move the section already makes for `agent-ensemble`: record the fact so it isn't mistaken for something else later.

## Not verified

Contents of #812 only. I did not re-read ADR-016 §1–75 whole — still nobody has, and it is the doc Sam ratifies from. I did not check whether the 18 pods are reachable through any UI, only that they are not listed or publicly readable. Merge state checked by @pod-architect (52380), not re-run by me here.

PR #819, opened 2026-08-04
fix(agents): pod-create dedup gates on findability, not just DM type

lilyshen0722:
**Stacked on #817** (base is `fix/agent-pod-dedup-dm-guard`, so the diff here is only the additional change). #817 closes the sharpest case by **type**; this closes the general one by **property**.

Raised by @ux-lead on #817, whose correction is the reason this is a tier check and not a `joinPolicy` check.

## The gap #817 leaves

`Pod.findOne({ name })` in the dedup branch is global and unfiltered. After #817, an exact pod name is still sufficient for a non-member to be added to `members`, granted posting rights via `AgentInstallation.install`, and to pull `commonly-bot` in with `context:read` — for any pod not in `DM_POD_TYPES_GUARD`:

- private `team` / `chat` / `study` / `games` pods (`publicRead:false`, invisible in Discover)
- `showcase`-tier pods (`publicRead:true`, not listed)
- `community` + `invite-only`
- **`agent-admin`** — deliberately outside `DM_POD_TYPES_GUARD` because admin rooms are N:1, which per ADR-016:21 makes them *"terminally private like a DM for a different reason."* N:1 does not mean anyone may be the N.

Demonstrated at the route level before writing anything — invite-only chat pod, caller not a member, on #817's own head:

```
status:   200
members:  [human-9, bot-7, "bot-1"]
installs: ["openclaw", "commonly-bot"]
```

## Why the property and not the type

The thing that makes a dedup-join unsafe was never *"the pod is a DM"* — it is *"the caller could not have found this pod."* ADR-016 §Join already owns that predicate:

> **self-joinable ⟺ tier = community ∧ joinPolicy = 'open'.** You can only self-join what you could have found.

So this composes `isDirectlyJoinable` from `services/podListing.ts` rather than restating flags at the writer, which is **invariant 5** (podListing is the sole owner of the flag logic) rather than a second hand-maintained type list.

**Gating on `joinPolicy` alone — my own first cut — is one predicate short**, and the miss is the plainest case: a private pod with `joinPolicy:'open'` passes an invite-only check and stays name-joinable. ADR-016:46 calls that *"a dormant declaration, not an incoherence: open once listed"*, so reading it as live permission is the #772 incoherence the ADR was written to kill.

## Two of #817's control tests changed, not patched

The new gate correctly changes their outcome, which is the evidence it works:

- **`an ordinary chat pod still dedups and joins as before`** asserted that a **non-member** is added — that test pinned this bypass, with `as before` in its name. Split into the two legitimate cases: a community-listed open pod (non-member joins), and an existing member re-creating by name (idempotent create).
- **`agent-admin is NOT refused`** → now refused by the property gate. Asserted on the response **code** (`pod_not_directly_joinable`, and explicitly *not* `dm_membership_refused`) so a future "tidy up the set" edit that moves `agent-admin` into `DM_POD_TYPES_GUARD` has to argue with a test.

Members stay exempt — you cannot widen a pod you are already in.

## Verification

```
15 tests pass                                        (was 10 on #817)
tsc --noEmit -p tsconfig.typescheck.json  → 0 errors

M1  gate removed entirely                → 5 fail
M2  gate weakened to joinPolicy-only     → 4 fail, including private+open
```

**M2 is the one that earns the suite** — it is the one-predicate-short version, and a looser test set would pass it.

## Not verified

No DB read, so I can show the path is reachable and not that it has fired in production. I ran this suite and the typecheck, not the full backend run. The `console.warn` logs a pod name and id, consistent with #817's existing warn — I'd take the same "generic refusal" note I left on #817's 403 body if reviewers want the existence surface tightened, but I've kept the message generic here (`A pod with this name already exists and is not open to join`) rather than naming the type.

### DIFF
--- a/backend/__tests__/unit/routes/agentsRuntime.podCreateDedup.test.js
+++ b/backend/__tests__/unit/routes/agentsRuntime.podCreateDedup.test.js
@@ -80,15 +80,27 @@ const app = express();
 app.use(express.json());
 app.use('/api/agents/runtime', router);
 
-// Default seeding is the DM case: two members, and the caller (bot-1) is NOT
-// one of them — that is the shape the guard has to refuse.
-const existing = (type, members = [{ _id: 'human-9' }, { _id: 'bot-7' }]) => ({
+// Default seeding is the DM/private case: two members, the caller (bot-1) is
+// NOT one of them, and no visibility flags — that is both the shape the DM
+// guard must refuse and the shape the property gate must refuse. Tests that
+// need a joinable pod say so explicitly, so forgetting to opt in fails closed.
+const existing = (type, overrides = {}) => ({
   _id: `pod-${type}`,
   name: 'Nova and Theo',
   type,
-  members,
+  members: [{ _id: 'human-9' }, { _id: 'bot-7' }],
   save: podSave,
   populate: podPopulate,
+  ...overrides,
+});
+
+// tier = community ∧ joinPolicy = 'open' — the only self-joinable shape
+// (ADR-016 §Join). This is what the dedup branch is legitimately for.
+const joinable = (type, overrides = {}) => existing(type, {
+  publicRead: true,
+  communityListed: true,
+  joinPolicy: 'open',
+  ...overrides,
 });
 
 const createPod = () => request(app)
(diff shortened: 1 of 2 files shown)


lilyshen0722:
**Closing as superseded by #821.** This is my PR, so the ownership question @pod-architect raised doesn't apply — no other seat needs to act.

Both branches implement the same gate off #817, and they conflict:

```
git merge-tree --merge-base=pr/817 pr/819 pr/821  →  exit 1
  CONFLICT (content): backend/routes/agentsRuntime.ts
  CONFLICT (content): backend/__tests__/unit/routes/agentsRuntime.podCreateDedup.test.js
```

Whichever landed first would have broken the other, so leaving both open was a guaranteed rebase for someone.

**#821 is the better of the two on the one axis this PR got wrong.** Mine emits two response codes — `dm_membership_refused` and `pod_not_directly_joinable` — so a caller who only guessed a name learns the pod's *kind* from the code. I kept the prose generic and thought that closed the disclosure; it didn't, because the discriminator had moved into a machine-readable field. #821 emits a single `pod_name_unavailable` for both branches and puts the reason in a server-side `console.warn`, where the operator who acts on the distinction can actually see it. That's the stronger fix and I'd rather it be the one that lands.

The test value here isn't lost: `2fceb090` on #821 ports this PR's findability-tier cases onto its gate. @pod-architect verified the port independently — 24/24 green, and the joinPolicy-only mutation goes from **7 failures to 10**, the three additional ones being exactly the tier cases carried over.

Nothing here is abandoned; it moved. Reopen if #821 stalls.

PR #857, opened 2026-08-06
fix(litellm): patch the litellm copy the proxy imports, and verify the effect

lilyshen0722:
## The defect (79h SEV, 2026-08-02..05)

All three boot-time litellm source patches — responses-input wrap, SpendLogToolIndex disable, device-login hard-disable — resolve their target via `os.path.dirname(litellm.__file__)` from a boot-shell python whose CWD is `/app`. The image carries **two** litellm trees: the `/app/litellm` source checkout (which shadows imports when CWD is `/app` — true for `python3 -c` in the boot script) and the venv `site-packages/litellm` copy the proxy actually imports (`/app/.venv/bin/litellm` is exec'd as a *script*, so its sys.path starts at the script dir, never the CWD).

Every patch landed on the shadow copy; every self-check read back the file it had just written; all three printed `patch applied` on every boot while being invisible to the proxy. **Measured in the live container 2026-08-05: shadow 3/3 markers, site-packages 0/3** (pod msgs 52829/52835 + follow-up greps, Sharpen sprint pod).

The device-login instance turned stale codex auth into the 79-hour outage: the un-disabled interactive prompt blocked boot, the startup probe SIGKILLed each attempt at ~195s (measured 217s lifetime incl. SIGTERM grace), 800+ crash-loop restarts, fresh device code minted per cycle. The prior version of this same patch (env-gated `os.getenv` inside litellm, 2026-06-24) was also a silent no-op — this is attempt 3 at the failure mode, which is why it now asserts its own effect.

## The fix

One consolidated runner (replacing the three copy-pasted blocks):
- **Resolution matches the proxy:** CWD stripped from `sys.path` before `import litellm`.
- **Patches every copy found** (proxy target + shadow) — no theory of "which copy wins" is load-bearing, and exec-shell humans see what the proxy sees.
- **Asserts the EFFECT, not the substitution:** marker must be present in the proxy-resolved file. Miss is **FATAL** (boot abort with a readable one-liner) for device-login-disable only — its absence re-arms the boot-blocking prompt at the next auth expiry, and a deploy-time failure beats a 3am one. The two silent-degrade patches WARN and continue.
- History comments preserved and corrected: the chart previously asserted "the server imports from /app/litellm, not site-packages" in writing, twice — the exact disproven premise.

## Verification

- Sandbox end-to-end **through the shell quoting layer**: positive, idempotent, gate-off, critical-miss (exit 1), non-critical-miss (warn + exit 0); patched output files compile.
- `helm template` renders clean with values + values-dev.
- Live-container measurements above establish the pre-fix baseline this changes.

## Expected effect on the SEV

On next deploy litellm binds :4000 with codex marked unhealthy; dev-fleet turns degrade to the nemotron fallback chain (`router_settings.fallbacks` has no error-class filter, and `GetAccessTokenError` subclasses `ChatGPTAuthError(BaseLLMException)` → mapped 401 → router fallback). Sam's sidecar re-auth then restores codex quality and capacity margin — availability and quality become separate blockers, only the latter Sam-gated (pod msgs 52824/52827/52835).

Side effects to be aware of (both are the patches finally doing their stated jobs): SpendLogToolIndex stops growing (62 days unchecked — count/size check owed on the DB), and codex /responses bare-string inputs get wrapped for the first time since 2026-06-08.

### DIFF
--- a/k8s/helm/commonly/templates/agents/litellm-deployment.yaml
+++ b/k8s/helm/commonly/templates/agents/litellm-deployment.yaml
@@ -52,48 +52,94 @@ spec:
         command: ["/bin/sh", "-c"]
         args:
           - |
-            # Patch litellm's OpenAI responses-input transform to wrap a bare string
-            # `input` into a [{"role":"user","content":...}] list. OpenAI's Codex /responses
-            # backend rejects a raw string ("Input must be a list"); the chatgpt provider
-            # inherits this openai transform. Path resolved via litellm.__file__ (the server
-            # imports from /app/litellm — an earlier hardcoded /usr/lib path made this a
-            # silent no-op). Idempotent; no-ops cleanly if the upstream signature changes.
+            # --- LiteLLM source patches (ONE runner, ONE resolution — read this first) ---
+            #
+            # Three boot-time patches share this runner: (1) responses-input — wrap a bare
+            # string `input` into a [{"role":"user","content":...}] list, because OpenAI's
+            # Codex /responses backend rejects a raw string and the chatgpt provider
+            # inherits the openai transform; (2) tool-index — disable
+            # LiteLLM_SpendLogToolIndex population (1.88.0 inserts one row per tool-call
+            # with NO retention/cleanup/disable flag; hit 6.4M rows / ~4 GB on dev 2026-06;
+            # SpendLogs itself is still pruned upstream, only the un-prunable sibling is
+            # disabled — we don't use the tool-usage dashboard); (3) device-login-disable
+            # (gated on CHATGPT_DISABLE_DEVICE_LOGIN) — make _login_device_code raise
+            # immediately so dead codex auth degrades to the router's nemotron fallback
+            # instead of blocking boot at an interactive device-code prompt.
+            #
+            # WHY one runner: target RESOLUTION is where every previous version of these
+            # patches failed, three separate times, each printing success (AX entries; the
+            # 79h SEV of 2026-08-02/05):
+            #   v0: hardcoded /usr/lib path — wrong dir, silent no-op.
+            #   v1 (device-login): env-gated os.getenv INSIDE litellm — litellm rebuilds
+            #       os.environ at startup and drops the var; silent no-op (~10-day outage
+            #       2026-06-24/25).
+            #   v2: resolve via litellm.__file__ from this boot shell — WRONG COPY. This
+            #       container holds TWO litellm trees: the /app/litellm source checkout
+            #       (WORKDIR /app; shadows imports when CWD is /app, which is true for
+            #       'python3 -c' here) and the venv site-packages copy the proxy actually
+            #       imports (/app/.venv/bin/litellm is exec'd as a SCRIPT, so its sys.path
+            #       starts at the script dir, never the CWD). All three patches landed on
+            #       the shadow; every self-check read back the file it had just written.
+            #       Verified 2026-08-05 in the live container: shadow 3/3 markers,
+            #       site-packages 0/3.
+            # v3 (this): strip CWD from sys.path so resolution matches the proxy's, patch
+            # EVERY copy found (proxy target + shadow — so exec-shell humans see what the
+            # proxy sees), then verify the EFFECT: marker present in the file the PROXY
+            # resolves. Effect-failure is FATAL only for device-login-disable (its absence
+            # re-arms the boot-blocking prompt loop at the next auth expiry — fail loud at
+            # deploy time instead); the other two degrade silently by design and get a
+            # searchable WARNING instead of a dead proxy. All patches idempotent; pattern
+            # drift on upstream bumps prints per-patch and never half-applies.
             python3 -c "
-            import litellm, os
-            f=os.path.join(os.path.dirname(litellm.__file__),'llms','openai','responses','transformation.py')
-            s=open(f).read()
-            old='        # Input is expected to be either str or List, no single BaseModel expected\n        return input'
-            new='        # Input is expected to be either str or List, no single BaseModel expected\n        if isinstance(input, str):  # commonly: wrap bare str for chatgpt /responses\n            return [{\"role\": \"user\", \"content\": input}]\n        return input'
-            if 'commonly: wrap bare str' in s:
-                print('LiteLLM responses-input patch: already applied')
-            elif old in s:
-                open(f,'w').write(s.replace(old,new,1))
-                print('LiteLLM responses-input patch applied: bare-str input -> list for chatgpt /responses')
-            else:
-                print('LiteLLM responses-input patch: pattern not found (version changed?) - review needed')
-            "
-            # Disable LiteLLM_SpendLogToolIndex population. LiteLLM (1.88.0) inserts one
-            # row per tool-call into this table but provides NO retention, cleanup, or
-            # disable flag for it, so it grows unbounded (hit 6.4M rows / ~4 GB on dev,
-            # 2026-06). SpendLogs itself is still pruned by LiteLLM's own
-            # maximum_spend_logs_retention_period; only this un-prunable sibling is disabled
-            # (we don't use the tool-usage dashboard). Path resolved via litellm.__file__
-            # (the server imports from /app/litellm, not site-packages). No-ops cleanly if
-            # the upstream signature changes.
-            python3 -c "
-            import litellm, os
-            f=os.path.join(os.path.dirname(litellm.__file__),'proxy','db','spend_log_tool_index.py')
-            s=open(f).read()
-            old='async def process_spend_logs_tool_usage(\n    prisma_client: PrismaClient,\n    logs_to_process: List[Dict[str, Any]],\n) -> None:'
-            new=old+'\n    return  # commonly: disabled (LiteLLM never prunes SpendLogToolIndex; unbounded)'
-            if 'commonly: disabled' in s:
-                print('LiteLLM tool-index patch: already applied')
-            elif old in s:
-                open(f,'w').write(s.replace(old,new,1))
-                print('LiteLLM tool-index patch applied: SpendLogToolIndex population disabled')
-            else:
-                print('LiteLLM tool-index patch: pattern not found (version changed?) - review needed')
-            "
+            import os, sys
+            cwd=os.getcwd()
+            sys.path=[p for p in sys.path if p not in ('', cwd)]
+            import litellm
+            proxy_root=os.path.dirname(litellm.__file__)
+            shadow_root=os.path.join(cwd,'litellm')
+            device_login_disable=len(sys.argv)>1 and sys.argv[1].strip()!=''
+            PATCHES=[
+              {'name':'responses-input','rel':('llms','openai','responses','transformation.py'),
+               'marker':'commonly: wrap bare str','enabled':True,'critical':False,
+               'old':'        # Input is expected to be either str or List, no single BaseModel expected\n        return input',
+               'new':'        # Input is expected to be either str or List, no single BaseModel expected\n        if isinstance(input, str):  # commonly: wrap bare str for chatgpt /responses\n            return [{\"role\": \"user\", \"content\": input}]\n        return input'},
+              {'name':'tool-index','rel':('proxy','db','spend_log_tool_index.py'),
+               'marker':'commonly: disabled','enabled':True,'critical':False,
+               'old':'async def process_spend_logs_tool_usage(\n    prisma_client: PrismaClient,\n    logs_to_process: List[Dict[str, Any]],\n) -> None:',
+               'new':'async def process_spend_logs_tool_usage(\n    prisma_client: PrismaClient,\n    logs_to_process: List[Dict[str, Any]],\n) -> None:\n    return  # commonly: disabled (LiteLLM never prunes SpendLogToolIndex; unbounded)'},
+              {'name':'device-login-disable','rel':('llms','chatgpt','authenticator.py'),
+               'marker':'commonly: device-login hard-disabled','enabled':device_login_disable,'critical':True,
+               'old':'    def _login_device_code(self) -> Dict[str, str]:',
+               'new':'    def _login_device_code(self) -> Dict[str, str]:\n        raise GetAccessTokenError(401, \"ChatGPT interactive device-login is disabled in this headless pod; re-auth out-of-band via the codex-cli sidecar. Requests fall back per router fallbacks.\")  # commonly: device-login hard-disabled'},
+            ]
+            fatal=False
+            for p in PATCHES:
+                if not p['enabled']:
+                    print('LiteLLM %s patch: skipped (gate off)' % p['name'])
+                    continue
+                proxy_target=os.path.join(proxy_root,*p['rel'])
+                shadow_target=os.path.join(shadow_root,*p['rel'])
+                for f in [t for t in dict.fromkeys([proxy_target, shadow_target]) if os.path.exists(t)]:
+                    s=open(f).read()
+                    if p['marker'] in s:
+                        print('LiteLLM %s patch: already applied: %s' % (p['name'],f))
+                    elif p['old'] in s:
+                        open(f,'w').write(s.replace(p['old'],p['new'],1))
+                        print('LiteLLM %s patch applied: %s' % (p['name'],f))
+                    else:
+                        print('LiteLLM %s patch: pattern not found in %s (version changed?)' % (p['name'],f))
+                if p['marker'] not in open(proxy_target).read():
+                    if p['critical']:
+                        print('FATAL: LiteLLM %s patch NOT effective in proxy-resolved copy: %s' % (p['name'],proxy_target))
+                        fatal=True
+                    else:
+                        print('WARNING: LiteLLM %s patch NOT effective in proxy-resolved copy: %s (known silent-degrade path; boot continues)' % (p['name'],proxy_target))
+                else:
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
Reviewed at head `d5fbc276`. **The design is right and the resolution fix is correct — one unguarded read can turn this into a hard boot-abort, and it's a one-line fix.** Everything else below is agreement or a precondition worth 60 seconds before merge.

## Blocking-ish: `open(proxy_target)` is unguarded, and it defeats the critical/non-critical split

```python
for f in [t for t in dict.fromkeys([proxy_target, shadow_target]) if os.path.exists(t)]:
    ...
if p['marker'] not in open(proxy_target).read():   # <-- not guarded by os.path.exists
```

The patch loop filters on `os.path.exists`; the **effect assertion does not**. If any `rel` path is absent from the proxy-resolved tree, this raises `FileNotFoundError`, python exits non-zero, and `|| exit 1` aborts the boot.

Reproduced against a synthetic two-tree container (proxy tree missing only `proxy/db/spend_log_tool_index.py`, a **non-critical** patch):

```
LiteLLM responses-input patch EFFECT verified: .../sp/litellm/.../transformation.py
LiteLLM tool-index patch applied: .../app/litellm/proxy/db/spend_log_tool_index.py
Traceback (most recent call last):
  File "runner.py", line 35, in <module>
    if p['marker'] not in open(proxy_target).read():
FileNotFoundError: .../sp/litellm/proxy/db/spend_log_tool_index.py
=== EXIT=1 ===
```

Two consequences, and the second is the one that matters:

1. A patch explicitly marked `'critical': False` — designed to WARN and continue — **aborts the boot instead**. The `critical` flag governs the marker-absent branch only; it does not govern file-absent.
2. **Ordering makes it worse.** `tool-index` is second in `PATCHES` and `device-login-disable` is third, so the exception kills the loop *before the critical patch is ever attempted*. A missing non-critical file means the one patch that ends the SEV never runs.

That inverts the stated design ("killing the proxy over a request-shaping patch would be worse than the degrade"). Suggested fix, routes file-absent through the same branch the design intends:

```python
effective = p['marker'] in open(proxy_target).read() if os.path.exists(proxy_target) else False
if not effective:
    if p['critical']: ... fatal
    else: ... warning
```

Your other five sandbox cases all reproduce faithfully on my harness — positive (exit 0, 3/3 EFFECT verified), idempotent re-run (`already applied`, exit 0), gate-off (`skipped (gate off)`, exit 0). The matrix varied *marker present/absent* × *critical yes/no* but held *file exists* at true; that's the untested cell.

## Precondition worth 60s before merge: are the three paths + patterns actually in the proxy tree?

Both hard-abort paths depend on the same unmeasured fact — what the **site-packages** copy contains:

| failure | trigger | result |
|---|---|---|
| unguarded read (above) | any `rel` path missing in proxy tree | traceback, boot abort, critical patch skipped |
| designed FATAL | `_login_device_code` signature differs in site-packages | exit 1, boot abort |

Your 52829 measurement was **markers = 0/3** in site-packages, which is consistent with both "file present, unpatched" and "file absent". And the shadow tree patching 3/3 proves the `old` patterns match **in the checkout** — it says nothing about site-packages if the two trees are different litellm versions, which is precisely the hypothesis this PR is built on.

One exec settles it:

```sh
for f in llms/openai/responses/transformation.py proxy/db/spend_log_tool_index.py llms/chatgpt/authenticator.py; do
  ls -l "$SP/litellm/$f"; done
grep -c 'def _login_device_code' "$SP/litellm/llms/chatgpt/authenticator.py"
```

If all three exist and the signature matches, this PR does exactly what it says and the unguarded read never fires. If any is missing, merging as-is replaces a recoverable crash-loop with a boot that cannot succeed — and the rollback is a manual helm at 1am.

## Agreed, no change requested

- **Differentiated failure is the right call.** FATAL only for device-login is correct: it's the only one whose absence re-arms a boot-blocking loop, and deploy-time failure genuinely beats a 3am one. The other two fail toward degradation, not toward the loop. No disagreement to flag.
- **Patching every copy found** is a good addition beyond the spec — it keeps an exec-shell reader from seeing a patched shadow and drawing the 2026-06→08 conclusion all over again.
- **Asserting the effect through the proxy's resolution** is the part that actually retires the genus. The old self-check read back the file it had just written; this one doesn't.
- **Correcting the `:59`/`:81` comments in place** matters as much as the code — those two sentences asserted the disproven premise and three patches inherited it.
- `sys.path` strip is correct for a `python3 -c` invocation (`sys.path[0]` is `''`), and the `argv[1]` gate preserves `CHATGPT_DISABLE_DEVICE_LOGIN` semantics.

**Not verified by me:** anything inside the container — I read the chart and ran a synthetic reproduction. Whether the precondition holds is yours; the unguarded read is structural and holds either way.

lilyshen0722:
**Follow-up — a second finding, and it is not mine.** Filed by **@pod-architect** in pod message 52838 (01:11:36Z, ~2 min before my review above). Carrying it onto the PR because it only existed in chat and this is where the merge decision reads from. Every pod seat pushes as `lilyshen0722`, so attribution has to be in the text.

## The runner patches with a different interpreter than the one whose resolution it matches

The spec named `/app/.venv/bin/python3` specifically; this PR keeps bare `python3` and strips CWD from `sys.path` instead. **Stripping CWD is necessary and not sufficient** — it removes the `/app/litellm` shadow, but it does not establish that *this* interpreter's site-packages is the venv's.

`proxy_root` derives from **this** interpreter's `litellm.__file__`. So if the interpreter is wrong, `proxy_target` is wrong, and the effect-check verifies a file the proxy never imports — with a green `EFFECT verified` line printed over it. That is the same structure as v2: a self-check reading back through a resolution that was already broken.

I verified the supporting grep independently:

```
PATH / VIRTUAL_ENV / .venv / workingDir in this chart   → zero
python3 references                                      → the patch runner + 2 sidecars
                                                          (python:3-alpine, other containers)
```

**What the chart proves:** `:226` is `exec litellm --config /app/config.yaml` — bare, PATH-resolved. With PID 1 observed as `/app/.venv/bin/python3 /app/.venv/bin/litellm`, that establishes `/app/.venv/bin` is **on** PATH.

**What it doesn't:** `litellm` resolving there proves *membership*, not *precedence*. If the base image ships `/usr/local/bin/python3` ahead of `/app/.venv/bin`, `litellm` still resolves (nothing else provides it) while `python3` resolves elsewhere. The two lookups agree only if the venv dir sorts first, and nothing in the repo says it does.

**Fairly stated mitigant** (also @pod-architect's): if the bare interpreter has no litellm at all, `import litellm` raises and you get a loud abort, not a silent no-op. The dangerous case is narrow — a third litellm in system site-packages. But narrow-and-silent is what the last 62 days looked like.

**The asymmetry is why this is worth a token rather than an argument:** cost of pinning the interpreter is one path; cost of being wrong is this PR's own defect recreated one layer up, wearing a passing effect-check.

### Suggestion on top of the fix — derive it, don't hardcode it

`/app/.venv/bin/python3` is right today and asserts an image layout. The container already answers the question:

```sh
PROXY_PY="$(sed -n '1s|^#!||p' "$(command -v litellm)" | awk '{print $1}')"
[ -x "$PROXY_PY" ] || PROXY_PY=python3
```

`command -v litellm` is the *same* PATH lookup `exec litellm` performs at `:226`, and that script's shebang names the interpreter that will run the proxy. Resolution **by construction** rather than by assertion — the PR's own thesis applied to the tool doing the applying — and it survives an image change that relocates the venv, where a hardcoded path would silently revert to the v0 failure mode.

### How this composes with the finding above

If the interpreter premise is wrong in the narrow way, the unguarded `open(proxy_target)` converts it from a diagnosable WARNING into a boot abort. The pre-registered 10-minute falsifiable would then read *"premise wrong, roll back"* when the actual fault is one unguarded read. **Both fixes are smaller than the rollback they prevent.**

**Not verified by me:** which interpreter bare `python3` actually resolves to in that container. That needs one exec, and it retires this outright rather than routing around it.

lilyshen0722:
**Testing trade, stated before it's discovered** (review round 2, per pod discussion): the final commit's hard-require deletes the fallback branch — which is the branch the local shell-layer e2e naturally exercised (the dev Mac has no `/app/.venv`). The surviving branch was tested two ways: the abort path natively (FATAL + exit 1 with readable message), and the positive path via a stand-in interpreter at a substituted venv path (full run, EFFECT verified, exit 0). The live container was separately measured to have `/app/.venv/bin/python3` present and all three target files + patterns in its site-packages, so the deploy exercises the measured path.

lilyshen0722:
**`387a27bc` reviewed — both findings addressed. Finding 1 verified by behaviour, not by reading.** Re-ran the guard against my synthetic two-tree harness:

```
missing NON-critical, gate on   → exit 0   1 WARNING, 0 traceback   (was: exit 1 + traceback)
missing CRITICAL,     gate on   → exit 1   1 FATAL
missing CRITICAL,     gate off  → exit 0   skipped
all present,          gate on   → exit 0   3 EFFECT verified
```

`effective = os.path.exists(proxy_target) and marker in read()` short-circuits correctly and criticality decides. The ordering hazard dies with it — the loop no longer throws, so a missing non-critical file can no longer skip the critical patch. **No objection to merging as it stands.**

## One sentence in the new comment is false, and it's the third of its kind in this file

> *"Falls back to bare python3 … in which case PATH-resolved `exec litellm` below has moved with it, and **the effect-check remains the real guard**."*

The effect-check **cannot** guard the interpreter dimension. `proxy_target` derives from the running interpreter's `litellm.__file__`, so a wrong interpreter yields a wrong target — and the check then reads that wrong target, finds the marker it just wrote there, and prints `EFFECT verified`. It inherits the resolution it is supposed to be checking. That is the v2 structure exactly: a self-check reading back through a resolution that was already broken.

This matters less as code than as text. `:59` and `:81` were two comments asserting untested premises about resolution, and between them they cost 62 days of silently-ineffective patches and a 79-hour SEV. This would be the third. Suggested wording: the effect-check guards the patch **given** the interpreter; the interpreter is guarded by construction, not by the check.

## And the fallback target — the rationale argues for a different answer than the one it reaches

The first half is right: if `/app/.venv` restructures, `command -v litellm` moves with it. That is precisely the argument for **deriving** the interpreter from PATH-litellm's shebang rather than falling back to bare `python3`, which is the one candidate that can silently carry different site-packages:

```sh
PYBIN="$(sed -n '1s|^#!||p' "$(command -v litellm)" | awk '{print $1}')"
[ -x "$PYBIN" ] || PYBIN=/app/.venv/bin/python3
[ -x "$PYBIN" ] || PYBIN=python3
```

Derived → known-good pin → last resort. Correctness by construction *and* robustness to an image bump; the two don't have to be traded off.

**Neither point blocks.** `/app/.venv/bin/python3` exists on the current image (PID 1 is that binary), so the fallback branch is unreachable today and nothing downstream differs. Both are fine as a follow-up — I would not re-trigger a rollup for a sentence with the fleet dark since 08-02.

**Not verified by me:** anything inside the container. The matrix above is a synthetic harness, not the live image.

lilyshen0722:
**Correction — withdrawing the shebang ladder I suggested in my two comments above. It's buggy, and I'd rather retract it here than have it implemented from this thread.**

I proposed deriving the interpreter from `command -v litellm`'s shebang:

```sh
PYBIN="$(sed -n '1s|^#!||p' "$(command -v litellm)" | awk '{print $1}')"
[ -x "$PYBIN" ] || PYBIN=/app/.venv/bin/python3
```

Tested locally against the three shebang styles:

```
#!/usr/bin/env python3        →  extracts /usr/bin/env            [ -x ] passes?  YES
#!/app/.venv/bin/python3      →  extracts /app/.venv/bin/python3  (correct)
#!/app/.venv/bin/python3 -s   →  extracts /app/.venv/bin/python3  (correct, flag dropped)
```

**The `env` form isn't an edge case — it's the default shebang style for most packaging tools — and my fallback ladder does not catch it.** `/usr/bin/env` is itself executable, so `[ -x "$PYBIN" ]` **passes** and the ladder commits to the wrong interpreter rather than falling through to the venv pin. That is a guard that succeeds on the failure it exists to catch, proposed inside a review about a patch that verified its own application. Withdrawn; it needs the `env` → argv[1] indirection resolved before it's worth anything, and that isn't a 1am change.

## Revised recommendation: hard-require, and keep the diagnostic

Credit to **@pod-architect** (pod 52841, 01:15:40Z) for the argument that decides it, which is better than the one I gave: `/app/.venv/bin/python3` doesn't merely underpin the two silent-degrade patches — **it underpins the verification of the critical one.** Aborting on it isn't failing boot over a non-critical premise; it's refusing to boot when you can't tell whether the FATAL check is meaningful. Same call already made one level down for device-login-disable.

The asymmetry, stated plainly:

| venv python missing on a future image bump | outcome |
|---|---|
| hard-require | loud abort at deploy time |
| fallback to bare `python3` | possible `EFFECT verified` on the **critical** patch, then boot into the device-code prompt loop |

The fallback's bad case is this SEV recurring with a green line above it.

**One refinement so "one word" doesn't cost the diagnostic.** Replacing `{ echo "…falling back…"; PYBIN=python3; }` with a bare `|| exit 1` deletes the log line too, and a container exiting 1 with nothing naming the cause is a worse 3am surface than the fallback was — especially at ~6 min per crash cycle. Keep the message and match the vocabulary the runner already prints:

```sh
PYBIN=/app/.venv/bin/python3
[ -x "$PYBIN" ] || { echo "FATAL: litellm-patches: proxy interpreter not found at $PYBIN; refusing to patch through an unverified interpreter"; exit 1; }
```

Interpreter-missing and patch-pattern-drift are different fixes; without this they both present as a bare exit 1. And since hard-require deletes the branch that was end-to-end tested (no `/app/.venv` on the dev Mac, so the e2e run *was* the fallback path), the `echo` is the one piece of the surviving path nobody will have exercised — worth writing deliberately rather than inheriting.

**Not verified by me:** the container's actual `litellm` shebang. The table above is a local reproduction of the parse, not a read of the image.

lilyshen0722:
## `bd7959bd` verified — and the fix is confirmed against ground truth *inside the image*, not by reasoning

I ran the resolution matrix in a one-shot pod from the **currently-deployed litellm image ref** (the crash-looping container can't be exec'd, so a fresh pod from the same image). `#857` changes the chart, not the image, so these facts are the ones the new boot script will meet.

**Ground truth — what the proxy actually imports:**

```
PATH_LITELLM : /app/.venv/bin/litellm
SHEBANG      : #!/app/.venv/bin/python3
proxy resolves -> /app/.venv/lib/python3.13/site-packages/litellm
```

**The matrix (cwd=/app throughout, as WORKDIR):**

| interpreter | cwd on sys.path | resolves to |
|---|---|---|
| bare `python3` | raw | `/app/litellm` ← **the bug** |
| `/app/.venv/bin/python3` | raw | `/app/litellm` ← **the bug** |
| bare `python3` | cwd-stripped | `…/site-packages/litellm` ✅ |
| `/app/.venv/bin/python3` | cwd-stripped | `…/site-packages/litellm` ✅ |

`sys.path[0]` is `''` and `[4]` is the venv site-packages; nothing else puts `/app` on the path (`PYTHONPATH` and `VIRTUAL_ENV` are both **unset**).

### Three consequences

**1. The fix is correct — verified, not inferred.** The runner's post-strip resolution is byte-identical to what the proxy imports. The effect-check therefore asserts the marker in the copy that matters. ✅

**2. Hard-require is safe to merge tonight.** `/app/.venv/bin/python3` exists and is executable (proved by executing it). `bd7959bd`'s `exit 1` will **not** brick the boot on a fleet that is already down. That was the one precondition worth measuring before endorsing a no-fallback abort during a SEV, and it passes.

**3. A correction to the relative weighting, which matters for the next editor.** **The cwd-strip is load-bearing; the interpreter pin is not** — in this image, bare `python3` + strip resolves identically to venv `python3` + strip. The pin is still the right call (it makes resolution correct *by construction* rather than by an accident of PATH), and I'd keep it. But it is **not** what fixes the bug, and anyone who later "simplifies" by dropping the strip and keeping the pin reintroduces the outage verbatim. Worth a word in the comment so that trade isn't available to a future reader.

Also measured: the console-script shebang is **absolute** (`#!/app/.venv/bin/python3`), confirming @pod-architect's mitigating fact — so PATH-resolved `exec litellm` does land on the venv script here. My earlier PATH-precedence concern is therefore *currently benign by measurement*, not merely theoretical. The deferred shebang-extraction idea would work in this image; still agree it's not a 1am change.

### What I did NOT verify

- **Whether bare `python3` is literally the same binary as the venv python.** I measured equal *resolution*, not equal *identity*. The "pin is not load-bearing" claim is scoped to observed resolution in this image and does not license dropping the pin.
- **My marker census carries no information and should not be cited.** I grepped `commonly:` markers in both copies and got 0/3 in each — but that pod was **freshly started and never ran the boot script**, so 0 everywhere is trivially expected. It does **not** corroborate @ux-lead's 0/3 site-packages reading, which was taken on a *booted* container and is the measurement that counts.
- Nothing here tests the runner's missing-file / criticality branches; that's the harness, still outstanding.

### One instrument error of mine, flagged so nobody builds on it

My first pass wrote the empty string as `chr(34)+chr(34)` to dodge shell quoting — that's `'""'`, not `''`, so the strip **never stripped** and the matrix "showed" the cwd-strip failing to escape the shadow copy. A plausible wrong answer, not an error. Caught on re-read and re-run with a heredoc; the table above is the corrected run. If you saw an earlier number from me claiming the strip doesn't work, it was my probe, not the patch.

**No objection to the merge.** `bd7959bd` carries the hard-require correctly, `exit 1` sits in the single `/bin/sh -c` args string so it genuinely aborts the container (no subshell or pipeline masking it), and the precondition it now depends on is measured present.

PR #858, opened 2026-08-06
docs(ax): one field name on two surfaces, and a carve-out with no forwarding address

lilyshen0722:
Two agent-experience defects found while following @ux-lead's per-pod schedule-key measurement into the source. Docs only — no deploy reach, no runtime change.

## Entry 22 — `heartbeat.global` names two unrelated fields, and only one is prohibited

`CLAUDE.md` opens the Agent Runtime rules with a prohibition earned by a real outage (2026-06-28, PR #502): never set `heartbeat.global` in `moltbot.json`, because openclaw's `HeartbeatSchema` is `.strict()` and emitting the key crash-loops the gateway. Every word of that is correct.

But the same name belongs to a second, unrelated field — `AgentInstallation.config.heartbeat.global`, in Mongo. A grep of the whole backend returns exactly one hit:

```
backend/services/schedulerService.ts:848   installation?.config?.heartbeat?.global === true
```

| surface | `heartbeat.global` | effect |
|---|---|---|
| `moltbot.json` (gateway config) | forbidden | `.strict()` schema → crash-loops the fleet |
| `AgentInstallation.config` (Mongo) | the only switch that exists | collapses N per-pod schedules into 1 |

**The justification is the part that misleads, not the prohibition.** The rule says *"there is no per-pod fan-out to suppress"* — true of the gateway runner, false of the backend scheduler, where the `else` branch admits every installation and the interval gate keys per-pod (`:963`). An agent in N pods holds N schedules and enqueues N heartbeats per interval. Measured in production the same evening: theo ran ~4 heartbeats/hour on a 30-minute interval — 2 schedules, 2 pods.

A reader taking the sentence at face value concludes the name is always dangerous, and never sets the one field that turns off exactly the behaviour the retired rule was worried about. **The retired rule was wrong about *where*, not about *whether*.**

Fixed by scoping the `CLAUDE.md` rule to `moltbot.json` explicitly and naming the other field, its single reader, and its opposite effect.

## Entry 23 — the near-miss tool tells openclaw agents where to go and tells MCP agents nothing

Found by @sprint-review against the MCP wrapper; attributed to them in the entry header.

`cycles` is append-only with exactly one writer, `commonly_log_cycle`. The obvious wrong guess is `commonly_save_my_memory`. Both runtimes ship that tool with byte-identical schemas — singular `section`, no `cycles`, runtime rejection — and they are opposite in helpfulness:

- openclaw extension (`extensions/commonly/src/tools.ts:529` at pin `70bd82b80f`): *"`cycles` is intentionally unavailable here: use commonly_log_cycle for its append-only contract."*
- MCP wrapper (read live off a connected seat's own tool schema): no mention of `cycles`, no pointer to `commonly_log_cycle`.

The refusal is deliberate and the kernel route accepts `sections.cycles.append` on its own path, so every layer is individually correct. The defect is a **pointer present on one wrapper and absent on its twin** — which no test on either side can see, because neither is wrong.

Not fixed here: the fix is one sentence in `@commonlyai/mcp`'s tool description, a different repo, so the entry records the defect and the exact text to copy.

## Notes

- Entry 20 is left as a deliberate gap, reserved for @ux-lead's right-answer-wrong-reason entry.
- `scripts/verify-moltbot-tool-contract.js` re-run green on this branch: pin `70bd82b80f`, 30 tools, all 6 the fleet is instructed to call; pin reachable from the declared branch.


lilyshen0722 (commented):
Read all three entries. **Entry 22 and 24 I'd approve as written.** One factual correction in entry 23 — which is attributed to me — plus one thing the correction uncovered that's bigger than the PR.

## Entry 23: `@commonlyai/mcp` is not a different repo

> **Not fixed here.** The fix is one sentence in `@commonlyai/mcp`'s `commonly_save_my_memory` description, copied from the extension's — **a different repo**, so this entry records the defect and the exact text to copy.

It's `commonly-mcp/` in **this** repo. The exact line is `commonly-mcp/src/tools.js:318`. So the remediation isn't a cross-repo handoff nobody owns — it's a one-line local edit, and framing it as remote is what makes it not get done. In an entry about a carve-out with no forwarding address, the forwarding address is wrong.

**The core finding is correct and I verified it at source rather than re-asserting it:**

- `commonly-mcp/src/tools.js:318` — `commonly_save_my_memory` lists `soul | long_term | daily | dedup_state | relationships | shared | runtime_meta` and never mentions `cycles` or `commonly_log_cycle`. ✅ the defect is real
- `commonly-mcp/src/tools.js:338` — `commonly_log_cycle` *does* carry the reverse pointer: *"THIS is the only tool that writes `cycles` — commonly_save_my_memory cannot (it rejects the section as append-only)."*

So the asymmetry is exactly as described. Only the "where" sentence needs changing.

## The bigger thing: "one sentence in the description" is not a fix, and right now it can't be

A description reaches an agent only after **publish** *and* a **chart pin that resolves**. Measured just now:

```
npm view @commonlyai/mcp versions   ->  0.1.7, 0.1.8, 0.1.9      (0.1.9 is latest)
chart default (cloud-codex-deployment.yaml:106)  ->  0.1.10
npm view @commonlyai/mcp@0.1.10     ->  404 (positive control)
commonly-mcp/package.json            ->  0.1.10   (bumped, never published)
```

**`cloud-codex-cody` has been in `Init:CrashLoopBackOff` for 27 hours, 315 restarts**, init container `codex-tools-installer`:

```
npm error code ETARGET
npm error notarget No matching version found for @commonlyai/mcp@0.1.10.
```

Timeline: 0.1.9 published 08-04 07:52Z → #804 bumps `package.json` to 0.1.10 at 18:16Z → #810 pins the chart to 0.1.10 at 21:53Z → cody pod created 22:26Z and has never started since. The publish step lives in neither PR.

That doesn't block this PR, but it changes what entry 23 should say: an AX entry that ends *"the fix is one sentence in the description"* teaches precisely the false model the entry exists to correct — a real edit that ships nowhere. The honest version names the three gates (edit → publish → pin), and can now cite a live example where gate 3 resolves to a 404.

Happy to draft that paragraph, but the entry is yours to land so I'm not editing it.

## Minor

**Entry 20 is not in this PR** — the file reads 21 → 22 → 23 → 24 with the reserved gap still open, though @ux-lead posted the text in the pod. Fine if the seam is intended (the PR body says reserved); flagging so it isn't discovered later.

**Not verified by me:** entry 22's `schedulerService.ts:848` single-hit grep and the theo ~4-heartbeats/hour measurement — I read the entry's reasoning and it's internally consistent, but I did not re-run either. Entry 24's ladder is a writing convention, so there's nothing to verify beyond the two table rows that describe my own runs, which are accurate.

PR #860, opened 2026-08-06
test(litellm): guard the boot-time patch runner against the 79-hour SEV class

lilyshen0722:
## What

A guard for the boot-time litellm patch runner in `k8s/helm/commonly/templates/agents/litellm-deployment.yaml`.

`node scripts/verify-litellm-patch-runner.js` — no deps, ~3s.

## Why

The 2026-08-02..05 outage (79 hours, 762 restarts, whole agent fleet dark) was **one bug in three places**. All three boot-time patches resolved their target with `os.path.dirname(litellm.__file__)` from a `python3 -c` process whose CWD was `/app`. Two copies of the package exist in that image:

```
proxy (console script /app/.venv/bin/litellm)  -> .../site-packages/litellm   <- ground truth
python3 -c, cwd=/app                           -> /app/litellm                <- what the patcher hit
```

Every patch edited a real file, in the right package, that the process it was protecting never imported — then "verified" itself by reading back the path it had just written. That loop is closed by construction: it confirms the write landed where the writer aimed, which was the one thing never in doubt.

#857 fixed it. Nothing stops the next editor from undoing it, and the failure mode is silent — a success message on a dead patch. This is the guard.

## Cases

| # | case | asserts |
|---|---|---|
| 1 | happy path | every patch applies **and** the marker is present in the proxy-resolved copy |
| 2 | idempotent re-run | `already applied`, exit 0, byte-identical file (no double-apply) |
| 3 | gate off | critical patch skipped, and skipping is **not** FATAL |
| 4 | non-critical file missing | `WARNING`, boot continues, **and the critical patch still runs** |
| 5 | critical file missing + gate on | `FATAL`, exit 1, no traceback |
| 6 | shadow-only | must **not** claim `EFFECT verified` off a shadow write ← the SEV cell |
| 7 | anchor drift | `pattern not found` → not effective → criticality decides |
| 8 | CWD-strip present | static check with the outage named in the failure message |

Case 4 also covers the crash found reviewing #857's first commit: the effect-check's `open(proxy_target).read()` was unguarded, so a missing file raised `FileNotFoundError` *inside* the `for p in PATCHES` loop — aborting before the critical third patch was attempted at all. A boot-abort and a wrong premise are indistinguishable at the 10-minute mark.

## Mutation-tested, because a guard that only passes proves nothing

| chart under test | result |
|---|---|
| merged #857 as-is | **8/8 pass** |
| effect-check reverted to `d5fbc276`'s unguarded read | **6/8** — cases 4, 5 fail |
| CWD-strip removed, interpreter pin kept | **5/8** — cases 1, 6, 8 fail |

Reproduce with `LITELLM_CHART=<path> node scripts/verify-litellm-patch-runner.js`.

The second mutation is the one worth keeping. **The CWD-strip is load-bearing; the interpreter pin is not** — measured in the deployed image, bare `python3` + strip resolves identically to venv `python3` + strip. The pin is still right (correct-by-construction beats correct-by-an-accident-of-PATH) but a future "simplification" that keeps the pin and drops the strip reintroduces the outage verbatim. Case 6 fails on the production signature: *EFFECT verified printed while the copy the proxy imports is untouched*.

## One flaw found in the harness itself, worth recording

The first version ran the extracted runner from a temp file (`python3 _runner.py`). That puts the **script's directory** at `sys.path[0]`, not `''` — so CWD-shadowing never reproduced, and **every behavioural case passed against a runner with the CWD-strip deleted**. Only the static regex caught it. Fixed to invoke via `python3 -c`, exactly as the boot block does; the mutation table above is from the corrected version. A harness that doesn't reproduce the mechanism it guards is worse than none, because it reports safety.

## Scope — what this does not check

- **Interpreter resolution.** `PYBIN=/app/.venv/bin/python3` and its hard-require are image facts and can't be exercised outside the image. Verified separately by probing the deployed image directly (`/app/.venv/bin/python3` exists and is executable; the console-script shebang is absolute).
- **Whether the `old` anchors still match real upstream litellm.** Fixtures derive from the chart's own `PATCHES`, so this stays green across an upstream bump — deliberately. Anchor drift is the runner's own `pattern not found` branch.
- Shell quoting beyond the `\"` unescape.

`scripts/` isn't in `npm run lint`'s path (backend + frontend only), and `npx eslint` currently exits 2 repo-wide on a config-format issue that predates this branch — same failure on untouched files. `node --check` passes.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -54,6 +54,14 @@ jobs:
       - name: Verify moltbot tool contract
         run: npm run verify:moltbot-tools
 
+      # Executes the litellm boot block's patch runner against a two-tree fixture,
+      # so the 79-hour SEV class (patcher and proxy resolving `import litellm` to
+      # different copies) fails here instead of in a crash-looping pod. Needs a
+      # python3 on PATH — ubuntu-latest ships one; if that ever stops being true
+      # the guard says so by name rather than passing quietly.
+      - name: Verify litellm patch runner
+        run: npm run verify:litellm-patch-runner
+
       - name: Backend TypeScript check
         run: |
           cd backend
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "lint:fix:backend": "cd backend && npm run lint:fix",
     "lint:fix:frontend": "cd frontend && npm run lint:fix",
     "verify:moltbot-tools": "node scripts/verify-moltbot-tool-contract.js",
+    "verify:litellm-patch-runner": "node scripts/verify-litellm-patch-runner.js",
     "prepare": "husky"
   },
   "devDependencies": {
--- a/scripts/verify-litellm-patch-runner.js
+++ b/scripts/verify-litellm-patch-runner.js
@@ -157,11 +157,26 @@ function introspectPatches(runnerSrc, fixture) {
  * matching `"$PYBIN" -c "…" "${CHATGPT_DISABLE_DEVICE_LOGIN:-}"`.
  */
 function runPython(src, fixture, argv) {
-  return spawnSync('python3', ['-c', src, ...argv], {
+  const res = spawnSync('python3', ['-c', src, ...argv], {
     cwd: fixture.app, // mirrors the container WORKDIR (/app)
     env: { ...process.env, PYTHONPATH: fixture.site, PYTHONDONTWRITEBYTECODE: '1' },
     encoding: 'utf8',
   });
+
+  // A missing interpreter has to name itself. Without this the first caller dies on
+  // `res.stdout` being undefined and the stack points at chart introspection — the
+  // guard blaming the artifact it guards for its own absent dependency. It already
+  // exited non-zero (so CI goes red either way, and this guard is not decoration);
+  // what was wrong was the cause it named.
+  if (res.error || typeof res.stdout !== 'string') {
+    const why = res.error ? res.error.code || res.error.message : 'produced no stdout';
+    fail(
+      `cannot run python3 (${why}).\n` +
+        '  This guard reproduces the boot block by executing it, so it needs a python3 on PATH.\n' +
+        '  Nothing was verified — treat this as UNKNOWN, not as a passing chart.'
+    );
+  }
+  return res;
 }
 
 /**
--- a/scripts/verify-litellm-patch-runner.js
+++ b/scripts/verify-litellm-patch-runner.js
@@ -1,0 +1,355 @@
+#!/usr/bin/env node
+/**
+ * Guard for the LiteLLM boot-time patch runner in
+ *   k8s/helm/commonly/templates/agents/litellm-deployment.yaml
+ *
+ * WHY THIS EXISTS
+ * ---------------
+ * On 2026-08-02..05 the litellm proxy crash-looped for 79 hours (762 restarts,
+ * whole agent fleet dark). Three boot-time source patches shared one resolution
+ * mechanism -- `os.path.dirname(litellm.__file__)` -- evaluated by a `python3 -c`
+ * process whose CWD was /app. Two copies of the package exist in that image:
+ *
+ *   /app/litellm                                       <- source checkout ("shadow")
+ *   /app/.venv/lib/python3.13/site-packages/litellm     <- what the proxy imports
+ *
+ * `python3 -c` puts '' (CWD) at sys.path[0], so the patcher resolved to the
+ * shadow. The proxy runs as the console script /app/.venv/bin/litellm, whose
+ * sys.path[0] is the script's directory, so it resolved to site-packages.
+ * Every patch edited a real file, in the right package, that the process it was
+ * protecting never imported -- and then "verified" itself by reading back the
+ * same path it had just written. That check is closed by construction: it
+ * confirms the write landed where the writer aimed, which was never in doubt.
+ *
+ * Measured in the deployed image (2026-08-06), for the record:
+ *
+ *   proxy (console script)          -> .../site-packages/litellm   <- ground truth
+ *   bare python3, raw,   cwd=/app   -> /app/litellm                <- the bug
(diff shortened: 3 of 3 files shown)


lilyshen0722 (commented):
Reviewed by **running it**, not reading it — the harness is sound, and I found one structural gap.

## Verified by execution

| check | result |
|---|---|
| as shipped, against merged `e55f0c21` chart | **8/8**, exit 0 |
| mutation: CWD-strip deleted, interpreter pin kept | **5/8** — cases 1, 6, 8. Matches your table exactly |
| case 6's failure text on that mutation | `claimed EFFECT verified while the proxy-resolved copy was absent` — the production signature |
| exit code, chart file absent | **1** |
| exit code, runner block deleted (anchor gone) | **1**, with a re-point-`START_RE` message |
| run from a foreign CWD (`/tmp`) | 8/8 — chart resolves off the script, not `$PWD` |

I went looking specifically for the failure mode you self-caught — a guard that reports safety when it never ran — and **it isn't there.** Both could-not-run paths exit 1 rather than passing vacuously. Reporting that as a checked negative, not an assumption.

Two things I'd call out as better than asked for:

- **Case 4 asserts the consequence, not the symptom.** `:305-309` requires `device-login-disable patch EFFECT verified` in stdout after a non-critical miss, with the failure string naming the loop-abort. That's the ordering correction, pinned — not "exit code was 0."
- **Case 6 carries a positive control.** `:328` asserts the shadow copy *was* still patched, so the case cannot pass by the runner doing nothing. That's the distinction between "correctly refused" and "never ran," and it's the one most refusal tests omit.

## Finding: the guard is wired to nothing

`scripts/verify-litellm-patch-runner.js` is referenced by **zero** files. Not in `package.json`, not in any workflow. The precedent it cites is wired at both tiers:

```
scripts/verify-moltbot-tool-contract.js
  -> package.json:16                "verify:moltbot-tools": "node scripts/..."
  -> .github/workflows/tests.yml:55 - name: Verify moltbot tool contract
```

The PR body's own claim is conditional — *"**Wired into CI** it becomes the guard that makes the next template edit fail red instead of shipping another silent no-op."* As merged, #860 doesn't satisfy its own antecedent: it's a script someone has to remember to run, which is the category of artifact it was built to replace. The three template edits that caused this SEV would each still ship green.

Two lines close it:

```diff
   "scripts": {
     "verify:moltbot-tools": "node scripts/verify-moltbot-tool-contract.js",
+    "verify:litellm-patches": "node scripts/verify-litellm-patch-runner.js",
```

```diff
       - name: Verify moltbot tool contract
         run: npm run verify:moltbot-tools
+
+      - name: Verify litellm patch runner
+        run: npm run verify:litellm-patches
```

**One constraint worth knowing before you push it:** a commit touching `.github/workflows/` needs the workflow-scoped token (samxu01), not the default account — switch for that push only and switch back, keeping authorship Lily. It fails at push time otherwise, which is a confusing 2am surface for a two-line change.

Approving on substance; the wiring is what makes it a guard rather than a document.

lilyshen0722:
**Wiring gap closed — the harness is now a gate, not a file.** `3221d412`.

@pod-architect named this in 52875 and offered to push it; taking it since #860 is mine. It's the same shape I've been filing all night — a real check that no gate consults ships nowhere, exactly like a description that never gets published or a patch applied to a copy nobody imports.

Wired the way the existing contract check already is, so there's one pattern and not two:

```
package.json          "verify:litellm-patch-runner": "node scripts/verify-litellm-patch-runner.js"
tests.yml     step 6  Verify litellm patch runner   (right after Verify moltbot tool contract)
```

**Verified in CI, not inferred from a green tick.** A passing workflow doesn't prove a specific step ran, so I pulled the step's own output out of run `31064601126`:

```
Verify moltbot tool contract: success
Verify litellm patch runner:  success

02:05:57.189Z  litellm patch-runner guard
02:05:57.406Z    ✓ 8. the CWD-strip is present — dropping it reintroduces the outage
02:05:57.406Z  8/8 cases passed
```

0.22s wall, no new dependencies — the script is node builtins only, and `ubuntu-latest` supplies the python3.

---

**Second change: the guard now names a missing python3 instead of blaming the chart.**

Checked what happens when the interpreter is absent, because a CI guard that passes without its dependency is decoration. Good news first: it already exited **1**, so CI would have gone red rather than silently green. But the message was wrong about the cause:

```
TypeError: Cannot read properties of undefined (reading 'split')
    at introspectPatches (scripts/verify-litellm-patch-runner.js:140:27)
```

`res.stdout` is `undefined` on ENOENT, and the first thing to touch it is chart introspection — so the stack points at the artifact being guarded, for the guard's own missing dependency. Someone hitting that in CI would go read the Helm chart. Now:

```
✗ cannot run python3 (ENOENT).
  This guard reproduces the boot block by executing it, so it needs a python3 on PATH.
  Nothing was verified — treat this as UNKNOWN, not as a passing chart.
```

The last line is the point: *no result* and *a passing chart* have to be distinguishable, or the guard has the same defect as `patch applied` — a message that's true about the wrong thing. Fixed in `runPython`, so all 8 cases and the introspection inherit it.

**What I verified where, since these aren't the same claim:**

| | verified | how |
|---|---|---|
| guard runs green in CI | ✅ | step log above, runner's own python3 |
| guard runs green locally | ✅ | 8/8, exit 0 |
| missing-python3 path | ✅ | `PATH=<node-only>` → exit 1 + the new message |
| guard goes **red** on a bad chart | ✅ **locally only** | mutation B (CWD-strip deleted) → 5/8, incl. case 6 on the production signature |

**The gap I'm stating rather than papering over:** the mutation testing that proves this guard *discriminates* was run on local python3, not on the runner's. The CI run proves it executes and passes there. `python3 -c` seeding `sys.path[0]` with `''` is stable across 3.x, so I expect no divergence — but I haven't measured CI's python doing the discriminating, and "it passed in CI" is not that measurement.

PR #891, opened 2026-08-11
docs(design): agent status honesty — brief for fleet design review

lilyshen0722:
Sam's call after the smoke test found `ngoc-tran` hitting silent mentions a day
after the fix merged: the #888 label was a patch — the real design is **agent
health as a first-class surface**, and it goes through the fleet for design
review before implementation.

Five honestly-derivable states (all from signals that exist today), four named
casualties grounding it, and one hard open problem: **Degraded** — the
hq-support state where an agent acks events but cannot act — has no clean
kernel signal, and both candidate fixes have costs.

The priority surface is deliberately not a dashboard. It is the **moment a
human @mentions an agent that cannot answer** — an inline explanation at send
time, in the room, which is the intervention that would have saved every user
named in the brief.

Review assignments: UX Lead (interaction, alarm/calm balance), Pod Architect
(state derivations, the Degraded signal), Sprint Review (adversary on
ownership-keyed alarming). Runs in the Sharpen pod under the 0.2.0 tone
contract — which makes it the first real test of that contract too.

Merging this PR = the brief is the agreed starting point; the reviewed spec
comes back to Sam for ratification before code.

### DIFF
--- a/docs/design/agent-status-honesty.md
+++ b/docs/design/agent-status-honesty.md
@@ -59,10 +59,19 @@ thresholds, and the only row that catches a dead-on-arrival install
   live.
 - **`runtimeType` goes through `normalizeRuntimeIdentity`** — it is unset on
   88% of stored installs; a rule reading the raw field inspects 24 of 200.
-- **States need a precedence rule.** Misconfigured (webhook, no URL) and
-  Never-connected fire together on the same 14 installs today — and two of
-  those 14 have the freshest `lastUsedAt` in the whole fleet. Structural
-  states outrank inferred ones.
+- **Liveness and config are orthogonal axes, not one enum.** The first
+  revision resolved the Misconfigured/Never-connected collision with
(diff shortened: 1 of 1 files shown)


lilyshen0722:
Review round complete — the fleet reviewed in Sharpen (msgs 52986–52991) and `d9cdbc4a` folds every accepted finding back into the brief. Headline changes: the key splits (alarm=ownership, explanation=dependency), the signal table goes per-runtime-class (AgentRun for native, route diversity for outbound-authenticating, lastUsedAt only for CLI wrappers), a sixth state **Unreachable** (declared trigger with no producer), and the Degraded problem is resolved without new telemetry. Five explicit decisions are queued under "Decisions for ratification."

The review also surfaced a live defect chain, already handled separately: #895 (native heartbeat opt-in projection — the 8-day first-party outage), #896 (boot-backlog ack-without-spawn), #897 (kernel cascade gap), #898 (pod.join producer / Unreachable implementation).

Issue #887, opened 2026-08-09
4 of 6 user @mentions in 10 days got no reply — two distinct causes, neither is user behaviour


lilyshen0722:
Measured on production 2026-08-09, prompted by the support bot flagging its own
broken tooling in a user-facing message.

## The number

Human `@mentions` in pods, last 10 days:

| | |
|---|---|
| mentions sent | 6 |
| got an agent reply | 2 |
| **silent** | **4** |

```
✗ Aug 09  user-8863   @opencode-test 开发好了吗
✗ Aug 09  user-8863   @opencode-test 当前配置接入commonly接入OK了吗
✗ Aug 07  spiderpc    @spiderpc-agent 如何付费
✗ Aug 06  Sam         @sprint-review what's a TLDR?
```

**This is not a user-behaviour problem.** These users did exactly the right
thing — they connected an agent and addressed it by handle. `user-8863` was
literally asking their agent *whether the Commonly connection was working*. It
could not answer, which is its own answer.

## Cause A — webhook installs that can never receive anything

`opencode-test` and `spiderpc-agent`:

```
runtime      : {"runtimeType":"webhook"}   <- no URL anywhere in the config
installation : active, in the right pod
tokens       : 1 issued
events       : NONE EVER
```

The mention resolves correctly — agent exists, is a pod member, install is
active, and `buildMentionMap` has no runtime filter, so `@opencode-test` maps to
a single active install. **Yet no `chat.mention` event was ever created for
either agent.** Zero events in that pod at all.

Across the whole instance in the last 5 days, no webhook-runtime agent appears
in `agentevents` — only `openclaw`, `claude-code` wrappers and `commonly-bot`.

**There are 13 active `webhook` installs.** If this reproduces for all of them,
that is 13 agents users believe they have connected, that can never answer.

Worth noting the likely product-level mismatch behind it: a user running
`opencode` locally **polls** CAP. A `webhook` runtime expects us to POST to
their endpoint. If the connect flow assigns `webhook` to someone who is
actually going to poll, the install is wrong from birth and no amount of
delivery logic saves it.

I have not pinned the exact branch that drops the enqueue — that wants a
reproduction (a test that mentions a webhook-runtime install and asserts an
event lands), not more code reading.

## Cause B — the support agent has no `commonly_*` tools

`hq-support` is the opposite failure: events **are** delivered and acked, but
the agent cannot act. It said so itself, in a message to users:

[...]

PR #925, opened 2026-08-13
feat(kernel): approval cards — payload column + ApprovalAction lifecycle + delegated create_pod (ADR-020)

lilyshen0722:
ADR-020 (ratified, #924) D2+D3+D7, first kernel PR of the build. Sharpen seats: **@sprint-review** this is yours to review; **@pod-architect** please check `ApprovalAction` + `approvalActionService` against ADR-017's invariants specifically.

**What lands:**
- **Structured message `payload`** — Mongo `Mixed` + PG `JSONB` (self-applying `schema.sql` ALTER, the `is_bot` pattern), threaded through every explicit whitelist: PG create/find column lists, agentMessageService's five literals, `normalizeMongo`, the `newMessage` socket literal. Bonus fix: the Mongo schema finally declares `metadata` — strict mode silently dropped it on the fallback path.
- **`ApprovalAction`** — ADR-017's lifecycle as code: `flagged → resolved/expired/moot`; only a human writes `resolved` (plain `auth` route, never `dualAuth`, plus service-level `isBot` defense); owner-only; atomic one-way transition; expiry fails closed; execute-at-most-once; execution failure recorded honestly on a kept decision. The resolved row IS D2's AuthorizedAction audit record.
- **`create_pod` executor** — delegated authority in practice: pod owned by the USER (`createdBy` + first member — never the bot), proposing agent joins with its installation cloned, PG mirrored FK-safe.
- **Card UI** — `V2ApprovalCard` renders five faces from payload; `messageCardUpdated` socket event (verbatim `messageReaction` pattern) patches one field on one message; no optimistic update. en + zh.

**Not in this PR** (next: producer PR): the `commonly_propose_action` native tool, the per-agent tool gating fix (recon found `TOOLS` goes unfiltered to every native agent), and the Guide prompt change. No card can exist in prod until the producer lands — this PR is inert until then by design.

Tests: 16 green locally (authorization matrix, lifecycle one-way-ness, D2 ownership assertions). Typecheck clean both tiers; frontend jest in CI.

### DIFF
--- a/backend/__tests__/unit/models/PgMessage.test.js
+++ b/backend/__tests__/unit/models/PgMessage.test.js
@@ -30,7 +30,7 @@ describe('PG Message model', () => {
     expect(pool.query).toHaveBeenNthCalledWith(
       1,
       expect.stringContaining('INSERT INTO messages'),
-      ['p', 'u', 'c', 'text', null],
+      ['p', 'u', 'c', 'text', null, null],
     );
     expect(pool.query).toHaveBeenNthCalledWith(
       2,
--- a/backend/__tests__/unit/services/approvalActionService.test.js
+++ b/backend/__tests__/unit/services/approvalActionService.test.js
@@ -1,0 +1,192 @@
+/**
+ * ADR-020 D2/D3 — the approval lifecycle's load-bearing invariants:
+ *  - only the owner decides; agents (isBot) never decide even as owner
+ *  - one-way transitions: resolve at most once, execute at most once
+ *  - expiry fails closed ("retiring an escalation is never an approval")
+ *  - decline never executes
+ *  - the executor creates resources owned by the USER, never the bot
+ */
+
+const mockApprovalFindById = jest.fn();
+const mockApprovalFindOneAndUpdate = jest.fn();
+const mockApprovalCreate = jest.fn();
+jest.mock('../../../models/ApprovalAction', () => ({
+  findById: (...args) => mockApprovalFindById(...args),
+  findOneAndUpdate: (...args) => mockApprovalFindOneAndUpdate(...args),
+  create: (...args) => mockApprovalCreate(...args),
+}));
+
+const mockPodCreate = jest.fn();
+const mockPodFindById = jest.fn();
+const mockPodUpdateOne = jest.fn();
+jest.mock('../../../models/Pod', () => ({
+  create: (...args) => mockPodCreate(...args),
+  findById: (...args) => mockPodFindById(...args),
+  updateOne: (...args) => mockPodUpdateOne(...args),
+}));
+
+const mockUserFindById = jest.fn();
+jest.mock('../../../models/User', () => ({
+  findById: (...args) => mockUserFindById(...args),
+}));
+
+const mockInstallFindOne = jest.fn();
+const mockInstallUpsert = jest.fn();
+jest.mock('../../../models/AgentRegistry', () => ({
+  AgentInstallation: {
+    findOne: (...args) => mockInstallFindOne(...args),
+    findOneAndUpdate: (...args) => mockInstallUpsert(...args),
+  },
+}));
+
+const mockGetOrCreateAgentUser = jest.fn();
+jest.mock('../../../services/agentIdentityService', () => ({
+  getOrCreateAgentUser: (...args) => mockGetOrCreateAgentUser(...args),
+  syncUserToPostgreSQL: jest.fn(),
+}));
+
+jest.mock('../../../config/socket', () => ({ getIO: jest.fn(() => null) }));
+
+const { resolveApproval } = require('../../../services/approvalActionService');
+
+const OWNER = 'aaaaaaaaaaaaaaaaaaaaaaa1';
+const STRANGER = 'bbbbbbbbbbbbbbbbbbbbbbb2';
+
+const flaggedRow = (over = {}) => ({
+  _id: 'appr-1',
+  podId: 'pod-1',
+  messageId: 'not-numeric-mongo-id',
+  ownerUserId: OWNER,
+  agentName: 'guide',
+  instanceId: 'udefault',
+  actionType: 'create_pod',
+  params: { name: 'Design Studio', type: 'chat' },
+  summary: 'Create a pod for design work',
+  status: 'flagged',
+  expiresAt: new Date(Date.now() + 60 * 60 * 1000),
+  save: jest.fn().mockResolvedValue(undefined),
+  ...over,
+});
+
+beforeEach(() => {
+  jest.clearAllMocks();
+  delete process.env.PG_HOST;
+  mockUserFindById.mockReturnValue({
+    select: jest.fn(() => ({ lean: jest.fn().mockResolvedValue({ isBot: false }) })),
+  });
+  // Mongo card-message update path (messageId non-numeric).
+  // eslint-disable-next-line global-require
+  jest.mock('../../../models/Message', () => ({ updateOne: jest.fn().mockResolvedValue({}) }), { virtual: false });
+});
+
+describe('resolveApproval authorization', () => {
+  test('404s on unknown approval', async () => {
+    mockApprovalFindById.mockResolvedValue(null);
+    const res = await resolveApproval({ approvalId: 'x', callerUserId: OWNER, decision: 'approved' });
+    expect(res.status).toBe(404);
+  });
+
+  test('403s a caller who is not the owner', async () => {
+    mockApprovalFindById.mockResolvedValue(flaggedRow());
+    const res = await resolveApproval({ approvalId: 'appr-1', callerUserId: STRANGER, decision: 'approved' });
+    expect(res.status).toBe(403);
+    expect(mockApprovalFindOneAndUpdate).not.toHaveBeenCalled();
+  });
+
+  test('403s a bot caller even when it is the owner — no agent may decide', async () => {
+    mockApprovalFindById.mockResolvedValue(flaggedRow());
+    mockUserFindById.mockReturnValue({
+      select: jest.fn(() => ({ lean: jest.fn().mockResolvedValue({ isBot: true }) })),
+    });
+    const res = await resolveApproval({ approvalId: 'appr-1', callerUserId: OWNER, decision: 'approved' });
+    expect(res.status).toBe(403);
+    expect(mockApprovalFindOneAndUpdate).not.toHaveBeenCalled();
+  });
+});
+
+describe('resolveApproval lifecycle', () => {
+  test('already-resolved returns 409 and never re-executes', async () => {
+    mockApprovalFindById.mockResolvedValue(flaggedRow({ status: 'resolved', decision: 'approved' }));
+    const res = await resolveApproval({ approvalId: 'appr-1', callerUserId: OWNER, decision: 'approved' });
+    expect(res.status).toBe(409);
+    expect(mockPodCreate).not.toHaveBeenCalled();
+  });
+
+  test('expiry fails closed: 410, row marked expired, nothing executes', async () => {
+    const row = flaggedRow({ expiresAt: new Date(Date.now() - 1000) });
+    mockApprovalFindById.mockResolvedValue(row);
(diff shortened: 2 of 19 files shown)


lilyshen0722 (commented):
Reviewed against ADR-020 D1–D3. **One finding worth fixing before merge**, one minor, and a list of what I did *not* check.

Scope note: @pod-architect has the ADR-017 lifecycle invariants (only-a-human-resolves / expiry-fails-closed / execute-at-most-once). I deliberately did not re-audit those — I took D2 delegated ownership, the route surface, the dual-DB payload column, and the frontend.

---

## Finding 1 — partial execution reports success, and the honest-failure UI can never fire for it

`approvalActionService.executeCreatePod` wraps the agent-join in a catch that only warns, and the success return sits *after* it, unconditional:

```js
} catch (agentErr) {
  console.warn('[approval] create_pod agent-join failed:', ...);
}
return { podId: String(pod._id), podName: pod.name, podType: type };
```

So the two paths diverge:

| what fails | `executionError` | card renders |
|---|---|---|
| `Pod.create` | set (executeAction throws) | `approvalCard.approvedButFailed` ✓ |
| agent-join | **never set** | `approvalCard.approved` — "✓ Approved — done." ✗ |

The machinery for the honest face already exists and works — `V2ApprovalCard.tsx:91` renders `approvedButFailed` off `payload.executionError`, and the total-failure case is tested ("execution failure keeps the decision and records the error honestly"). Only the routing is missing, so this is a small fix, not a redesign.

**Why it matters more than a warning log:**

- This function's own comment says membership without an `AgentInstallation` **cannot post — the runtime 403s**. A half-joined Guide is a Guide that silently cannot speak in the pod it just asked to create.
- D2 names "Guide added as a member" as part of what an approved action *is*. Returning success without it means the approval record claims an outcome that didn't happen.
- **It is not recoverable.** Execution is at-most-once by design — re-resolving returns 409 "Already decided". The card is spent, there's no retry path, and no surface reports the gap.
- The end state is a user who was told their agent is in the room, watching it not respond. That is precisely the #887 silent-agent failure the status-honesty brief exists to eliminate, reintroduced one layer down.

The PG-mirror catch above it has the same shape but is materially less severe — a mirror is derivable and repairable out of band.

**Suggested fix**, cheapest version: let the agent-join failure reach `executionError` (or add an `executionWarnings` field if you'd rather keep "approved + working pod" distinct from "approved + broken pod"). Throwing instead would surface it but orphans the created pod, so I'd not do that without a compensating delete.

Stated as read from source — the control flow is unambiguous, so I did not build a repro for it.

## Finding 2 (minor) — `MessageType` union and schema enum are twinned by a comment

`models/Message.ts` carries: *"The union and the schema enum below are declared independently — change BOTH."* A comment is the only thing holding those in sync, and TS erasure means a test can't compare them directly.

Worth closing structurally rather than by discipline:

```ts
export const MESSAGE_TYPES = ['text', 'image', 'system', 'card'] as const;
export type MessageType = typeof MESSAGE_TYPES[number];
// ...
type: { type: String, enum: [...MESSAGE_TYPES], ... }
```

One declaration, no drift possible, comment unnecessary.

---

## What's right, specifically

- **The PG migration is done correctly** — this is the failure mode I went looking for and it's handled. `ALTER TABLE messages ADD COLUMN IF NOT EXISTS payload JSONB` lives in `schema.sql` following the existing `is_bot` pattern, and `init-pg-db.ts` applies the whole file unconditionally at boot (`server.ts:313`). Existing deployments get the column, not just fresh installs. A `schema.sql`-only change that silently skipped live databases would have passed every test here (pg-mem builds from the same file) and broken in production.
- **Route is human-only by construction** — plain `auth`, never `dualAuth`, with a service-layer `isBot` check as defence in depth, and the limiter *before* auth. The comment explaining why is worth keeping.
- **D2 conformance holds**: `createdBy: ownerUserId`, owner seeded into `members`, agent joined *with* an installation rather than bare membership.
- **`CREATABLE_POD_TYPES` excludes DM-shaped types**, protecting the §3.10 two-member invariant — good catch to make that explicit rather than relying on the caller.
- **i18n parity is clean**: 15/15 keys, and zh-CN is genuinely translated rather than English copied forward.
- The `metadata` schema declaration fixes a real latent bug in passing — strict-mode Mongoose was silently dropping a field `agentMessageService` had been passing all along.

## What I did NOT verify

- **The three ADR-017 invariants** — @pod-architect's assigned lane. I observed tests covering all three and passing, but did not audit them independently.
- **Anything visual.** No browser check. `v2.css` +129 lines and five card states are unverified for layout; per CLAUDE.md v2 CSS needs a real-browser pass because jsdom has no layout engine.
- **`updateCardEverywhere` socket fanout** — not exercised by me; D3 calls for "server-state change → socket fanout" and I only read it.
- **Full suite.** CI's Test & Coverage was still pending when I reviewed. I ran `approvalActionService.test.js` locally on node@22: **9/9 pass**.

lilyshen0722 (commented):
Second pass on the amendment (`2b77db63` + `0f722a30`). **My finding is properly fixed and the trap @pod-architect pre-registered was avoided.** One non-blocking note, one minor still open, and one thing I could not verify locally.

Reviewed the two amendment commits specifically — the range also contains a main merge (#923/#924), which isn't this PR's work.

## Finding 1 — fixed, and fixed at the right layer

`agentJoined` / `agentJoinError` are now carried out of `executeCreatePod` instead of swallowed, the `else` branch also catches the unresolvable-bot-user case I hadn't called out separately, and the card renders `--attention` rather than a clean done:

```
agentJoinFailed = execution?.agentJoined === false
```

Strict `=== false` is the right choice — `undefined` on pre-existing rows and on non-`create_pod` actions won't false-positive into an attention face.

The copy names the fix, which is what D3 asks for: *"the pod was created, but {agent} could not join it. Open the pod and re-add the agent."* en/zh parity held (both locales got both new keys). Regression test present and named.

## @pod-architect's 53054 trap — avoided

They flagged that the decidable-expiry path only works if the resolve path *stops* writing `status='expired'`, since the atomic filter matches on `status:'flagged'`. Checked: that write is gone entirely, and there is no remaining writer of `'expired'` for `ApprovalAction` anywhere in `backend/` outside the type and enum declarations. So the transition still matches and the decidable path doesn't 409.

Also good that `0f722a30` amends ADR-020 D3 in the same PR rather than leaving code and spec disagreeing — the conflict was the actual defect, and it's resolved in the artifact that carries it.

## Non-blocking note — the dead `expired` branch is a loaded gun

Two of `ApprovalStatus`'s four states now have no writer: `moot` (as @pod-architect established) and, after this change, `expired`. That's correct for the current design — the age warning is keyed on `flagged` + past `expiresAt`, not on status.

The risk is forward-looking. `V2ApprovalCard:36` still reasons about "an explicitly-transitioned `expired` status," so if anyone later adds an expiry sweep that writes `status='expired'` — an obvious-looking piece of housekeeping — the card renders dead buttons again and the fail-silent behavior ADR-017:201 just removed comes straight back, with no test failing.

Cheapest guard: drop the enum value, or leave a comment at the model saying nothing may write it and why. Same class as the Unreachable state in the status-honesty brief — a declared value with no producer, except here the producer is the thing to prevent rather than to build.

## Minor from my first pass, still open

`MessageType`'s union and the schema enum in `models/Message.ts` are still twinned by a comment. Non-blocking; the `const MESSAGE_TYPES = [...] as const` shape closes it whenever convenient.

## Verified / not verified

**Verified by me, on `0f722a30`:**
- `approvalActionService.test.js` — **10/10 pass** locally on node@22, including `agent-join failure is carried in the result, never a clean "done"` and the inverted expiry test.
- **Frontend typecheck: clean.**
- No remaining `status='expired'` writer; card face, i18n parity, strict `=== false` guard.

**Not verified — and the failure was mine, not the PR's:** my local backend `tsc:check` reported two `@sentry/node` errors in `instrument.ts`. That's my instrument: I symlinked `node_modules` from a checkout 44 commits behind, `@sentry/node` **is** declared in this PR's `backend/package.json`, and `instrument.ts` predates this PR. Flagging it only so nobody re-runs it and thinks they found something. CI's Test & Coverage is the authority and was still pending when I finished.

**Also not verified:** anything visual — still no browser check on the card faces, now including the new attention state; and socket fanout.

**Position:** my findings are cleared. I'd merge on CI green — the backend typecheck is the one claim in the ship message I couldn't independently confirm.

lilyshen0722 (commented):
Third pass, one **docs-only** finding — non-blocking for behaviour, but it should land before merge because it's two lines and it sits in the file that defines the lifecycle.

**@pod-architect's misquote catch was applied in one of three places.** They flagged (53047) that the expiry branch borrowed `moot`'s rule, against ADR-017:228 — *"expired is TTL only; nothing that means don't look may borrow it."* The service was fixed:

```
"never an approval" in approvalActionService.ts:   before 1  →  after 0   ✓
```

Two verbatim copies survive, and both now teach the behaviour this PR removed:

**1. `models/ApprovalAction.ts:16-17`** — inside the header block introduced as *"Invariants the route layer enforces and tests pin"*:

```
 *   - past expiresAt, resolution fails closed (410) and the row is marked
 *     expired on read — "retiring an escalation is never an approval"
```

All three clauses are now false: resolution past `expiresAt` is honored, nothing marks the row expired on read (no reader and no writer — that's @pod-architect's 53048, and I confirmed no `status='expired'` writer remains anywhere in `backend/`), and it still borrows `moot`'s rule. The correction exists — but 30 lines down at `:48-50`, attached to `decidedAfterExpiry`, where a reader of the header contract never reaches it.

**2. `__tests__/unit/services/approvalActionService.test.js:5`** — the suite header:

```
 *  - expiry fails closed ("retiring an escalation is never an approval")
```

The suite below it now asserts the opposite: *"a decision past expiresAt is honored and stamped decidedAfterExpiry (ADR-017:201)"*. A file describing itself as pinning X while pinning ¬X is the worst version of this, because the header is what someone reads to find out what's covered.

**Why flag a comment at all:** this is the #861/#863 dead-cue class the team already has a rule for — the correction landed at the site that raised it, while verbatim copies kept the dead rule alive. It matters more than usual here because `ApprovalAction.ts`'s header is the natural place someone lands before adding the expiry sweep, and that header currently tells them the row is *supposed* to be marked expired — which is exactly the change that would silently restore fail-silent behaviour (the loaded-gun note from my second pass).

Suggested: replace both with the ADR-017:201 rule and, at the model, state that nothing may write `status: 'expired'` and why.

Position unchanged otherwise: my findings are cleared, merge on CI green.

lilyshen0722 (commented):
`87ce000d` fixes both stale copies correctly — control confirms zero occurrences of the borrowed `moot` quote remain in `backend/`, and the test-suite header now matches what the suite actually asserts. That closes my third-pass finding.

**One thing it sharpens rather than closes**, and it's the loaded-gun note from my second pass — non-blocking, but the new wording points *at* the hazard:

```
 *     ... Nothing writes status 'expired' in v1 — the state exists for
 *     future sweeps only.
```

Accurate, and it converts my warning into an invitation. Measured what a sweep would actually do:

```
V2ApprovalCard.tsx:73    buttons render ONLY under status === 'flagged'
V2ApprovalCard.tsx:130   status === 'expired' → terminal face, no buttons
```

So the moment anything writes `status: 'expired'` — the housekeeping this comment now signposts — the card drops to a button-less terminal face and the late decision is refused again. That is precisely the fail-silent behaviour ADR-017:201 removed and this PR just spent two commits eliminating, restored by a change the model file recommends, with no test failing.

Also worth noting for consistency: `approvalCard.expired` is currently an unreachable face. That's the exact criterion used to delete the `moot` face in `2b77db63` — applied to one and not the other.

Two cheap options, either fine:
- Delete the `expired` face and key everything on age (consistent with the `moot` deletion), or
- Keep it, and make the comment state the constraint rather than the invitation: a future sweep may mark rows for housekeeping but **must never render the card undecidable** — expiry is age, not refusal, at every layer.

The second is probably better since a sweep is genuinely wanted eventually; it just has to not touch decidability.

**Position: cleared on `87ce000d`, merge on CI green.** Test & Coverage and E2E were still pending when I checked.

PR #928, opened 2026-08-13
style(v2): align accents and pod-list geometry

lilyshen0722:
## Summary
- remove the remaining message accent rail while preserving the mention wash; unify quote edges at 3px
- keep Community in a capped 3+auto filter grid; give every pod row one reserved preview line and a fixed time slot
- route the live pulse and board focus treatment through shared semantic tokens

## Shadow audit
Classified all 42 v2 declarations: floating menus/popovers/modals and focus rings remain; resting surfaces inherit the existing no-shadow V2 tokens.

## Verification
- `npm test -- --watchAll=false --runInBand src/v2/__tests__/v2-layout-invariants.test.ts` (28 passed)
- `npm run typecheck`
- `npm run build`
- `npm run lint` remains blocked by the pre-existing `PersonalityBuilder.tsx:440` unescaped-entity error

No browser MCP is available in this runtime, so the specified visual pass remains for review.

### DIFF
--- a/frontend/design-system/tokens.css
+++ b/frontend/design-system/tokens.css
@@ -40,6 +40,8 @@
   /* ---------- Semantic ---------- */
   --c-success:         #10b981;
   --c-success-soft:    #e3f6ec;
+  --c-success-ring:    rgba(16, 185, 129, 0.40); /* live-activity pulse */
+  --c-success-ring-strong: rgba(16, 185, 129, 0.50);
   --c-warning:         #f4a23a;
   --c-warning-soft:    #fdefdc;
   --c-info:            #0891b2;   /* cyan — distinct from accent blue */
--- a/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
+++ b/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
@@ -30,7 +30,11 @@ const read = (rel: string): string =>
 // Grab the body of the first `<selector> { ... }` block. Selectors here have no
 // nested braces, so a naive slice to the next `}` is sufficient.
 const ruleBody = (css: string, selector: string): string => {
-  const start = css.indexOf(`${selector} {`);
+  // Prefer a selector at the start of a CSS line. A descendant selector can
+  // contain the same text (`.parent .target {`) and is not the rule being
+  // pinned.
+  const lineStart = css.indexOf(`\n${selector} {`);
+  const start = lineStart === -1 ? css.indexOf(`${selector} {`) : lineStart + 1;
   if (start === -1) return '';
   const end = css.indexOf('}', start);
   return end === -1 ? '' : css.slice(start, end);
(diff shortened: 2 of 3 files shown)


lilyshen0722 (commented):
Ran the real-browser check that wasn't available on your side (CLAUDE.md requires one for v2 layout, since jsdom has no layout engine). **Most of it verifies. One regression: the new fixed time slot clips the two most common timestamp formats.**

Harness: real `v2.css` + `tokens.css` from this PR head, served over http, markup replicated from `V2PodsSidebar.tsx:658-693`, 300px rail. Instrument control first — v2.css applied (897 rules) and `.v2-pods__item` computed to exactly `64px`.

## Verified working

| claim | measured |
|---|---|
| fixed 64px pod rows | all 5 rows exactly **64px** |
| reserved one-line snippet | snippet **16px** even when empty |
| stable title slot | title **20px** |
| no clipping of row content | `scrollHeight > clientHeight` false on every row |
| capped Community grid | `fit-content(92px)` resolves to **66.05px** — under the cap, label renders in full |

The reserved-slot behaviour is genuinely right: an empty snippet still holds its 16px, so rows don't jump.

## The regression — `.v2-pods__item-time { width: 44px }`

New in this PR; base `#927` had no width constraint on that rule at all, so the time sized to content.

Measured on the real elements at the real inherited font (SF Pro Text 11px), constraint temporarily relaxed:

```
"Yesterday"   needs 51.7px   slot 44px   CLIPS  (−7.7px)
"11:59 PM"    needs 51.2px   slot 44px   CLIPS  (−7.2px)
"2:04 PM"     needs 44.2px   slot 44px   grazes — renders "2:04 P…"
"Aug 12"      needs 37.6px                fits
"Wed"         needs 23.3px                fits
```

Visually the first row renders **`11:59 …`** — the AM/PM marker is the part elided, so a user cannot tell 11:59 in the morning from 11:59 at night.

`formatRelativeTime` (`v2/utils/grouping.ts:73`) has four branches. **Two of them clip, and they are the two most recent:** today (`h:mm AM/PM`) and `"Yesterday"`. The weekday and month-day branches — older pods, further down the list — are the ones that fit.

**Suggested fix:** ~52px clears both at this font. But the better shape may be to drop the fixed width and let the *title* absorb the shrink — it already carries `min-width: 0` + ellipsis and degrades gracefully, and a truncated pod name costs far less than a truncated timestamp.

## Why 28/28 passed, and a trap in the fix

The guard asserts presence, not geometry:

```js
expect(time).toContain('width: 44px');
```

Green is fully consistent with the clip — the test confirms the constraint exists and cannot know 44px is too small. This is exactly the gap CLAUDE.md describes.

**More importantly: that assertion now pins the bug.** Widening or removing the cap turns the test red, so the fix has to update `v2-layout-invariants.test.ts:183` in the same change. Worth knowing before someone tries the one-line fix and reads the red as their own mistake.

## Instrument notes — two of my three methods were wrong

Recording these because the numbers above would be wrong if I'd stopped at either:

1. **`scrollWidth > clientWidth` under-reports.** With `text-overflow: ellipsis` the rendered content fits by construction, so the two are equal. It reported `"2:04 PM"` as fine; the screenshot showed it elided.
2. **A cloned-span probe over-reported** (63px vs 51px) — `getComputedStyle().font` returned empty, so it fell back to a default font.
3. **The reported numbers use the real elements** with `width`/`overflow` temporarily relaxed and restored, so the true inherited font applies.

## Not verified

- Mobile width, quote edges, mention-rail removal, pulse/focus tokens — I only measured pod-row geometry and the filter rail.
- Font is macOS **SF Pro Text**; Linux CI and other platforms will differ. The margin is ~7px so the direction should hold, but the exact pixels won't, and a longer localized `"Yesterday"` would be worse.
- My per-tab clip probe on the filter rail **did not discriminate** — `width: auto` is inert on grid items, so those `CLIPS: false` results carry no information. The Community conclusion rests on the resolved track (66.05px < 92px cap) and the screenshot, not on that probe.

Issue #915, opened 2026-08-13
Your Team card shows 'Never connected' for native-runtime agents (Guide replied 2 minutes earlier)


lilyshen0722:
On `/v2/agents` (Your Team), the Guide's card reads **'Never connected'** — captured 2026-08-12 two minutes after the Guide answered a message via wake→claim→native→deepseek. The Pod Tools sidebar simultaneously showed the correct 'Active 1m ago'.

Cause: the card's status derives from BYO wrapper check-ins (runtime-token lastUsedAt). Native-runtime agents never use a wrapper token, so the card lies for exactly the agent every new user meets first.

This is #891 (agent status honesty) surface 2: the Your Team card should derive from the same `agentStateService` state as the composer/typeahead — for native agents, last `AgentRun` activity is the truthful signal. A signup-day user seeing their Guide as 'Never connected' right after it talked to them reads it as broken.

PR #935, opened 2026-08-13
feat(guide): Scout — the named first teammate that sets up your local agents

lilyshen0722:
### 1. The Guide is now **Scout**
The Cindy-in-Raft decision. `displayName` flips Guide → Scout everywhere human-facing; `agentName` stays `'guide'` — it's the stable identity key (memory envelopes, AGENT_TYPES, manifests, starter tasks). Handles: **@scout** resolves via the mention map's existing `displaySlug` index, **@guide** keeps working via the bare-agentName rule; the frontend handle preference for opaque-token agents becomes *displayName slug → agentName* (moltbot handles like @aria untouched — pinned both directions in tests). The signup opener now introduces Scout by name. `rename-guide-to-scout.ts` migrates the two DB name sources for existing rows (only rows still named exactly "Guide" — user renames survive).

### 2. Scout sets up local agents — `connect_local_agent` actionType
Scout proposes; the owner approves; the executor creates the seat **owned by the user** (ephemeral registry row + installation, ADR-006 self-serve shape) and the approved card deep-links to the connect page with `?pod=&name=` prefilled.

**The D1 line, drawn deliberately:** the executor creates the *seat only*. The runtime token is never minted on this path — the connect page force-rotates on issue, in the user's own browser session, exactly as today. No payload field ever carries a credential (chat history is member-readable and dual-store mirrored). Pinned by test: `JSON.stringify(executionResult)` must not match `/token/i`.

Guards: #609 foreign-name check at **propose time** (Scout gets an immediate error, picks a new name) *and* re-checked at **execute time** (state can change between propose and approve). Existing own install = success (`alreadyInstalled`), matching the BYO page's identity-continuity behavior.

### 3. `commonly_agent_status` + `agentStateService` (Raft P4)
One derivation of proof-of-life — heartbeats ∪ AgentRuns ∪ token lastUsedAt — **extracted from the pod-agents route**, which now consumes it (no copy left behind). One bucket vocabulary: `active / idle / stale / never-connected / ready` — `ready` is the honest word for native agents (the #915 lesson: they have no connection to make). Scout answers "is my agent connected?" from the same truth the roster shows, and its prompt maps each bucket to the action that fixes it.

## Proof
- 15 approval-service cases (3 new: owner-owned seat + no-credential invariant, identity-continuity success, #609 re-check failure face)
- agentStateService bucket suite; toolGating extended (status tool declaration-gated; propose schema teaches both actionTypes — schema-lag would mute the feature silently)
- **The #915 regression suite caught a real bug in this PR** (named-export `buildAgentUsername` dropped the token source during extraction) — fixed before ship
- Delivery-hint suite: @scout for persona-named opaque agents, @aria unchanged, raw tokens never surface
- Both typechecks clean; en+zh strings shipped together

Post-merge: deploy → `rename-guide-to-scout.ts` + `refresh-native-agent-configs.ts` on dev → live E2E (ask Scout to set up a seat, approve, land on prefilled connect page).


lilyshen0722:
Second commit (`ed84b07b`) extends this to the **full DB rename** per Sam's call: `agentName` is now `'scout'` everywhere, not just the displayName. The persona/package split was a permanent two-name tax and pre-GTM is the only cheap moment. `migrate-guide-to-scout.ts` moves users (+PG resync), installations, **memory envelopes** (identity continuity), registry (delete-not-rename vs the boot seeder's fresh row), flagged approvals, runs, and pending events; `MENTION_ALIASES scout:['guide']` keeps the real user's @guide muscle memory landing; chat history stays frozen. Run order after merge: deploy → migrate → config refresh.

PR #938, opened 2026-08-13
docs(adr): ADR-021 — pi turn engine spec + staged OpenClaw retirement (Proposed)

lilyshen0722:
The wiring spec Sam asked for, and the retirement he called — one ratifiable artifact, because the second is gated on the first being live (house rule: never deprecate until the replacement is live).

**Part A** — D1 spec grounded in the real SDK (types read from the pinned 0.84.1 dist, not spike memory): `noTools: 'all'` so a pi session has zero builtins, commonly tools mapped via `defineTool` onto the one existing dispatcher, session-per-run + `dispose()` + ephemeral agentDir (first concrete D5 step), caps → same `errorKind` vocabulary so AgentRun metrics compare across engines, per-manifest `engine: 'pi'` + `PI_ENGINE_ENABLED` kill-switch, merge-dark rollout.

**Part B** — retirement in 5 gated phases: pi soak → install freeze → persona disposition (25 live gateway state dirs, grouped with per-row recommendations) → infra excision (gateway deployment down, `_external/clawdbot` submodule REMOVED — the five-bump lineage saga ends — provisioner/presets/tool-contract-CI deleted) → docs sweep. Identity continuity absolute: no User rows, memory, or history deleted, ever.

**Open decisions flagged for Sam:** community-pod re-seating, Phase-3 timing vs GTM, and whether the Scout flag flip waits for a quiet window.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XeUH4HVDsDHYPsHJthXjB8

Issue #898, opened 2026-08-12
pod.join has no producer — pod-welcomer's declared trigger has never fired once ("Unreachable" install class)


lilyshen0722:
From the #891 review thread (Sharpen msgs 52990/52991, both claims verified on two instruments):

- `pod-welcomer` declares `triggers: ['pod.join']` (`backend/config/native-agents/pod-welcomer.ts:60` — the load-bearing copy; a duplicate manifest also lives at `packages/commonly-apps/src/pod-welcomer/index.ts:52`).
- **Nothing in production emits `pod.join`.** `podController.joinPod` enqueues no agent event; the only dynamic-type enqueue is a dev route (`routes/dev.ts:106`). The enum exists (`types.ts:18`, `AgentRun.ts:27,111`) — the producer doesn't.
- Confirmed against run history: **0 of pod-welcomer's 1,727 recorded runs were pod.join-triggered** — all were heartbeats it never declared (pre-#833 default-on dispatch). Post-#833 (and post-#895, which deliberately does not re-enable undeclared heartbeats) the app is dead-on-arrival: installed, active, error-free, and structurally unable to ever run.

Two deliverables:

1. **The producer.** Emit `pod.join` from the join path (controller or a pod-membership hook) to installs that declare it — the welcome-someone-who-just-joined behavior is the app's entire design, and it has never actually existed.
2. **The "Unreachable" state** (proposed in 52990 for the #891 status-honesty table): a component that declares a trigger no producer emits, derivable **statically at install time** from `triggers[]` against the set of emitted event types. No telemetry, no thresholds — the cheapest row in the table and the only one that catches a dead-on-arrival install. Worth wiring as an install-time warning even before the honesty surfaces ship.

Also worth folding in: the duplicate manifest (two copies of `triggers[]` for the same app) is its own drift defect — one should derive from the other.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XeUH4HVDsDHYPsHJthXjB8

lilyshen0722:
**Triage trap for whoever picks up deliverable 1** — the nearest string match to `pod.join` is a decoy, and it's wrong in two independent ways.

Grepping `pod.join` lands on:

```
backend/gateway/index.ts:38    POD_JOINED: 'pod.joined',
```

That reads like "the producer exists, it's just misnamed — one-line rename." It isn't:

[...]

PR #939, opened 2026-08-13
docs(plans): agent-platform consolidation — the connective plan (for fleet review)

lilyshen0722:
One map connecting ADR-020 (shipped), ADR-021/#938 (proposed + ratified build spec), the approval-producer CAP gap (W1 — found today: only native agents can propose; BYO/MCP/future-pi agents have no path), the OpenClaw retirement phases, and the onboarding-funnel follow-ups. Ordering constraint worth review: **W1 is upstream of agent-runtime M1** — the runtime's tool wire is the MCP client, so the propose tool must exist in @commonlyai/mcp first or cloud agents are born unable to propose. Decision ledger separates ratified-do-not-relitigate from Sam's five open calls. Peer review requested from the Sharpen seats (asks in the doc's final section).

### DIFF
--- a/docs/plans/2026-08-13-agent-platform-consolidation.md
+++ b/docs/plans/2026-08-13-agent-platform-consolidation.md
@@ -48,16 +48,35 @@ Producer surface today is an in-process native tool only. Order within W1:
    (actionType + key params — already on the wire in `buildCardPayload`,
    currently unrendered) as the PRIMARY consent line, proposer prose demoted
    to pitch. "Inert without the owner" covers execution, not mislabeling.
+   **Round-2 upgrade (sprint-review, verified at source): the mislabeling gap
+   is LIVE today, not BYO-only** — summary is caller-supplied (empty-reject +
+   500-truncate is the entire validation) and renders as the headline, while
+   Scout runs on injectable workspace input. Frontend-only fix (pod-architect:
+   params/actionType already on the wire) — it does not block the route and
+   may land ahead of it.
+2b. **Decider lifecycle (round 2, ux-lead + pod-architect):** membership churn
+   recreates undecidability regardless of derivation — nothing cascades pod
+   membership onto ApprovalAction, so an owner leaving strands cards. Three
+   agreed semantics: (i) an explicit **empty-decider face** (attention tone,
+   names the fix) — needed regardless; (ii) per-viewer decidability computes
+   in the **GET pending route** (per-caller), NEVER broadcast in CardPayload
+   (the reaction-`mine` lesson, already inscribed in that file); (iii)
+   decide-time re-derivation **widens** as well as repairs (a later-joining
+   human becomes decider) — allow and **stamp the audit row** (the
+   `decidedAfterExpiry` precedent), don't refuse.
 3. `commonly_propose_action` into `@commonlyai/mcp` → **0.4.0** — publish is a
    **Sam-only web-OTP step on the W2 critical path** (moved to the Open ledger).
 4. Later, same pattern for `commonly_agent_status`.
 
 **Why first:** ADR-021's tool wire is the MCP client — built before W1, cloud
 agents are born unable to propose. W1 is upstream of W2-M1.
-**Acceptance:** a BYO wrapper proposes; the owner approves; **the card visibly
-attributes origin and renders the server's parse**; pinned by a
-`reactionController.test.js`-shaped service-tier suite (dual-auth + install
-gate + decider derivation incl. bot-created-pod refusal).
+**Acceptance:** a BYO wrapper proposes **through the PUBLISHED MCP package**
+(sprint-review: an HTTP-only E2E passes while the npm tool is unpublished —
+the exact artifact-vs-source class the smoke rule exists for); the owner
+approves; **the card visibly attributes origin and renders the server's
+parse**; pinned by a `reactionController.test.js`-shaped service-tier suite
+(dual-auth + install gate + decider derivation incl. bot-created-pod refusal).
+Remedy verified on all four installedBy write paths (sprint-review).
 
 ### W2 — agent-runtime (ADR-021 Part A.1) · GATED ON #938 MERGE
 Milestones amended after review: **M1 = skeleton + smoke Scout + the CLAIMS
--- a/docs/plans/2026-08-13-agent-platform-consolidation.md
+++ b/docs/plans/2026-08-13-agent-platform-consolidation.md
@@ -28,47 +28,97 @@ backfill + npm).
 ## Workstreams
 
 ### W1 — Approvals become CAP (the universality gap) · SMALL, FIRST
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
Reviewed my four asks. **Three sequencing holes (all verified at source), a test-tier gap per workstream, one undefined failure semantic, and four ledger corrections.** One thing I expected to be a blocker turned out not to be, and I say so below rather than leaving the scary version standing.

## 1. Sequencing holes

**(a) W2-M1 is gated on a human-only npm publish that isn't in the Open column.**

W1 step 2 → `@commonlyai/mcp` **0.4.0**, "publish needs Sam's web-OTP." W2 "depends on W1 for a complete toolset at M1." Measured:

```
npm view @commonlyai/mcp version   → 0.3.0
npm view @commonlyai/mcp@0.4.0     → does not exist
```

So a Sam-only action sits on the W2 critical path, and the ledger's Open column — the doc's own instrument for exactly this — doesn't list it.

This isn't hypothetical on this surface: the same publish gate produced a **28-hour Cody `Init:CrashLoopBackOff`** when a chart pinned an mcp version that was bumped but never published. Version-bumped ≠ published ≠ consumed.

Sharper: **W1 step 1 (the HTTP route) can ship without the publish, and the stated acceptance won't notice.** "A BYO wrapper proposes a card that the owner approves, E2E" passes over the route alone while `commonly_propose_action` is still unpublished — so W1 can read done with cloud agents still unable to propose via MCP, which is the exact state W1 exists to prevent. The acceptance criterion needs a second clause naming the published version.

**(b) W3 Phase 3 turns `main` red.**

```
tests.yml:73                       run: npm run verify:moltbot-tools
scripts/verify-moltbot-tool-contract.js   reads _external/clawdbot
                                          process.exit(2) on "submodule not initialised"
```

Phase 3 is "gateway down, submodule removed." Exit 2 is non-zero, so that CI step fails the moment the submodule goes. Nothing in the plan pairs retiring the guard with the excision — and the guard is load-bearing precisely *because* of the submodule, so it should die in the same PR, not before or after.

**(c) W5's trigger has no producer, so the park is permanent.**

"Parked until trigger"; the trigger is a workspace gaining a second human. Nothing watches human-membership transitions (swept `backend/**/*.ts` for `isBot: false` / human-count predicates — 9 hits, all a type default and preset prompt text; control grep alive). Same class as `pod.join` in #898: a declared state whose producer doesn't exist. The condition will be met and nobody will learn. Cheapest fix is to derive it where D6's first half already runs.

## 2. Test tiers per workstream

Only W1 carries an acceptance criterion. W2/W3/W5 name none; W4 names one (the `host: 'byo'` stamp).

- **W1** — the doc cites the reactions precedent for the dual-surface *rule* but not for its *test shape*, which is the reusable half: `reactionController.test.js` is the service-tier test that pins dual-auth + the active-installation gate. The propose route needs that same tier, or the "safe by construction" claim is unpinned.
- **W2** — M1/M2 have no tier. The missing one is service-tier on the queue-transport swap, which is exactly where a partial flip shows up. "Soak" at M3 is dev-env and doesn't substitute.
- **W3** — the "14-day quiet gate" is an observation window, not an assertion. Nothing *tests* the gateway is unused before teardown. A cluster-tier check (zero events dispatched to moltbot runtimes across the window) makes the gate falsifiable instead of a waiting period.
- **W4** — the opener-copy experiment touches v2. jsdom has no layout engine; a real-browser check is required per CLAUDE.md, and this week #928's 44px time slot clipped two of four timestamp formats while 28/28 presence guards stayed green.

## 3. W2-M1 failure semantics vs the manual fallback

**"Manual fallback" is in the Decided column, but only the name is decided** — no trigger, no mechanism, no owner, no granularity anywhere in the doc.

What I verified as *fine*: the double-dispatch worry doesn't apply. `agentEventService` creates native events pre-claimed (`status: 'delivered'`, `attempts: 1`), so an external poller can't also take them. That's the scary version and it's handled.

What's actually open: `runtimeType` is **per-installation**, and Scout is stated at 96 active installs. The plan doesn't say whether M1's flip is per-install or fleet-wide, what a half-flipped fleet means, or how 96 rows get reverted under pressure. "Manual fallback" as a safety net needs a named procedure, because the moment it's needed is the moment nobody wants to discover it's undefined.

**A trap on the migration path — and I checked, it does NOT hit Scout.**

```
server.ts:277                  seedNativeAgents()   ← runs at boot
seed-native-agents.ts:~305     $set config.runtime.runtimeType = 'native'
                               comment: "so every deploy refreshes it"
```

I expected this to be a blocking W2 finding. It isn't: **`apps.ts` seeds only `podSummarizerApp`, `podWelcomerApp`, `taskClerkApp` — Scout is absent**, and `agentBootstrapService` writes no `runtimeType` (file exists; verified negative). So Scout's cloud flip will not be reverted at boot.

It still belongs in the plan as a named trap, because those three are the reference implementations most likely to follow Scout onto the runtime — and if they do, every restart silently reasserts `native`, with no test failing and no symptom except the new runtime going quiet.

## 4. Ledger — what's actually still open

**Move Decided → Open:**
- **The 0.4.0 publish (Sam's OTP)** — blocking, human-only, on the W2 critical path, currently listed nowhere.
- **"manual fallback"** — decided in name only; either define trigger/mechanism/owner or move it to Open.

**Add to Open (currently in neither column, both unowned):**
- W3 Phase 3 ↔ retiring `verify:moltbot-tools` in the same change.
- W5's trigger detector.

The rest of the Decided column reads genuinely settled to me, and "smoke the shipped artifact, never repo source" is the one I'd most want kept — it's what makes (a) above findable at all.

## Not verified / out of my lane

- **ADR-021 and #938 themselves** — I reviewed the plan's sequencing, not the transport design. The AgentEvent-queue coupling is @pod-architect's ask.
- **The "96 active installs" and "25 dormant state dirs" figures** — no DB access here; taken as given.
- **W4 priorities, opener conversion, card-face trust cues** — @ux-lead's.
- I cleared Scout against the boot seeder specifically; I did not audit Scout's full install path for other boot-time writers beyond confirming `agentBootstrapService` writes no `runtimeType`.

lilyshen0722:
**All three fleet reviews folded** (commit `c70dbb32`; ADR corrections on #938 `e2443cd5`). Every load-bearing claim was verified before folding — the 63/261 bot-created-pod measurement and the 30-min GC `deleteMany` both reproduced exactly. Headline changes: W1's first commit is the decider-derivation fix (blocking, pre-solves W5's multi-human case); the card's consent face gains the ADR-001 origin cue + server-parsed action line **before** `commonly_propose_action` ships; claims protocol moved into W2-M1; hosted events get a visible-failure TTL; manual fallback is defined, not named; W4 runs instrument-first with the stalled-connect spec claimed by ux-lead; the moltbot CI guard dies in the same PR as the submodule. Ledger updated — the mcp 0.4.0 web-OTP publish is now correctly on the critical path as a Sam-only step.

PR #944, opened 2026-08-14
feat(approvals): POST /pods/:podId/propose-action — the CAP producer surface (W1 step 1)

lilyshen0722:
**W1 step 1** of `docs/plans/2026-08-13-agent-platform-consolidation.md`.

## What it closes

Approval-card **consumers** were already universal — any human decides via `/api/approvals`. **Producers** were native-only, because `proposeAction` had exactly one caller: the in-process native runtime. That's the CAP universality gap.

A BYO wrapper or a hosted runtime can now ask for consent through the same service, with the same validation, as a first-party agent. This is also upstream of W2: per ADR-021, cloud agents built before this exist unable to propose.

## Conventions followed

`agentRuntimeAuth` + `phase4RateLimit` (propose is more expensive than post — the sibling mutating-route convention), pod-scope check, and the **active-installation gate** the claim route already applies. Token scope alone is not permission to act in a pod; an uninstalled agent must not keep proposing.

## Where the logic lives, and why

The novel rules are in `proposeActionForRuntime` **in the service**, not in the handler. The route is a five-line adapter.

That placement is the interesting part. My first draft of the test re-declared the handler inside the test file — it passed while proving nothing about the route. That's the copy-drifts-from-its-source failure this repo has already paid for more than once, and on a consent surface it's exactly the wrong place to accept it. The test now imports the function that actually runs.

## Validation is deliberately not duplicated

Unknown `actionType`, param rules, summary presence, and the decider derivation with its refusal all stay in `proposeAction`. Two validators on a consent surface is precisely how the HTTP and in-process producers drift apart.

Refusals map to **400 carrying the service's own sentence** — an agent told *"no accountable human owns this pod, so an approval card here would be undecidable"* can act on that; a bare 400 teaches it nothing.

## Tests

Six, covering the route's own contract rather than re-testing the service:

- authorized agent with an active install proposes
- **principal comes from the token, never the body** (no proposing as someone else by naming them)
- token scoped to another pod → 403, before any DB work
- inactive installation → 403 even with a valid pod-scoped token
- the decider refusal reaches the caller intact
- garbage body still reaches the service — the route does not pre-judge it

33 pass across this suite plus the two approval service suites and the intro suite.

## Next in W1

`commonly_propose_action` into `@commonlyai/mcp` → **0.4.0**, whose publish is a Sam-only interactive web-OTP step, then acceptance E2E through the **published** package (an HTTP-only E2E passes while the npm tool is unpublished — the artifact-vs-source class).

### DIFF
--- a/backend/__tests__/unit/routes/agentsRuntime.propose-action.test.js
+++ b/backend/__tests__/unit/routes/agentsRuntime.propose-action.test.js
@@ -1,0 +1,145 @@
+/**
+ * W1 step 1 — the CAP producer surface for approval cards.
+ *
+ * Consumers were already universal (any human decides via /api/approvals);
+ * producers were native-only. `POST /pods/:podId/propose-action` closes that
+ * gap, and its novel rules live in `proposeActionForRuntime` so this test
+ * exercises the SAME function the route calls.
+ *
+ * That placement is the point. The first draft of this file re-declared the
+ * handler inside the test — it passed while proving nothing about the route,
+ * which is the copy-drifts-from-its-source failure this repo has already paid
+ * for more than once. Importing the real export is the fix.
+ *
+ * Validation is deliberately NOT re-tested here: it belongs to `proposeAction`
+ * and is covered by approvalActionService.decider.test.js. Two validators on a
+ * consent surface is how the HTTP and in-process producers drift apart.
+ */
+
+const mockInstallFindOne = jest.fn();
+jest.mock('../../../models/AgentRegistry', () => ({
+  AgentInstallation: { findOne: (...args) => mockInstallFindOne(...args) },
+  AgentRegistry: { getByName: jest.fn(), create: jest.fn() },
+}));
+
+const mockApprovalCreate = jest.fn();
+jest.mock('../../../models/ApprovalAction', () => ({
+  create: (...args) => mockApprovalCreate(...args),
+  findById: jest.fn(),
+  findOneAndUpdate: jest.fn(),
+}));
+
+const mockPodFindById = jest.fn();
+jest.mock('../../../models/Pod', () => ({
+  findById: (...args) => mockPodFindById(...args),
+  create: jest.fn(),
+  updateOne: jest.fn(),
+}));
+
+const mockUserFindById = jest.fn();
+jest.mock('../../../models/User', () => ({ findById: (...args) => mockUserFindById(...args) }));
+
+const mockPostMessage = jest.fn();
+jest.mock('../../../services/agentMessageService', () => ({
+  postMessage: (...args) => mockPostMessage(...args),
+}));
+
+jest.mock('../../../services/agentIdentityService', () => ({
+  getOrCreateAgentUser: jest.fn(),
+  syncUserToPostgreSQL: jest.fn(),
+}));
+
+jest.mock('../../../config/socket', () => ({ getIO: jest.fn(() => null) }));
+
+const { proposeActionForRuntime } = require('../../../services/approvalActionService');
+
+const POD = '6a692a1be833c668acdb84cf';
+const HUMAN = 'aaaaaaaaaaaaaaaaaaaaaaa1';
+
+const installation = (over = {}) => ({
+  agentName: 'scout', instanceId: 'default', displayName: 'Scout', config: null, ...over,
+});
+
+const call = (over = {}) => proposeActionForRuntime({
+  podId: POD,
+  installation: installation(),
+  podAuthorized: true,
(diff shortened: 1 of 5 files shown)


lilyshen0722 (commented):
Adversarial review, both asks answered plus one finding you didn't ask about.

## Ask 1 — the split is real, and validation is not duplicated

Confirmed. The route is a genuine adapter: extract `podId`, resolve the installation, call `proposeActionForRuntime`, spread the verdict. **Zero validation in the handler.** The service holds the novel HTTP-surface rules (active-install gate, principal-from-token, refusal mapping) and delegates every domain rule — actionType, params, summary, decider derivation — to the shared `proposeAction`. That's the right seam and it does what the comment claims.

## But the seam has one hole, and it is the same shape as the bug you were fixing

`podAuthorized` is a **boolean computed in the route** by `ensurePodMatch(...)` and handed to the service:

```js
// route
podAuthorized: ensurePodMatch(req.agentInstallations || installation, podId, req.agentAuthorizedPodIds)

// service
if (!podAuthorized) return { status: 403, ... }
```

So the one rule that decides *may this token act on this pod at all* lives on the route side. And the test:

```js
const call = (over = {}) => proposeActionForRuntime({ ..., podAuthorized: true, ...over });
```

It calls the service directly with the boolean supplied. `:110` covers `podAuthorized: false`. Together those prove **the service honours the boolean** — never that the route computes it correctly. A route that returns `true` when it should return `false` passes this suite untouched.

That's your original failure one layer over: you moved logic into the service so "the thing under test is the thing that runs," and the authorization decision is the one piece that stayed on the other side.

It's worse than local, and this is the part I'd act on:

```
ensurePodMatch — references in backend/**/__tests__ : 0
callers: agentsRuntime.ts:1298, :1365, and now propose-action
```

**Zero test coverage repo-wide.** It's not that this PR failed to test it; nothing ever has, and this PR just made it load-bearing for a consent surface.

Cheap fix that keeps your principle intact: pass `installations` + `authorizedPodIds` into `proposeActionForRuntime` and let the *service* call `ensurePodMatch` (export it, or move it). Then the derivation is inside the tested unit and `podAuthorized` stops being caller-asserted.

## Finding you didn't ask for — the active-install query omits `instanceId`

```js
// :205, this PR
const active = await AgentInstallation.findOne({ agentName: installation.agentName, podId, status: 'active' });

// :130, same file, existing
const foreign = await AgentInstallation.findOne({ agentName, instanceId: 'default', installedBy: { $ne: ownerUserId } });
```

The sibling pins `instanceId`; this one doesn't. So the gate asks "is **some** install of this agentName active in this pod," not "is **this caller's** install active."

Your test at `:118` — *"an inactive installation is refused even with a pod-scoped token"* — passes because **no** active install exists. It would not catch the case it reads like it covers: caller's instance inactive, a *different* instance of the same agentName active in that pod. The query finds the sibling and returns 200. With per-user Scout seats keyed `u<sha256(userId)[0:10]>`, several instances of one agentName sharing a pod is a real shape, not a hypothetical.

**Reachability I did not establish, stated as a gap rather than an exploit:** I didn't trace how `req.agentAuthorizedPodIds` is populated. If it derives from *active* installs only, an inactive caller never holds a pod-scoped token and this is unreachable. That's the one join that decides severity, and it's the natural thing to check before changing anything.

## Ask 2 — "8 existing routes" is no longer true; it's 9

Measured on this PR's head:

```
agentRuntimeAuth, phase4RateLimit   (flagged order)  → 9
phase4RateLimit, agentRuntimeAuth   (correct order)  → 3   ← includes this new route
```

```
:316  POST   /messages/:messageId/claim
:362  DELETE /messages/:messageId/claim
:908  POST   /agent-dm
:1749 POST   /pods/:podId/typing
:1804 POST   /pods/:podId/messages
:2320 PUT    /memory
:2450 POST   /memory/sync
:3505 POST   /pods/:podId/ask
:3566 POST   /asks/:requestId/respond
```

Your new route is correctly on the safe side, and starting there rather than joining the queue is the right call.

**Is one diff the right shape?** For the move itself, yes — it's mechanical, single-file, identical transformation, and CodeQL verifies it. But the count in that comment is what just drifted from 8 to 9, which is the actual lesson: **a number in a prose note is not a guard.** The durable form is a test asserting zero routes match the flagged order. Then the count can't age and a new route can't quietly join the queue — same pattern as `v2-layout-invariants`, which this repo already uses for exactly this "presence test until a better tier exists" job.

I'd do both in one PR: the 9-route reorder plus the guard that keeps it at zero.

## Not verified

- I did not run the suite; findings are from source.
- `agentAuthorizedPodIds` population (above) — the join that decides the instanceId gap's severity.
- Nothing about #943, #940, #941 — other lanes.

PR #954, opened 2026-08-15
feat(onboarding): alert when a newcomer types and nothing answers (W4 item 2)

lilyshen0722:
Closes W4 item 2. The instrument, not another fix.

## Why

Every onboarding defect found on 2026-08-14 was found by a human deciding to read production transcripts. The fixes were real; the detection was a person choosing to look. `pod-summarizer` failed every six hours for a month and nobody noticed. This is the part that does not require anyone to choose.

A cron asks one question every five minutes: **did someone who signed up recently say something in a room that could have answered, and did nothing answer?**

An **episode** is one `(user, pod)` run of silence, not one message — someone typing four times into a dead room is one failure, not four.

## Calibration

Threshold and semantics were calibrated against production by @pod-architect rather than guessed:

| decision | value | why |
|---|---|---|
| threshold | **15 min, wall-clock** | `AGENT_EVENT_REQUEUE_DELIVERED_MINUTES` is 10, so under ~12 fires inside a legitimate retry; `AGENT_EVENT_STALE_PENDING_MINUTES` is 30, past which the evidence is deleted along with the answer |
| not idle-gated | — | idle would key on `lastActive`, which over-counts: `V2PodChat.tsx:279` polls every 60s, so an abandoned tab reads as presence. It would also make the rate incomparable across cohorts, killing the alert's second job |
| scope | pod holds **≥1 active `AgentInstallation`** | not membership — a member without an installation 403s on post, so it was never going to answer either. Same predicate the posting path uses |
| reply | **bot authors only** | the claim is *the platform did not answer*; a human answering is recorded as `human-rescued` rather than folded in |
| unit | per-incident + hourly collapse | base rate is ~1/day on a working system, but the failure this watches for is a regression, and regressions strand users in bulk |

The observed data agrees independently: across 21 days, **every genuine reply landed inside 107 seconds** (5s, 15s, 40s, 48s, 73s, 103s, 107s) and the next cluster was 10+ hours. Nothing lands in between.

## The event snapshot is not optional

`AgentEvent` pending rows are deleted at 30 minutes; we fire at 15. That gap is the **only** window in which two failures that look identical afterwards can be told apart:

- `never-enqueued` → a producer bug (the write path skipped `enqueueMentions`)
- `enqueued-never-answered` → a runtime bug

They need opposite fixes. Captured at fire time or the alert reports that something is wrong with no way left to say what.

## Delivery refuses to be silently inert

Detection and delivery are separate services because they fail differently and only one is allowed to be quiet. With no `ONBOARDING_ALERT_EMAIL` set, the service **says so loudly on every pass that had something to report** rather than computing correctly and returning success. Three of the eight defects found on 08-14 were that exact shape; an alerting service is the worst possible place to add a fourth. A machine-readable `[onboarding-silence] ALERT …` line is always emitted regardless of email config.

`GET /api/admin/analytics/silence` is the pull half — an alert you can only receive is one you cannot go back and check, and it makes "did the fix work" answerable without re-reading transcripts by hand.

## `lastLogin` is deliberately not built

W4 item 2 was written as "`lastLogin` + alert". `lastLogin` does not exist — but the instrument it was meant to provide already does. `User.lastActive` is written on every authenticated request at a 15-minute throttle (`middleware/auth.ts:14`) and is **live in production**: DAU 9, WAU 24, returnedD1 5, which would be structural zeros if it were frozen.

With a 7-day JWT, a login timestamp samples an engaged daily user at weekly granularity — misleading rather than merely redundant. Dropping it was confirmed by @ux-lead and @pod-architect. The real caveat runs the other way: `lastActive` *over*-counts, and that is now documented where it is read.

## Verification

- **43 tests**, all green, including the 12 pre-existing analytics tests after `HUMAN_FILTER`/`BOT_FILTER` moved into `services/userClassification.ts` (one definition, not two copies of a subtle rule)
- `tsc:check` clean — baseline-compared against `main` by stash, identical output
- **Ran the real service against production data, read-only** (episode model stubbed in `require.cache`; zero writes): 54 messages scanned, 1 skipped for having no agent, **16 raw hits collapsing to 10 episodes with 6 absorbed** — matching an independently written probe. The collapse rule is doing real work.

## Not in scope

This does not nudge the user — that is W4 item 3 (@ux-lead's stalled-connect spec), which fires on install state and speaks to the user. They are complementary: that one catches a seat whose token was never used before anybody types; this one catches a Scout that breaks tonight, and knows nothing about tokens.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -120,6 +120,25 @@ TELEGRAM_BOT_TOKEN=
 # SENDGRID_FROM_EMAIL=noreply@yourdomain.com
 
 # -----------------------------------------------------------------------------
+# Onboarding-silence alert  (OPTIONAL, but the alert is log-only without it)
+# -----------------------------------------------------------------------------
+# Where "a new user typed and nothing answered" is sent. The detection cron
+# runs regardless and always logs a machine-readable line; set this and it also
+# reaches a person. Leaving it unset is a supported configuration, not a
+# mistake — but then the only reader is `kubectl logs`, and the whole point of
+# this alert is that nobody has to decide to go look.
+# ONBOARDING_ALERT_EMAIL=ops@yourdomain.com
+#
+# Tunables (defaults are calibrated against production — see
+# backend/services/onboardingSilenceService.ts before changing them):
+# ONBOARDING_SILENCE_THRESHOLD_MINUTES=15   # past the 10-min event requeue,
+#                                           # inside the 30-min pending GC
+# ONBOARDING_SILENCE_WINDOW_DAYS=7          # how long after signup counts as onboarding
+# ONBOARDING_SILENCE_LOOKBACK_HOURS=24      # how far back one pass judges
+# ONBOARDING_SILENCE_ROLLUP_THRESHOLD=5     # per-hour count above which alerts collapse
+# ONBOARDING_SILENCE_RESOLUTION_WINDOW_DAYS=7
+
+# -----------------------------------------------------------------------------
 # X / Twitter integration  (OPTIONAL)
 # -----------------------------------------------------------------------------
 # X_API_KEY=
--- a/backend/__tests__/unit/routes/admin.analytics.silence.test.js
+++ b/backend/__tests__/unit/routes/admin.analytics.silence.test.js
@@ -1,0 +1,111 @@
+const request = require('supertest');
+const express = require('express');
+
+// GET /api/admin/analytics/silence — the pull half of the onboarding-silence
+// alert (W4 item 2). The cron pushes; this exists so "did the fix work" is
+// answerable without hand-reading production transcripts, which is how every
+// onboarding defect was found on 2026-08-14.
+
+let mockRole = 'admin';
+jest.mock('../../../middleware/auth', () => (req, res, next) => {
+  req.user = { id: 'admin1' };
+  req.userId = 'admin1';
+  next();
+});
+jest.mock('../../../middleware/adminAuth', () => (req, res, next) => {
+  if (mockRole !== 'admin') return res.status(403).json({ message: 'Admin access required' });
+  return next();
+});
+jest.mock('../../../middleware/ipRateLimit', () => ({
+  cloudflareIpRateLimitKeyGenerator: () => 'test-key',
+}));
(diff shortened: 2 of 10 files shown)


lilyshen0722:
## `diagnose()` is now validated on real production data, both classes

I flagged in review that the dry run could not exercise the thing the event snapshot exists for: every historical episode came back `noneEnqueued: true`, which reads as "producer bug" but was almost certainly just GC. @pod-architect pointed out the fix — the `pgMessageController` route bug is itself a reliable `never-enqueued` producer, and it is about to be patched, so the fixture had to be captured before that happens.

Done, in the `adr-006-smoke` pod, using the real `snapshotAgentEvents`:

| arm | write path | handle | snapshot | diagnosis |
|---|---|---|---|---|
| **control** | socket `sendMessage` | `@commonly-bot` | `total:1, {acked:1}, targets:[commonly-bot/default], noneEnqueued:false` | `acked-but-no-reply` |
| **treatment** | `POST /api/pg/messages/:podId` | `@commonly-bot` | `total:0, {}, noneEnqueued:true` | `never-enqueued` |

Same pod, same handle, two minutes apart. The only variable is the write path.

**The first attempt was invalid and is worth recording.** I ran it with `@commonly-summarizer` — which is the installation's *displayName*; the resolvable handle is the username `@commonly-bot`. Both arms came back empty, and the negative arm looked like a clean confirmation of the route bug. It was not: the control was broken, not the route. Re-running control-first with the resolvable handle is what made the result mean anything.

So the snapshot separates the two classes on live data, which was the open question.

## Also confirmed as a real defect, separately

`POST /api/pg/messages/:podId` writes the row, returns 200, renders in the pod, and never calls `enqueueMentions`. Only four write paths do: `server.ts:711`, `postController:295`, `messageController:280`, `agentsRuntime:1836`.

@pod-architect established the consequence is larger than dropped mentions: `enqueueWakeOnMessage` is called from *inside* `enqueueMentions` (`agentMentionService:1032`, `:1343`), so in My Workspace — where Scout runs `wakeOnMessage: true` and is meant to answer every message — a message written over that route is invisible to Scout entirely. Verified at source.

That is not fixed here and wants its own PR. This one only makes it *detectable*.

PR #958, opened 2026-08-15
docs(adr): ADR-023 — the hosted agent-runtime substrate (Proposed, gated on one spike)

lilyshen0722:
**Proposed, not decided.** The whole ADR turns on one spike that takes a day, and "no change, withdraw this" must stay a live outcome.

## The argument in one paragraph

ADR-021 Part A.1 says plainly why the hosted runtime is a multiplexed Deployment rather than per-agent:

> **NOT a container per agent (does not scale past dozens)**

It *wanted* isolation and settled, because Kubernetes cannot give it at that granularity. A Durable Object is one addressable, single-threaded, stateful object per key — exactly `(agentName, instanceId)` — with its own storage and idle hibernation. The tradeoff the ADR was forced into stops existing.

And it removes work rather than adding capability:

- the shared-PVC blast radius that motivated the multiplexing caveat (the 08-13 clobber incident) doesn't exist when each agent owns its storage
- DOs are single-threaded per object, so *"two engines must never interleave one conversation"* becomes a platform property instead of a protocol to implement and keep correct. Claims stay necessary **across** engines (BYO wrapper vs hosted); they stop being necessary **within** the hosted one — which is the half @pod-architect pulled into M1.

## Why this isn't a detour from activation

W2 is unbuilt, so this is a substrate choice for greenfield work, not a migration.

More importantly: the funnel's biggest drop is the "install a CLI and run a command on your machine" step, and a hosted runtime is what removes it. Measured against production this session — **103 real humans, 28 ever typed, 4 ever had a genuine back-and-forth, 1 ever returned on a second day.** `user-fbd3` went signup → BYO install in 77 seconds, faster than almost anyone in the dataset, hit that step, gone in 72 minutes.

So substrate work and activation work are the same work here, not competitors for it.

## The gate

**Does `@earendil-works/pi-coding-agent` run under `workerd`'s Node compat?**

- yes → Durable Objects, hibernation, per-agent isolation
- no → Containers. Better isolation than the multiplexed Deployment, but idle is no longer free and half the argument weakens — at which point re-examine honestly whether it beats the GKE Deployment ADR-021 already specifies.

Tracked separately as the blocking spike. **Do not amend ADR-021 before it returns.**

## Credits — and what they don't buy

$10k available. That removes a cost objection; it does not buy engineering time, which is the scarce resource. D3 records three uses needing no substrate decision at all:

- **R2 as an `ObjectStore` driver.** Attachments are bytes inside MongoDB today (`OBJECT_STORE_DRIVER` unset → `mongoDriver`, the only one that exists). ADR-002 planned an external store; it never landed. The interface is narrow and documented as open for extension — one new file. Gets bytes off Atlas, the most expensive storage in the stack and the subject of the 07-23 tier incident.
- **Workers AI via LiteLLM** — `cloudflare` is already a configured provider in the live config. ConfigMap change, converts credit into runway on the largest variable cost.
- **Pages for the frontend** — static SPA, separable, reversible.

D4 states the thing I'd most want on the record: **twelve months is a credit term, not a deadline and not a moat.** The moat, if there is one, is portable identity + memory + the social graph. Cloudflare hosts that; it doesn't create it.

## What this explicitly is not

Written into the ADR because it's the failure mode of every "let's move to X", and because the cheap parts get held hostage by the expensive one:

- **not a backend migration** — Express/Mongo/Postgres/Socket.io stay on GKE
- **not a reason to delay the staging environment** — hours of work, independent, and this session wrote to production three times to verify things because there's nowhere else
- **not a re-litigation of ADR-021** — pi stays the turn engine; only the substrate under Part A.1 is in question

## Requested review

@pod-architect — D1 and the claims claim especially. You pulled claims into M1 for the requeue-interleave reason; I'm asserting DO single-threading covers the within-engine half. If that's wrong the ADR loses one of its two structural arguments.

@sprint-review — the risks section, and whether "no change" is genuinely still reachable after an ADR is written. Writing the document is itself a commitment pressure and I'd rather you name that than have it operate quietly.

### DIFF
--- a/docs/adr/ADR-023-agent-runtime-substrate.md
+++ b/docs/adr/ADR-023-agent-runtime-substrate.md
@@ -1,0 +1,69 @@
+# ADR-023: the hosted agent-runtime substrate — Durable Objects vs a multiplexed Deployment
+
+- **Status:** Proposed (drafted 2026-08-15). **Gated on one spike — see D1.** Not a decision yet.
+- **Amends, if accepted:** ADR-021 Part A.1 (the `agent-runtime` build spec). Everything else in ADR-021 — pi as the turn engine, staged OpenClaw retirement, the M1–M4 ladder, the fallback discipline — stands unchanged.
+- **Does NOT touch:** the shell (Express, MongoDB, PostgreSQL, Socket.io, the v2 frontend). See "What this is not" below, which is the most important section in this document.
+
+## Context
+
+Four things converged on 2026-08-15.
+
+**1. W2 is unbuilt, and ADR-021 settled for a design it did not want.** Part A.1 specifies:
+
+> A dedicated `agent-runtime` Deployment … multiplexing **persistent sessions keyed by (agentName, instanceId)** … with state on a PVC. **NOT a container per agent (does not scale past dozens)** and NOT in the backend process (shared blast radius — the 2026-08-13 clobber incident is the standing example).
+
+The ADR wanted per-agent isolation. It rejected it because Kubernetes cannot provide isolation at that granularity past dozens of agents. That is a property of the substrate, not of the design.
+
+**2. The funnel's binding constraint is the step a hosted runtime removes.** Measured this session against production: 103 real humans ever; 28 ever sent a message; 4 ever had a genuine back-and-forth; **1 ever returned on a second day**. The largest single drop is people who arrive, are asked to install a CLI and run a command on their own machine, and leave. `user-fbd3` (2026-08-15) went signup → BYO install in 77 seconds — faster than almost anyone in the dataset — hit that step, and was gone in 72 minutes.
+
+Hosted agents are how that step disappears. So the runtime substrate is not infrastructure housekeeping running parallel to activation work; on current evidence it *is* the activation work.
+
+**3. Cloudflare's primitives map onto the rejected design.** A Durable Object is one addressable, single-threaded, stateful object per key — exactly `(agentName, instanceId)` — with its own storage, that hibernates when idle. Per-agent isolation, scale, and near-zero idle cost are properties of the primitive rather than things to engineer.
+
+Two consequences worth stating separately, because they remove *work* rather than adding capability:
+
+- **Isolation for free.** The shared-PVC blast radius that motivated the multiplexing caveat does not exist when each agent owns its own storage.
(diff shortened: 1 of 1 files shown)


lilyshen0722:
Blocking spike tracked as #959 — that issue decides D1, and this ADR should not be merged as anything other than **Proposed** until it returns.

lilyshen0722:
## D4 rewritten — I had the economics backwards

The first draft said *"twelve months is a credit term, not a deadline."* That's advice for a funded company. Commonly is unfunded — no raise, no YC — so the credits are substantially the budget and the term is a real clock. Withdrawn.

The replacement argues the same conclusion from a stronger place. **The economics that matter are per-user, not infrastructure:**

| | marginal cost | activates? |
|---|---|---|
| BYO | ~$0/user — they bring the compute | **no** — 28 of 103 ever typed, 1 ever returned |
| hosted | inference, per user, on us | **yes** — it removes the step the funnel dies on |

Unfunded, that's backwards by construction: **the better hosted works, the faster it burns.**

Which is the actual case for the substrate, and it isn't "GKE is expensive." DO hibernation plus Workers AI on credits puts the marginal cost of a hosted agent at approximately zero for the term. The experiment that fixes activation is **unaffordable on GKE without funding and affordable on credits.** Cheaper infrastructure is a rounding error beside that.

Two things are now conditions of accepting this ADR rather than notes:

1. **Pricing ships with hosted agents, not after.** Free-on-credits is coherent only if something charges before the term ends.
2. **Credits get spent toward a returning user, not toward completeness.** They reward using Cloudflare, not using it for everything.

## Two risks this exposed

- **The cliff at term end is real and dated.** Zero marginal cost is a property of the credits, not the architecture. Succeed at hosted agents with no pricing and the reward is a bill with no revenue — worse than today, because by then users depend on it.
- **Opportunity cost is the largest risk in this ADR and it isn't technical.** Twelve months on substrate arrives at month twelve with better architecture, the same 4%, and the same wedge YC declined. Every week has to be defensible as the fastest path to a second returning user.

## One open question I can't answer and nobody has written down

**What is the monthly burn, and what dominates it — GKE, Atlas, or inference?** Every economic claim in D4 bends on that number, and it's known to the operator and to no one else. Added as open question 5.

@sprint-review — this is the second version of D4 written in the same session by someone who was argued into the conclusion between drafts. Worth checking whether the new framing is genuinely better reasoning or the same conclusion with better rhetoric.

PR #964, opened 2026-08-16
fix(security): stop lending the server GitHub credential to any agent token

lilyshen0722:
## Summary

Three GitHub routes were reachable by **any `cm_agent_*` token** — which every agent, installed by any user on the instance, holds — and each of them spent our shared server-side `GITHUB_PAT`:

| route | what a caller could do |
|---|---|
| `POST /api/github/token` | receive `process.env.GITHUB_PAT` **in plaintext**, or mint an App installation token for a repo of their choosing |
| `GET /api/github/pulls/:number/diff` | read the diff of any repo that PAT could reach, including private ones |
| `POST /api/github/pulls/:number/review` | post a review — including `APPROVE` — onto any such repo |

`owner`/`repo` were **caller-supplied**; `Team-Commonly/commonly` were only defaults. So the caller chose the target and our credential supplied the authority. The rate limiter on the `/pulls` routes bounded volume, not authority.

## Why nothing happened, and why that matters

The PAT has been 401-dead for weeks. Every call failed upstream, so **the broken credential was doing the access control**. That is also why this stayed invisible: it was filed as an ops annoyance (AX #9 flagged the credential as unverified) rather than an authorisation gap, because a failing call and a forbidden one look identical from outside.

The practical consequence: **rotating the PAT would have re-armed all three at once.** This lands first for that reason.

## What changed

- **Deleted** all three routes and the `commonly_pr_diff` / `commonly_pr_review` MCP tools they backed.
- **Pinned** the issue routes to `Team-Commonly/commonly` server-side. They had the same caller-supplied-target shape.
- **Retired** the `VALID_NAME` regex on those inputs — validating a caller's repo name made it well-formed, never authorised, and there is no caller-supplied value left to interpolate.
- **Redirected the agent-facing guidance** (heartbeat preset + bundled `github` skill) to `gh` instead of recommending the removed tools.

## No capability is lost

Every runtime that used these has a shell and its own GitHub identity:

- local wrappers use the operator's `gh` keyring auth — **PR #963 was opened that way**, not through this proxy
- cluster runtimes get `gh` plus a git credential helper at image build

`gh` is also strictly more capable here: it supports line-level review comments, which `commonly_pr_review` never did (it took only `{event, body}`).

The tools existed for ergonomics — `docs/audits/ui-smoke-2026-05-23` records agents having to "know the gh CLI is available". That is a real complaint, and it justified building a credential proxy without anyone re-asking whose authority it spent.

## Deliberately left open

A shell-**less** runtime — native Tier-1, or the hosted Code Reviewer persona in ADR-022's v1 cast — still has no GitHub path. That case needs per-user GitHub App auth, because reviewing a user's repository under *our* identity is wrong even when it works. A shared PAT cannot express it, so no amount of gating on these routes would have delivered it.

## Verification

`backend/__tests__/unit/routes/github.upstreamErrorRoutes.test.js` — 10 passing:

- each removed route asserted **unroutable against the mounted app**, with `isPatConfigured: true` (the strongest configuration for the defect — precisely when `/token` used to answer with the secret)
- separately asserted that **no GitHub service call is reached** for any of them, so a 404 arriving for an unrelated reason cannot satisfy the test
- issue pinning asserted against a caller-supplied `{owner: 'attacker', repo: 'private-repo'}`
- the existing source-level route-contract check still passes over the remaining 5 routes

`commonly-mcp` — 46 passing. Tool count 29 → 27, plus a by-name assertion that the two tools are absent, because a future tool would restore the count and silently re-satisfy a bare length check.

`npx tsc --noEmit` clean.

Note: `backend/__tests__/unit/routes/github.upstreamErrors.test.js` fails to run on this branch — it also fails identically on `main` (verified by restoring the pristine file and re-running; the stack differs only by the line number of an import). Pre-existing, untouched here.

## Follow-ups not in this PR

1. **Do not rotate the PAT until this merges.** Rotation is a privilege-escalation event while these routes exist.
2. The issue routes remain reachable by any agent token — now only against our own repo, but `POST /issues/:number/close` is still a write any agent can perform. Worth an explicit authorisation decision.
3. AX audit entry 25 records the rules earned, including "a dead credential is not a closed door, and it hides one."

### DIFF
--- a/backend/__tests__/unit/routes/github.upstreamErrorRoutes.test.js
+++ b/backend/__tests__/unit/routes/github.upstreamErrorRoutes.test.js
@@ -77,12 +77,6 @@ const githubRouteHandlers = (source) => {
 // valid request that reaches its service boundary.
 const PROXYING_ROUTE_CASES = [
   {
-    name: 'POST /token',
-    service: 'getInstallationToken',
-    send: (client) => client.post('/api/github/token'),
-    assertResponse: (res) => expect(res.body.message).toBe(res.body.error),
-  },
-  {
     name: 'GET /issues',
     service: 'listOpenIssues',
     send: (client) => client.get('/api/github/issues'),
@@ -102,14 +96,22 @@ const PROXYING_ROUTE_CASES = [
     service: 'closeIssue',
     send: (client) => client.post('/api/github/issues/1/close'),
   },
-  {
-    name: 'GET /pulls/:number/diff',
-    service: 'getPullDiff',
-    send: (client) => client.get('/api/github/pulls/1/diff'),
-  },
+];
+
+// These three routes were removed because they lent our server-held GitHub
+// credential to any caller holding an agent token. `/token` returned it
+// outright; the two `/pulls` routes spent it on a caller-chosen `owner`/`repo`.
+//
+// The test is written against the mounted app rather than the source text on
+// purpose: a source-level assertion ("the file no longer contains
+// `/pulls/:number/review`") passes just as happily if the route moves to
+// another router and stays reachable. What matters is that nothing answers
+// these paths.
+const REMOVED_CREDENTIAL_ROUTES = [
+  { name: 'POST /token', send: (client) => client.post('/api/github/token') },
+  { name: 'GET /pulls/:number/diff', send: (client) => client.get('/api/github/pulls/1/diff') },
   {
     name: 'POST /pulls/:number/review',
-    service: 'createPullReview',
     send: (client) => client.post('/api/github/pulls/1/review').send({ event: 'APPROVE' }),
   },
 ];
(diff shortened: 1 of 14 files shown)


lilyshen0722:
Added a fourth instance found while sweeping for siblings — worth calling out because it isn't a GitHub route and wouldn't be found by grepping for one.

**The task board.** `githubIssueNumber` is caller-supplied on task creation and validated only as a positive integer (`tasksApi.ts:155`), so it can name any issue in the repo. Two writes trusted it, both under the shared PAT and both reachable by any agent token:

- completing a task auto-closed that issue, with the caller-supplied `prUrl` interpolated into the closing comment
- creating a sub-task commented on the parent's issue, with `title` as the body

So an agent could create a task naming someone else's issue, complete it, and have the server close it as us with attacker-chosen text.

Gated on a new `githubIssueOwned` flag, set only where this server itself opened the issue. The number remains display metadata; it is no longer authority to write. Legacy tasks predate the flag and no longer auto-close — a write we can't prove we own should fail closed.

Note the search that found it: grepping route files for `process.env.*TOKEN` came back clean. The credential use was **service-mediated** (`tasksApi` → `GitHubAppService` → `_apiHeaders`), which is the shape worth remembering — the direct-env grep is the one that gives false comfort.

PR #970, opened 2026-08-18
feat(reactions): notify agents of new acknowledgements

lilyshen0722:
## Summary
- enqueue the reacted-to bot author a prompt-compatible acknowledgement when a reaction is newly inserted
- keep acknowledgements unclaimable and prevent self-reaction and duplicate-retry wakes
- preserve socket fan-out and cover the inbound agent path

## Verification
- `cd backend && node_modules/.bin/jest __tests__/unit/controllers/reactionController.test.js __tests__/unit/controllers/reactionController.emojiValidation.test.js --runInBand` (19 passed)
- `git diff --check`


lilyshen0722 (commented):
Reviewed at `8fd5403a` against the full diff (4 files, +238/−26). Good shape overall — the `added` boolean threaded from `ON CONFLICT DO NOTHING`'s `rowCount` is the right way to make the receipt fire once, the self-reaction guard is there, and omitting `payload.messageId` to dodge ADR-018's claim gate is a real insight with a test pinning it.

One defect, in the fallback branch the comment exists to justify.

### The `username` fallback enqueues events that can never be delivered

`reactionController.ts`, `enqueueReactionAcknowledgement`:

```js
const agentName = String(author.botMetadata?.agentName || author.username || '').trim().toLowerCase();
```

The consume side matches exactly. `agentEventService.ts:1091-1094`:

```js
const query = { agentName: safeAgentName, instanceId: safeInstanceId, status: 'pending' };
```

and the poller's own key comes from `middleware/agentRuntimeAuth.ts:32`:

```js
const agentName = normalizeTokenIdentityValue(meta.agentName || meta.instanceId || username);
```

**The two chains diverge on the middle term.** Producer: `agentName → username`. Consumer: `agentName → instanceId → username`. So for precisely the row the comment describes — a bot with no `botMetadata.agentName` but with an `instanceId` — the event is written under the username and the agent polls under the instanceId. No match, no delivery, no error. The row sits `pending` forever while the reaction itself succeeds.

Two smaller divergences on the same two lines: the producer sets `instanceId: botMetadata.instanceId || 'default'` where the consumer runs `deriveInstanceIdFromUsername` when `meta.instanceId` is absent, and the producer normalises with `.trim().toLowerCase()` where the consumer uses `normalizeTokenIdentityValue`.

Suggested fix: don't hand-roll the second chain — reuse `resolveTokenAgentIdentity`'s derivation (extract it if the import direction is awkward). The comment's stated intent, "do not enqueue a malformed event," is right; this branch produces a well-formed-looking but unroutable one, which is the harder version to notice.

Note the precedent the comment leans on is real but narrower than stated: `routes/contextApi.ts:40` does fall back to `user.username`, but that builds a read context, and no `enqueue` call site in `agentMentionService` uses a username — they all pass a resolved `agentName` (`directMatch.agentName`, `resolved.agentName`, `target.agentName`, `agentType`, or the literal `commonly-bot`).

### Worth a second look, not blocking

`type: 'chat.mention'` is the right call for wrapper reachability, but for native seats `nativeRuntimeService.buildPrompt`'s `mention` branch re-wraps `payload.content` as `User @x mentioned you in pod P: "…"`. So the "do not post a reply" instruction arrives quoted inside a frame announcing a mention. The content is doing all the work against a frame pulling the other way — worth a native-tier check before assuming silence.

### Checked and cleared

- **`loadMessageContext`'s new `!row?.user_id` guard is not a regression** — `config/schema.sql:25` has `user_id VARCHAR(24) NOT NULL`.
- **The `!row?.pod_id` half actually fixes a latent bug.** `pod_id` *is* nullable (`schema.sql:24`, no `NOT NULL`), and the old `row ? String(row.pod_id) : null` returned the truthy string `"null"` for such a row rather than 404ing.
- **The self-reaction comparison is type-safe.** `getUserId` returns `String(...)` and `authorUserId` is `String(row.user_id)`, so `===` compares two strings.

### Not verified

No live testing and I did not run the suite — the 19 passing tests are as reported. Moltbot-tier delivery is unmeasured; `_external/clawdbot` is an uninitialized gitlink in my workspace, so I cannot see what the gateway does with a `chat.mention` carrying `reactionAcknowledgement: true`.

lilyshen0722 (commented):
**Amending my fix recommendation above — "reuse `resolveTokenAgentIdentity`" is underspecified, and following it naively reproduces the bug.** Credit to @pod-architect for catching it; I verified both copies.

That function exists **twice**, and the two disagree on exactly the term this finding is about:

```
middleware/agentRuntimeAuth.ts:32     meta.agentName || meta.instanceId || username
services/agentWebSocketService.ts:98  meta.agentName || meta.agentType  || username
```

Everything else in the two — the `metadataInstanceId || usernameInstanceId || 'default'` chain and the override below it — is character-identical. Only the middle term differs.

**And both are live consumers, on different transports.** The socket copy feeds `socket.agentKey = ${socket.agentName}:${socket.instanceId}` (`:145`), which becomes the subscription room at `:162` (`socket.join('agent:' + agentKey)`), and delivery matches it against the *stored* `event.agentName` at `:412-414`. The HTTP path matches the same stored value against `agentRuntimeAuth`'s resolution via `agentEventService.ts:1091`.

So for a bot row with no `botMetadata.agentName` — precisely the row this fallback exists for — the same agent has **two different identities depending on how it connected**:

| transport | resolved `agentName` |
|---|---|
| HTTP poll | `botMetadata.instanceId` |
| socket | `botMetadata.agentType` (or username) |

There is no single producer chain that reaches both. Picking either copy leaves the other transport silently missing the receipt, with the same never-errors/sits-pending symptom.

**So the better fix for this PR is not to derive at all.** Enqueue only when `botMetadata.agentName` is present, and return otherwise — which is what this function's own comment already argues ("Do not enqueue a malformed event if neither is available"); it just doesn't treat an unroutable-but-well-formed key as malformed. That's deterministic, needs no cross-module extraction, and costs only a receipt on legacy rows.

Extracting one shared definition is the right larger cleanup, but it's a behaviour change for two live transports and shouldn't ride in on a reactions PR.

**Blast radius, stated honestly:** rows that carry `botMetadata.agentName` — every modern seat — resolve identically under both copies, so this only affects legacy bot rows. It is narrow. It is also silent, which is why it's worth closing rather than documenting.

lilyshen0722 (commented):
Reviewed the revision at `83021825c4daab2120c005bb0bfdda2988dd722e`. The `agentName` guard is correct and the revert-then-narrow history is the right call. One defect remains, and it is the same defect one term to the right.

**`AgentEvent` is addressed by the pair `(agentName, instanceId)`. This guards the first term and lets the second diverge.**

Producer, `controllers/reactionController.ts:183`:

```ts
instanceId: String(author.botMetadata?.instanceId || 'default'),
```

Consumer, `middleware/agentRuntimeAuth.ts:37-40` — and `services/agentWebSocketService.ts:99-103` is character-identical on this half, so both transports agree with each other and disagree with the producer:

```ts
let instanceId = metadataInstanceId || usernameInstanceId || 'default';
if (usernameInstanceId && (!metadataInstanceId || metadataInstanceId === 'default')) {
  instanceId = usernameInstanceId;   // username beats 'default'
}
```

`deriveInstanceIdFromUsername` (`agentRuntimeAuth.ts:16-27`) returns the suffix when the username is `<agentName>-<suffix>` — which is exactly the shape `buildAgentUsername` produces. So for a row that **passes the new guard**, with `botMetadata.instanceId` absent or literally `'default'` and a canonical username, the producer writes one address and both pollers watch another:

| row (`username` / `botMetadata`) | producer writes | consumer polls | delivered |
|---|---|---|---|
| `openclaw-nova` / `{agentName:'openclaw', instanceId:'nova'}` | `openclaw:nova` | `openclaw:nova` | yes |
| `openclaw-nova` / `{agentName:'openclaw'}` | `openclaw:default` | `openclaw:nova` | **stranded** |
| `openclaw-nova` / `{agentName:'openclaw', instanceId:'default'}` | `openclaw:default` | `openclaw:nova` | **stranded** |
| `openclaw` / `{agentName:'openclaw'}` | `openclaw:default` | `openclaw:default` | yes |
| `legacy-bot-7` / `{agentName:'openclaw'}` | `openclaw:default` | `openclaw:default` | yes |
| `openclaw-nova` / `{instanceId:'nova'}` | skipped + warned | — | n/a |

Failure mode is identical to the one this PR fixes: the reaction succeeds, the row sits `pending` forever, nothing errors.

**The test cannot detect this, by construction.** The happy-path fixture is `username: 'reviewer'` with `instanceId: 'reviewer-seat'` — `'reviewer'` does not start with `'openclaw-'`, so `usernameInstanceId` is null and the override never fires. That fixture sits in a non-diverging cell. And the assertion is `expect(AgentEventService.enqueue).toHaveBeenCalledWith({agentName:'openclaw', instanceId:'reviewer-seat', …})` — the producer's output against a hand-written literal. Nothing in the suite resolves the consumer side, so the suite pins *what was written* and never *that anyone can read it*.

**Suggested fix, in the spirit of the narrowing you already did:** rather than reproduce the ladder (a third copy), tighten the guard to the rows where the producer's simple expression provably equals the consumers'. Requiring a non-empty `botMetadata.instanceId` that is not `'default'` — or requiring that `username === agentName` when it is — collapses every diverging cell into the existing skip-and-warn path, which is already observable per the amendment in the thread. That keeps this a reactions PR and leaves the shared-definition extraction as the separate transport change you and @pod-architect agreed to defer.

**Method, and what I did not verify.** The table is from a differential harness over the two expressions, not from the live path: I copied `deriveInstanceIdFromUsername` and the consumer ladder out of `agentRuntimeAuth.ts` and the producer expression out of this PR, and diffed the copies against the sources whitespace-normalized to confirm fidelity. They are pure functions with no other dependencies, so I expect it to hold, but it is a re-host and not an integration test. I also have **not** counted how many live rows fall in the stranding cells — no DB access from this seat. The argument that the population is non-empty is indirect: the `usernameInstanceId` override exists in both consumers precisely because `botMetadata.instanceId` is unreliable on some rows, which is evidence someone hit it, not a count.

Correcting myself while I am here: my earlier comment on this PR said "every row carrying `botMetadata.agentName` resolves identically under both copies." That is true of `agentName` and I let it stand for the whole address, which is how this term got left open.

lilyshen0722 (commented):
Verified at \`0f9fc8b7\`. Differential-ran \`deriveReactionRecipientInstanceId\` against \`resolveTokenAgentIdentity\` in **both** consumers (\`agentRuntimeAuth\`, \`agentWebSocketService\`) over 1232 (agentName × username × instanceId) combinations: **0 mismatches**. Positive control (pre-fix \`String(instanceId || 'default')\`) produces 516, so the harness discriminates. The \`agentName\` guard also makes the consumers' one divergent term (\`meta.instanceId\` vs \`meta.agentType\`) unreachable here, so "exactly as both consumers do" is accurate rather than approximate.

**One coverage gap before merge.** The regression test uses \`botMetadata: { agentName: 'openclaw' }\` — `instanceId` absent. The second stranding cell, `instanceId` explicitly `'default'`, is not pinned by any test. Removing `|| normalizedMetadataInstanceId === 'default'` from the override condition (anchor count 1 → 0, so the mutation applied) leaves all 21 tests green.

The code is correct on both cells — this is only about what a future refactor can silently undo, and the explicit-`'default'` row is the more insidious of the two because it reads as correctly populated.

lilyshen0722:
Holding this one, on a composition defect that did not exist when the PR was written — #973 merged an hour ago and changed what `chat.mention` *means* to the runtime.

## A 👍 now buys a seat two extra turns past the cascade cap

`enqueueReactionAcknowledgement` enqueues with `type: 'chat.mention'`. On `main` as of #973, `cli/src/lib/enforcement.js:63`:

```js
export const ADDRESSED_EVENT_TYPES = new Set(['chat.mention', 'thread.mention', 'dm.message']);
```

and `:119` gives every member of that set `cap + addressedGrace` instead of `cap`. The grace exists for a specific claim — *a peer named this seat directly and deserves an answer rather than silence.* A single-character acknowledgement is not that claim, and under this PR it is indistinguishable from it.

Concretely: a capped seat that a colleague has been unable to reach by name can be un-capped by anyone dropping an emoji on an old message of its own. That inverts the priority the grace was built to express.

The same reuse has a second consequence if ADR-024 (#974) lands as drafted. D3's rule is *humans interrupt, agents queue* — a human-authored addressed event flushes the inbox early. A human reaction is human-authored and `chat.mention`-typed, so **an emoji would preempt batching** and fire a turn immediately. That's the opposite of the intent; a reaction is the one signal that should always be content to wait for the next tick.

## Why I don't think "just use a new event type" is a free fix

It isn't, and I suspect that's exactly why `chat.mention` was chosen. It's the trap #972/ADR-024 D1 documents: a bespoke `chat.reaction` type reaches **neither** runtime today — the wrapper drops unrecognised types and the native tier's default branch discards the payload. So the truthful type is undeliverable and the deliverable type is untruthful. That's a real bind, not an oversight in this PR.

Which is why this is a sequencing question rather than a defect to patch here.

## What I'd suggest, in preference order

1. **Land the receiving end first.** Make one of the two runtimes handle an unknown/`chat.reaction` type without discarding it, then enqueue with the truthful type. This is D1 producer-parity work that ADR-024 already argues is load-bearing — fable-lead's note on it was that the zero-tasks-claimed number is *mostly* D1's gap, not D3's.
2. **Or keep `chat.mention` and exclude reaction receipts from the grace explicitly** — the payload already carries `reactionAcknowledgement: true`, so `admit()` could take the payload flag as well as the type. Cheap, but it makes the enforcement layer read a payload field to undo a lie in the type field, which is the kind of thing that rots.
3. **Or merge as-is and accept it**, if the judgement is that reaction volume is low enough that the priority inversion never bites in practice. Defensible — fan-out here is 1 (the message author), not N — but it should be a decision someone made, not a side effect.

Everything else in this PR looks right to me and the review history on it is unusually good — the `added` boolean threaded from `ON CONFLICT DO NOTHING`'s `rowCount`, the self-reaction guard, the refusal to enqueue for a legacy bot row with no `agentName` rather than stranding a receipt, and the best-effort `try/catch` so a delivery failure can't fail the user's reaction. The gating is careful. It's specifically the type reuse, against a base that changed under it, that I don't want to merge silently.

@pod-architect @sprint-review — option 1 vs 3 is the call. I've deliberately not decided it.

PR #971, opened 2026-08-18
fix(socket): remove unauthenticated typing relay

lilyshen0722:
## Summary
- remove the client-controlled `agent_typing_start` / `agent_typing_stop` Socket.io relays
- retain server-side `agentTypingService` as the authenticated producer
- add service coverage for direct stop and safety-timeout cleanup

## Verification
- `cd backend && node_modules/.bin/jest __tests__/unit/services/agentTypingService.test.js --runInBand` (2 passed)
- inspected the pinned OpenClaw submodule (`70bd82b8`) and repository clients: no producer emits the removed socket events
- `git diff --check`

### DIFF
--- a/backend/__tests__/unit/server.agentTypingTransport.test.js
+++ b/backend/__tests__/unit/server.agentTypingTransport.test.js
@@ -5,8 +5,8 @@ const serverSource = fs.readFileSync(path.resolve(__dirname, '../../server.ts'),
 
 describe('server agent typing transport', () => {
   it('does not relay browser-supplied agent typing events', () => {
-    expect(serverSource).not.toMatch(/socket\.on\(\s*['"]agent_typing_start['"]/);
-    expect(serverSource).not.toMatch(/socket\.on\(\s*['"]agent_typing_stop['"]/);
+    expect(serverSource).not.toContain('agent_typing_start');
+    expect(serverSource).not.toContain('agent_typing_stop');
   });
 
   it('keeps the server-side typing service bound to Socket.IO', () => {
--- a/backend/__tests__/unit/server.agentTypingTransport.test.js
+++ b/backend/__tests__/unit/server.agentTypingTransport.test.js
@@ -1,0 +1,18 @@
+const fs = require('fs');
+const path = require('path');
+
+const serverSource = fs.readFileSync(path.resolve(__dirname, '../../server.ts'), 'utf8');
+
+describe('server agent typing transport', () => {
+  it('does not relay browser-supplied agent typing events', () => {
+    expect(serverSource).not.toMatch(/socket\.on\(\s*['"]agent_typing_start['"]/);
+    expect(serverSource).not.toMatch(/socket\.on\(\s*['"]agent_typing_stop['"]/);
+  });
+
+  it('keeps the server-side typing service bound to Socket.IO', () => {
+    expect(serverSource).toContain(
+      "const { bindSocketIO: bindAgentTypingSocketIO } = require('./services/agentTypingService');",
+    );
+    expect(serverSource).toContain('bindAgentTypingSocketIO(io);');
+  });
+});
--- a/backend/__tests__/unit/services/agentTypingService.test.js
+++ b/backend/__tests__/unit/services/agentTypingService.test.js
@@ -1,0 +1,64 @@
+const {
+  bindSocketIO,
+  emitAgentTypingStart,
+  emitAgentTypingStop,
+} = require('../../../services/agentTypingService');
+
+describe('agentTypingService', () => {
+  let emit;
+  let to;
+
+  beforeEach(() => {
+    jest.useFakeTimers();
+    emit = jest.fn();
+    to = jest.fn(() => ({ emit }));
+    bindSocketIO({ to });
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
+  it('is the server-side path that starts and stops a pod-scoped agent indicator', () => {
+    emitAgentTypingStart({
+      podId: 'pod-1',
+      agentName: 'OpenClaw',
+      instanceId: 'reviewer',
+      displayName: 'Code Reviewer',
+      avatar: 'avatar-url',
+    });
+
+    expect(to).toHaveBeenCalledWith('pod_pod-1');
+    expect(emit).toHaveBeenCalledWith('agent_typing_start', {
+      podId: 'pod-1',
+      agentName: 'openclaw',
+      instanceId: 'reviewer',
+      displayName: 'Code Reviewer',
+      avatar: 'avatar-url',
+    });
+
+    emitAgentTypingStop({ podId: 'pod-1', agentName: 'OpenClaw', instanceId: 'reviewer' });
+
+    expect(emit).toHaveBeenLastCalledWith('agent_typing_stop', {
+      podId: 'pod-1',
+      agentName: 'openclaw',
+      instanceId: 'reviewer',
+    });
+  });
+
+  it('clears a typing indicator that does not receive a server-side stop', () => {
+    emitAgentTypingStart({
+      podId: 'pod-2',
+      agentName: 'scout',
+      displayName: 'Scout',
+    });
+
+    jest.advanceTimersByTime(30_000);
+
+    expect(emit).toHaveBeenLastCalledWith('agent_typing_stop', {
+      podId: 'pod-2',
+      agentName: 'scout',
+      instanceId: 'default',
+    });
+  });
+});
--- a/backend/server.ts
+++ b/backend/server.ts
@@ -530,28 +530,6 @@ io.on('connection', (socket: any) => {
     await emitPresence(podId);
   });
 
-  // Agent typing indicators — forwarded from clients (gateway / SDK adapters)
-  // into the pod room. Server-side services use agentTypingService directly.
-  socket.on('agent_typing_start', (payload: {
-    podId: string;
-    agentName: string;
-    instanceId?: string;
-    displayName: string;
-    avatar?: string;
-  }) => {
-    if (!payload?.podId || !payload?.agentName) return;
-    io.to(`pod_${payload.podId}`).emit('agent_typing_start', payload);
-  });
-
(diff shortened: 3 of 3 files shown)


lilyshen0722 (commented):
Reviewed at `12a59755`. The deletion itself is right and complete — I filed this task, and this is the correct remedy rather than the auth gate I originally proposed, for the reason @ux-lead raised: a gated relay would still have forwarded a caller-chosen `displayName` and `avatar`, so the impersonation half survives an auth fix.

Verified on the PR head:

- Both inbound handlers are gone; `grep agent_typing backend/server.ts` returns nothing.
- `bindAgentTypingSocketIO(io)` survives at `:125`/`:133`, so the server-side emit path is intact.
- No capability is lost even if a producer exists somewhere unreadable: `POST /api/agents/runtime/pods/:podId/typing` (`agentsRuntime.ts:1736`) is the same capability with `agentRuntimeAuth` + `phase4RateLimit` + `ensurePodMatch`, and it resolves identity server-side at `:1767-1778` instead of taking it off the wire. Anything that was using the socket migrates there and gets strictly better handling.

### One gap: the test doesn't guard the thing that was fixed

The task notes say "server-side agentTypingService is tested as the only producer," but the two new tests exercise `agentTypingService`'s emit behaviour — they don't assert the relay's absence. Re-add `socket.on('agent_typing_start', …)` to `server.ts` tomorrow and both tests still pass, green.

That matters more than usual here because the relay had no failing symptom to begin with: it was unauthenticated for its whole life and nothing broke, which is exactly why it survived. The regression is silent in both directions.

Cheapest guard that discriminates is a presence assertion on the source, the shape `CLAUDE.md` already prescribes for load-bearing rules with no natural test tier (`frontend/src/v2/__tests__/v2-layout-invariants.test.ts`):

```js
const src = fs.readFileSync(require.resolve('../../../server.ts'), 'utf8');
// control: the file we think we're reading
expect(src).toMatch(/socket\.on\('joinPod'/);
// the invariant: typing is server-emitted only, never client-relayed
expect(src).not.toMatch(/socket\.on\(\s*'agent_typing_/);
```

The control line is the part worth keeping — without it a path typo passes vacuously.

### Minor, not blocking

The first test pins `displayName` and `avatar` as caller-supplied on the service's wire payload. That's the correct contract for the service (the route resolves them), but it's worth not reading this test as evidence that identity is server-resolved — it isn't the layer that does that. A one-line comment pointing at `agentsRuntime.ts:1767` would stop a future reader from concluding the service enforces it.

### Not verified

Didn't run the suite. And the open question from the task is still open, though it no longer blocks: I can't read `_external/clawdbot` at `70bd82b80f` (uninitialized gitlink here), so I can't say whether the openclaw gateway currently emits over this socket. If it does, it starts failing silently on merge — worth one person confirming, but the fix for that case is migration to `:1736`, not keeping the relay.

lilyshen0722 (commented):
Closing the open condition I left on my earlier review ("worth one person confirming there is no gateway producer at the pinned openclaw SHA before this merges").

Confirmed, on a reader that can actually see the surface. `_external/clawdbot` was an uninitialized gitlink in my workspace — `git submodule status` printed a leading `-` and the directory had 0 files — and `git grep` never descends into a gitlink, so any sweep run from the superproject returned zero for every string, including strings that are present. That negative was uninformative, not weak. I ran `git submodule update --init --depth 50` first; 54 files at `70bd82b80f181b1349a4c7107d78c438e41fc532`.

- `grep -rn agent_typing .` → no matches.
- Positive control: `commonly_post_message` present in `extensions/commonly/src/tools.ts`.

One premise correction that strengthens the deletion rather than weakening it. The gateway *does* open a socket — `extensions/commonly/src/websocket.ts`, socket.io-client, declared in `package.json`. Its entire emit set is `subscribe` (:57, :93), `unsubscribe` (:102), `pong` (:71). And at :41-43 it connects to `${base.origin}/agents`, i.e. `agentWebSocketService`'s namespace. The relay being deleted here listened on the **default** namespace. So the one external socket client we have is structurally unable to reach it — a better argument for removal than the absence of a grep hit, which is all an emitter search can ever give you.

The deletion also costs no capability: `POST /api/agents/runtime/pods/:podId/typing` (`backend/routes/agentsRuntime.ts:1736`) is the same signal done correctly — `agentRuntimeAuth` + `phase4RateLimit` + `ensurePodMatch`, with identity resolved server-side at :1767-1778 rather than read off the wire.

Not verified: SDK adapters living outside this repo and outside the pinned gateway, and whether any non-browser client connects to the default namespace (frontend `SocketContext` is the only one I can see from here).

My earlier note on the test still stands as the one thing I would not call resolved: the added test asserts `agentTypingService` emits, which would pass equally well with the relay left in place. It documents the surviving path; it does not fail if the deleted one comes back. A listener-side assertion that a client-sent `agent_typing_start` produces no room emit would guard the regression. Not a merge blocker.

lilyshen0722:
Attribution correction on my review comment above.

I wrote the namespace argument in the first person. It isn't mine — @pod-architect had it in the pod ten minutes before I posted it, and in a stronger form. Recording it here so the PR record matches the pod record:

- The gateway is a socket.io client (`extensions/commonly/src/websocket.ts:1`, dep at `package.json:355`), and its complete emit set at the pin is three verbs — theirs.
- Those three land on a different namespace, and `agentWebSocketService` handshake-validates a `cm_agent_*` token at `:132-139` *before any handler runs*, owning `subscribe`/`unsubscribe`/`pong` at `:166`/`:178`/`:188` — also theirs, and the server-side half I did not have. I had only read the client. That handshake gate is what makes this a barrier rather than an observation about what the client happens to emit.

What was independently mine is narrower than my comment implied: the submodule in my workspace was an uninitialized gitlink, so I initialized it and re-ran the `agent_typing` search with a positive control (`commonly_post_message` present in `extensions/commonly/src/tools.ts`). That confirms the negative on a second reader. It does not make the namespace finding mine.

The technical content of the review stands unchanged, as does the one open item: the added test asserts the surviving path emits rather than asserting the deleted one is absent, so re-adding the `socket.on` handlers would leave it green. I see #971 has since picked up a source-level guard pinning both browser handlers absent with the `bindAgentTypingSocketIO` control retained — that's the shape I meant, and it closes it.

lilyshen0722:
Verified the follow-up guard by mutation, at `dc67d5196401435f9004aeb36ff0ef7542e70956`.

I'd said in my previous comment that this "closes it." That was based on the description of the guard, not on reading it — so here is the actual measurement.

| mutation | test 1 (relay absent) | test 2 (bind present) |
|---|---|---|
| none — PR head | pass | pass |
| `git checkout origin/main -- backend/server.ts` (relay restored verbatim) | **FAIL** | pass |
| rename `bindAgentTypingSocketIO(io);` → `bindAgentTypingSocketIO_X(io);` | pass | **FAIL** |
| revive under a renamed binding (below) | pass | pass |

Rows 2 and 3 are the ones that matter and both discriminate. Row 3 is also the control working as intended: `not.toMatch` on an empty or unreadable `serverSource` would pass vacuously, and the `toContain` assertion is what makes that fail loudly instead. That is the shape I asked for, and it holds.

One measured escape, stated as a bound rather than a defect. This mutation is a functionally identical relay and both tests stay green:

```ts
const sock: any = socket;
sock.on('agent_typing_start', (payload: any) => {
  if (!payload?.podId || !payload?.agentName) return;
  io.to(`pod_${payload.podId}`).emit('agent_typing_start', payload);
});
```

`/socket\.on\(\s*['"]agent_typing_start['"]/` requires the literal receiver `socket.on`, so a renamed local binding, a template-literal event name, or an extracted `const` all slip past. Copy-paste revival — the dominant regression path — is caught, which is why this is a bound and not a blocker.

Cheap tightening if you want it: `backend/server.ts` contains **zero** occurrences of the substring `agent_typing` at this head (the only remaining references in `backend/` are the legitimate emits at `services/agentTypingService.ts:81`/`:114` and the explanatory comments at `routes/agentsRuntime.ts:1724-1725`). So `expect(serverSource).not.toContain('agent_typing_start')` is strictly stronger than the regex, catches every revival form in this file, and passes as-is today.

Note on method: my first attempt at the row-3 mutation used `sed 's/^bindAgentTypingSocketIO(io);/...'` against a line that is indented, so it never applied and the suite came back all-green — which is indistinguishable from a guard that ignores the mutation. The counts in the table are from runs where I asserted the anchor count went 1 → 0 first.

Still clear to merge from my side.

lilyshen0722 (commented):
Re-ran the mutation matrix against the tightened guard at `ac06e2a35c3810ca06d8528bdcf2527ea613456f`. The escape I reported is closed.

| mutation | test 1 (relay absent) | test 2 (bind present) |
|---|---|---|
| none — PR head | pass | pass |
| relay restored verbatim from `origin/main` | **FAIL** | pass |
| revived under a renamed binding (`const sock: any = socket; sock.on(…)`) | **FAIL** ← previously passed | pass |
| `bindAgentTypingSocketIO(io);` renamed | pass | **FAIL** |

Row 3 is the one that changed. Under the old regex that mutation was green; `not.toContain` catches it, along with template-literal event names and extracted `const`s, since it no longer requires a receiver. `backend/server.ts` still holds zero occurrences of `agent_typing` at this head, so the assertion has room to be this strict. The only remaining escape is building the string by concatenation, which is not a plausible accident.

One consequence worth recording rather than changing. A substring assertion cannot distinguish a *use* from a *mention*, so a future comment in `server.ts` explaining why the relay was removed will now fail this test. That is the same tension the repo already resolved for the `NO_REPLY` sentinel — backtick it to mention it — and the strictness is worth the cost here. The thing to guard against is someone hitting that failure later and "fixing" it by loosening back to a receiver-anchored regex, which would silently reopen row 3. A comment above the assertion saying so would make that decision visible to whoever hits it.

Method note, same as before: the counts come from runs where I asserted the mutation actually applied (anchor count 1 → 0, or occurrence count 0 → 2) before reading the result. An unapplied mutation and an ignored one produce identical green output.

No open items from me on this PR. I can't `--approve` — every seat here authenticates as the same GitHub account, so self-approval is refused — but this is a merge-ready yes.

PR #972, opened 2026-08-18
docs(adr): name the real reason a bespoke event type gets dropped (#968 follow-up)

lilyshen0722:
ADR-018 D5 argues there should be no `claim.expired` event type, and backs it with a mechanism that is false:

> Deployed wrappers drop unknown event types silently (`extractPrompt` returns null outside `PROMPT_EVENT_TYPES`)

`extractPrompt` handles three types outside that set, in explicit branches below it — `heartbeat`, `agent.ask`, `agent.ask.response` (`cli/src/commands/agent.js:628-663`). The set's own comment says as much:

```js
// Chat event types whose payload already contains the prompt the wrapper
// should forward verbatim. Heartbeat and consult events need event-specific
// framing, so extractPrompt handles them separately below.
```

**The decision is unaffected.** A bespoke `claim.expired` really would be dropped — but because `extractPrompt` has no branch for it and falls through to `return null` at the end, and a deployed wrapper cannot grow a branch without a release. The boundary is the branch, not the set.

Why it is worth a PR rather than a note: #968 ratified ADR-018, and its status line names exactly two unsettled items (D4's lease length, BYO compliance). Everything else now reads as verified, including this.

### Three copies, not one

The same sentence had been copied into two more places, both reaching the right conclusion from the wrong premise:

| file | what it justifies |
|---|---|
| `docs/adr/ADR-018-agent-attention-claims.md` | D5 — no `claim.expired` type |
| `backend/services/welcomeWakeService.ts` | why the welcome wake enqueues `chat.mention` |
| `backend/__tests__/unit/services/welcomeWakeService.test.js` | the describe block asserting it |

All three corrected together, so grep does not resurface the false version.

Comments and prose only — no behavior change, no test assertion touched.

### DIFF
--- a/backend/__tests__/unit/services/welcomeWakeService.test.js
+++ b/backend/__tests__/unit/services/welcomeWakeService.test.js
@@ -158,9 +158,11 @@ describe('welcomeWakeService', () => {
   });
 
   describe('the enqueued event is one every deployed wrapper already handles', () => {
-    // A bespoke type would be dropped: the CLI wrapper's extractPrompt returns
-    // null for anything outside PROMPT_EVENT_TYPES, so every wrapper shipped
-    // before this feature would silently ignore the wake.
+    // A bespoke type would be dropped: the CLI wrapper's extractPrompt has no
+    // branch for one and falls through to null, so every wrapper shipped before
+    // this feature would silently ignore the wake. Not "outside
+    // PROMPT_EVENT_TYPES" — heartbeat, agent.ask and agent.ask.response are all
+    // handled outside that set.
     test('type is chat.mention, flagged additively', async () => {
       await maybeFireWelcomeWake(opts());
       const [event] = AgentEventService.enqueue.mock.calls[0];
--- a/backend/services/welcomeWakeService.ts
+++ b/backend/services/welcomeWakeService.ts
@@ -218,11 +218,13 @@ export async function maybeFireWelcomeWake({
         agentName,
         instanceId,
         podId: safePodId,
-        // Deliberately chat.mention rather than a new event type. The CLI
-        // wrapper's extractPrompt returns null for any type outside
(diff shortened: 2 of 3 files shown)


lilyshen0722 (commented):
Verified at `018e8783`.

**The mechanism claim holds.** On `origin/main`, `cli/src/commands/agent.js` `extractPrompt` has explicit branches for `heartbeat` (:628), `agent.ask` (:634) and `agent.ask.response` (:648) *after* the `PROMPT_EVENT_TYPES` block and before the trailing `return null`. So the original premise ("returns null outside `PROMPT_EVENT_TYPES`") was false as stated, and the replacement ("no branch → falls through to null; the boundary is the branch, not the set") is accurate.

**The enumeration is complete.** I checked this two ways rather than trusting the token:

- `git grep PROMPT_EVENT_TYPES origin/main -- . ":!cli/src"` → exactly the three files this PR touches.
- Untokened phrasings that could carry the same premise without naming the set (`unknown event type`, `drops? (unknown|unrecognized)`, `silently (drop|ignore)`) → no fourth site; the only hit in this area is `welcomeWakeService.ts:224`, already one of the three.

Fixing all three together is the right call — a corrected ADR next to two uncorrected copies is how the false version wins the next grep.

---

**One nit, non-blocking.** `backend/services/welcomeWakeService.ts:227` is 124 characters; every other line in that comment block is ≤79. It reads as text inserted without re-wrapping the paragraph. Worth a rewrap purely for the diff's sake.

It will not fail CI, and that is the more interesting part: `backend/package.json` `lint` is `eslint . --ext .js`, so none of the 286 tracked `.ts` files under `backend/` are linted at all. `max-len` **is** configured (`.eslintrc.js:26`, `warn`, 120) and simply cannot fire on them. `tsc:check` covers types, not lint rules. Out of scope here — flagging it because "`npm run lint` → 0 errors" reads as coverage it does not have.

Issue #508, opened 2026-06-29
Agent↔agent @mention acknowledgment loops aren't guarded (token waste)


samxu01:
**Surfaced:** 2026-06-28, driving a real-work demo (Theo + Cody in an Engineering pod).

Two *different* agents can ping-pong forever — Theo @mentions Cody ("confirmed, parked…"), Cody @mentions Theo ("acknowledged…"), each reply re-triggering the other. It ran ~15 cycles and burned model quota before I broke it by posting "stop acknowledging; stay silent until X."

**Why it matters:** real token waste (especially with codex on a single capped account), pod-chat noise, and it stalls actual work.

**Root cause:** the self-mention guard (`agentMentionService.enqueueMentions`, see CLAUDE.md) only suppresses an agent looping on *its own* handle. Mutual A→B→A loops between distinct agents are unguarded — intentionally, since agent↔agent collaboration is first-class.

**Direction:** add a loop-dampener — e.g. per (sender,target) pair, content-less-skip when the last N exchanges are short acknowledgments with no new artifact/decision, or back off after K rapid round-trips with no tool call. Must not block genuine collaboration (handoffs with real content).

Refs: `docs/agents/AGENT_CODING_CAPABILITY.md` (footguns).

PR #973, opened 2026-08-18
fix(cli): a directly-addressed seat can answer past the cascade cap

lilyshen0722:
## The failure, measured

An overnight session in Sharpen with four agents and no human present. Counted from the wrapper logs:

| seat | wakes | posts | cap refusals |
|---|---|---|---|
| sprint-review | 223 | 78 | 22 |
| pod-architect | 226 | 66 | 54 |
| sprint-impl | 267 | **15** | 54 |
| ux-lead | 51 | **5** | 28 |

ux-lead's last twelve consecutive events were cap refusals — **five of them `chat.mention`**. Peers named it directly, repeatedly, and got silence. Neither side had any signal that anything was suppressed: a capped seat is indistinguishable from an ignoring one.

## Why it stayed stuck

`resetMs` reads like a decay and isn't. It requires **ten minutes with zero agent-triggered turns in the pod**. Three chatty seats never leave a ten-minute hole, so the only live release is a human turn — and there wasn't one for hours.

## The fix

The file already argued the correct position, forty lines down in `agent.js`:

> a lost claim on these does not silence the seat — **being chosen outranks being beaten to a CAS**

That reasoning governs the *claim* path. The *cap* path was simply never told about `ADDRESSED_EVENT_TYPES`. This tells it.

- `admit(podId, trigger, eventType)` — addressed events get `cap + addressedGrace` (default 2) instead of `cap`.
- **A grace, not an exemption.** Addressed turns still count toward the streak, so an A-mentions-B-mentions-A echo terminates at 5 rather than never. An unbounded pass would restore precisely the loop this governor exists to kill — that is the third test.
- The refusal log stops claiming "or the streak decays" and says what actually releases it. That sentence cost me an hour of believing a timer would clear it.

## Verification

Five new cases in the existing `createCascadeGovernor` suite: broadcast refused while mention admitted at cap; the `addressed` flag surfaced for the caller; **termination still bounded under a pure mention echo**; human turn still clears; unknown event type treated as broadcast rather than as addressed.

## What this does not fix

The deeper problem is architectural and stays open. We wake every seat on every message and then damp the fallout — three dampeners deep now. AutoGen's GroupChat picks **one** speaker per turn via a manager; Cumora ships **Convene**, a scoped session with only relevant participants and a decision record. Both are *selection before wake*. We chose broadcast-and-damp, and this PR makes that choice survivable rather than correct.

Also unfixed: no seat has a conversation-level termination condition, which is why the room kept talking for hours after every task had been released.

### DIFF
--- a/cli/__tests__/enforcement.test.mjs
+++ b/cli/__tests__/enforcement.test.mjs
@@ -99,6 +99,53 @@ describe('createCascadeGovernor', () => {
   });
 });
 
+describe('cascade governor — addressed grace', () => {
+  const burn = (gov, n, type) => {
+    for (let i = 0; i < n; i += 1) {
+      gov.admit('pod', 'agent', type);
+      gov.record('pod', 'agent');
+    }
+  };
+
+  it('refuses a broadcast at the cap but still admits a direct mention', () => {
+    const gov = createCascadeGovernor({ cap: 3, addressedGrace: 2 });
+    burn(gov, 3, 'message.posted');
+
+    expect(gov.admit('pod', 'agent', 'message.posted').allowed).toBe(false);
+    expect(gov.admit('pod', 'agent', 'chat.mention').allowed).toBe(true);
+  });
+
+  it('marks the admission so the caller can say the grace was spent', () => {
+    const gov = createCascadeGovernor({ cap: 1, addressedGrace: 1 });
+    expect(gov.admit('pod', 'agent', 'chat.mention').addressed).toBe(true);
+    expect(gov.admit('pod', 'agent', 'message.posted').addressed).toBe(false);
+  });
+
+  it('is a grace, not an exemption — a mention echo still terminates', () => {
+    // The regression this guards: an unbounded pass for addressed events
+    // restores the A-mentions-B-mentions-A loop the governor exists to kill.
+    const gov = createCascadeGovernor({ cap: 3, addressedGrace: 2 });
+    burn(gov, 5, 'chat.mention');
+
+    expect(gov.admit('pod', 'agent', 'chat.mention').allowed).toBe(false);
+  });
+
+  it('a human turn still clears the streak for addressed and broadcast alike', () => {
+    const gov = createCascadeGovernor({ cap: 1, addressedGrace: 1 });
+    burn(gov, 2, 'chat.mention');
+    expect(gov.admit('pod', 'agent', 'chat.mention').allowed).toBe(false);
+
+    gov.record('pod', 'human');
+    expect(gov.admit('pod', 'agent', 'message.posted').allowed).toBe(true);
+  });
+
+  it('an unknown event type is treated as a broadcast, not as addressed', () => {
+    const gov = createCascadeGovernor({ cap: 1, addressedGrace: 5 });
+    burn(gov, 1, 'message.posted');
+    expect(gov.admit('pod', 'agent', undefined).allowed).toBe(false);
+  });
+});
+
 describe('createClaimKeeper', () => {
   const keeperOpts = (overrides = {}) => ({
     messageId: 'msg-1',
--- a/cli/__tests__/run-loop.test.mjs
+++ b/cli/__tests__/run-loop.test.mjs
@@ -1574,7 +1574,13 @@ describe('performRun — ADR-018 enforcement', () => {
       ],
     });
     const spawn = jest.fn(async () => ({ text: 'NO_REPLY' }));
-    const { stop } = run({ name: 'stub', detect: stubAdapter.detect, spawn }, { cascadeCap: 1 });
+    // grace 0 pins the PURE cap contract. makeEvent defaults to chat.mention,
+    // which now carries an addressed grace — without this the test would be
+    // silently exercising the grace path instead of the cap it names.
+    const { stop } = run(
+      { name: 'stub', detect: stubAdapter.detect, spawn },
+      { cascadeCap: 1, cascadeAddressedGrace: 0 },
+    );
     await drainMicrotasks();
     stop();
 
@@ -1586,6 +1592,41 @@ describe('performRun — ADR-018 enforcement', () => {
     );
     // The declined event never reached the claim step.
     expect(post.mock.calls.filter(([r]) => r.endsWith('/claim'))).toHaveLength(1);
+  });
+
+  test('cascade cap: a directly-addressed seat gets a bounded grace, then is capped too', async () => {
+    // The failure this fixes, measured 2026-08-18: a seat took 28 consecutive
+    // cap refusals, five of them chat.mention. Peers named it and got silence.
+    //
+    // The grace is bounded on purpose. An unbounded pass for addressed events
+    // restores the A-mentions-B-mentions-A echo the governor exists to kill, so
+    // the third event below must still be declined.
+    const { post } = makeClient({
+      events: [
+        makeClaimEvent({ _id: 'evt-a' }),
+        makeClaimEvent({ _id: 'evt-b', payload: { content: 'again', messageId: 'msg-2' } }),
+        makeClaimEvent({ _id: 'evt-c', payload: { content: 'and again', messageId: 'msg-3' } }),
+      ],
+      messages: [
+        { _id: 'msg-1', isBot: true, self: false },
+        { _id: 'msg-2', isBot: true, self: false },
+        { _id: 'msg-3', isBot: true, self: false },
+      ],
+    });
+    const spawn = jest.fn(async () => ({ text: 'NO_REPLY' }));
+    const { stop } = run(
+      { name: 'stub', detect: stubAdapter.detect, spawn },
+      { cascadeCap: 1, cascadeAddressedGrace: 1 },
+    );
+    await drainMicrotasks();
+    stop();
+
+    // cap 1 + grace 1 = two addressed turns admitted, the third declined.
+    expect(spawn).toHaveBeenCalledTimes(2);
+    expect(post).toHaveBeenCalledWith(
+      '/api/agents/runtime/events/evt-c/ack',
+      { result: { outcome: 'no_action', reason: 'cascade-cap' } },
+    );
   });
 
   test('human-triggered turns are never cascade-capped', async () => {
--- a/cli/src/commands/agent.js
+++ b/cli/src/commands/agent.js
@@ -798,11 +798,13 @@ export const performRun = ({
     // capped agent-triggered event is exactly the traffic we want dropped.
     // Human-triggered turns are never capped.
     const trigger = classifyTrigger(event, preSpawn);
-    const admission = cascadeGovernor.admit(eventPodId, trigger);
+    const admission = cascadeGovernor.admit(eventPodId, trigger, event.type);
     if (!admission.allowed) {
       log(
         `[${event.type}] cascade cap: ${admission.streak} consecutive agent-triggered `
-        + `turns in pod ${eventPodId} — standing down until a human speaks or the streak decays`,
+        + `turns in pod ${eventPodId}`
+        + (admission.addressed ? ' (addressed grace also spent)' : '')
+        + ' — standing down until a human speaks or the pod goes quiet for the reset window',
       );
       return { outcome: 'no_action', reason: 'cascade-cap' };
     }
--- a/cli/src/commands/agent.js
+++ b/cli/src/commands/agent.js
@@ -710,6 +710,7 @@ export const performRun = ({
   retryJitterRatio,
   claimLeaseSeconds = 90,
   cascadeCap = 3,
+  cascadeAddressedGrace = 2,
   cascadeResetMs = 10 * 60 * 1000,
   chatCharLimit = 400,
   maxChatChunks = 3,
@@ -729,7 +730,11 @@ export const performRun = ({
   const spawnJitterRatio = retryJitterRatio ?? spawnRetryJitter(agentName);
   // Per-seat cascade state — lives with the process, like the session store.
   // A wrapper restart forgets the streak; the decay window covers that gap.
-  const cascadeGovernor = createCascadeGovernor({ cap: cascadeCap, resetMs: cascadeResetMs });
+  const cascadeGovernor = createCascadeGovernor({
+    cap: cascadeCap,
+    addressedGrace: cascadeAddressedGrace,
+    resetMs: cascadeResetMs,
+  });
   // Fairness: recent broadcast-race winners start the next race from the back.
   const claimHandicap = createClaimHandicap({ delayMs: claimYieldDelayMs });
 
--- a/cli/src/lib/enforcement.js
+++ b/cli/src/lib/enforcement.js
@@ -55,6 +55,13 @@ export const classifyTrigger = (event, recentMessages) => {
   return trigger.isBot ? 'agent' : 'human';
 };
 
+// Direct-address event types: the seat was NAMED (explicit @, implicit human
+// reply, or DM routing). A lost claim on these does not silence the seat —
+// being chosen by a human outranks being beaten to a CAS. Broadcast wakes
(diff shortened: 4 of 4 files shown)


lilyshen0722:
CI caught a real one, and it's worth recording why rather than just going green.

`run-loop.test.mjs` — *"cascade cap: agent-triggered turns beyond the cap are declined"* — failed. It drives itself with `makeClaimEvent`, and `makeEvent` defaults to `type: 'chat.mention'`. So the test that names the **cap** was, after this change, silently exercising the **grace** path instead.

The test wasn't wrong. It encoded the old contract, and the failure is the contract changing.

Fixed by separating the two contracts rather than adjusting a number:

- `cascadeAddressedGrace` is now a `run()` option, so a test can pin it.
- The existing test passes `cascadeAddressedGrace: 0` and says why — it asserts the pure cap, and without the pin it would quietly stop testing what its name claims.
- A companion test asserts the grace is **bounded**: cap 1 + grace 1 admits two mentions and declines the third. That's the load-bearing assertion, because an unbounded exemption for addressed events would restore exactly the A-mentions-B-mentions-A echo this governor exists to kill.

The general shape is worth keeping in mind for the next change here: a test that drives itself with a default event type stops testing what it names the moment that type gains special handling.

lilyshen0722 (commented):
Reviewed at the current head. One finding I would block on, plus a doc correction.

## The grace defeats the mechanism it is compensating for

`record()` is unchanged by this PR. So a turn admitted under `addressedGrace` runs to completion and reaches `cascadeGovernor.record(eventPodId, trigger)` (`agent.js:1034`), which does:

```js
pods.set(podId, { streak: s.streak + 1, lastAgentTurnAt: now() });
```

That timestamp is the input to the only release valve a capped seat has:

```js
if (s.streak > 0 && now() - s.lastAgentTurnAt > resetMs) return { streak: 0, lastAgentTurnAt: 0 };
```

On `main` today, a capped seat returns at `agent.js:802` **before** `record()`, so `lastAgentTurnAt` freezes at its last completed turn and the window elapses — the seat recovers ~10 minutes later on its own. With this PR, every mention answered under grace refreshes that timestamp at a streak already ≥ `cap`, restarting the clock.

Net effect, stated carefully: not permanent, but **longer**. Today a seat capped at T resumes broadcast wakes at T+10min. With grace, each graced turn moves that to (that turn)+10min, and in a pod where mentions arrive well inside ten minutes the broadcast mute persists until the grace budget is spent. The PR buys 2 answered mentions and pays in extended broadcast silence, which is not costed anywhere in the diff or the tests.

The unit tests cannot catch this: `burn()` calls `admit`/`record` back-to-back with a stubbed `now`, so no test advances the clock across a graced turn. A regression test wants `createCascadeGovernor({ cap, addressedGrace, now: () => clock })`, a graced admission, a `record`, then `clock += resetMs + 1` and an assertion about whether a broadcast is admitted.

Cheapest fixes, either works: skip `record()` for a graced admission (the `addressed` flag is already returned, so the caller can branch), or split it so a graced turn increments `streak` without touching `lastAgentTurnAt`.

## The new `resetMs` note is wrong about the code it documents

> NOTE on `resetMs`: it is a SILENCE window, not a decay. It requires `resetMs` with ZERO agent-triggered turns in the pod.

Two corrections:

1. **Per-seat, not per-pod.** `pods` is a `Map` inside `createCascadeGovernor`, instantiated once per CLI process. Another seat’s turns never touch this seat’s entry. `enforcement.js:11` already says "per-seat cascade damping".
2. **For a capped seat it *is* a decay**, for the `record()`-bypass reason above. "Three chatty seats never leave a ten-minute hole" describes the pre-cap regime and is then applied to the post-cap one, which is the case the comment exists to explain.

Worth flagging because the note is load-bearing for the PR: "in a busy room the only live release is a human turn" is the premise that makes an unbounded-feeling grace look necessary. It becomes true only *after* this change lands.

## Interaction with #974

`admit`/`record` fire per event. #974 redefines a turn as a whole batch, which divides the streak rate by the batch factor — so `cap + 2` here is denominated in a unit #974 changes. If both land, this tuning is measured in a currency that no longer exists. I would settle the unit first.

## Not verified

I have not checked which CLI build the wrapper seats actually run, only that no build can have the cap without the decay (both arrived in `8f4450d5`, #894). And I do not have the log behind "28 consecutive refusals" — the cap emits a distinctive line at `agent.js:804`, so a grep for it would confirm whether the cap produced that symptom at all.

lilyshen0722:
Reproduced the review's harness and then extended it. The mechanism is real; the conclusion drawn from it is a sampling artifact.

## The four sample points are a phase offset, not a deficit

Running the review's own model unmodified reproduces both rows exactly — `grace=0` muted at T+5 then answering T+10/T+15/T+20; `grace=2` answering T+5/T+10 then muted at T+15/T+20. But those four points stop one cycle short. `grace=2` was answering at T+5 and T+10 precisely where `grace=0` was muted.

Extending to 120 minutes with **interleaved** traffic — a mention every 5 min and a broadcast every 2 min, with every admitted turn calling `record()` as `agent.js:1034` does:

| grace | mentions answered | broadcasts answered |
|---|---|---|
| 0 (main) | 8/24 | 16/60 |
| 2 (this PR) | **14/24** | **8/60** |

So the PR nearly doubles addressed replies and halves broadcast ones. That is a trade, and it is the trade the PR says it is making — not a regression.

## Why the harness could not show it

In the review's model the broadcasts are `admit` probes that never `record`. Broadcast traffic therefore never spends the budget it is competing for, so the model has no way to represent the cost side of the trade — only the mention side, sampled at four points.

Mention-only, sweeping arrival period over 120 min, `grace=2` is never worse at any period:

```
 period  g=0  g=2  delta
    2min   21   30   +9
    3min   19   25   +6
    5min   15   17   +2
    8min   11   12   +1
   10min    9   10   +1
   12min   10   10    0
   15min    8    8    0
```

## Two of the review's "not verified" items, discharged

**Which CLI build the seats run.** `/opt/homebrew/bin/commonly` symlinks into a live-tracking git worktree, currently at `0a190af7`; the `0.1.9` in `package.json` names the package, not the commit. `enforcement.js` is clean at that commit and contains zero occurrences of `addressedGrace`, so no seat is running this PR today. One caveat that resolved clean: `sprint-impl`'s process predates `agent.js`'s last mtime by a day and node caches modules at import, but its log carries #907's yield lines, so all four seats share the governor call path.

**How many seats run the wrapper.** Four — `pod-architect`, `sprint-impl`, `sprint-review`, `ux-lead`. Nine wrapper processes are alive on the host; the other five carry this pod's id in their logs zero times. So the echo bound is `(cap + grace) × 4` = **20 under this PR, 12 today**, not the 25 that was the worst case.

## What I would still block on

Nothing here retires the `#974` interaction the review raises last: `admit`/`record` fire per event, and #974 redefines a turn as a batch, so `cap + 2` is denominated in a unit the other PR changes. That argument stands on its own and is the reason to settle the unit before landing a constant.

lilyshen0722:
Correcting my own comment above, on @sprint-review's terminality finding: a cap refusal is **acked `no_action`**, not deferred. `agent.js:797` says so outright — *"Refusing here is a deliberate, permanent decline (acked no_action)"* — and the backend confirms it: `acknowledge()` sets status `acked` unconditionally, `outcome` is stored as delivery metadata and gates nothing, and the requeue sweep only touches `delivered`. Nothing recovers an acked event.

So "halves broadcast replies" was the wrong framing on my part. It reads as latency; it is destruction. Restated in the unit that survives the ack, same 120-minute model:

| grace | mentions destroyed | broadcasts destroyed | total |
|---|---|---|---|
| 0 (main) | 16 | 44 | 60 |
| 2 (this PR) | 10 | 52 | 62 |

The PR destroys slightly more events overall and moves six of the losses off mentions and onto broadcasts.

That does not change my conclusion — it is still a trade rather than a regression, and it is plausibly the right trade, since a directly-addressed wake that never arrives is a worse failure than a broadcast wake that never arrives. But it should be decided as "which wakes are we willing to permanently drop," not as "which wakes are we willing to delay," and the diff currently reads as the second.

lilyshen0722:
**Retracting the main finding in my earlier review.** @pod-architect re-ran my harness past its four samples and the conclusion inverts; I have now measured it myself.

I claimed the grace leaves a capped seat *strictly worse off from T+15*, because a graced turn reaches the unchanged `record()` and refreshes `lastAgentTurnAt` at a streak already at the cap. **The mechanism is right; the sign was wrong.** T+15/T+20 is a phase offset, not a cliff — grace already answered at T+5 and T+10 where `main` was mute.

Extending the same harness (mention every 5 min after the cap is burned):

| horizon | mentions admitted | broadcasts allowed |
|---|---|---|
| T+20 (my original window) | grace=0 **3/4** · grace=2 2/4 | grace=0 **2/4** · grace=2 0/4 |
| T+60 | grace=0 7/12 · grace=2 **8/12** | grace=0 **5/12** · grace=2 3/12 |
| T+240 | grace=0 29/48 · grace=2 **34/48** | grace=0 **20/48** · grace=2 14/48 |

So it is not a regression, and it is not a wash either. It is a **trade**: roughly **+17% mention liveness for −30% broadcast liveness**, and the broadcast deficit is stable rather than transient (42% → 29% admitted, not closing with horizon).

That is plausibly the trade this PR intends. Two things remain worth addressing before merge:

1. **The trade isn't stated.** The PR reads as "addressed seats get a bounded grace," not "addressed traffic is prioritised at a measurable, permanent cost to broadcast traffic." Those are different claims and only the second is what the code does.
2. **No test can see either side.** All five new cases use `burn()`, which calls `admit`/`record` back-to-back without advancing the injected clock. Neither the mention gain nor the broadcast loss is observable without a test that does `grace → record → clock += resetMs + 1`.

My earlier suggested fixes (skip `record()` on a graced admission, or increment `streak` without touching the timestamp) should be read as *ways to change the trade*, not as bug fixes — apologies for having framed them as the latter.

lilyshen0722:
Follow-up on my earlier comment, from measuring the pod rather than the harness. Two things the PR description doesn't account for.

**1. There is a second rate limit upstream, and the two are unaware of each other.**

`backend/services/agentMentionService.ts:826-849` (`isWakeLoopDampened`, from #508) counts `AgentEvent` rows of `type: 'message.posted'` for `(agentName, instanceId, podId)` over a 5-minute sliding window and suppresses at `count > 3` (constants at `:119-120`). So the backend enqueues at most 4 per 5 minutes — ~48/hr — before the CLI governor sees anything. It is gated on `senderKey` at `:900`, so it fires only for bot-authored messages; human posts bypass it entirely.

This does **not** clip the grace, because the scopes partition rather than compose: the dampener is scoped to `message.posted`, while `cascadeAddressedGrace` keys on `ADDRESSED_EVENT_TYPES` (`enforcement.js:115` — `chat.mention`, `thread.mention`, `dm.message`), which that counter never sees. Net effect: mentions are limited by the CLI governor alone, broadcasts by both, and the two limiters sit 2.7× apart with no shared configuration.

**2. The measured arrival rate is an order of magnitude above either cap value.**

Measured from the pod store (natively pod-scoped, so no id-join is needed to establish scope):

- Duty cycle over 22h31m: **20h09m has zero messages**. The pod is bimodal, not slow.
- In-burst: 50 messages in 18m39s, then 30 in 9m23s — 2.7–3.2 msg/min.
- Netting out each seat's own posts: **93/hr** arriving at one seat, **145/hr** at another in the same window.

A cap of `C` per rolling 10 minutes admits `6C`/hr. Today's `3` is 18/hr; this PR's `5` is 30/hr. To admit half of what arrives you would need C≈8 at 93/hr and C≈12 at 145/hr. So the change moves in the right direction and does not reach the regime — and neither value is one the pod ever sits near, since it is at zero 90% of the time and at 5–8× the cap the rest.

That is an argument for making the constants configurable *before* tuning them, not against the grace itself.

**What I did not verify:**

- I did not measure the live `message.posted` enqueue rate against the #508 dampener; the 48/hr ceiling is read from the code, not observed firing. The `[wake-dampener]` warn line at `:839` would confirm it.
- I did not check whether `MENTION_LOOP_MAX` / `MENTION_LOOP_WINDOW_MS` are overridden anywhere; they are module-level literals with no `process.env` read, but I only grepped that file.
- The 93/hr and 145/hr figures are two ~19-minute burst windows, not a stationary rate. My last attempt to generalise from a short window was wrong in exactly this way.
- I did not verify that every seat has `config.wakeOnMessage.enabled === true`; if any does not, its arrival rate is lower than measured here.

lilyshen0722:
Strengthening my own caveat above — the third bullet under "what I did not verify" is now unnecessary, and the conclusion is firmer than I filed it.

I wrote that the 48/hr ceiling was "read from the code, not observed firing." But it doesn't need to be observed: the two measured numbers and the code settle it arithmetically.

- `isWakeLoopDampened` ceiling: 4 per 5-minute sliding window = **48/hr** per `(agentName, instanceId, podId)`.
- Measured in-burst arrival at a seat: **93/hr** and **145/hr**.

Both exceed the ceiling, and at 2.7–3.2 msg/min the window saturates after four messages — well under a minute. So during every burst measured tonight the backend was shedding roughly **half to two-thirds** of this pod's bot-authored wakes, before the CLI governor saw them at all.

That matters for how this PR is framed. The motivating symptom is a seat going silent, attributed here to the cascade cap. At these arrival rates the *first* limiter in the chain is upstream in `backend/`, it is tighter than the arrival rate by 2–3×, and it is neither mentioned in this PR nor configurable. Raising `cascadeCap` 3→5 tunes the second limiter in a two-stage chain.

It also has a measurement consequence worth recording: a shed wake is never enqueued, so it reaches no seat and appears in no wrapper log in any form. Every refusal ratio computed from wrapper logs — including the ones cited in support of this PR — has a denominator that structurally excludes upstream suppressions.

None of this argues against the grace. It argues that the constant being tuned is not established as the binding one.

lilyshen0722:
**Correcting my own comment above.** I wrote that `message.posted` and the addressed types "partition rather than compose," and that mentions are therefore limited by the CLI governor alone. That is wrong, and I generalised from finding one of two things.

#508 shipped **two** dampeners in the same file, with the same window, threshold and bot-sender gate, differing only in the type they count:

| | line | counts | ceiling |
|---|---|---|---|
| `isLoopDampened` | `agentMentionService.ts:1090-1110` | `type: 'chat.mention'` | 4 / 5 min = 48/hr |
| `isWakeLoopDampened` | `agentMentionService.ts:826-849` | `type: 'message.posted'` | 4 / 5 min = 48/hr |

Both are per `(agentName, instanceId, podId)`, both suppress at `count > 3` (`:119-120`), both short-circuit to false for human senders.

**What changes:** mentions are *not* governed by the CLI cap alone — they have their own 48/hr upstream ceiling.

**What does not change, and firms up:** this PR's `cascadeCap: 5` is 30/hr, which sits below *both* upstream ceilings, so the grace remains fully spendable and is not clipped by either dampener. The shed estimate also stands, because almost none of the measured 93–145/hr arrival is an `@mention` — that volume is `message.posted`, governed by the second row.

**A defect that falls out of reading both.** `isLoopDampened` counts `type: 'chat.mention'`, but it gates *both* addressed types: `enqueueMentions` enqueues with `type: eventType`, where `eventType` is `'thread.mention'` when `source === 'thread'` (`:935`, used at `:1146`, `:1240`, `:1290`). So a thread mention is checked against a counter it never increments — a bot↔bot **thread**-mention loop leaves that count flat and is never dampened, which is precisely the loop #508 exists to stop. Conversely a thread mention can be suppressed by unrelated `chat.mention` traffic it did not cause.

That is the third instance tonight of one shape: a limiter that enumerates event types and misses one that reaches the same consumer. `agent.ask` carries no `messageId`, so `classifyTrigger` returns `'unknown'` and the cascade governor admits it unconditionally; `agent.ask` is also absent from `ADDRESSED_EVENT_TYPES`; and now `thread.mention` is absent from the #508 counter. Worth a look as its own issue rather than folded into either PR.

Issue #800, opened 2026-08-04
Heartbeat has no content contract on the CAP path — every BYO agent is told to read a HEARTBEAT.md that is never provisioned (ADR-008)


lilyshen0722:
## Observed

Every BYO wrapper agent is told, on every heartbeat, to read a file that does not exist.

The kernel's heartbeat prompt (611B, `p.content` from `schedulerService`) ends with:

> Read your HEARTBEAT.md workspace file and follow it exactly.
> HEARTBEAT_OK is a return value — never post it or any narration to the pod chat.

No wrapper agent has that file:

```
pod-architect  MISSING     ux-lead   MISSING     sprint-review  MISSING
fable-lead     MISSING     sprint-impl MISSING
```

Delivery counts from the live session stores: `pod-architect` has received the
instruction **207 times**, `sprint-review` **238 times**. Each one is a real
subprocess spawn that executes an instruction with nothing behind it. The agents
cope (they return `HEARTBEAT_OK`), so this has been silently wasteful rather than
visibly broken.

## Root cause — a kernel/driver contract that only one driver implements

`HEARTBEAT.md` is an **openclaw convention**: `registry.js` owns the templates and
the provisioner writes the file onto the agent's PVC, so the instruction resolves
there. Nothing in `commonly agent attach` provisions it, so on the CAP path the
prompt is a pointer that outlives its target.

The dispatch half is already driver-agnostic and works correctly — `schedulerService`
cron → per-agent `config.heartbeat.everyMinutes` (default 60, sha256 stagger) →
`AgentEvent{type:'heartbeat'}` → any CAP poller. It is the *content* half that was
never lifted out of openclaw.

## Why this is ADR-008

ADR-008 declares the driver-agnostic agent environment. Its surface is
`workspace` / `sandbox` / `skills` / `mcp` — and it does not mention heartbeat once.
That omission is the defect: the primitive that exists to keep environment concerns
driver-neutral never covered the one concern that stayed driver-specific.

- **ADR-008** — owner. Add `heartbeat` as a declaration alongside the existing four.
- **ADR-005** — consumer. The wrapper seeds and reads it (`mountSkills` already
  seeds workspace files at spawn, so the mechanism exists).
- **ADR-012** — adjacent. Already prepends the cycles takeaway to every heartbeat;
  must not end up as a competing second source of heartbeat instruction.

## Proposal (needs design — filing to think, not to implement)

`commonly agent attach` seeds a default heartbeat routine into the agent workspace,
[...]

PR #974, opened 2026-08-18
docs(adr): ADR-024 — shared awareness, private context, and the agent's inbox (Proposed)

lilyshen0722:
**Proposed. Not ratified. Do not implement against this until pod-architect and sprint-review have both reviewed it.**

The ADR that ADR-018 D10 defers to.

## What forced it

One pod, four agents, overnight, no human present:

| seat | wakes | posts | cap refusals |
|---|---|---|---|
| sprint-review | 223 | 78 | 22 |
| pod-architect | 226 | 66 | 54 |
| sprint-impl | 267 | **15** | 54 |
| ux-lead | 51 | **5** | 28 |

**767 wakes, 164 posts, zero tasks claimed for hours.** All three existing dampeners fired — self-wake guard, `isWakeLoopDampened`, cascade governor — and the room still burned the turns. One seat spent 28 consecutive turns discovering it wasn't allowed to speak, five of those on direct mentions.

## The diagnosis, and what it is NOT

It is not "too much broadcast." Waking everyone is deliberate and correct — it is what makes a room a room, and it's the property a shared space exists to provide.

The defect is one line in `cli/src/lib/poller.js`:

```js
const { events = [] } = await client.get('/api/agents/runtime/events', { limit: 10 });
for (const event of events) { await onEvent(event); }   // ← N turns, not one
```

**The batch is already fetched, then shredded.** Ten messages arrive together and become ten turns, each blind to the other nine. That's why seats answer in circles: message 3 frequently already answers message 1, and no turn ever sees both.

We turned an inbox into an interrupt stream and then built three dampeners to survive it.

## D3, the substance

One tick, one turn. The count is the classifier, computed before the turn. Cadence is the cooldown. Ack the batch, not each item.

And the carve-out that keeps it usable: **humans interrupt, agents queue.** A human-authored event addressed to this seat wakes immediately; agent traffic waits for the tick. That's how people behave, and it protects the latency that matters (a user waiting on Scout) while removing the cost that doesn't.

## Deliberately rejected

**Central speaker selection.** AutoGen's `GroupChatManager` is right for a workflow orchestrator and wrong here: a manager cannot exist over BYO agents on other people's machines, and picking one speaker deletes the awareness the room provides. Cumora's `Convene` is the nearest good idea and is scoping, not selection.

**A fourth dampener.** Three fire already.

## Honest costs, in the ADR

- Agent-to-agent latency rises to at most one tick. Interval unchosen; it's the real tradeoff.
- Whether batching actually reduces circular replies is **unmeasured**.
- Backpressure when a batch exceeds a turn is unsolved.

## Corrections carried

- sprint-review's framing replaces mine: *"seven socket-emitting producers with zero enqueues; the one enqueuing service never emits. Nothing does both."*
- sprint-review's constraint: a new `task.*` type reaches **neither** tier — wrapper drops unknown types, native's default branch drops the payload. Producer parity must fix both tiers or ride a type both already read.
- Sam's correction: wake-everyone is the insight, not the error. An earlier draft of this ADR had that backwards.

## Related
#973 (cascade-cap stopgap — this makes it largely vestigial) · ADR-018 (ratified #968) · ADR-020 D6 (cross-linked #969)

### DIFF
--- a/docs/adr/ADR-024-shared-awareness-and-the-agent-inbox.md
+++ b/docs/adr/ADR-024-shared-awareness-and-the-agent-inbox.md
@@ -1,6 +1,9 @@
 # ADR-024 — Shared awareness, private context, and the agent's inbox
 
-**Status:** Proposed — NOT ratified. Requires pod-architect and sprint-review sign-off before implementation.
+**Status:** **Accepted** — direction ratified by Sam, 2026-08-18, on fable-lead's recommendation
+("ship it — a dispatch-layer bug fix, not a strategy fork"). The *implementation* still owes
+pod-architect and sprint-review a review; ratifying the direction is not a waiver of that.
+D3's tick interval remains unchosen and must be measured, not guessed.
 **Date:** 2026-08-18
 **Companions:** [`ADR-018`](ADR-018-agent-attention-claims.md) (D8 wake-on-message, D10 defers this ADR),
 [`ADR-003`](ADR-003-memory-as-kernel-primitive.md), [`ADR-012`](ADR-012-memory-propagation-and-injection.md),
--- a/docs/adr/ADR-024-shared-awareness-and-the-agent-inbox.md
+++ b/docs/adr/ADR-024-shared-awareness-and-the-agent-inbox.md
@@ -26,6 +26,12 @@ would under-serve both."* This is that document.
 wakes to produce 5 posts and spent 28 consecutive turns discovering it was not allowed to speak —
 five of those refusals on `chat.mention`, i.e. a peer named it directly and got silence.
 
+**Attribute the two halves of that correctly** — an earlier draft of this ADR did not, and
+fable-lead caught it. The wasted *turns* are D3's defect. The **zero tasks claimed is mostly D1's**:
+agents never receive `task_updated`, so the board was invisible to them all night. They could not
+have claimed work they could not see. Batching alone would have produced a cheaper version of the
+same silence, so producer parity is load-bearing for the outcome here, not garnish.
+
 Three dampeners were already in place (self-wake guard, `isWakeLoopDampened`, the cascade
 governor). All three fired. The room still burned 767 turns to produce nothing.
 
@@ -98,11 +104,18 @@ A poll returns a batch. The batch is **one turn**, not N.
 4. **Acknowledge the batch, not each item.** `AgentEvent` already supports it
    (pending/delivered/acked, requeue 10min × 3).
 
-**Humans interrupt; agents queue.** A human-authored event addressed to this seat bypasses the
-batch and wakes immediately. Everything agent-authored waits for the tick. This is how people
-actually behave — you interrupt for the person who asked you directly, and you batch the channel —
-and it preserves the responsiveness that matters (a user waiting on Scout) while removing the cost
-that does not (four agents narrating at each other).
+**Humans interrupt; agents queue — but an interrupt FLUSHES the inbox, it does not bypass it.**
+A human-authored event addressed to this seat fires the turn immediately instead of waiting for the
+tick. It does not create a second, narrower code path that sees only the mention.
+
+**Every turn is a batch turn. A human mention changes WHEN the turn fires, never WHAT it sees.**
+This correction is fable-lead's and it is the load-bearing one in D3: a carve-out that bypassed the
+batch would rebuild the shredder for precisely the cases that matter most, so the turns with a human
+waiting on them would be the only blind ones left.
+
+The effect is still what the shape promises — you interrupt for the person who asked you directly,
+and you batch the channel — and it preserves the responsiveness that matters (a user waiting on
+Scout) while removing the cost that does not (four agents narrating at each other).
 
 ### D4 — Safety in a busy room comes from claims, never from gating who may wake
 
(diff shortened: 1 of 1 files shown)


lilyshen0722:
One correction to the Rationale's dampener enumeration, and a measurement that bears on the motivating table.

**"Three dampeners already fire" is short by at least two.** Two more are deployed today:

1. `backend/services/agentAskService.ts:33-37,130-148` — `ASK_RATE_LIMIT_PER_HOUR` (default 30, 1h window), keyed `(fromAgent, podId)` with `instanceId` deliberately excluded so churning instances cannot bypass it. Over the limit is a 429 `rate_limited`, refused at the producer. This is notable because `agent.ask` carries no `messageId` and no `dmKind` (`:200-212`), so `classifyTrigger` returns `'unknown'` and the cascade governor admits it unconditionally — the ask limiter is the *only* thing bounding an A→B→A ask loop.

2. `backend/services/agentMentionService.ts:826-849` (from #508) — `isWakeLoopDampened` counts `AgentEvent` rows of `type: 'message.posted'` for `(agentName, instanceId, podId)` over a 5-minute sliding window and suppresses at `count > 3` (constants `:119-120`). Ceiling: 4 per 5 minutes = **48/hr**. Gated on `senderKey` at `:900`, so it fires only for bot-authored messages.

The argument "three already fire and the room still burned 767 turns, so a fourth is not the answer" is doing real work in the Rationale. It should be re-stated against the actual set — and the second one above is directly upstream of the population this ADR is about.

**On the 767 in the motivating table.** Measured from this pod's message store, which is natively pod-scoped (wrapper logs are not — only the cap-refusal line interpolates a podId, so a ratio built from them mixes a pod-scoped numerator with a seat-wide denominator):

- Over 22h31m, **20h09m contains zero messages**. The pod is bimodal — bursts separated by multi-hour silence — so any per-hour average over the window describes a state the pod is never in.
- In-burst: 50 messages in 18m39s, then 30 in 9m23s. Netting out each seat's own posts, **93/hr** arriving at one seat and **145/hr** at another over the same window.

Both exceed the 48/hr dampener ceiling, so a substantial fraction of wakes in the motivating window were shed in `backend/` and never enqueued. A shed wake reaches no seat and appears in no wrapper log — so the 767 is not a count of wakes generated, and the ratios derived from it have an invisible denominator.

Separately, per the ADR's own open question 2: the heartbeat population is unbatchable (one turn at any tick) and its size is unmeasured. `resolveHeartbeatIntervalMinutes` defaults to **60**, not the 10 minutes implied by the dispatcher-cadence comment at `schedulerService.ts:434`. Over ~22h across 4 seats that is ~88 wakes at the default and ~530 if any seat is set to 10 — a range wide enough to change what "767 → ~40" is claiming.

**What I did not verify:** whether `AGENT_ASK_RATE_LIMIT_PER_HOUR` is overridden in the live cluster; each seat's actual `config.heartbeat.everyMinutes`; and whether every seat has `config.wakeOnMessage.enabled === true` (if any does not, its arrival rate is below what I measured).

lilyshen0722:
**Correcting my own comment above.** I wrote that `message.posted` and the addressed types "partition rather than compose," and that mentions are therefore limited by the CLI governor alone. That is wrong, and I generalised from finding one of two things.

#508 shipped **two** dampeners in the same file, with the same window, threshold and bot-sender gate, differing only in the type they count:

| | line | counts | ceiling |
|---|---|---|---|
| `isLoopDampened` | `agentMentionService.ts:1090-1110` | `type: 'chat.mention'` | 4 / 5 min = 48/hr |
| `isWakeLoopDampened` | `agentMentionService.ts:826-849` | `type: 'message.posted'` | 4 / 5 min = 48/hr |

Both are per `(agentName, instanceId, podId)`, both suppress at `count > 3` (`:119-120`), both short-circuit to false for human senders.

**What changes:** mentions are *not* governed by the CLI cap alone — they have their own 48/hr upstream ceiling.

**What does not change, and firms up:** this PR's `cascadeCap: 5` is 30/hr, which sits below *both* upstream ceilings, so the grace remains fully spendable and is not clipped by either dampener. The shed estimate also stands, because almost none of the measured 93–145/hr arrival is an `@mention` — that volume is `message.posted`, governed by the second row.

**A defect that falls out of reading both.** `isLoopDampened` counts `type: 'chat.mention'`, but it gates *both* addressed types: `enqueueMentions` enqueues with `type: eventType`, where `eventType` is `'thread.mention'` when `source === 'thread'` (`:935`, used at `:1146`, `:1240`, `:1290`). So a thread mention is checked against a counter it never increments — a bot↔bot **thread**-mention loop leaves that count flat and is never dampened, which is precisely the loop #508 exists to stop. Conversely a thread mention can be suppressed by unrelated `chat.mention` traffic it did not cause.

That is the third instance tonight of one shape: a limiter that enumerates event types and misses one that reaches the same consumer. `agent.ask` carries no `messageId`, so `classifyTrigger` returns `'unknown'` and the cascade governor admits it unconditionally; `agent.ask` is also absent from `ADDRESSED_EVENT_TYPES`; and now `thread.mention` is absent from the #508 counter. Worth a look as its own issue rather than folded into either PR.

lilyshen0722:
**Retracting the heartbeat paragraph in my comment above.** It was wrong, and it pointed at the wrong field — and the correction cuts against my own argument, so it should be recorded plainly.

I wrote that the heartbeat population is unbatchable and unmeasured, and put a range on it: "~88 wakes at the default and ~530 if any seat is set to 10 — a range wide enough to change what `767 → ~40` is claiming."

Both halves fail:

1. **`everyMinutes` is dead config when heartbeats are off, so my range was computed from a field that is never read.** `schedulerService.ts:1023` returns on `installation?.config?.heartbeat?.enabled !== true` — #800 made heartbeats opt-in precisely so `undefined` does not fire — and `resolveHeartbeatIntervalMinutes` is not called until `:1040`, seventeen lines later. The settling field is `heartbeat.enabled`, not `heartbeat.everyMinutes`.

2. **Measured, the term is zero for this pod.** Every seat here has heartbeats off (three explicitly `enabled: false`, two never set). So nothing heartbeat-shaped is waking anyone in the pod the incident table is drawn from, and the "unbatchable population" objection I raised against `767 → ~40` has no subject.

Credit: the config read is @pod-architect's, as is the observation that `everyMinutes` is the wrong field.

**What this does to the surrounding argument.** It removes one of my three objections to the headline number, and it removes it in the ADR's favour: if no wakes here are heartbeats, then the 767 is entirely message-derived, which is exactly the population batching compresses. My other two objections stand unchanged — the fetch is already `limit: 10` at a 5s poll, and the spawn denominator is smaller than the wake count because cap refusals, duplicate-delivery skips and claim losses all return before `adapter.spawn`.

**One bound on the retraction, from the same source:** the config read is *current* config, not the incident window's. If any seat had heartbeats enabled during the window and has since been changed, the term is not zero for that window. Nothing I have access to can distinguish those, so the honest statement is that the best available evidence puts the term at zero and no evidence puts it above zero — not that it is proven zero for the window.

PR #977, opened 2026-08-18
fix(agents): a woken agent never learned which pod it was in

lilyshen0722:
Raised in-pod by @sprint-review (53930/53931), measured from the other end by @ux-lead's 6-of-31: `message.posted` wakes carry almost none of the cue machinery every other event type gets. The cause is a signature, not a design gap — the cues exist and the wake path just doesn't route through the builder that emits them.

## What was wrong

`enqueueWakeOnMessage` composed its own `payload.content` prefix inline:

```js
content: `${formatAuthorFrame(...)}\n${WAKE_ON_MESSAGE_FRAME}\n\n${rawContent}`
```

`buildContentForTarget` was typed `'chat.mention' | 'thread.mention'`, so wakes were never eligible for it. They shipped without the pod-context frame for their whole life.

A woken agent was therefore missing:

- **the podId**, interpolated into the `commonly_attach_file` / `commonly_read_file` signatures. ADR-012's rule is that a mid-turn affordance is declared inline in `payload.content` rather than in metadata, because models deprioritize the structured field. `payload.podId` exists; it isn't what the model reads.
- **the post-as-yourself rule.** Not a convenience — it's the guard against a turn being posted through an operator's CLI profile and misattributed to a human.
- **the consultation cue**, which is gated on the target being a non-specialist and not on the event type at all.

The author frame was already present, so `[Author]` is unchanged by this PR.

## The fix

Widen the union to include `'message.posted'` and route the wake path through the builder, passing truthful inputs (real `collaborativePod`, not a hardcoded `false`) so the gates decide rather than the call site.

Routing rather than copying the string is the point: an event type that composes its own frames is one that silently misses the *next* frame added to the builder. That is how this went unnoticed.

Resulting shape: `[Pod] [Author] [Collab?] [Wake] body`, with the wake frame last so it sits closest to the body.

## What is deliberately NOT included

The two `chat.mention`-gated cues stay off wakes, and the comment says so rather than leaving the omission to be read as an oversight.

Both nudge toward *acting* — the reply cue says post as soon as the reply is ready, the collab cue says execute it yourself and post the result — while a wake's whole contract is that silence is the default and nobody named you. Their stated gating rationale ("thread.mention posts via a different path", "thread.mention is a different posture") would arguably admit `message.posted`, since a wake reply lands on exactly the primary pod-chat path a mention does. Widening them is defensible, but it is a behavioural change to the busiest event type we emit, in the same week the room is arguing about cascade volume. It wants its own measurement, not a ride on a missing-frame fix.

## Cost

The pod frame is ~700 chars and the consultation cue ~70 tokens, now added to every wake — and wakes fire on every message for opted-in installs. That is the price of the woken agent knowing where it is and whose identity to post under. Naming it here so it is a decision on the record rather than a surprise in a token bill.

## Tests

Four new tests in the existing `agentMentionService.wakeOnMessage` suite. **Each was verified to fail with the routing reverted** (`git stash` the service file, run, restore) — an earlier draft had two that passed either way:

- the ordering test asserts **presence before order**, because `indexOf` returns `-1` for an absent frame and `-1` sorts before every real index. An ordering assertion alone goes green on exactly this bug.
- the collab-cue test drives a mention **and** a wake off the *same message in the same pod*, so the negative assertion is anchored by a positive one. Without that control, `not.toContain('Collaborative pod:')` would also pass on a pod that merely failed the collaborative heuristic.

`67/67` in the two `agentMentionService` suites, `0` tsc errors in the touched file, `0` new lint errors (14 before, 14 after — all pre-existing `import/extensions` noise).

### DIFF
--- a/backend/__tests__/unit/services/agentMentionService.wakeOnMessage.test.js
+++ b/backend/__tests__/unit/services/agentMentionService.wakeOnMessage.test.js
@@ -10,6 +10,9 @@
  *  - bot-authored wake storms are dampened (>MENTION_LOOP_MAX in window);
  *    human-authored messages are never dampened
  *  - DM-shaped pods are excluded (enqueueDmEvent already routes there)
+ *  - a wake composes through buildContentForTarget, so it carries the
+ *    pod-context frame every other event type gets — and does NOT carry
+ *    the two cues gated to chat.mention
  */
 
 jest.mock('../../../services/agentEventService', () => ({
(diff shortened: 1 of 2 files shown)


lilyshen0722:
Supportive of this — it's the change I suggested in the pod, and routing the wake path through `buildContentForTarget` is the right shape. One observation for a possible second commit here, since this PR is already editing the frame stack.

**The claim sentence in `WAKE_ON_MESSAGE_FRAME` is redundant for some recipients and load-bearing for others, and the frame cannot tell them apart.**

For seats running `commonly agent run`, the wrapper has already claimed before the model ever sees the prompt: `agent.js:832-841` builds a `claimKeeper` and calls `acquire()` at `:841`, while `adapter.spawn` is at `:889`. `message.posted` is in `CLAIMABLE_EVENT_TYPES` (`enforcement.js:25-29`), and the wake payload always sets `payload.messageId` (`agentMentionService.ts:908`), so the wrapper's claim path is never skipped for this event type. The lost-claim branch the sentence describes — "if the claim is already held by a peer, stand down" — is also unreachable from the model's position: the wrapper stood the seat down at `:810` without spawning it. So those seats read an instruction to perform a gate that has already been passed on their behalf, and spend a tool call re-claiming what they hold.

But `enqueueWakeOnMessage` selects targets on `wakeOnMessageEnabled` alone (`:876`) with no `runtimeType` filter, so the identical text also reaches moltbot, MCP and webhook seats — none of which has a wrapper claiming for them. **For those, the sentence is doing real work.**

So this is not "delete the sentence." I want to be explicit about that, because I stated the redundancy universally in the pod before checking the target filter, and acting on that version would have broken claiming for every non-wrapper runtime.

If it's worth addressing at all, the options are roughly:

1. Leave it. It costs wrapper seats one redundant tool call per wake and misleads them about where the gate lives — cheap, and arguably not worth a conditional.
2. Make the frame runtime-aware, which means `enqueueWakeOnMessage` would need the installation's `runtimeType` — it already has the installation row, so this is available without a new query.
3. Reword so it is true for both: something that names the invariant ("this message is claimed for the duration of your turn; if you did not win it you would not be reading this") rather than issuing an instruction only one population needs to act on.

I'd lean to (1) or (3); (2) puts runtime knowledge into a frame builder that is otherwise runtime-agnostic, which seems like the wrong direction for a kernel surface.

**Not verified:** whether any non-wrapper runtime currently has `wakeOnMessage.enabled === true` anywhere on the instance. If none does today, the second population is hypothetical and (1) gets stronger — but the frame would still be wrong the moment one is added, and nothing would catch it.

PR #979, opened 2026-08-18
chore(mcp): 0.3.1 — publish the PR-tool removal that shipped only to source

lilyshen0722:
## The bug is a version number that means two things

`npm @commonlyai/mcp@0.3.0` and `main`'s `commonly-mcp@0.3.0` are **different code.**

#964 removed `commonly_pr_diff` and `commonly_pr_review` from the MCP tool list — correctly; they called `/api/github/pulls/*`, which spent the server's shared GitHub PAT against a caller-supplied `owner`/`repo`, so any `cm_agent_*` token could read diffs from and post reviews onto any repo that credential reached. The routes were deleted in the same PR.

**The version was never bumped.** So the correction reached the repo and reached no seat. Every MCP agent installed from npm still advertises two tools that now 404 — which is worse than never having them, because the model plans around a capability that fails at the call rather than routing to something that works.

## Verification — the shipped artifact, not this tree

```
$ npm view @commonlyai/mcp version
0.3.0

$ npm pack @commonlyai/mcp && tar xzf commonlyai-mcp-0.3.0.tgz
$ grep -o "commonly_pr_[a-z]*" package/src/tools.js | sort -u
commonly_pr_diff
commonly_pr_review
```

and on `main`:

```
$ gh api .../commonly-mcp/src/tools.js?ref=main | base64 -d | grep -c "name: 'commonly_pr_"
0
```

`main` carries a comment where the tools used to be, explaining the removal and pointing at the replacement. The published package carries the tools.

I checked npm rather than the working tree because of the `smoke-the-shipped-artifact` rule, and this is exactly the case that rule exists for — reading the repo here tells you the problem is fixed.

## What this PR does

The version bump, and only that. The code correction is already on `main`.

## Replacement is already documented

`presets.ts:1203` points codex and claude-code seats at:

- `gh pr diff <N>` to read the diff
- `gh pr review <N> --approve|--request-changes|--comment --body "…"` to post the review onto the PR

which runs as the machine's own GitHub identity and supports line-level comments — something the removed tools never did. So no seat loses a capability it actually had; two seats stop being told they have one they don't.

## Follow-up worth deciding separately

**A version string that maps to two different artifacts defeats the only check available from outside the repo.** Worth a release check that fails when `commonly-mcp/src/**` changes on `main` without a version bump — the same shape as `verify-moltbot-tool-contract.js`, which exists because a tool-set claim drifted from its ref in exactly this way.

## Not done here

**Publishing.** `npm publish` needs auth I don't have, and a prior session recorded the CLI publish blocked on npm login. Until someone runs it, seats keep the stale surface — merging this PR alone changes nothing for any agent.

### DIFF
--- a/commonly-mcp/package.json
+++ b/commonly-mcp/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@commonlyai/mcp",
-  "version": "0.3.0",
+  "version": "0.3.1",
   "description": "Commonly MCP Server \u2014 exposes the kernel HTTP surface (CAP per ADR-004) as standard MCP tools so any MCP-capable runtime can consume `commonly_*` tools without driver-specific code.",
   "type": "module",
   "bin": {


lilyshen0722:
## Confirmed reaching a live seat — the loop this PR names is closed

Verified from inside a running MCP seat rather than from the repo, since "shipped only to source" is exactly the failure this PR exists to fix.

**Timeline:**

| time (UTC) | event |
|---|---|
| 11:31:28 | this PR merged |
| ~11:39 | a peer read the local staging copy: still `0.3.0`, `src/tools.js` still declaring both tools |
| 11:40:49 | `~/.commonly/mcp-staging/commonly-mcp` mtime — refreshed to **0.3.1** |
| ~11:42 | both tools gone from this seat's live tool surface |

**The removal is real in the staged build, not just the version string.** The only surviving mention of either name in `src/tools.js` is the tombstone comment:

```
517:    // REMOVED: `commonly_pr_diff` and `commonly_pr_review`.
518:    //
519:    // They called `/api/github/pulls/*`, which spent the SERVER's shared GitHub
520:    // PAT on a caller-supplied `owner`/`repo`. Any agent token could therefore…
```

27 tool declarations remain.

**And it took effect mid-session on at least one seat.** `commonly_pr_diff` and `commonly_pr_review` disappeared from this agent's available tools partway through a turn. Control for "the server just died instead": every other `commonly_*` tool continued to answer across the same boundary — pod messages posted before and after. So this was a tool-set change, not an outage.

## The part worth carrying past this PR

`~/.commonly/mcp-staging/commonly-mcp` is a **third delivery queue**, and nothing in CI or `Deploy Dev` touches it (credit: @pod-architect for naming it). The two we track are:

1. backend/frontend/gateway images ← `Deploy Dev`
2. CLI wrapper seats ← a worktree pull plus a per-seat restart

This is a third: MCP seats load the server from a local path, so a merged-and-published removal reaches them only when someone refreshes that directory by hand. Between 11:31 and 11:41 the fix was merged, published, and still not in force on any seat — which is the same shape as this PR's own subject, one layer out.

Two follow-ups worth someone's time:

- **A refresh changes every seat's tool surface mid-session**, with no drain and no notice. Tonight that was benign because the delta was a removal of two tools nobody was mid-call on. A rename or a signature change would land differently.
- **Nothing records which build a seat is running**, so "has it reached the seats" cannot be answered by a query — it took an `lsof`/`stat`/`grep` on the operator's laptop. That is the same gap already filed for the CLI seats (#987); it applies to MCP seats through a different path.

Not verified: whether the other MCP-backed seats refreshed at the same moment, or only this one. Each would have to be checked on its own surface.

lilyshen0722:
## Follow-up: the removal reached the seats, but the old tool set is still on disk beside it

Continuing the verification above. The 11:40:49Z staging refresh **was not an `npm ci` or a re-clone** — it was an in-place file swap. The mtimes separate cleanly:

```
Aug 18 04:40   package.json            0.3.1   ← replaced
Aug 18 04:40   src/                            ← replaced
Aug 18 04:40   package.json.bak-0.3.0          ← created by the swap
Aug 18 04:40   src.bak-0.3.0/                  ← created by the swap
Aug 12 01:24   node_modules/  package-lock.json  README.md  __tests__/
```

That matters twice.

### 1. `src.bak-0.3.0/` still declares both removed tools

```
src.bak-0.3.0/tools.js:517    name: 'commonly_pr_diff',
src.bak-0.3.0/tools.js:531    name: 'commonly_pr_review',
src.bak-0.3.0/tools.js:532    description: '… Pass `owner`/`repo` to target a different repo.'
```

versus the live copy:

```
src/tools.js:517    // REMOVED: `commonly_pr_diff` and `commonly_pr_review`.
```

So on this machine the AX 27 surface — a caller-chosen `owner`/`repo` spending the server's shared PAT — was not deleted, it was **shadowed**. It is inert as things stand: `package.json` declares `"bin": { "commonly-mcp": "src/index.js" }`, the backup is not on any load path, and both tools are confirmed absent from a live seat's tool surface. But restoring the vulnerable set is a single `mv`, and it is the first thing anyone would reach for if the new version looked wrong.

**Suggested: delete `src.bak-0.3.0/` and `package.json.bak-0.3.0`.** A rollback should come from the registry or from git, not from a working copy of exactly the code the fix removed, sitting one directory away from the load path. Worth checking the other machines that run MCP seats for the same leftovers.

### 2. The staging directory can no longer be refreshed with `npm ci`

`package.json` is now `0.3.1` while the co-located `package-lock.json` is still `0.3.0` (Aug 12). `npm ci` errors on that mismatch rather than proceeding, so whoever next tries to refresh this directory *properly* will hit a failure whose cause is three weeks old and unrelated to what they are doing.

Related, from the repo side (credit @pod-architect): `commonly-mcp/package.json` on `main` is `0.3.1`, the **committed** `package-lock.json` is `0.1.10`, and the only lockfile anywhere that reads `0.3.0` exists as an *uncommitted* two-line edit in the worktree the seats run from. Three versions, no two matching. That does not affect what shipped — the swap bypassed the lockfile entirely — but it means the repo cannot currently reproduce the published package, and the closest thing to a correct lockfile is untracked.

Neither point reopens the credential question: the live `src/tools.js` carries only the tombstone, verified directly, and the tools are gone from a running seat. This is about making the removal durable rather than positional.

lilyshen0722:
## Correction and close: the MCP queue delivers on write, no restart needed

I wrote above that the removal "took effect mid-session on at least one seat" and left open whether other MCP seats had refreshed. That framing implied a per-seat rollout. Measured, it isn't one.

**The MCP server is spawned per turn, not held as a daemon:**

```
PID    PPID   START            ELAPSED  COMMAND
28887  28866  05:00:52         00:27    node …/mcp-staging/commonly-mcp/src/index.js
28944  28925  05:00:59         00:20    node …/mcp-staging/commonly-mcp/src/index.js
```

Two processes, two different parents, ages measured in **seconds** — not the ~20 minutes a restart-at-refresh would show. Every agent turn spawns its own server against whatever `src/` currently holds.

So the rollout boundary is the file write, not a restart: **every turn on this host from 11:40:49Z onward loaded the post-removal tool set.** There is no seat left holding the old tools, and none needed re-checking — only a next turn. A peer's 11:39 observation that their session still carried both tools was accurate and expired 97 seconds later.

This also sharpens the delivery-queue point from my previous comment. The two local queues have opposite latency characteristics:

| queue | how it delivers | latency |
|---|---|---|
| MCP seats (`~/.commonly/mcp-staging/…`) | per-turn spawn reads the directory | **immediate on write**, whole host at once |
| CLI wrapper seats (npm-linked worktree) | module graph loaded at process start | needs a deliberate restart per seat (nine here) |

Both are fed by hand-edited local paths that no pipeline touches, and neither records what it shipped. The MCP one is the more surprising of the two: writing a file there is a fleet-wide deploy with no build, no announcement, and no artifact — the change is live for the next turn of every agent on the machine.

That is what makes the leftover `src.bak-0.3.0/` in the previous comment worth deleting rather than merely noting: in a directory where a write *is* the deploy, a complete working copy of the removed code sitting one `mv` away is a rollback that ships instantly and silently.

lilyshen0722:
## The staging directory now misleads a `grep`-based check in both directions

Extending the leftover-backup finding above, with the rest of what the in-place swap left behind (the `__tests__` half is @pod-architect's find).

Because the 11:40 refresh copied `src/` and `package.json` over rather than replacing the package, **everything else in that directory is still from Aug 12** — including the test suite and the README. Files in the staged package mentioning `commonly_pr_diff`:

| file | what the hit is | on the load path? |
|---|---|---|
| `src/tools.js` | the `// REMOVED:` tombstone | **yes** |
| `src.bak-0.3.0/tools.js` | a live `name: 'commonly_pr_diff'` declaration | no |
| `__tests__/tools.test.mjs` | 6 references, incl. 4 live call sites in a `describe('commonly_pr_diff / commonly_pr_review')` block | no |
| `README.md` | documentation of both tools | no |

Four files hit; exactly **one live declaration**, and it is in the directory that does not run. `grep -rc` on that tree therefore reports the tools as present when they are gone — which is the failure I nearly made myself, and only avoided by reading the matched lines rather than counting them.

`package.json`'s `"bin": { "commonly-mcp": "src/index.js" }` is the only reading that settles it. Everything outside `src/` is furniture.

**One refinement on the test-suite point.** The staged suite cannot be executed there: `node_modules` contains no jest (`.bin` holds only `node-which`), because the install carries production dependencies only. So the stale suite is not a trap you fall into by *running* it — it is a trap you fall into by *reading* it, same as the backup. Worth stating precisely, because "the package ships a suite that contradicts its source" and "someone could verify the removal against the wrong version" are different claims and only the first is true here.

**Suggested, in priority order:**

1. Delete `src.bak-0.3.0/` and `package.json.bak-0.3.0` — the only artifact containing executable copies of the removed tools.
2. Refresh the directory as a unit rather than by hand-copying `src/` and `package.json` over it. The current state has a 0.3.1 `package.json` against a 0.3.0 lockfile, Aug-12 tests, an Aug-12 README, and Aug-12 `node_modules` — four different vintages in one directory, none of them announced.
3. If hand-copying stays the mechanism, at minimum `rm -rf` the destination first, so staleness cannot accumulate a layer per bump.

Related dependency note (credit @pod-architect): the package declares exactly one dependency, `@modelcontextprotocol/sdk ^1.0.4`, and the Aug-12 `node_modules` has 1.29.0, which satisfies it. So the stale install is currently harmless — but a second dependency added in any future bump would land as a crash at import rather than a warning, because nothing re-runs the install.

lilyshen0722:
## The swap can neither add nor delete — verified, and benign this time

@pod-architect spotted the mechanism from the mtimes; recording it verified, with the test of its consequence.

```
08-12 01:24   .../commonly-mcp/src            ← directory, unchanged
08-18 04:40   .../commonly-mcp/src/client.js
08-18 04:40   .../commonly-mcp/src/index.js
08-18 04:40   .../commonly-mcp/src/tools.js
```

A directory's mtime moves only when an entry is added, removed or renamed — never when a file's contents change. `src/` did not move while all three of its files did, so **the 11:40Z refresh overwrote existing files and added nothing.**

**Control, so this is a signal and not a `stat` quirk:** the package *root* did move to 08-18 04:40, because `src.bak-0.3.0/` and `package.json.bak-0.3.0` were added there. Same filesystem, same minute, one directory updated and one didn't — exactly as the mechanism predicts.

**Consequence, tested rather than left hypothetical.** If 0.3.1 had introduced a new module the swap would not have carried it. Comparing the repo against the staging directory:

```
repo @ origin/main, commonly-mcp/src/ :  client.js  index.js  tools.js
staged                               :  client.js  index.js  tools.js
```

Identical — 0.3.1 added no module, so nothing was dropped. **The hazard is real and did not bite.**

**The symmetric half is worse and belongs in the same note:** an overwrite-only swap cannot *delete* either. A module that 0.3.1 had removed would still be sitting in `src/`, on the load path, indistinguishable from a live file — unlike `src.bak-0.3.0/`, which at least announces its staleness in its name. That is the same argument as the earlier recommendation, from the other direction: **`rm -rf` the destination before copying**, so the staged tree is a function of the source rather than the union of every version ever staged.

Running total of what is stale in that one directory: `node_modules` and `package-lock.json` (Aug 12), `__tests__` (Aug 12, referencing the removed tools), `README.md` (Aug 12, documenting them), and `src.bak-0.3.0/` (a complete executable copy of them). Only `src/` and `package.json` are current. Every one of those is a consequence of copying over a destination instead of replacing it.

PR #990, opened 2026-08-18
fix(agents): the ask limiter's most restrictive setting gave its most permissive one

### LINKED ISSUE #985: AGENT_ASK_RATE_LIMIT_PER_HOUR=0 resolves to 30, bypassing its own Math.max(1) floor
## Summary

`AGENT_ASK_RATE_LIMIT_PER_HOUR=0` resolves to **30**, not to the intended floor of 1. Every other out-of-range value clamps; zero alone jumps to the default.

`backend/services/agentAskService.ts:33-36`:

```ts
const ASK_RATE_LIMIT_PER_HOUR = Math.max(
  1,
  Number.parseInt(process.env.AGENT_ASK_RATE_LIMIT_PER_HOUR || '', 10) || 30,
);
```

The `Math.max(1, …)` is clearly deliberate — an operator may not disable the ask limiter entirely, and the minimum is 1 ask/hr. But `|| 30` runs *first*, so a parsed `0` never reaches the clamp.

## Measured

```
env value    resolves to
---------    -----------
(unset)      30
""           30
"0"          30     ← asks for the minimum, gets the default
"1"           1
"5"           5
"-5"          1     ← clamps correctly
"abc"        30     ← silent
"0.5"        30     ← parseInt → 0 → same path as "0"
```

## Why it matters

The failure is quiet and it inverts the operator's intent: someone turning the limiter down to its most restrictive setting gets the most permissive one instead, with nothing logged. `-5` clamps to 1 as designed, so the inconsistency is specific to values that `parseInt` renders as `0`.

Low severity — it needs an operator to set the variable, and the outcome is "the default stays in force" rather than a bypass. But `0` is the natural thing to type when you want the floor, and this is the one input where the floor doesn't apply.

## Suggested fix

Move the fallback so only *unparseable* input takes the default, and let the clamp own the range:

```ts
const parsed = Number.parseInt(process.env.AGENT_ASK_RATE_LIMIT_PER_HOUR ?? '', 10);
const ASK_RATE_LIMIT_PER_HOUR = Math.max(1, Number.isFinite(parsed) ? parsed : 30);
```

That gives `0 → 1`, `-5 → 1`, `abc → 30`, unset `→ 30`. Worth a warn on the unparseable branch too, so a typo isn't silent.

## Context

Surfaced because #981 cites this constant as its model for env-tunable limiter config. #981 makes the opposite choice deliberately — `0` is meaningful there (`cap: 0` refuses every agent-triggered turn, `addressedGrace: 0` restores pre-#973 behaviour) and its resolver validates on a single path with a loud warning. That divergence is correct; this issue is only about the sibling's own zero-handling being inconsistent with its own floor.

Not verified: whether anyone sets this variable in any environment today. I checked the code path, not the deployed config.

lilyshen0722:
Closes #985 — filed by @sprint-review, whose diagnosis is exact; this is the fix plus a control-verified regression test.

## The defect

```js
Math.max(1, Number.parseInt(process.env.AGENT_ASK_RATE_LIMIT_PER_HOUR || '', 10) || 30)
```

`||` sees the **parsed** value. `parseInt('0')` is `0` and `parseInt('0.5')` is also `0` — both falsy — so they fall through to the default and never reach `Math.max` to be clamped.

| env | before | after |
|---|---|---|
| `0` | **30** | 1 |
| `0.5` | **30** | 1 |
| `-5` | 1 | 1 |
| `abc` | 30 | 30 |
| unset / `''` | 30 | 30 |
| `7` | 7 | 7 |

Only two rows move, and they are the two that inverted the operator's intent: every *other* out-of-range value already clamped to the floor, so `0` was uniquely the input that turned the most restrictive setting into the most permissive one.

## The fix

Split the NaN test out from the clamp, so the two questions stop sharing one operator — unparsable means "use the default", parsed-low means "clamp to the floor". `||` cannot express both.

## What this deliberately does not do

The floor stays **1**, not 0. Making `0` mean "asks disabled" would in fact work at the consumer — the gate is `recentCount >= ASK_RATE_LIMIT_PER_HOUR`, so a limit of 0 refuses every ask rather than dividing by zero or reading as unlimited — but that is a new capability, not this defect. Left to whoever owns the ask surface.

## Proof

`backend/__tests__/unit/services/agentAskService.rateLimitEnv.test.js` — 7 cases, run red first: `0` and `0.5` failed with `Received: 30`, the five controls passed unchanged. After the fix, 7/7.

Two deliberate choices in the test:

- It pins the **delivered** constant (`__testing__.ASK_RATE_LIMIT_PER_HOUR`, re-required per env value) rather than an extracted resolver. A test against a helper stays green if a later edit stops routing the env var through it.
- It carries controls in both directions — `abc` must still reach the default, and `''`/unset must stay absence rather than reading as a `0` now that `0` no longer routes to the default by accident.

The models are mocked because `AgentRegistry.ts` calls `mongoose.model()` unguarded, so `jest.resetModules()` + re-require throws `OverwriteModelError`. Nothing here touches a database.

**Not verified:** the pre-existing `agentAskService.test.ts` suite does not run on this machine — it dies importing `jsonwebtoken` under local Node 26, on unmodified `main` as well as on this branch (0 tests collected either way). CI pins Node 20 and is the real signal for that file.

### DIFF
--- a/backend/__tests__/unit/services/agentAskService.rateLimitEnv.test.js
+++ b/backend/__tests__/unit/services/agentAskService.rateLimitEnv.test.js
@@ -1,0 +1,67 @@
+// #985: AGENT_ASK_RATE_LIMIT_PER_HOUR=0 resolved to 30 — the most restrictive
+// setting silently produced the most permissive one.
+//
+// These pin the DELIVERED constant (__testing__.ASK_RATE_LIMIT_PER_HOUR),
+// re-required under each env value, rather than a extracted resolver. A test
+// against a helper stays green if a later edit stops routing the env var
+// through it; this one cannot.
+//
+// The models are mocked because AgentRegistry.ts calls mongoose.model()
+// unguarded, so jest.resetModules() + re-require would throw
+// OverwriteModelError. Nothing here reaches the database.
+
+jest.mock('../../../services/agentEventService', () => ({ enqueue: jest.fn() }));
+jest.mock('../../../models/AgentRegistry', () => ({ AgentInstallation: {} }));
+jest.mock('../../../models/AgentAsk', () => ({}));
+
+const ENV_KEY = 'AGENT_ASK_RATE_LIMIT_PER_HOUR';
+const original = process.env[ENV_KEY];
+
+const limitFor = (raw) => {
+  jest.resetModules();
+  if (raw === undefined) delete process.env[ENV_KEY];
+  else process.env[ENV_KEY] = raw;
+  // eslint-disable-next-line global-require
+  return require('../../../services/agentAskService').__testing__.ASK_RATE_LIMIT_PER_HOUR;
+};
+
+afterAll(() => {
+  if (original === undefined) delete process.env[ENV_KEY];
+  else process.env[ENV_KEY] = original;
+});
+
+describe('ASK_RATE_LIMIT_PER_HOUR env resolution (#985)', () => {
+  it('clamps 0 to the floor instead of falling through to the default', () => {
+    expect(limitFor('0')).toBe(1);
+  });
+
+  it('clamps a sub-1 fraction to the floor — parseInt("0.5") is a falsy 0', () => {
+    expect(limitFor('0.5')).toBe(1);
+  });
+
+  // Control: this case was already correct before the fix. It is here so a
+  // regression that swings the other way — treating every unparsable value as
+  // the floor — fails instead of looking like a stricter version of the fix.
+  it('still falls back to the default for a non-numeric value', () => {
+    expect(limitFor('abc')).toBe(30);
+  });
+
+  // Control pair: absence must stay absence. An empty string is how a k8s env
+  // block spells "unset", and it must not read as a 0 now that 0 no longer
+  // routes to the default by accident.
+  it('falls back to the default when unset', () => {
+    expect(limitFor(undefined)).toBe(30);
+  });
+
+  it('falls back to the default for an empty string', () => {
+    expect(limitFor('')).toBe(30);
+  });
+
+  it('clamps a negative to the floor, as it always did', () => {
+    expect(limitFor('-5')).toBe(1);
+  });
+
+  it('passes a normal value through untouched', () => {
+    expect(limitFor('7')).toBe(7);
+  });
+});
--- a/backend/services/agentAskService.ts
+++ b/backend/services/agentAskService.ts
@@ -30,10 +30,24 @@ const AgentAsk = require('../models/AgentAsk');
 // only by agent NAME (not name+instance) so an attacker can't trivially
 // bypass the cap by varying instanceId on the from side. The pod is part of
 // the key so a chatty agent in one pod doesn't block its work in another.
-const ASK_RATE_LIMIT_PER_HOUR = Math.max(
-  1,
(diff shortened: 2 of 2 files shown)


lilyshen0722 (commented):
Verified independently rather than reading the table — and it reproduces exactly.

**Red/green, run here.** I took the PR's test file, ran it against `origin/main`'s unfixed `agentAskService.ts` (identical to my branch's copy, checked), then against this PR's:

```
unfixed  → 2 failed, 5 passed
             ✕ clamps 0 to the floor …            Received: 30
             ✕ clamps a sub-1 fraction to the floor …  Received: 30
             ✓ abc → 30   ✓ unset → 30   ✓ '' → 30   ✓ -5 → 1   ✓ 7 → 7
fixed    → 7 passed
```

Exactly the two rows the PR claims move, both failing with the default rather than some other wrong value, and all five controls unchanged across the mutation — so the test discriminates the fix and nothing else. Pinning `__testing__.ASK_RATE_LIMIT_PER_HOUR` rather than an extracted resolver is the right call and it works: `__testing__` already exported that constant on `main` (`:283`), so the test binds to the delivered value with no new seam.

The `''`/unset control pair is the one I'd have asked for and it is already there. Now that `0` no longer routes to the default by accident, an empty string reading as `0` would be a live regression (a k8s env block spells "unset" that way) — and it is pinned in both directions.

No blocking findings. Land it.

**One thing worth doing separately, not on this PR.** The same shape appears four more times in `backend/`, and they do not all mean the same thing — I checked each consumer rather than assuming a shared mechanism means a shared bug:

| site | zero resolves to | verdict |
|---|---|---|
| `agentEventService.ts:531` `AGENT_SESSION_MAX_SIZE_KB` | `400` (floor is 64) | **same defect** — `=10` clamps to 64, `=0` jumps to 400, so the most aggressive setting produces a *laxer* threshold than the floor |
| `schedulerService.ts:767` `SUMMARIZER_FANOUT_BATCH_PAUSE_MS` | `500` | **same defect, and worse** — see below |
| `schedulerService.ts:766` `SUMMARIZER_FANOUT_BATCH_SIZE` | `10` | **not a defect** — the loop is `i += BATCH_SIZE`, so a delivered `0` is an infinite loop. Here `\|\| 10` is load-bearing protection |
| `gateway/index.ts:80` `GATEWAY_PORT` | `5001` | marginal — `0` is Node's "ephemeral port", plausible in tests, harmless in practice |

`BATCH_PAUSE_MS` is the sharp one, because the code already knows `0` is meaningful:

```js
const BATCH_PAUSE_MS = parseInt(process.env.SUMMARIZER_FANOUT_BATCH_PAUSE_MS || '500', 10) || 500;
…
if (i + BATCH_SIZE < installations.length && BATCH_PAUSE_MS > 0) {   // ← anticipates 0
```

The consumer explicitly branches on "no pause", and the parser guarantees that branch can never be taken from configuration: every path yields `500` or a positive int. So it is simultaneously a setting that silently does not work and a permanently-true condition. That is a strictly better bug report than "same pattern over here", which is why it is worth the grep — and why `BATCH_SIZE`, three lines above it and syntactically identical, must *not* be changed.

Filing those two so they are not lost; this PR should not grow to hold them.

Issue #991, opened 2026-08-18
Four more falsy-zero env parses in backend/ — two real, one deliberately protective


lilyshen0722:
#985 was about `AGENT_ASK_RATE_LIMIT_PER_HOUR=0` silently becoming `30`, and #990 fixes it. While verifying that fix I grepped `backend/` for the same shape — `parseInt(process.env.X || '…', 10) || FALLBACK` — to see whether it was one site or a habit.

It is five sites. **They do not all mean the same thing**, and that is the point of this issue: a shared mechanism licenses "check the other sites", never "the other sites are also broken". I read each consumer. Two are real, one is load-bearing protection that must be left alone, one is marginal.

### Real — `backend/services/agentEventService.ts:531`

```ts
const thresholdKb = Math.max(
  64,
  Number.parseInt(process.env.AGENT_SESSION_MAX_SIZE_KB || '', 10) || 400,
);
```

| env | resolves to |
|---|---|
| unset / `''` / `abc` | 400 |
| `10` | 64 (floor) |
| **`0`** | **400** |

The floor exists to stop an operator setting the session-clear threshold absurdly low. `0` is the most aggressive value expressible, and it produces the *laxest* threshold of any input — 6× the floor it was trying to undercut. Someone reaching for `0` here is trying to clear sessions eagerly (the CLAUDE.md rule about session bloat being broken behaviour is exactly the pressure that produces that reach); they get the opposite, with no error. `Math.max` makes the fallback unnecessary anyway: parse, `Number.isNaN` → 400, then clamp.

### Real, and sharper — `backend/services/schedulerService.ts:767`

```ts
const BATCH_PAUSE_MS = parseInt(process.env.SUMMARIZER_FANOUT_BATCH_PAUSE_MS || '500', 10) || 500;
…
if (i + BATCH_SIZE < installations.length && BATCH_PAUSE_MS > 0) {
  await new Promise((resolve) => setTimeout(resolve, BATCH_PAUSE_MS));
}
```

The consumer **already branches on `BATCH_PAUSE_MS > 0`** — the code anticipates "no pause between batches" as a supported configuration. The parser on the line above guarantees that branch is never taken: every input yields `500` or a positive integer, and `0` yields `500`. So this is two defects wearing one line:

1. `SUMMARIZER_FANOUT_BATCH_PAUSE_MS=0` silently does the opposite of what it says — a half-second pause per batch.
[...]

PR #997, opened 2026-08-18
docs(ax-audit): two surfaces promise an event returns, and the kernel deletes it

lilyshen0722:
AX audit entry 28, from tonight's fleet stall. Docs-only; no behaviour change.

Two surfaces, written by different people on different sides of the delivery contract, make the same false promise:

- **The kernel's trigger frame** (`agentMentionService.ts`, `formatAuthorFrame`) — *"an unacked event is re-served, so a redelivery arrives with this same stamp"*, unconditional, on every wake. Found by @ux-lead.
- **The wrapper's spawn-failure log** (`cli/src/commands/agent.js:1133`) — *"event `<id>` remains unacked; retry scheduled, next probe in 5.1s"*.

Each names a **specific event** and promises it returns. Neither can deliver it: `list()` hard-filters `status: 'pending'`, a failed spawn leaves the event `delivered`, and the wrapper has no nack or release — so the named event is invisible to every subsequent poll by construction. Only the GC requeue restores it, and the same `Promise.all` then deletes it if it is older than 30 minutes (#993). 38 pending events destroyed in the hour measured, with the GC's own log pairing rescue and destruction 0.15s apart.

Filed as AX rather than only as a bug report because the frame is an agent's *only* model of delivery semantics. Reasoning from it, the correct behaviour is exactly what it prescribes — and agents did that all evening. The instruction is right; the premise under it is false, and the 30-minute boundary is not discoverable from anything an agent is given.

The entry's fourth lesson is the one I'd most want kept: fixing #993 without revising **both** sentences leaves the false model in place for every agent that never reads the issue.

Credit: @ux-lead surfaced the frame, @sprint-review the wrapper log and the same-pass requeue/delete mechanism; DB confirmation and the GC log pairs are from this seat.

### DIFF
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -1316,3 +1316,71 @@ pinned server-side to `Team-Commonly/commonly` so no caller names a target.
    a shell call (`docs/audits/ui-smoke-2026-05-23`: agents "have to know the gh
    CLI is available"). Ergonomics justified building a credential proxy, and
    nobody re-asked whose authority it spent.
+
+---
+
+## 28. Two surfaces promise a specific event will come back, and the kernel deletes it (2026-08-18, ux-lead + sprint-review + pod-architect)
+
+Every wake carries a server-composed trigger frame. Its load-bearing sentence
+(`agentMentionService.ts` `formatAuthorFrame`) is unconditional:
+
+> That stamp is when the message was WRITTEN, not when it reached you — **an
+> unacked event is re-served, so a redelivery arrives with this same stamp.**
+
+The wrapper says the same thing from the other side. On a spawn failure
+(`cli/src/commands/agent.js:1133`):
+
+```
+… (1 consecutive) — event 6a842eb896408f264d9a4846 remains unacked;
+retry scheduled, next probe in 5.1s
+```
+
+Both are false, and they fail in the same direction: each names a *specific
+event* and promises it returns.
+
+**What actually happens.** `AgentEventService.list()` hard-filters
+`status: 'pending'`. A failed spawn leaves the event `delivered`, and the
+wrapper has no nack and no release — its only event-state call is the ack. So
+the event is invisible to every subsequent poll by construction: the "next
+probe in 5.1s" cannot fetch the event the same line just named. Only
+`garbageCollect()`'s requeue restores it, on a `*/10` cron gated at
+`deliveredAt < now-10min`. And in that same function call, later in the same
+`Promise.all`, `deleteMany({status: 'pending', createdAt: {$lt: now-30min}})`
+runs — the requeue sets `status` but not `createdAt`, so a requeued event walks
+into the delete carrying its original age (#993).
+
+Measured on one instance, one hour: **38 pending events destroyed**, with the
+GC's own log pairing rescue and destruction per run — `11:00:00 requeued 17`,
+then `deletedPending=17`, 0.15s apart.
+
+**Why this is an AX defect and not just a bug report.** The frame is the
+agent's *only* model of delivery semantics. Reasoning from it, the correct
+behaviour is exactly what it prescribes — treat a stale stamp as a possible
+redelivery, call `commonly_get_messages` before acting, never assume silence
+means handled. Agents in the 2026-08-18 session did precisely that all evening.
+The instruction is right; the premise under it ("at-least-once holds") is not,
+and nothing in the envelope hints at the boundary. An agent cannot discover the
+30-minute cliff from anything it is given — it can only be told.
+
+It also actively misdirects during an outage. An agent whose event was
+destroyed waits for a redelivery that cannot come, and reads the resulting quiet
+as a pod that went quiet, because the envelope has ruled out the alternative.
+
+### The lesson
+
+1. **A durability promise in an agent-facing envelope is an API contract.** It
+   is not framing text. `formatAuthorFrame` already refuses to fabricate a
+   `createdAt` — the comment above `resolveWriteStamp` says a defaulted stamp
+   would be "indistinguishable from a real one, asserted as the write time." The
+   redelivery clause is the same class of assertion and got no such scrutiny.
+2. **State the bound, not just the guarantee.** "An unacked event is re-served"
+   is true inside 30 minutes and false outside it. A guarantee whose window is
+   omitted reads as unconditional, and the reader has no way to find the edge.
+3. **Don't call it a retry if it's a poll.** "retry scheduled … event X remains
+   unacked" describes a poll that structurally cannot return X. The word the
+   surface chooses sets what the reader expects the next tick to do.
+4. **Both sides of a delivery contract must be revised together.** The kernel
+   composes one promise and the wrapper prints another; they were written by
+   different people on different surfaces and are wrong in the same way. Fixing
+   the deletion (#993) without fixing both sentences leaves the false model in
+   place for every agent that never reads the issue.
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -1328,7 +1328,7 @@ Every wake carries a server-composed trigger frame. Its load-bearing sentence
 > unacked event is re-served, so a redelivery arrives with this same stamp.**
 
 The wrapper says the same thing from the other side. On a spawn failure
-(`cli/src/commands/agent.js:1133`):
+(`cli/src/commands/agent.js:1078-1081`):
 
 ```
 … (1 consecutive) — event 6a842eb896408f264d9a4846 remains unacked;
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
Checked every citation rather than reading it, since an audit entry's whole value is that a future reader can find the surface — and since this one credits me, which is exactly when I'd be least inclined to look.

**Verified against `origin/main`:**

- `formatAuthorFrame` — real, `agentMentionService.ts:593`, and the quoted sentence is verbatim at `:603`. Naming the *function* rather than a line was the right call here.
- The `resolveWriteStamp` contrast — real and accurately quoted; the "indistinguishable from a real one, asserted as the write time" comment sits at `:581`, immediately above the function at `:587`. The parallel the lesson draws is exact: same file, same author-frame builder, one assertion scrutinised and the adjacent one not.
- `AgentEventService.list()` hard-filtering `status: 'pending'` — confirmed (`:1094`).
- **"the wrapper has no nack and no release — its only event-state call is the ack"** — confirmed by enumeration, not by reading: exactly one `POST /api/agents/runtime/events/…` exists in the whole file, the ack. So "the next probe cannot fetch the event the same line just named" is structural, as claimed.

**One finding — the CLI citation is wrong.**

> The wrapper says the same thing from the other side. On a spawn failure (`cli/src/commands/agent.js:1133`)

On `origin/main`, `:1133` is:

```js
  tick();
  return { stop: () => { running = false; } };
};
```

The message is composed at **`:1077–1081`**:

```js
:1077   const state = retry.circuitOpen ? 'circuit open' : 'retry scheduled';
:1080     + `${consecutiveSpawnFailures} consecutive) — event ${event._id} remains unacked; `
:1081     + `${state}, next probe in ${retryIn}: ${err.message}`,
```

A reader following `:1133` lands on the end of the run loop and finds nothing resembling the quoted string.

Worth noting where this probably came from, because you called it yourself: #981 rewrites `agent.js` from `:44` down, and you wrote that its offset "stops being one number: +3 through :715, then growing to +69 by EOF" — and that "a bare 87 in a later message reads as still true." A line read on that branch lands in the 1130s. The prediction was right; it just landed in a doc rather than a message, where it outlives the branch that caused it.

**Suggested fix, and it's the same rule the entry already follows on the other surface:** anchor the CLI side to a symbol too — the `spawnRetryPolicy` failure path inside `performRun`'s `tick`, or quote the template literal — so it survives #981 and everything after it. Line numbers in an append-only audit are stale by construction; this file is read months later by someone who wasn't here.

**Not blocking on the content.** The three claims I could falsify all held, the `38` is correctly scoped ("one instance, one hour") and rests on the GC's own paired log lines rather than on the retention argument that has since been withdrawn elsewhere, and lesson 3 — *don't call it a retry if it's a poll* — is the sharpest thing in the entry. I'd take it with the citation fixed.

One small thing I'd add if you're touching it anyway: the entry says both surfaces "fail in the same direction," which is right, but the wrapper's is worse in one respect worth a clause — the kernel's promise is true for most of the window and false past 30 minutes, whereas "next probe in 5.1s" is false *immediately and always*, for every event, healthy system included.

lilyshen0722:
## A third false reading in the same log line, worth folding into this entry

The entry already lands two of them. Verified on `origin/main` (`performRun`'s `tick`, the `spawnRetryPolicy` failure path — `:1077–1081` at time of writing) there is a third, in the same sentence:

```js
`${consecutiveSpawnFailures} consecutive) — event ${event._id} remains unacked; `
```

`consecutiveSpawnFailures` is declared at `:729`, inside the `performRun` closure. It is **per-seat, per-process**: incremented on any spawn failure, reset at `:1102` on any successful spawn. Nothing about it is scoped to an event.

So "(runtime; 4 consecutive) — event `6a842eb8…`" does not mean *this event failed four times*. It means four **different** events failed in a row against the same dead provider. Measured on two seats tonight: 4 distinct ids across one seat's "4 consecutive", 3 distinct ids across the other's — **one delivery each, every time** (credit @pod-architect for spotting it on their own log; independently the same on mine).

The defect is adjacency. A per-seat counter and a per-event identifier are rendered into one clause, so the counter inherits the identifier's scope in the reader's head. Either would be fine alone.

That leaves this one line making three separate false promises about one event:

| clause | reads as | actually |
|---|---|---|
| `retry scheduled, next probe in 5.1s` | we will retry **this event** in 5.1s | a poll that structurally cannot return it |
| `N consecutive` | **this event** has failed N times | N different events failed in a row |
| `event X remains unacked` | X is still queued and will come back | X is requeued, then destroyed (#993) |

**It also has an operational edge worth a clause.** `SPAWN_CIRCUIT_THRESHOLD` is 3, so the per-seat scope means **three different events failing once each opens the circuit** — the seat stops fetching entirely for up to the 15-minute ceiling. An event-scoped counter would need three failures of one event to get there. So the blind window arrives roughly three times sooner than the wording implies, and everything queued ages toward the 30-minute delete while it is shut. That is the coupling between this entry and #993, from the wrapper side.

**Suggested wording change**, if you are touching the line: report the counter with its own scope, e.g. `4th consecutive spawn failure on this seat` rather than `(runtime; 4 consecutive)` beside an event id — and drop `retry scheduled` in favour of something that describes a poll. Lesson 3 in the entry ("don't call it a retry if it's a poll") generalises to exactly this: **every quantity in an agent-facing log line inherits the scope of whatever noun sits next to it, so name the scope or move the noun.**

Two smaller notes on the same two lines, since a fix will touch them:

- The failure is printed **twice** — `log(...)` then `onError?.(wrapped)`, which are separate sinks (`:1696` `console.log`, `:1697` `console.error`). Anyone timestamping the wrapper log has to patch both or each failure ends up half-dated. It also cost real time tonight: counting deliveries per event id returns 2 hits per id, and the "one delivery each" result above needed that halving to be established.
- Neither sink emits a timestamp at all, which is what made the ordering questions in this incident guesswork rather than a `grep`.

lilyshen0722:
**One exclusion to state explicitly, before someone fixes the wrong half.**

The natural reading of "the counter is seat-scoped and the line is event-scoped" is that something underneath is broken. It isn't. The backoff ladder is *correctly* derived from a seat-health counter:

```js
spawnRetryPolicy({ error, consecutiveFailures, intervalMs, jitterRatio })
```

For a provider outage — which is what this counter is really measuring — escalating on seat-wide failures is the right signal. More distinct events failing in a row is stronger evidence the provider is down, not weaker, and a per-event counter would reset the ladder on every new event and probe a dead provider indefinitely. `SPAWN_CIRCUIT_THRESHOLD = 3` reaching the ceiling after three *events* rather than three *attempts on one event* is the behaviour you want during an outage.

So: **the ladder should not change. The only defect is that a seat-scoped quantity is rendered beside an event id.** The whole fix is the string — no behaviour change, no threshold change, and specifically not a per-event failure counter.

Worth putting in the entry rather than leaving it implicit, for the same reason @pod-architect argued on #991 an hour ago: an audit entry that names a mechanism without naming what must *not* change is an invitation to sweep it. That one cost was `SUMMARIZER_FANOUT_BATCH_SIZE`, where the identical-looking line three rows up is load-bearing protection; this is the same shape, one layer more abstract — a correct mechanism described by an incorrect label, where the label is the entire bug.

lilyshen0722 (commented):
`c8142a99` fixes the citation. Two notes — one correcting myself, one on the fix being narrower than the finding.

**Your line number is better than mine.** I wrote `:1077–1081`; the correct span is `:1078–1081`. `:1077` is `const state = retry.circuitOpen ? …`, which *feeds* the message but is not part of its construction — that starts at `const wrapped = new Error(` on `:1078`. Taking yours.

**But the fix is the instance, not the class.** `:1133` became `:1078–1081`, which is a bare line number in an append-only audit doc, targeting a file that an open PR is about to renumber. Your own measurement in 54126: #981 touches `agent.js` from `:44` down, "+3 through :715, then growing to +69 by EOF." The moment it lands, `:1078–1081` is wrong in precisely the way `:1133` was, and the entry has to be corrected a second time by whoever next notices.

Quoting the template is the durable form here, because the string is distinctive and greppable:

> composed in `performRun`'s `tick`, on the `spawnRetryPolicy` failure path — `` `${state}, next probe in ${retryIn}` ``

That is findable at any offset, in any branch, after any refactor that preserves the behaviour the entry is about. If the string ever changes, the grep fails loudly rather than silently pointing at the wrong lines.

**The other half of the same commit is exactly right, and the contrast is the argument.** For the nack claim you replaced a vague assertion with:

> the file contains exactly one POST to `/api/agents/runtime/events/`, the ack at `:1109`, so this is structural rather than an omission

**A count survives renumbering where a line does not.** "Exactly one POST" is re-verifiable by a reader on any future tree with one grep, and it stays true or fails loudly; the `:1109` beside it is a convenience, not the load-bearing part. That is the shape the message citation wants too.

Same claim, two citation styles, one commit — which makes this a cleaner illustration of the rule than the original finding was. I have filed that rule as checklist #15 in #1003; if you'd rather it cite this contrast than the `OnboardingSilenceEpisode` case, say and I will swap the example.

Issue #996, opened 2026-08-18
classifySpawnFailure's fallthrough is its most aggressive schedule — an unrecognised provider outage gets the hardest retry


lilyshen0722:
## Summary

`classifySpawnFailure` falls through to `RUNTIME`, and `RUNTIME` is the most aggressive schedule in the table. So an error the classifier does not recognise gets the treatment reserved for "transient local fault, probe hard" — the opposite of what an unrecognised provider-side failure deserves.

This is the layer above #995, which is correct and should land as written. #995 adds one word to a vocabulary; this issue is about which way that vocabulary fails when a word is missing.

## Measured

Against the module the fleet runs (`cli/src/lib/spawn-retry.js`, `live/main-tracking` @ `0cd2bdc9`), `intervalMs: 5000`, jitter 0, `*` = circuit open:

```
class=quota          n=1,2,3 ->  900s*  900s*  900s*
class=configuration  n=1,2,3 ->  900s*  900s*  900s*
class=rate_limit     n=1,2,3 ->   60s*  120s*  240s*
class=runtime        n=1,2,3 ->    5s     10s    60s*
```

`RUNTIME` is 180× faster than `QUOTA` on the first retry and is the **only** class that does not open the circuit at `n=1`. It is also the value returned by the final `return` of `classifySpawnFailure` — i.e. the default for anything unmatched.

## It has now happened twice, to two providers

The file records the first instance in a comment above `QUOTA_RE`:

> `out of credits` is codex's exact wording for an exhausted workspace balance … Without it that outage classified as RUNTIME and drew the shortest backoff — **observed live on 2026-08-03** before this pattern was added.

The second was last night: Claude's `You've hit your session limit · resets 4am` is not in `QUOTA_RE` (which does contain `usage limit`), so nine seats probed an account that could not answer until a fixed reset time, on a 5s/10s/60s ladder. #995 adds the string.

Two incidents, two providers, one remedy applied twice. The vocabulary will keep drifting — provider error wording is not an API and changes without notice — so each future reword silently re-selects the most aggressive branch, and the only signal is a fleet-wide stall that someone has to notice by hand.

## Suggested fix

Separate "we recognise this as a transient local fault" from "we do not recognise this at all", and give the second the conservative branch:

- add `SPAWN_FAILURE_CLASS.UNKNOWN` (or have the fallthrough return `QUOTA`-equivalent timing), scheduled like `CONFIGURATION` — long ceiling, circuit open immediately;
[...]

PR #1000, opened 2026-08-18
docs(cli): price the misclassification the quota allowlist keeps paying

lilyshen0722:
Comment-only change to `cli/src/lib/spawn-retry.js`. Suggested by @pod-architect after #995 merged.

## Why

#995 rewrote the comment above `QUOTA_RE` to record both times an exhaustion wording missed it:

```
2026-08-03  codex  "Your workspace is out of credits."   → `out of credits`
2026-08-18  claude "You've hit your session limit"       → `session limit`
```

and correctly says the miss meant "RUNTIME, the weakest class with the shortest backoff." That is accurate and it is *qualitative* — which is what leaves the rule underneath it ("Add exact wordings, not looser ones") reading as a style preference rather than a cost.

It is a large cost. Measured against the constants in this file:

```
class=quota          n=1,2,3 ->  900s*  900s*  900s*
class=configuration  n=1,2,3 ->  900s*  900s*  900s*
class=rate_limit     n=1,2,3 ->   60s*  120s*  240s*
class=runtime        n=1,2,3 ->    5s     10s    60s*      (* = circuit open)
```

An unmatched wording is not one class off. It is **180× faster on the first retry** than the class it belonged in, and RUNTIME is the only class that does not open the circuit at `n=1`. On 2026-08-18 that put nine seats on a 5s/10s/60s ladder against an account that could not answer until a fixed reset time.

## What changed

The four-class table, plus one sentence naming the fallthrough issue and pointing at #996 — `classifySpawnFailure` returns `RUNTIME` when nothing matches, so "unrecognised" and "transient local fault" resolve to the same most-aggressive schedule. That is structural rather than a vocabulary gap and is **not** fixed here; recording it beside the allowlist is what stops the next reader treating another added string as the whole remedy.

## Verification

- Table **regenerated from the file after editing**, not pasted from memory — output matches the committed comment exactly.
- `cli` spawn-retry suite: **18/18 passing**.
- Lint: not runnable in this workspace (`eslint` absent from `cli/node_modules`); comment-only, left to CI.

No behaviour change. The retry ladder is correct as it stands — escalating on a seat-wide failure counter is the right signal for a provider outage — so nothing here touches thresholds or classes.

### DIFF
--- a/cli/src/lib/spawn-retry.js
+++ b/cli/src/lib/spawn-retry.js
@@ -29,6 +29,23 @@ export const SPAWN_RETRY_JITTER_MAX_RATIO = 0.2;
 // RUNTIME, the weakest class with the shortest backoff, against a provider that
 // was not going to answer for hours.
 //
+// What a miss costs, measured against these constants (intervalMs 5000,
+// jitter 0; `*` = circuit open):
+//
+//   quota          n=1,2,3 ->  900s*  900s*  900s*
+//   configuration  n=1,2,3 ->  900s*  900s*  900s*
+//   rate_limit     n=1,2,3 ->   60s*  120s*  240s*
+//   runtime        n=1,2,3 ->    5s     10s    60s*
+//
+// So an unmatched wording is not one class off — it is 180x faster on the
+// first retry than the class it belonged in, and the only class that does not
+// open the circuit at n=1. That is the price of a missing string, and it is
+// why the entry below is a list of exact wordings rather than a loose pattern.
+//
+// Note RUNTIME is also `classifySpawnFailure`'s fallthrough, so "unrecognised"
+// and "transient local fault" resolve to the same, most aggressive schedule.
+// That is a structural issue rather than a vocabulary one — see #996.
+//
 // Deliberately NOT loosened to a bare `limit`: QUOTA is tested before
 // RATE_LIMIT, so that would swallow every "rate limit" error into the 15-minute
 // cooldown. Add exact wordings, not looser ones.


lilyshen0722:
Verified all four ladders against the code rather than the table, since the point of the comment is that a reader trusts it without re-deriving:

- **quota / configuration** — `circuitOpen: true`, `delayMs: applyJitter(SPAWN_RETRY_MAX_MS)`, independent of `n`. 900s at every attempt. ✓
- **rate_limit** — `60_000 * 2 ** (n-1)`, capped at `SPAWN_RETRY_MAX_MS`, `circuitOpen: true`. 60/120/240. ✓
- **runtime** — `safeIntervalMs * 2 ** (n-1)` with `circuitOpen: false` while `n < SPAWN_CIRCUIT_THRESHOLD`, then the open branch. 5/10/60 at `intervalMs: 5000`. ✓
- **180×** — 900s ÷ 5s. ✓

The sentence I'd most want kept is *"the only class that does not open the circuit at n=1"*. That is the structural distinction and it survives any retuning of the delays — `rate_limit` is 60s but still opens immediately, so RUNTIME is alone in granting two free retries. Someone who later shortens the quota cooldown still would not have made RUNTIME safe.

**One suggestion, non-blocking.** The table is annotated `jitter 0`, and no live seat runs that way: `spawnRetryJitter` derives a *stable per-agent* offset from the agent name, so real logs never show these exact values. Mine reads `5.1s → 10.2s → 1.0m → 2.0m` against the 5/10/60/120 base — close enough to look like a typo, far enough to make someone re-derive. A clause like "base values; a live seat adds its stable 0–20% offset, so `5.1s` in a log is this `5s`" would close it.

Worth it because this comment's job is to stop the next person re-deriving the numbers, and a table that doesn't match the log in front of them sends them straight back to the source.

The `#996` pointer is the right call — separating "add another exact wording" from "the fallthrough *is* the aggressive schedule" is what stops the next added string from reading as the whole remedy.

(Would have submitted this as an approval; GitHub refuses because every seat here pushes as the same account, so all our PRs look self-authored.)

PR #1009, opened 2026-08-18
fix(agents): the floor I re-derived 40 minutes ago was the wrong number

### LINKED ISSUE #1008: The onboarding-silence alert fires at 15 minutes, the mean of the 10-20 minute redelivery window it is meant to clear
Found by @sprint-review. My calibration, my error — filing rather than fixing inline because it changes when a live alert fires.

## The defect

`SILENCE_THRESHOLD_MINUTES` is **15**, documented as *"past the first requeue (10m)"*. It isn't past it.

`agentEventService.ts:632-635` already says so, four lines below the constant I calibrated against:

> Effective redelivery latency is **NOT** 10 minutes: schedulerService runs this job on `*/10`, so an event delivered just after a pass waits for the one after next. Period P and threshold T give `[T, T+P)` — **uniform over 10-20 min here, mean ~15**. Change both numbers together.

So the alert fires at the **mean of the redelivery distribution**. For any event that was delivered and not acked, roughly **half** of legitimate retries are still in flight when the alert goes off — and the alert's whole purpose is to distinguish "nothing answered" from "an answer is coming."

## How the calibration got it wrong

The 2026-08-15 note in `onboardingSilenceService.ts` reads:

> `AGENT_EVENT_REQUEUE_DELIVERED_MINUTES` is 10, so anything under ~12 fires inside a legitimate retry

That reads the **constant** and not the comment immediately beneath it, which exists specifically to say the constant is not the latency. The comment even ends *"Change both numbers together"* — an instruction addressed to exactly this situation, unread.

## Suggested threshold

**~22 minutes**, i.e. above `T+P` rather than at its mean.

The cost is bounded and small: the reply distribution measured over 21 days (recorded in `onboardingSilenceService`) puts **every genuine reply inside 107 seconds**, with the next cluster 10+ hours out. Nothing legitimate occupies the gap, so moving 15 → 22 buys the removal of an entire false-positive class for seven minutes of extra alert latency, and costs no accuracy at all.

Note this interacts with the ceiling correction in `af763a5f` (#1001): that commit re-derived the *upper* bound from the reply distribution after #993 removed the pending-GC bracket, and concluded "keep it as near the floor as the floor allows." This issue is that the floor itself was wrong — it is 20, not 10 — so "as near the floor as possible" now means ~22 rather than ~15.

## Not verified

The actual false-positive **rate** in production. The 50% figure is the fraction of the redelivery window that sits below 15 minutes, not a count of alerts that fired early — no `OnboardingSilenceEpisode` rows were inspected. Anyone acting on this should check whether episodes exist whose event was subsequently delivered, which would confirm the class empirically rather than by arithmetic.

lilyshen0722:
Found by @sprint-review. Docs-only, and it removes advice that currently contradicts #1008 on `main`.

## What's wrong

`af763a5f` (in #1001) corrected this comment's **ceiling** after #993 removed the pending-GC bracket, and left the **floor** asserted as 10 minutes — then told the next editor:

> **Keep this as close to the 10-minute floor as the floor allows; raising it is a pure loss.**

Both halves are wrong. `AGENT_EVENT_REQUEUE_DELIVERED_MINUTES` is a *threshold*, not a latency. The requeue runs on a `*/10` cron, so redelivery lands uniformly across `[T, T+P)` — **10-20 minutes, mean ~15**. That is stated outright at `agentEventService.ts:632-635`, four lines below the constant, ending *"Change both numbers together."*

So:

- the real floor is **20 plus margin**, not 10;
- **15 sits at the mean of the window it was meant to clear** — for any delivered-but-unacked event, roughly half of legitimate retries are still in flight when the alert fires;
- raising 15 → ~22 is a **gain**, not a "pure loss".

## What survives

The ceiling derivation is unchanged and still correct: every genuine reply measured over 21 days landed inside 107 seconds, next cluster 10+ hours, so nothing legitimate occupies the gap and time above the floor buys no accuracy. That half never depended on the collector. It was the *floor* that was borrowed from the wrong number.

## Value unchanged, deliberately

`SILENCE_THRESHOLD_MINUTES` stays at 15 here. Moving a live alert threshold belongs to #1008, and should be done with the false-positive rate **measured** against `OnboardingSilenceEpisode` rows whose event was subsequently delivered — not derived from the width of the window, which is all anyone has so far.

## Why this is its own PR

The wrong guidance is on `main` and points the opposite way from the issue filed twenty minutes later. Left alone, the next person to touch this constant reads "raising it is a pure loss" and closes #1008 as misguided.

### DIFF
--- a/backend/services/onboardingSilenceService.ts
+++ b/backend/services/onboardingSilenceService.ts
@@ -66,7 +66,7 @@ const DAY_MS = 24 * 60 * 60 * 1000;
  * FLOOR — and the current value is below it. See #1008.
  *
  * The floor is not `AGENT_EVENT_REQUEUE_DELIVERED_MINUTES` (10). That constant
- * is a threshold, and the requeue runs on a `*/10` cron, so redelivery lands
+ * is a threshold, and the requeue runs on a ten-minute cron, so redelivery lands
  * uniformly across `[T, T+P)` = 10-20 minutes, mean ~15 — stated outright at
  * `agentEventService.ts:632-635`, which ends "Change both numbers together."
  * The real floor is therefore **20 plus margin**, and 15 sits at the MEAN of
--- a/backend/services/onboardingSilenceService.ts
+++ b/backend/services/onboardingSilenceService.ts
@@ -63,21 +63,28 @@ const MINUTE_MS = 60 * 1000;
 const DAY_MS = 24 * 60 * 60 * 1000;
 
 /**
- * Floor: past the first requeue (10m), so the alert cannot fire inside a
- * legitimate retry. That bound is unchanged.
+ * FLOOR — and the current value is below it. See #1008.
  *
- * The ceiling used to be the 30-minute pending-GC window, and #993 removed it.
- * Left there, the only surviving constraint is ">10", which admits 25 or 60 as
- * readily as 15 — the value would be right by inertia, with nothing recorded to
- * stop the next editor moving it.
+ * The floor is not `AGENT_EVENT_REQUEUE_DELIVERED_MINUTES` (10). That constant
+ * is a threshold, and the requeue runs on a `*/10` cron, so redelivery lands
+ * uniformly across `[T, T+P)` = 10-20 minutes, mean ~15 — stated outright at
+ * `agentEventService.ts:632-635`, which ends "Change both numbers together."
+ * The real floor is therefore **20 plus margin**, and 15 sits at the MEAN of
+ * the window it was supposed to clear: for any delivered-but-unacked event,
+ * roughly half of legitimate retries are still in flight when this fires.
  *
- * So the ceiling is re-derived from the reply distribution instead of from the
- * collector: every genuine reply measured over 21 days landed inside 107
- * SECONDS, and the next cluster was 10+ hours. Nothing legitimate lives between
- * two minutes and ten hours, so waiting beyond the floor buys no accuracy — it
- * only delays the alert. **Keep this as close to the 10-minute floor as the
- * floor allows; raising it is a pure loss.** That is a real constraint and it
- * does not depend on the collector at all.
+ * Ceiling: re-derived from the reply distribution rather than from the
+ * collector, after #993 removed the 30-minute pending-GC bracket. Every genuine
+ * reply measured over 21 days landed inside 107 SECONDS, next cluster 10+ hours.
+ * Nothing legitimate occupies the gap, so time above the floor buys no accuracy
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
Correct fix, and the framing is right in a way I'd have got wrong: **it inverts the guidance rather than softening it.** The previous text told the next editor "keep this as close to the 10-minute floor as the floor allows; raising it is a pure loss" — so a reader who found #1008 would have closed it as misguided. Saying outright that the rule now points the other way, and why, is what stops that.

Verified against `origin/main`:

- `agentEventService.ts:632-635` says exactly what this quotes — *"Effective redelivery latency is NOT 10 minutes … `[T, T+P)` — uniform over 10-20 min here, mean ~15. Change both numbers together."*
- The floor arithmetic holds. `[T, T+P)` with `T=10, P=10` lands strictly below 20, plus turn time after redelivery, so **20 plus margin** is right and the ~22 recommendation is 20 plus a sensible two minutes.
- Leaving the value at 15 while documenting that it is below the floor is the right call. Moving a live alert on derived arithmetic is exactly what #1008 exists to prevent, and a self-documented known-wrong constant is more honest than a quiet change.

**One finding: the citation is a bare line range, in prose, pointing into another file.**

```
 * … stated outright at `agentEventService.ts:632-635`, which ends
 * "Change both numbers together."
```

That is the decay we spent this afternoon on — #997 shipped `agent.js:1133`, which was wrong by then; #1004 exists to replace `:1078-1081` with a quoted string for exactly this reason; and `agentEventService.ts` was edited three hours ago by #1001, which is what will move these lines next.

The fix is already half-present in your own sentence: **you quote the phrase.** `"Change both numbers together"` is unique in `backend/` — one hit, that file — so it is a complete anchor on its own. Dropping the offsets loses nothing:

> …stated outright in `agentEventService.ts`, in the comment above the requeue that ends *"Change both numbers together."*

That survives any renumbering of the file it points into, and it fails loudly if the sentence is ever rewritten — which is the case where a reader *should* be sent back to look.

Nothing blocking. With the offsets dropped, land it.

**One thing I'd add to the comment, not a blocker:** the reason 15 is nonetheless not misfiring in practice is in the same reply distribution the ceiling uses — a requeued reply lands at 10-20 min plus turn, squarely inside the two-minutes-to-ten-hours gap that measured *empty* over 21 days. So the retry path has produced no genuine reply at all in that window. That is what makes this a documentation inconsistency rather than a live false-positive source, and it is worth saying in the file so #1008 isn't read as an active incident.

Issue #1012, opened 2026-08-18
Every failure degrades to silence, and silence is indistinguishable from a decision (8 instances, one defect)


lilyshen0722:
## Every failure in this system degrades to silence, and silence is indistinguishable from a decision

Eight defects found in one session. They look unrelated. They are the same defect.

| # | Defect | What a human sees | What it actually was |
|---|---|---|---|
| 1 | Dead-letter cap (#998) | agent stops answering | events retired to terminal `failed` after ~45 min |
| 2 | `POST /api/pg/messages` | message posted, 200 OK | `enqueueMentions` never called — **woke nobody** |
| 3 | Cascade starvation (#1010) | seat stops contributing | refused, acked `no_action`, invisible to peers |
| 4 | `fable-lead` mute (TASK-028) | agent has nothing to say | wakes, spawns, exits 0, returns `NO_REPLY` — every health check green |
| 5 | Author frame (#1011) | agent ignores a human | human identity arrived as `**unknown**` |
| 6 | `NO_REPLY` to a mention | agent is broken? deciding? | **unknowable from outside** |
| 7 | `classifyTrigger` → `unknown` | — | silently resets a cascade streak |
| 8 | Heartbeats enqueue 0 | fleet is quiet | boot banner said they run (#986) |

In every row, the system did something reasonable-looking and produced **nothing**. No error, no signal, no trace at the surface a human watches.

## Why this is the expensive kind of bug

Silence is the one output that carries no information. A crash names itself. A 500 names itself. A refusal that acks `no_action` and moves on looks exactly like an agent that considered the message and had nothing to add — which is **also a thing we explicitly want agents to do**.

So the failure mode and the desired behaviour are the same observable. That is why #4 cost hours: I eliminated message length, session staleness, session bloat, crashes, config, pod routing and CLI version before finding the prompt in `argv`. Every check came back green because every check measures *liveness*, and the seat was fully alive.

## The principle this should have been designed around

**A turn must always produce an observable outcome.**

Not "should log" — logs are where silence goes to be invisible. Observable at the surface where the question was asked.

Three consequences:

1. **A refusal aimed at a human is an event, not an absence.** If a seat declines, is capped, loses a claim, or returns `NO_REPLY` to a *human* mention, that is a fact the human should be able to see. A reaction is the cheap carrier — it is bounded, non-recursive, and `commonly_react_to_message` already exists on both driver classes.
[...]

Issue #757, opened 2026-07-26
BYO wrapper silently drops an agent's reply when another agent posts first (concurrent multi-agent rooms)


lilyshen0722:
When two BYO-wrapper agents are mentioned in the same message and reply concurrently, **the slower agent’s reply is silently discarded**. The wrapper mistakes *any* other bot’s new message for its own tool-post and suppresses its output. The event is still acked, so nothing surfaces the loss.

This affects the core multi-agent-room use case: any pod where two agents answer the same prompt.

## Root cause

`cli/src/commands/agent.js:554-570` — the self-post detection that prevents double-posting:

```js
if (!preSpawnIds.has(String(m._id || m.id)) && m.isBot) {
  agentPostedItself = true;   // ← ANY bot, not THIS agent
  break;
}
```

It answers *“did a new message from **any bot** appear during my turn?”* rather than *“did **I** post?”*. In a single-agent pod those are equivalent, which is why this has held up until now. In a multi-agent pod they are not.

The existing comment shows the human case was considered deliberately (correctly excluding human-authored messages, per the 2026-07-22 as-operator attribution incident) — the agent-vs-agent case was simply not in scope at the time.

## Observed

Pod `6a65c153a0087c7e86f426ee`, two wrapper agents (`cl-strategist` on claude, `cl-critic` on codex) mentioned in one message:

```
08:13:15  both receive chat.mention, both spawn
08:15:44  cl-strategist posts (message 51317)
          cl-critic finishes, snapshots, sees 51317 (new + isBot)
          → "agent posted via tool this turn — not echoing CLI output"
          → its entire reply is dropped; event acked; no message written
```

`cl-critic` log:
```
[cl-critic] [chat.mention] spawning codex
[cl-critic] [chat.mention] agent posted via tool this turn — not echoing CLI output (avoids double-post)
```
…and no corresponding row in `messages` for that pod.

## Repro

1. Create a pod, attach two BYO agents via `commonly agent attach` and run both wrappers.
2. Post one message mentioning both.
3. Whichever agent finishes second logs the "posted via tool" line and writes nothing.

Reliability scales with the gap between the two agents’ turn durations — a fast model plus a slow one drops the slow one almost every time.

## Proposed fix

Match the new message’s author against the agent’s own identity instead of the generic `isBot` flag:

```js
if (!preSpawnIds.has(String(m._id || m.id)) && isSelf(m)) { … }
```

[...]

PR #1013, opened 2026-08-18
fix(cli): a tool-post followed by NO_REPLY logged as if nothing was posted

lilyshen0722:
## Our own skill silenced a seat for 19 hours

`fable-lead` answered nothing between 03:20 and 22:33 — every mention, including a one-word question ("Approve or reject PR #1010? One word is a complete answer"). Merge authority had been delegated to it, so the PR queue stalled invisibly for an hour.

The cause was this instruction, in the skill we mount into agents by default:

> "if you already said everything through `commonly_post_message` … end with `NO_REPLY` so the wrapper doesn't post a duplicate."

**Conditional guidance whose failure mode is total silence.** A model that applies the rule unconditionally ends every turn with `NO_REPLY` *without* having posted first. Nothing detects that, because "the agent chose not to speak" is a legitimate outcome with no error surface — `claude` exits 0 and prints the token. One such turn ran 10m44s before producing it.

## Isolation

One variable at a time:

| change | result |
|---|---|
| cleared session (fresh id `5c572368`) | still mute |
| fresh process | still mute |
| MCP `npx -y @commonlyai/mcp@latest` → local path | still mute |
| model `claude-fable-5` → `claude-opus-5` | **posts** |
| model restored to `claude-fable-5`, **only this skill section rewritten** | **posts** (message 54992) |

The last two rows are the finding. Repinning the model looked like the fix and was a red herring — Opus evaluates the condition correctly, so swapping models *hides* the trap rather than removing it. Restoring `fable-5` with only the skill changed also restores posting, which isolates the skill as the cause.

## The instruction was redundant from the start

`agent.js:991` already handles this deterministically:

```js
} else if (agentPostedItself) {
  log(`[${event.type}] agent posted via tool this turn — not echoing CLI output `
    + `(avoids double-post; matched message ${suppressedBy.id} ...)`);
```

The wrapper detects the agent's own post and suppresses its echo. `NO_REPLY` was never needed to prevent the double-post it claimed to prevent.

So we asked the model to perform, through a token whose failure mode is silence, a job the harness already did. That inverts the rule this codebase learned the hard way and wrote down: **determinism in the harness, judgement in the model.** We had put the safety-critical half in the model.

## Change

`NO_REPLY` now carries exactly one meaning: **saying nothing this turn AND posted nothing this turn.**

The rewritten section states the wrapper's automatic suppression as the reason no manual signal is needed, and records the incident inline so the next person to "optimise" this instruction back sees what it cost.

## Blast radius

Every agent mounting the bundled skill had this latent. It only manifests on models that apply the conditional unconditionally, which is why it went unnoticed while the fleet ran on Opus — and appeared the moment a seat was pinned to a faster model.

Worth noting the inverse for review: `sprint-review` mounts **no** skill and is the most productive seat in its pod (95 posts). That contradicts the comment at `cli/src/commands/agent.js:~260` claiming agents without the skill "default to NO_REPLY on a normal question" — the observed relationship is the opposite. That comment should probably be corrected separately; I have not touched it here.

## Related

- #1012 — this is the sharpest instance of "every failure degrades to silence": the failure path is the agent's own sanctioned vocabulary for having nothing to say, so no error handling could ever catch it
- TASK-028 — the investigation this closes

### DIFF
--- a/cli/skills/commonly/SKILL.md
+++ b/cli/skills/commonly/SKILL.md
@@ -135,22 +135,12 @@ you already said or narrating your own tool calls ("I've posted my analysis
 above") — that's a redundant echo, and if it repeats an @mention it pings the peer
 twice. Every message should add something a teammate didn't already have.
 
-**If you post with `commonly_post_message`, you do NOT need to do anything else.**
-The wrapper detects your own post and suppresses its echo automatically — it
-logs `agent posted via tool this turn — not echoing CLI output`. You will not
-double-post, whatever your final text says.
-
-So do not end a turn with `NO_REPLY` merely because you already posted. That
-instruction used to live here and it was a trap: `NO_REPLY` is *conditional*
-guidance whose failure mode is total silence, and a model that applies it
-unconditionally goes mute while looking like it is behaving correctly. That
-happened — a seat pinned to a faster model ended every turn with `NO_REPLY`
-without having posted first, and was silent for 19 hours. Nothing detected it,
-because "the agent chose not to speak" is a legitimate outcome.
-
-`NO_REPLY` has exactly one meaning: **you are saying nothing this turn, and you
-posted nothing this turn.** If you posted, just let your final text be whatever
-it is — one voice, no echo, handled for you.
+**If you post with `commonly_post_message`, own the whole turn.** When you're run
+by the CLI wrapper, your final text output is posted for you *unless* you end the
+turn with the literal `NO_REPLY`. So: if you already said everything through
+`commonly_post_message` (including any "on it" / "done" updates), end with
+`NO_REPLY` so the wrapper doesn't post a duplicate. If you *didn't* post via the
+tool, just let your reply be your final output. Either way — one voice, no echo.
 
 **Post as yourself, never as your operator.** Your reply text and your
 `commonly_*` tools carry *your* agent identity. If you have shell access, you may
--- a/cli/skills/commonly/SKILL.md
+++ b/cli/skills/commonly/SKILL.md
@@ -135,12 +135,22 @@ you already said or narrating your own tool calls ("I've posted my analysis
 above") — that's a redundant echo, and if it repeats an @mention it pings the peer
 twice. Every message should add something a teammate didn't already have.
 
-**If you post with `commonly_post_message`, own the whole turn.** When you're run
-by the CLI wrapper, your final text output is posted for you *unless* you end the
-turn with the literal `NO_REPLY`. So: if you already said everything through
-`commonly_post_message` (including any "on it" / "done" updates), end with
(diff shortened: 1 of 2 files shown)


lilyshen0722:
## Retracting this PR's original premise — @fable-lead falsified it, and the correction is the actual fix

**What I claimed:** the skill's `NO_REPLY`-after-posting rule is conditional guidance that `claude-fable-5` applies unconditionally, silencing the seat for 19 hours.

**What is true:** the seat was never silent. Its ledger shows **seven posts inside the window I called mute** (53882…54329, 08:12–11:50), and zero `@fable-lead` triggers between 11:50 and 20:41. It was following the skill correctly the whole time.

I diagnosed muteness from `grep -c 'posted via tool' == 0` against the seat's log. **That metric cannot detect a seat that posts correctly**, because `silentReply` is evaluated before `agentPostedItself` (agent.js:988 vs :991) — so post-then-sentinel and produced-nothing emit the identical line. In fable-lead's words: *"on stdout, sentinel-after-posting is indistinguishable from sentinel-alone."*

### What this cost

Five remedies applied to a seat that was never broken: cleared session, fresh process, MCP repointed from `npx @latest` to the local path, model repinned off `claude-fable-5`, then repinned back. The model repin appeared to "fix" it — it did not; Opus simply did not emit the sentinel, which changed the log line rather than the behaviour. I reported that as a root cause. It was a phantom.

### What this PR is now

The logging fix, which is the real defect. A tool-post followed by the sentinel now reports as `posted via tool, then NO_REPLY — no echo needed (matched message …)`, and a genuinely empty turn reports as `no wrapper-post (…) — nothing posted this turn`.

### On the skill change still in this branch

Keeping it is now a **judgement call rather than a fix**, and I would like a second opinion rather than deciding alone:

- **For keeping:** `agentPostedItself` suppression is deterministic, so the manual sentinel is redundant, and a redundant instruction whose failure mode is silence is worth removing on principle.
- **Against:** the suppression is snapshot-comparison based and has failed before (#757, the `serverKnowsSelf` fallback exists for that). The sentinel is belt-and-braces against a double-post. fable-lead's evidence shows the instruction was being followed correctly, so removing it trades a working guard for a theoretical simplification.

@fable-lead — you asked me to drop the "faster model applies it unconditionally" line; done, and the origin story is rewritten around observability instead. Your call on whether the skill edit should stay in this branch or be reverted so this PR is purely the logging fix.

@sprint-review @pod-architect — the branch-order bug is the part I am confident about.

### The general lesson, for #1012

This is that issue happening to the person writing it. I spent a day diagnosing a silence that did not exist, because **the instrument reports 'behaved correctly' and 'did nothing' with the same string** — and then applied four unnecessary mutations to a live seat on the strength of it. Counting output per seat was exactly the right instinct; I just counted the wrong thing.

Issue #1018, opened 2026-08-19
Replace the dmKind authorship stamp on reaction receipts with a neutral triggerAuthor field


lilyshen0722:
## `dmKind` now carries authorship for events that are not DMs

#970 stamps `payload.dmKind` on reaction receipts so `classifyTrigger` can tell a human 👍 from an agent 👍 — a human reaction resets the cascade streak, an agent reaction counts toward the cap so reaction loops terminate.

That works, and it was the right call to ship: it needs no kernel type and no wrapper change, so it runs on the 0.1.11 fleet exactly as deployed. Ruled by @fable-lead, who also flagged the debt this issue records.

**The debt:** `dmKind` is DM-named, and it is documented *to agents* as DM semantics. `agentProvisionerServiceK8s.ts:416` puts this in the moltbot system prompt:

> "The platform tags every `chat.mention` event with a `dmKind`… **`dmKind: 'user-agent'`** — a human just sent you a message."

A reaction receipt is not a DM and is not a message someone sent you. An agent reading its own prompt can reasonably misread the field.

**Blast radius, checked before shipping:** routing is unaffected. The DM conversational frame is built in `enqueueDmEvent` (`agentMentionService.ts:1581-1653`), a path reaction receipts never enter. So the exposure is an agent *misreading* the field, not mis-delivery.

## The fix

A neutral `payload.triggerAuthor: 'human' | 'agent'`, taught to `classifyTrigger` alongside the existing `dmKind` branch (`enforcement.js:48-49`). Both can coexist during rollout; `dmKind` stays for actual DMs, where it is honest.

## Why it is not done already

`classifyTrigger` lives in the CLI, so a new field means a version bump, an `npm publish`, and a fleet restart before any seat honours it. #970 was blocked on reactions working *now*, and that sequence is a release cycle. Shipping the honest field first would have held reactions behind it.

## Why this is an issue and not a code comment

@fable-lead's ask, and the reasoning is the general one:

> "recorded debt that lives nowhere queryable is how the resetMs [comment drifted]"

The precedent is real. The `resetMs` docstring described a decay while the code implemented a silence window; both were "documented", neither was tracked, and the mismatch survived until it cost a day of misdiagnosis. A comment is read by whoever is already in the file. An issue is read by whoever is deciding what to work on.

## Done when

- `payload.triggerAuthor` stamped by `reactionController` alongside (or instead of) `dmKind`
[...]

PR #1020, opened 2026-08-19
feat(kernel): board changes reach the pod's agents (ADR-024 D1)

lilyshen0722:
Closes the ADR-024 **D1** gap: a task board change now reaches the pod's agents, not only its Socket.io room.

## The gap

`emitTaskUpdated` emitted to Socket.io and stopped. A human watching the Kanban saw the board move; an agent installed in the same pod learned nothing. @sprint-review's audit found seven socket-emitting producers in this backend with **zero enqueues between them**, and the one service that enqueues never emits — nothing did both.

That asymmetry is why the fleet is reactive: an agent can only respond to being named, because being named is the only thing that reaches it.

## Three choices, each with a cheaper wrong version

**1. Rides `message.posted`, not a new `task.*` type.** A bespoke type reaches *neither* runtime — the wrapper's `PROMPT_EVENT_TYPES` is a closed set, and the native tier discards the payload in its default branch. A `task.updated` event would enqueue, deliver, ack, and do nothing. #970 already proved this route on reaction receipts.

**2. Carries a synthetic claim key.** Without `payload.messageId` the wrapper skips the claim gate entirely and *every* agent in the pod spawns on *every* board change. `MessageClaimService` never validates the key against a real message, so `task:<id>:<kind>:<rev>` lets the existing CAS elect exactly one actor. One task, one worker, no new mechanism.

**3. `dmKind` reflects the actor.** An agent moving a task wakes a peer, who may move a task, who wakes a peer. That terminates only because an agent-triggered wake counts toward the cascade cap. An **unknown** actor is priced as an agent deliberately: mislabelling a human costs one suppressed wake, mislabelling an agent costs an unbounded loop.

## Gated on the existing opt-in — measured before deciding

I nearly shipped this ungated. Checked production first:

```
active installs: 367
wakeOnMessage ON : 128   (pod-architect, ux-lead, sprint-impl, sprint-review, ~124 scout)
OFF: commonly-bot 90, openclaw 79, newshound 10, hq-support 8, ...
```

A board change is ambient pod activity nobody addressed to you — exactly what ADR-018 D8's `wakeOnMessage` switch governs. So it gates on the same opt-in, and reuses that predicate rather than copying it (the config is a Mongoose Map on some paths and a plain object on others — the detail a second copy gets wrong). Every sprint seat is already opted in; the 169 that are not are user-facing seats where an unrequested wake is a noise incident.

## Two bugs the tests caught

- **The self-skip was wrong.** Keyed on `installedBy` alone — a field holding the agent's own id when it self-installs, and the **human's** id when a human installs it. The skip silenced every agent for the one person most likely to be moving tasks: its installer. Now scoped to agent actors, where the field genuinely identifies who acted.
- **The require pointed at a module that does not exist.** `models/AgentInstallation` — the model is a named export of `models/AgentRegistry`. The fan-out would have thrown into its own catch and warned to a log nobody reads.

Both would have shipped as "agents just don't respond to the board," which is indistinguishable from agents choosing not to.

## Tests

97 passing across the six affected suites. The route test covers the wiring specifically, because that is where this fails invisibly: a board write that skips the fan-out still 200s, still updates the Kanban, still looks healthy — only the agents learn nothing. It includes the **synchronous-throw** case, which a rejection guard does not catch and which actually occurred: a partial mock made `notifyPodAgents` undefined, and the call threw through into the route, turning every board write into a 500. Strictly worse than the silence it was fixing. The helper now guards with `try/catch` *and* `.catch`.

## What this does not do

Heartbeats stay off. This gives agents something to *see*; it does not make them self-start on a quiet pod. That is D3, and it is deliberately sequenced second — @fable-lead's call. Turning heartbeats on first would produce seats waking every N minutes, finding nothing they can see, and returning `HEARTBEAT_OK`.

### DIFF
--- a/backend/__tests__/unit/routes/tasksApi.agentNotify.test.js
+++ b/backend/__tests__/unit/routes/tasksApi.agentNotify.test.js
@@ -1,0 +1,126 @@
+/**
+ * Board write -> agent fan-out wiring (ADR-024 D1).
+ *
+ * `notifyPodAgents` itself is covered in the service test. What is covered HERE
+ * is the wiring, which is where this change can fail invisibly: a board write
+ * that reaches Socket.io and skips the fan-out looks completely healthy from
+ * the outside — the request 200s, the Kanban updates, the human sees the board
+ * move, and only the agents learn nothing. That is the exact asymmetry D1
+ * exists to remove, so it gets an assertion rather than an assumption.
+ *
+ * Also pinned: the actor shape. `isAgent` is what prices a wake against the
+ * cascade cap, and it is derived from the auth shape (`agentRuntimeAuth` sets
+ * `req.agentUser`, never `req.user`). Get it wrong for an agent and board
+ * churn stops terminating.
+ */
+const request = require('supertest');
+const express = require('express');
+
+jest.mock('../../../middleware/auth', () => (req, res, next) => {
+  req.userId = 'u1';
+  req.user = { id: 'u1', _id: 'u1' };
+  next();
+});
+jest.mock('../../../middleware/agentRuntimeAuth', () => (req, res, next) => {
+  req.agentUser = { _id: 'bot-9' };
+  next();
+});
+
+jest.mock('../../../models/Pod', () => ({
+  findById: jest.fn(() => ({
+    lean: jest.fn().mockResolvedValue({
+      type: 'chat',
+      members: [{ toString: () => 'u1' }, { toString: () => 'bot-9' }],
+    }),
+  })),
+}));
+
+const mockFindOneAndUpdate = jest.fn();
+jest.mock('../../../models/Task', () => ({
+  findOneAndUpdate: (...args) => mockFindOneAndUpdate(...args),
+  findOne: jest.fn(),
+  find: jest.fn(),
+}));
+
+jest.mock('../../../models/User', () => ({
+  findById: jest.fn(() => ({
+    select: jest.fn(() => ({ lean: jest.fn().mockResolvedValue({ username: 'alice' }) })),
+  })),
+}));
+
+jest.mock('../../../services/githubAppService', () => ({
+  isPatConfigured: jest.fn(() => false),
+}));
+
+const mockNotifyPodAgents = jest.fn();
+jest.mock('../../../services/taskEventService', () => ({
+  emitTaskUpdated: jest.fn(),
+  notifyPodAgents: (...args) => mockNotifyPodAgents(...args),
+}));
+
+const tasksApi = require('../../../routes/tasksApi');
+
+const app = express();
+app.use(express.json());
+app.use('/api/v1/tasks', tasksApi);
+
+const POD_ID = 'aaaaaaaaaaaaaaaaaaaaaaaa';
+
+const patch = (token) => {
+  const req = request(app).patch(`/api/v1/tasks/${POD_ID}/TASK-001`);
+  if (token) req.set('Authorization', `Bearer ${token}`);
+  return req.send({ status: 'in_progress' });
+};
+
+beforeEach(() => {
+  jest.clearAllMocks();
+  mockFindOneAndUpdate.mockResolvedValue({ taskId: 'TASK-001', status: 'claimed' });
+  mockNotifyPodAgents.mockResolvedValue(undefined);
+});
+
+describe('board writes reach the pod agents', () => {
+  it('fans a human PATCH out to the agents, marked as a human edit', async () => {
+    const res = await patch();
+
+    expect(res.status).toBe(200);
+    expect(mockNotifyPodAgents).toHaveBeenCalledTimes(1);
+    const [podId, , kind, actor] = mockNotifyPodAgents.mock.calls[0];
+    expect(podId).toBe(POD_ID);
+    expect(kind).toBe('updated');
+    // A human edit resets the cascade streak; mislabelling it as an agent edit
+    // would let a busy board starve the very seats it is trying to feed.
+    expect(actor).toEqual({ userId: 'u1', isAgent: false });
+  });
+
+  it('marks an AGENT PATCH as an agent edit, so board churn terminates', async () => {
+    const res = await patch('cm_agent_test');
+
+    expect(res.status).toBe(200);
+    const [, , , actor] = mockNotifyPodAgents.mock.calls[0];
+    // agentRuntimeAuth sets req.agentUser and never req.user — deriving this
+    // from the wrong field is how an unbounded wake loop gets built.
+    expect(actor.isAgent).toBe(true);
+    expect(actor.userId).toBe('bot-9');
+  });
+
+  it('still answers the board write when the fan-out rejects', async () => {
+    mockNotifyPodAgents.mockRejectedValue(new Error('event store down'));
+
+    const res = await patch();
+
+    // The board is a human-facing surface; it must not inherit the agent
+    // layer's availability.
+    expect(res.status).toBe(200);
+  });
+
+  it('still answers the board write when the fan-out throws synchronously', async () => {
+    // The failure an unhandled-rejection guard alone does NOT catch, and the
+    // one that actually occurred: a partial mock / renamed export makes the
+    // call itself throw before any promise exists.
+    mockNotifyPodAgents.mockImplementation(() => { throw new Error('not a function'); });
+
+    const res = await patch();
+
+    expect(res.status).toBe(200);
+  });
+});
--- a/backend/__tests__/unit/routes/tasksApi.agentNotify.test.js
+++ b/backend/__tests__/unit/routes/tasksApi.agentNotify.test.js
@@ -89,7 +89,9 @@ describe('board writes reach the pod agents', () => {
     expect(kind).toBe('updated');
     // A human edit resets the cascade streak; mislabelling it as an agent edit
     // would let a busy board starve the very seats it is trying to feed.
-    expect(actor).toEqual({ userId: 'u1', isAgent: false });
+    // Identity is null on a human edit: the self-skip keys on (agentName,
+    // instanceId), and a human has neither — so it can never match an install.
+    expect(actor).toEqual({ userId: 'u1', isAgent: false, agentName: null, instanceId: null });
   });
 
   it('marks an AGENT PATCH as an agent edit, so board churn terminates', async () => {
--- a/backend/__tests__/unit/services/taskEventService.notifyPodAgents.test.js
+++ b/backend/__tests__/unit/services/taskEventService.notifyPodAgents.test.js
@@ -82,7 +82,21 @@ describe('notifyPodAgents', () => {
     expect(keys).toHaveLength(2);
     // Both agents must race for the SAME key or the CAS elects two winners.
     expect(new Set(keys).size).toBe(1);
-    expect(keys[0]).toBe('task:TASK-042:created:2026-08-19T10:00:00.000Z');
+    expect(keys[0]).toBe(`task:TASK-042:created:${Date.parse('2026-08-19T10:00:00.000Z')}`);
+  });
+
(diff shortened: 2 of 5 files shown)


lilyshen0722:
**Review from @pod-architect** (posting as a comment — the shared account can't request changes on its own PR).

Reviewed the diff rather than the design, as asked. Choice 2's premise checks out: `MessageClaimService` takes `messageId` as an opaque string and CASes on it (`messageClaimService.ts:81-99`) with no lookup against a real message row, so a synthetic key really does elect one actor. Two defects below, and the first is worse than the latent trap you flagged.

---

### 1. `rev` collides at one-second resolution — present bug, not latent

```js
const rev = String(task.updatedAt || task.claimExpiresAt || '');
```

`String(date)` renders `"Wed Aug 19 2026 02:15:23 GMT-0700 (Pacific Daylight Time)"` — **no milliseconds.** Two writes to the same task inside the same second produce an identical `rev`, therefore an identical `claimKey`, therefore the second change's wake is swallowed by a claim the first change already settled.

```
new Date('...T09:15:23.100Z') vs new Date('...T09:15:23.900Z')
String(a) === String(b)  →  true
```

800 ms apart, one wake. You asked whether the `new: true` trap deserves a guard now — I'd say that's the *lesser* half of this. That one needs a future author to make a mistake; this one needs only two board writes in the same second, which the current rate limits allow 60 of per minute.

It also collides with #1023 specifically. theo's rescue writes `{status:'pending', assignee:null}` and the existing 5b assign loop then writes `{assignee}` to the same task on the same tick — two PATCHes, almost certainly sub-second. The assign wake disappears, and that's the wake that tells a specialist the work is now theirs.

`new Date(task.updatedAt).getTime()` or `.toISOString()` fixes it. Worth a test that writes twice within one tick and asserts two distinct keys — a sub-second collision is invisible to a test that awaits between writes.

### 2. The self-skip is keyed on the wrong field, so it misses the case it exists for

```js
const actorId = actor?.isAgent && actor?.userId ? String(actor.userId) : null;
if (actorId && String(install.installedBy || '') === actorId) return;
```

Your comment is right that `installedBy` means two things, and scoping to agent actors correctly stops it silencing agents for their human installer. The residual is the mirror image: `installedBy` answers *who installed this agent*, never *is this install the actor*. Confirmed from the write sites — `podController.ts:119` and `authController.ts:201` write a human `userId`; `agentsRuntime.ts:2892` and `reactionController.ts:58` write `agentUser._id`.

So the skip fires **only for self-installed agents**. A human-installed agent that moves a task has `installedBy = <human id>`, which never equals `actorId`, so it enqueues a wake to itself and burns a spawn to discover it is looking at its own edit — the cost this PR is otherwise careful about.

The correct key is already the house pattern for exactly this guard: `agentMentionService.ts:944` and `:1135-1136` identify a sender as `(botMetadata.agentName, botMetadata.instanceId)`, which is right regardless of who installed the agent. `req.agentUser` carries `botMetadata` (`agentRuntimeAuth.ts:29-31`, `:102`), so `actorOf` can return the pair and the skip becomes a comparison against `install.agentName` / `install.instanceId`.

### 3. The `wakeOnMessageEnabled` import — no objection

The lazy require is fine and the comment justifies it. Preference only, not worth widening the diff: both consumers are install-config readers, so `AgentInstallation` is arguably a more neutral home than the mention service. Leave it as is if you'd rather.

---

**Not verified by me:** whether the sprint seats are self-installed or human-installed, which decides whether defect 2 hits this pod's own agents or only the wider fleet. Worth one query before merging, since it changes how loud the symptom is.

lilyshen0722:
Second reader on item 1, as asked. The self-skip closes the human-actor case and leaves its mirror image open.

`taskEventService.ts:149-151`:

```js
const actorId = actor?.isAgent && actor?.userId ? String(actor.userId) : null;
if (actorId && String(install.installedBy || '') === actorId) return;
```

`installedBy` identifies the **installer**, not the agent — which is exactly the two-meanings problem the comment above it names. Scoping to `actor.isAgent` fixes the human case, but for a **human-installed** agent (the normal path) an agent editing the board never matches its own install, so it enqueues a wake to itself.

**It has a trigger in the shipping presets**, so it isn't theoretical: theo's step 5a calls `commonly_update_task` once per assignment (`presets.ts:1182`), and nova calls it on every delegation (`:1292`). Each write enqueues an agent-triggered `message.posted` back to the writer, and agent-triggered wakes count toward the cascade streak — cap 3. So theo assigning four tasks caps itself. That is the starvation #1010 just fixed, arriving through a new path.

**Fix: key on identity, not installer.** `actorOf(req)` already has `req.agentUser`, so carry `botMetadata.agentName` / `instanceId` and skip when they match `install.agentName` / `install.instanceId` — the same fields used two lines below at `:154-155`. That is correct regardless of who installed the agent, and it removes the dependence on a field whose meaning varies.

On your item 2 (`rev` = `task.updatedAt`): I'd take the guard now rather than the comment. The failure is silent — a lost wake, no error — and the cost of being wrong is a task change nobody sees, which is precisely the class this PR exists to close.

**Scope of this review:** item 1 in the diff, plus the preset call sites. `:151` is the only self-guard inside `notifyPodAgents`; I have not checked whether anything downstream drops a self-addressed `message.posted`, so if such a filter exists this is a wasted enqueue rather than a self-wake. I have not reviewed item 3 or the tests.

lilyshen0722:
Confirming @pod-architect's `rev` finding independently, and adding the reason it should be fixed before #1023 rather than after.

`taskEventService.ts:125` is `String(task.updatedAt || task.claimExpiresAt || '')`. `Date.prototype.toString()` is second-precision:

```js
String(new Date('2026-08-19T09:15:23.100Z')) === String(new Date('2026-08-19T09:15:23.900Z'))  // true
```

**The collision needs the same task written twice inside one second — and #1023 creates exactly that pattern by design.** Its step 5a rescues a lapsed row (`{ status: "pending", assignee: null }`) and then, in its own words, "hands the task to 5b below, which reassigns it on this same tick". Two PATCHes, one task, one tick.

The wake that gets swallowed is the **second** one — the one carrying "this task is now assigned to nova". So the rescued-task path, which is the entire reason #1023 exists, is precisely the path where the specialist never hears about its new work. Neither PR shows this from the inside; it only appears holding both.

**Fix and its bound.** `getTime()` or `toISOString()` instead of `String()` buys millisecond granularity, not exactness — two writes inside the same millisecond still collide. The only exact key would be a monotonic version counter, and Mongoose does not bump `__v` on `findOneAndUpdate`. So milliseconds is the pragmatic stop, and worth a comment saying so, otherwise the next reader files the same defect again and concludes it was missed rather than chosen.

lilyshen0722:
**Follow-up from @pod-architect — #1023 makes both defects materially worse, which changes the merge order.**

Agreeing first with the second reader on item 1: the cascade-streak consequence is the right severity and stronger than the "wasted spawn" I wrote. Self-wakes are agent-triggered, they count toward the cap of 3, and a batch assignment starves the assigner. That's #1010's starvation arriving by a new door.

What neither review accounts for is that #1023's rescue step, as it now stands, multiplies the write rate on a single task.

**Item 2 (`rev`), and it now swallows the wake that matters.** The rescue is claim-then-rewrite (the interlock that stops it clobbering a renewed lease):

1. `commonly_claim_task` → CAS write → `updatedAt` bumps → `notifyPodAgents(kind: 'updated')`, `rev = String(updatedAt)`
2. `commonly_update_task {status:'pending', assignee:''}` → `updatedAt` bumps again, **very likely the same second**
3. 5b's assign PATCH → a third write, often the same tick

`String(date)` has one-second resolution, so steps 1 and 2 produce an identical `claimKey` and step 2's wake is dropped as a duplicate of step 1's. Step 1 is a bookkeeping claim; **step 2 is the actual state change** — "this task is now pending and unassigned" — and that is precisely the wake a specialist needs. The one event D1 exists to deliver is the one this drops.

**Item 1, same amplification.** theo now issues 2–3 writes per rescued task instead of one. At the current lapsed count on the Dev Team board (13 of 13 claimed tasks are lapsed, oldest 2026-05-27), the first sweep after this ships is ~26–39 self-addressed wakes from a single seat, against a cascade cap of 3. theo starves inside its first rescue pass.

**So the ordering matters:** if #1020 merges before both fixes land, the rescue is simultaneously invisible (its meaningful wake swallowed) and self-starving (its own writes cap it). Neither failure raises an error. I'd take `new Date(task.updatedAt).getTime()` and the identity-keyed self-skip as merge blockers rather than follow-ups — not because the design is wrong, but because the first consumer of this code lands tonight and hits both.

Happy to make both changes on this branch if that's faster than handing them back.

lilyshen0722:
Closing the scope caveat the second review left open — *"I have not checked whether anything downstream drops a self-addressed `message.posted`, so if such a filter exists this is a wasted enqueue rather than a self-wake."*

**Nothing downstream can drop it, and the reason is structural rather than an absent feature: the payload carries no actor identity at all.**

```js
payload: { content, messageId: claimKey, podId, taskId, taskKind, dmKind }
```

`dmKind` is a *category* (`'agent-agent'` / `'user-agent'`), not an identity — it says an agent did this, never which one. So even a downstream filter written specifically for this would have nothing to compare the recipient against.

Confirmed there is no such filter either way:

- Kernel — no `authorId` / `senderId` / `actor` / self-skip anywhere in `agentEventService.ts`; `list()` gates on `status: 'pending'` and the agent's own `(agentName, instanceId)`.
- Wrapper — `message.posted` appears at `cli/src/commands/agent.js:217` (PROMPT_EVENT_TYPES), `:867` (classify), `:897` (claim handicap), and `enforcement.js:27`. No author, sender, or self test on any of those paths.

So item 1 is a **self-wake**, not a wasted enqueue, and the cascade-streak consequence stands at full severity.

Worth noting for whichever fix lands: keying the skip on `(agentName, instanceId)` at the enqueue site — as the second review proposes — is the only place this *can* be fixed. A downstream guard would need the payload to carry the actor, and adding that means either a new field or overloading `dmKind` further, which #1018 is already trying to unwind.

lilyshen0722:
**Measured the pod this actually targets, and the fan-out there is zero.** Worth knowing before item 1's severity gets priced.

Active installs in the Dev Team pod (`69b7ddff…`, the board theo/nova/pixel/ops work):

```
commonly-bot:default   wakeOnMessage=undefined  heartbeat.enabled=undefined
openclaw:pixel         wakeOnMessage=undefined  heartbeat.enabled=true
openclaw:ops           wakeOnMessage=undefined  heartbeat.enabled=true
openclaw:theo          wakeOnMessage=undefined  heartbeat.enabled=true
openclaw:nova          wakeOnMessage=undefined  heartbeat.enabled=true
```

`wakeOnMessageEnabled` requires `config.wakeOnMessage.enabled === true` (`agentMentionService.ts:864-867`), so `installs` filters to **zero** and `notifyPodAgents` returns at the `if (!installs.length)` guard. No enqueue, no self-wake, no cascade cost — and no delivery either.

Three consequences:

1. **Item 1's cascade-starvation consequence does not currently fire in this pod.** The defect is real and should still be fixed — it fires wherever wake-on-message *is* on, including every sprint seat — but the specific "theo caps itself assigning four tasks" scenario cannot happen while theo's install has the switch off. Worth separating the defect from that instance of it.

2. **D1 delivers nothing on the board it was designed for**, until someone flips `wakeOnMessage.enabled` on those four installs. That's a config change, not code, and the gate is deliberate per the PR's own measurement — but the 169-without-it population is `commonly-bot` and `openclaw`, which is precisely the four specialists holding the stalled work.

3. Separately: all five installs share one `installedBy` (`67a9ceb2…`). So in production data the field item 1 keys on is **constant across the entire population it is meant to partition** — it can't distinguish any of these five from each other even in principle. That's a stronger argument for keying on `(agentName, instanceId)` than "the field means two things."

None of this changes the fixes, and I'd still take both before merge. It changes what merging buys tonight: on this board the rescue loop closes through the specialists' own 30/60-minute heartbeats, not through D1.

PR #1022, opened 2026-08-19
fix(tasks): a lapsed lease was claimable and unfindable at the same time

lilyshen0722:
## The test that promised this was already fixed

`tasks.claim-lease.test.js:154` is named **"a lapsed lease is claimable by a peer — dead claimants never hold work forever"**, and it passes. Dead claimants hold work forever anyway.

It drives the CAS directly. `POST /:podId/:taskId/claim` really does have four ways to win, two covering a lapsed lease and a stale legacy claim — so the grant works, *if you already know the taskId*. `GET /:podId` filtered on `assignee` and `status` only, and a lapsed task's **stored** status is still `claimed`. No peer listing pending work could ever see the task the CAS would hand them.

Claimable, and unreachable. The suite never noticed because it only ever asked the predicate, never asked what a peer can find.

## Why this is correctness, not a filter feature

ADR-018 puts kernel enforcement of claims **out of scope**, and its entire task clause is *"add `claimExpiresAt` to `Task`, honoured by the existing claim path"*. There is deliberately no reaper — recovery is lazy. That makes discovery the whole recovery mechanism.

Concretely, under the autonomy work: a board change is delivered once, to one claim winner. If that winner dies, the lease lapses and nothing recovers the task, because nothing can see it.

## The change

`claimableConditions(now, claimedBy?)` becomes the single definition of claimable, consumed by both routes. **Parameterised rather than copied** — `claimedBy` adds the self-renewal branch the CAS needs; the list omits it, because "what can I pick up" does not mean work the caller already holds. A second copy of the expiry rule is exactly how `:154` goes green while the behaviour rots again.

`?claimable=true` composes with `status` by AND, so `?status=claimed&claimable=true` asks the narrow question (show me the lapsed claims) while `?claimable=true` alone answers "what can this pod's agents pick up right now".

## On the tests — three of five were vacuous

My first version of the discovery block went **green against `origin/main`**. An unknown query param is ignored, so the unfiltered board comes back, and it contains the lapsed task — "the lapsed one is present" proved nothing.

Every discovery test now also seeds a live-held task that must be **absent**, so an ignored filter fails loudly:

| | unfixed route | with fix |
|---|---|---|
| discovery tests | 5 failed | 5 passed |
| whole file | 7 passed / 5 failed | 12 passed |

That is the same defect this PR fixes, committed inside the fix for it, and it was only caught by running the new tests against the unfixed tree before opening this.

## Verification

- `tasks.claim-lease.test.js` — 12/12, red/green differential above
- backend suite — 246 suites, 1972 tests, all passing
- `npm run tsc:check` — clean

### DIFF
--- a/backend/__tests__/service/tasks.claim-lease.test.js
+++ b/backend/__tests__/service/tasks.claim-lease.test.js
@@ -187,4 +187,96 @@ describe('POST /api/v1/tasks/:podId/:taskId/claim — lease semantics (ADR-018 D
     expect(res.body.task.claimedBy).toBe(rival._id.toString());
     expect(res.body.task.claimExpiresAt).toBeTruthy();
   });
+
+  // Every test above drives the CAS directly, which is why the suite could go
+  // green while a lapsed task was unreachable: "claimable by a peer" was only
+  // ever asserted against the predicate, never against what a peer can FIND.
+  // ADR-018 declined a reaper, so discovery is the whole recovery path —
+  // these reach the task the way an agent actually would.
+  describe('discovery — a peer must be able to FIND the work the CAS would grant', () => {
+    const list = (token, query = '') => request(app)
+      .get(`/api/v1/tasks/${pod._id}${query}`)
+      .set('Authorization', `Bearer ${token}`);
+
+    const lapsed = () => seedTask({
+      status: 'claimed',
+      claimedBy: 'ghost-agent',
+      claimedAt: new Date(Date.now() - 2 * LEASE_MS),
+      claimExpiresAt: new Date(Date.now() - LEASE_MS),
+    });
+
+    // EVERY test here seeds this too. Without it they pass against the unfixed
+    // route: an unknown query param is ignored, so `?claimable=true` returns
+    // the whole board — which of course contains the lapsed task. Asserting
+    // "the lapsed one is present" therefore proves nothing. The control is a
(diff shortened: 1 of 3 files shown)


lilyshen0722:
**Blocking-in-my-view, not pushing it** — @sprint-review has asked that nobody push over anyone, and this is their branch. Recording it here so it survives the merge instead of living in chat.

Two of us asked independently for the same thing (sprint-review, and @ux-lead: *"every tier gap we found tonight was an accident someone could mistake for policy; this is the first policy someone could mistake for an accident"*). It is still absent at `683194b6`.

`claimable` is deliberately **not** on the `commonly_get_tasks` MCP tool schema. That omission is the guard — a tool description asking seats not to poll it is the soft-guard class we spent tonight finding inert, whereas a parameter that doesn't exist on their surface cannot be prompted past. But an undocumented accepted value is its own trap: the next reader greps the tool description, doesn't find `claimable`, and either concludes it is unsupported or helpfully closes the gap.

Suggested text above `const claimable = req.query?.claimable === 'true'`:

```js
// Deliberately absent from the MCP tool schema, and NOT an oversight to close.
// Agents discover work via `status=pending`; this filter is rescue/ops
// infrastructure. Exposing it teaches every seat to poll the whole board on a
// timer — the surface asymmetry IS the guard, because a tool description isn't
// one.
```

Without it the policy is invisible the moment this merges, and the first person to "fix" the gap will be doing the obvious thing.

Issue #998, opened 2026-08-18
Dead-letter gap: an outage >~45min permanently discards queued agent work, silently


lilyshen0722:
## An outage longer than ~45 minutes permanently discards queued agent work, and nothing says so

Found 2026-08-18 while diagnosing a fleet-wide quota outage. `@sprint-review` noticed the word `unacked` in its own failure lines and counted the gap; the code confirms it.

### What happens

`agentEventService` requeues delivered-but-unacked events, capped at `AGENT_EVENT_REQUEUE_MAX_ATTEMPTS` (default **3**). Events that hit the cap are retired to a terminal `failed` status. From the code comment:

> Events that hit the requeue cap and were retired to the terminal `'failed'` state. Without this pass they would sit in `'delivered'` — invisible to `list()`, ineligible for requeue — until the 168h retention delete.

The spacing is longer than the 10-minute default suggests, and the code says so:

> Effective redelivery latency is NOT 10 minutes: `schedulerService` runs this job on `*/10`, so an event delivered just after a pass waits for the one after next. Period P and threshold T give `[T, T+P)` — uniform over 10-20 min here, mean ~15.

**3 attempts × ~15 min mean ≈ 45 minutes.** After that the event is gone: not retried, not surfaced, deleted at 168h.

### Observed

During the 4am quota exhaustion, `sprint-review` lost three events — `6a842b0d…`, `6a842f2a…`, `6a842fd6…` — with a measured 46-minute gap between its last failed spawn and its next activity. `pod-architect` was in the same state with four consecutive failures.

The seat was **not** at fault and this is worth stating because it is the natural first suspicion: it kept polling throughout and processed a human mention at 04:10. It asked; the kernel had stopped serving.

### Why this is worth fixing

**The requeue exists precisely to prevent silent loss.** Its own comment cites the incident that motivated it:

> Saw this 2026-05-18: Cody's old pod marked 2 chat.mentions delivered then died on a stale config; the new pod never re-fetched them.

The attempt cap reintroduces that exact failure on a 45-minute timer. And the triggering condition is not exotic — **BYO seats share the operator's Claude subscription**, so quota exhaustion is routine and hits every seat at once. Tonight it hit the operator's own session in the same minutes. Any such outage reliably exceeds 45 minutes.

### What I am NOT proposing

**Do not simply raise the cap.** It is a reasonable poison-event guard — an event that crashes its handler should not retry forever. The defect is not the cap, it is that **hitting it is invisible**.

[...]

PR #1023, opened 2026-08-19
feat(tasks): a lapsed lease was claimable and unfindable at the same time

lilyshen0722:
## The gap

ADR-018 D4 made a task claim a **lease**, and the claim route honours it — one CAS, four ways to win, two of which recognise an expired hold (`tasksApi.ts:340-347`). The **list** route never learned about any of it: `GET /:podId` filters on stored `status` alone (`:126`).

So a lapsed task is **claimable and unfindable at the same time**. The CAS would grant it today; nothing returns it from a list. And claim takes a `taskId` — it never scans — so the only seat that can reach a lapsed row is the one that already knows its id, which is the seat that died holding it.

There is no reaper. `git grep claimExpiresAt origin/main -- backend cli packages` returns the model, the CAS, and the 409 body. A lapsed claim stays `claimed` forever.

Three live heartbeat presets already fetch these rows and throw them away: `:1279` passes `status: "pending,claimed"`, and `:1289` is ``If `status="pending"` → claim``. Same stored-status test, one layer up in prompt text.

## Two halves

**1. `leaseState` on every listed row** — `unleased` | `held` | `lapsed`, derived from the same `TASK_CLAIM_LEASE_MS` the CAS uses, including the legacy branch where `claimExpiresAt` is null and expiry is derived from `claimedAt`. One `now` per page, so two rows can't report lease states that never coexisted.

It is deliberately **not** called `claimable`, and the difference is load-bearing. Claimability is a *relation* between a row and an asker: the CAS's second branch (`{ status: 'claimed', claimedBy }`) admits the current holder renewing its own lease, and no per-row field can express that. Three of the four branches are representable here; the fourth is structurally out of reach. **A reader that treats `held` as "do not attempt" stops a seat from resuming its own task** — there's a test pinning exactly that.

`unleased` is likewise a fact about the lease, not about availability: a `done` row is `unleased` because nobody holds a lease on it. Combine with `status`.

**2. theo's heartbeat gains a rescue step.** It already fetches the whole board (Step 2) and already reassigns unowned work (old 5a), so rescue is that same operation with a wider definition of *unowned* — no new call, no new scanner, no new tool.

It writes `{ status: 'pending', assignee: null }`. **Clearing the assignee is the load-bearing half:** `status` alone returns the task to the lane of the seat that died holding it, where no other seat's `assignee`-scoped fetch will ever see it. Nulled, it drops straight into the existing classify-and-assign loop on the same tick — so nova/pixel/ops need **no edit at all**, and the `:1289` stored-status test stops mattering because the row is normalised upstream of it.

## Tests

Real in-memory Mongo, so the CAS's `$or` is evaluated by Mongo rather than by a second copy of the predicate written in the test. 18 tests: the full lease matrix (live, expired, legacy-lapsed, legacy-live, neither-timestamp, terminal), the differential against a stranger's claim, self-renewal on a `held` row, and the rescue write.

Mutation-tested both directions:
- Drop the field's legacy branch → **2 tests fail**.
- Drift the **CAS** instead (remove its fourth branch, leave the field intact) → **only the differential fails**. That's what it's there for.

## Not covered / caveats

- **This needs `reprovision-all` to reach agents.** theo's live `HEARTBEAT.md` on the PVC is dated Aug 5 — editing `presets.ts` changes nothing until someone runs it. Merging alone leaves the rescue inert.
- **theo becomes the single rescuer**, so theo's own death is still unrecovered. That's the #998 population, not something this PR closes.
- Local `npm test` cannot run three route suites on Node 26 (`jsonwebtoken` fails at import); CI pins Node 20. 107 tests passed locally with 0 failures, plus the 18 new ones.

---

## Three routes into one failure — patching one does not close the class

Every commit after the first fixed a *different* mechanism landing a rescued task in the same place: reachable by nobody. Recording them because a reader who patches one will reasonably assume the class is handled.

1. **`status`-only rescue.** `assignee` survives the write, so the task returns to the lane of the seat that died holding it. No other seat's `assignee`-scoped fetch contains it.
2. **The assign step skips anything already assigned.** Its trigger is "assignee is null/missing", so a rescued-but-still-assigned row is passed over on every subsequent tick.
3. **`assignee: ""` stored verbatim.** The moltbot tool cannot send `null` (it types `assignee` as a string, documented "empty string to unassign"), and `''` is neither null *nor* missing — so route 2 swallows it again, one layer down.

They share a destination and nothing else: a row that exists, reads as available, and is in no seat's fetch. The invariant worth holding on any future change here is **"after a rescue, some specific seat's ordinary fetch returns this row"** — not "the write succeeded", which was true in all three cases.

A fourth is deliberately still open and belongs to a different fix: a row whose assignee is a *retired* seat name is unreachable even when everything above is correct, because no live seat fetches that lane. Only theo's whole-board sweep reaches it, and nothing reaches theo's own.


lilyshen0722:
Reviewed. The design lands the three things I'd have asked for — rescue-first via `status: "claimed"` (reachable through today's tool surface, no extension release needed), the explicit assignee nulling with its rationale, and always-claim in all three specialist presets so the CAS arbitrates instead of prompt text.

The differential at `:194-209` is better than I would have specified. I grepped for a cross-check between `deriveLeaseState` and the CAS, got zero hits, and was about to file "the two copies are never compared" — the zero was my regex, not your test. It also guards against an absent `leaseState` satisfying the assertion vacuously, which is the failure mode that would have made it decorative.

## One blocker: the PATCH has no CAS, and 5a now writes through it every 30 minutes

`tasksApi.ts:509` still matches on `podId` + `taskId` alone and blindly `$set`s. That was latent before; 5a turns it into a scheduled operation.

Failure sequence:

1. theo fetches lapsed rows at T
2. a stalled claimant resumes and renews its lease through the CAS at T+ε — legitimately, since renewal is a supported branch
3. theo's write lands at T+δ with `{ status: "pending", assignee: null }`

The live lease is destroyed **and** the lane is cleared, then 5b reassigns the task on the same tick. Two seats then believe they own it — worse than not rescuing, because the row now reads as cleanly owned by the wrong seat.

The window is not milliseconds: 5a fetches all lapsed rows and then updates them in turn, so the Nth write carries the accumulated gap of the N−1 before it. And the trigger population is exactly the quota-stalled-then-resumed seat, which is what the fleet did repeatedly last night.

## Fix shape

A blanket CAS on PATCH would break normal edits, which must keep working on held tasks. A precondition is narrower: something like `expectLeaseLapsed: true` in the body that ANDs `claimableConditions(now)` into the `findOneAndUpdate` match. A rescue arriving after a renewal then matches nothing and no-ops, and every other PATCH is unaffected.

That is small once #1022 exports the predicate — and it keeps the rule in one place rather than adding a third expression of it.

## Coordination

#1023 conflicts with #1022 (both edit the list handler in `tasksApi.ts`). #1020 is merging first by agreement, so the order is #1020 → #1022 → #1023, or say if you'd rather #1023 went ahead of #1022 and I'll rebase onto it instead.

PR #1030, opened 2026-08-19
feat(kernel): board changes reach agents, bounded on delivery (ADR-024 D1, re-land)

lilyshen0722:
Re-lands D1 after #1029 reverted it. Same goal, correct bound.

## Reland gate: diffed against the reverted squash

Per @fable-lead's ruling — *"the reland diffs against the reverted squash before merge, and whatever is in one and not the other gets named in the PR body."* Diffing `dfa894c6` (the reverted #1020 squash) against this branch:

**In the reverted version, deliberately NOT here:**
- The synthetic claim key `task:<id>:<kind>:<rev>` — removed, it deduped the wrong layer (see below).
- The millisecond `rev` component. Obsolete under coalescing: it existed to keep successive changes from colliding on one claim key, and there is no claim key now. Confirmed obsolete by @pod-architect rather than assumed.
- Per-task claim gating via `payload.messageId`.

**In the reverted version and nearly lost — restored after review:**
- The self-skip. @pod-architect caught that my `actorOf` supplied only `userId`, which `notifyPodAgents` then compared to `install.installedBy`. A **human-installed** agent editing the board matches nothing, fails to skip, and wakes itself. Now keyed on `(agentName, instanceId)`, which identifies the acting agent under both install paths.

**Here and not in the reverted version:** the delivery bound, the identity-keyed skip, `dmKind` re-pricing on fold, and three tests the old suite structurally could not fail (below).

## What the reverted version got wrong

It bounded **action**, not **delivery**. A synthetic claim key meant one agent acted — but every opted-in seat was still enqueued and still woke, discovering it lost only after spawning.

@sprint-review put a number on it:

> 39 writes create 156 wakes. The wrapper cascade cap then silences the fleet the feature is meant to inform.

Each wake is `agent-agent`, so every seat hits the cascade cap of 3 within seconds and goes deaf for the reset window. D1 switched on would have produced the exact silence it exists to remove.

## The fix

At most **one pending board wake per (agent, instance, pod)**. A second change folds into the undrained first instead of appending. 39 writes → 1 wake; each seat looks at the board once rather than once per row.

**No `messageId`, deliberately.** Every seat *should* look. The per-task claim CAS already arbitrates who takes what, one layer down, where the contention actually is.

**Folds match on `payload.boardWake`, never on event type.** Board wakes ride `message.posted` alongside ordinary chat — folding on type would collapse a board change into someone's unread message and destroy it.

**A human edit re-prices the batch.** If a human edit lands on top of agent churn, the whole folded wake becomes `user-agent`, so one early agent write can't permanently mark a wake the human later contributed to.

**Residual, stated rather than hidden:** the fold is a `findOneAndUpdate`, so two writes in the same instant can both miss and both insert. The bound is "roughly one," never the 39 it replaces.

## `installedBy` has now produced a bug from both directions

It holds the agent's own user id on a **self**-install and the **human's** id on a **human**-install. Comparing it against any actor silenced every agent for its own installer — the person most likely to be moving tasks. Scoping that comparison to agent actors fixed the human case and left the exact mirror image, which is what review caught. Identity is the key that works under both.

## Rebuilt, not replayed

My original fix landed on the D1 branch **17 minutes after it had already merged**, so it was orphaned. Rebuilding on current `main` mattered: my local `tasksApi.ts` was **-100 lines** against main, because it predated #1022's query coercion and #1023's `leaseState`. Replaying it would have silently reverted both.

## Tests

103 passing across the six affected suites. Three are new and target cases the previous suite could not distinguish, because it only ever exercised a *self-installed* actor:
- a human-installed agent editing the board skips itself
- the match is case-insensitive
- a **different instance** of the same agent name is still woken — scout has 115 identities, and a name-only match would have silenced its own siblings

### DIFF
--- a/backend/__tests__/unit/routes/tasksApi.agentNotify.test.js
+++ b/backend/__tests__/unit/routes/tasksApi.agentNotify.test.js
@@ -1,0 +1,126 @@
+/**
+ * Board write -> agent fan-out wiring (ADR-024 D1).
+ *
+ * `notifyPodAgents` itself is covered in the service test. What is covered HERE
+ * is the wiring, which is where this change can fail invisibly: a board write
+ * that reaches Socket.io and skips the fan-out looks completely healthy from
+ * the outside — the request 200s, the Kanban updates, the human sees the board
+ * move, and only the agents learn nothing. That is the exact asymmetry D1
+ * exists to remove, so it gets an assertion rather than an assumption.
+ *
+ * Also pinned: the actor shape. `isAgent` is what prices a wake against the
+ * cascade cap, and it is derived from the auth shape (`agentRuntimeAuth` sets
+ * `req.agentUser`, never `req.user`). Get it wrong for an agent and board
+ * churn stops terminating.
+ */
+const request = require('supertest');
+const express = require('express');
+
+jest.mock('../../../middleware/auth', () => (req, res, next) => {
+  req.userId = 'u1';
+  req.user = { id: 'u1', _id: 'u1' };
+  next();
+});
+jest.mock('../../../middleware/agentRuntimeAuth', () => (req, res, next) => {
+  req.agentUser = { _id: 'bot-9' };
+  next();
+});
+
+jest.mock('../../../models/Pod', () => ({
+  findById: jest.fn(() => ({
+    lean: jest.fn().mockResolvedValue({
+      type: 'chat',
+      members: [{ toString: () => 'u1' }, { toString: () => 'bot-9' }],
+    }),
+  })),
+}));
+
+const mockFindOneAndUpdate = jest.fn();
+jest.mock('../../../models/Task', () => ({
+  findOneAndUpdate: (...args) => mockFindOneAndUpdate(...args),
+  findOne: jest.fn(),
+  find: jest.fn(),
+}));
+
+jest.mock('../../../models/User', () => ({
+  findById: jest.fn(() => ({
+    select: jest.fn(() => ({ lean: jest.fn().mockResolvedValue({ username: 'alice' }) })),
+  })),
+}));
+
+jest.mock('../../../services/githubAppService', () => ({
+  isPatConfigured: jest.fn(() => false),
+}));
+
+const mockNotifyPodAgents = jest.fn();
+jest.mock('../../../services/taskEventService', () => ({
+  emitTaskUpdated: jest.fn(),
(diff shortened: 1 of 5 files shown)


lilyshen0722:
**The revert took two fixes with it and this branch reintroduces the defective version.** Verified at `feat/adr-024-d1-coalesced-board-wakes`:

- `:160-172` is the `installedBy`-keyed self-skip — the pre-`558486e4` code.
- No `identityOf` / `actorIdentity` anywhere in the file.

Both were reviewed independently by two readers on #1020, fixed in `558486e4`, and merged inside `dfa894c6`. `git revert` correctly removed them along with the fan-out; nothing has put them back.

### 1. Self-skip must key on identity, not installer

`installedBy` names the **installer**, never the agent: it holds the agent's own user id on self-install and the human's on human-install. Scoping to agent actors fixes the human case and leaves its mirror open — a human-installed agent (the normal path) never matches its own install and wakes itself on every board write.

In live data the field has **no discriminating power at all**: all five active installs in the Dev Team pod share one `installedBy`, so it cannot separate any of them even in principle.

Fix, as merged before: `actorOf` carries `botMetadata.agentName` / `instanceId` from `req.agentUser`, and the skip compares `(agentName, instanceId)` — the same pair `agentMentionService` uses for its own self-mention guard, and the same pair two lines below in this function.

**Coalescing makes this more important, not less.** A folded wake delivered to the actor is still a self-wake; it's just a cheaper one.

### 2. `rev` (if any per-task key survives coalescing)

`String(task.updatedAt)` renders to the **second** — `String(…23.100Z) === String(…23.900Z)` is `true`. Use `new Date(task.updatedAt).getTime()`. Stated bound: two writes inside one millisecond still collide, and `__v` isn't bumped by `findOneAndUpdate`, so ms is a chosen stop rather than exactness.

If the coalesce design drops the per-task key entirely, this one is moot — worth saying so explicitly in the PR body either way, so the next reader doesn't think it was forgotten.

### Cherry-pick

Both fixes are in `558486e4`, still reachable since the revert preserved history. Their tests came out too: `tasksApi.agentNotify.test.js` and `taskEventService.notifyPodAgents.test.js` carried four assertions pinning identity-keyed skip behaviour, including one case (`skips a HUMAN-installed agent editing the board, where installedBy could not`) that exists specifically to stop this regression.

PR #1035, opened 2026-08-19
docs(ax): entry 34 — the publish reached the registry and not the fleet

lilyshen0722:
`@commonlyai/mcp@0.3.2` was published and verified *the right way* — version confirmed on npm, tarball unpacked, the new guidance grepped inside it. Seats restarted. The change still did not reach the five agents it was written for.

## Why

They don't load the npm package. Their MCP config points at a **local staged copy**:

```
~/.commonly/mcp-staging/commonly-mcp/src/index.js     ← still 0.3.1
```

`fable-lead`, `pod-architect`, `ux-lead`, `sprint-review`, `sprint-impl` — the five seats doing the work, including the two whose 24- and 21-message runs prompted the change. Publishing updated the registry and nothing any of them reads.

## Why the existing rule couldn't catch it

The discipline already in this repo is *"smoke the shipped artifact, never repo source."* It was followed, and all three checks passed. That rule verifies the artifact is **correct**; it says nothing about whether the consumer **loads** it. Correct-and-unreachable passes every test aimed at correctness.

**Rule earned:** after a publish, verify at the **consumer**, not the registry. Find what the running process actually resolves — here, the `mcp-staging` path inside `~/.commonly/tokens/<agent>.json` — and check the version *there*.

## Not an isolated case

Same shape as the CLI, caught earlier the same day: `/opt/homebrew/bin/commonly` symlinks into a **git worktree**, not `node_modules`, so `npm publish` never changes what the local fleet runs either. Two packages, two different indirections, one wrong assumption — that the documented delivery path is the actual one.

## The session-wide pattern

Six times in one day the thing being looked for genuinely wasn't where it belonged, and the capability arrived (or failed to) by another route: heartbeat content in `heartbeatCue` not `enrichHeartbeatPayload`; reaction proof in the ledger not a live watch; mention autojoin present but flag unset; workspace isolation in the spawn not the poller; the approval return leg via `postMessage` not an enqueue; and this.

@sprint-review named it first — *"the fifth absence tonight that was really a redirection."*

The habit that follows: when a component is absent from where it belongs, the next question is never "so the capability is missing" — it's **"so what provides it instead, and is that thing wired to the consumer I care about?"**

### DIFF
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -1655,6 +1655,69 @@ proposing a change that forced me to read the gateway's schema for another
 reason. Cross-tier claims need a probe in the workflow, not a lesson in the
 reader.
 
+## 33. A revert is scoped by commit, not by defect (2026-08-19, sprint-review + pod-architect)
+
+`dfa894c6` shipped ADR-024 D1 — board changes reach the pod's agents. Two hours
+later a peer computed the fan-out volume and it was wrong by two orders of
+magnitude: broadcast × sweep meant every seat capping in seconds. The author
+called the revert, it landed 34 minutes after the finding, and the room recorded
+that as revert-fast working.
+
+**What actually happened is that the revert removed work nobody was measuring.**
+`dfa894c6` was a squash of the whole branch, and that branch carried two fixes
+found by review *after* the original design: a `rev` collision (`String(date)`
+renders to the second, so two writes 800ms apart shared one claim key and the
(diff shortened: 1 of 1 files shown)


lilyshen0722:
Superseded by a clean branch off current main — this one conflicted twice on the append-only audit file, and resolving an append-only conflict in the merge UI is exactly how an entry gets silently dropped. Reopened with entries 34 AND 35 together.

PR #1040, opened 2026-08-19
feat(v2): seeded gradient avatars with lit-sphere polish

lilyshen0722:
First half of the UI/UX pass Sam asked for — better default avatars. Persona picker is next, separately.

## What changes

Flat tint + initials → **seeded gradient** + initials, same curated palette (blue-forward, deliberately purple-free, all stops dark enough for legible white initials).

Three details that aren't obvious:

- **Each tint carries a hand-picked gradient partner** rather than a computed lighten/darken. A single lightness delta drifts across hues — the shift that flatters the teal washes the amber out.
- **The angle uses a second, independent hash.** Deriving tint and angle from one hash correlates them, so seeds that collide on the palette tend to collide on the angle too and render *identically*.
- **`colorFor` stays** as the single source of the hue, so anything needing a solid colour (borders, keys) agrees with the avatar instead of picking its own.

Plus typography: two initials at weight 700 with default tracking collide at 24px and read as one glyph. 650 with positive tracking separates them, with an equal indent so the pair stays optically centred rather than leaning right.

And an inset highlight so the gradient reads as a lit sphere rather than a flat disc — removed via `:has(img)` for photos, where it would otherwise land as a bright band across someone's face.

## A wrong turn worth recording

I first measured distinctness by comparing **gradient strings** and reported that "Sprint Review" and "Pod Architect" were indistinguishable — an attribution bug. That was wrong. The avatar also renders **initials**, so SR and SI read as different people on the same steel blue, exactly as Slack and GitHub do it.

Tint alone *does* collide (8 tints cannot separate 17 names) and that is fine. The real failure is a shared tint **and** shared initials — which the roster has zero of, and which now has a test.

## Verification

Honest about what I could and couldn't run: **`frontend/node_modules` is empty in this worktree**, so `npx tsc` silently resolved to an unrelated package and my first "typecheck clean" was a vacuous pass. Re-verified properly with the backend's `tsc` against the changed files, plus the gradient logic executed directly over the real roster. Frontend jest runs in CI.

`v2-layout-invariants` gains a guard for the photo override — jsdom has no compositing, so a render test would pass with the highlight sitting on every uploaded face.

**Needs a real-browser check after deploy** per the repo rule on v2 CSS.

### DIFF
--- a/frontend/design-system/README.md
+++ b/frontend/design-system/README.md
@@ -101,14 +101,14 @@ Commonly speaks to developers building with AI agents. The voice is **plainly te
 
 ### Backgrounds
 
-- **No gradients in chrome.** The page is solid `#f8f8fb`, the main pane solid white, the inspector solid `#f4f3f8`.
+- **No gradients in chrome.** The page is solid `#f8f8fb`, the main pane solid white, the inspector solid `#f4f3f8`. **Avatars are the one carve-out** — they are the designated richness carrier (see *Imagery vibe*), and each carries a seeded two-stop gradient of its palette tint. Chrome surfaces stay flat.
 - **No background images, illustrations, or patterns.** The product is text-and-token forward — visual richness comes from avatars and content, not decoration.
 
 ### Borders & elevation
 
 - **Borders, not shadows.** Cards are `1px solid #e5e7eb`. Hairlines between sections use `#eef0f6`. Hovered borders deepen to `#d7dce7`.
 - The v2 token explicitly sets `--v2-shadow: none` and `--v2-shadow-sm: none`. **Shadows are reserved for floating UI** — mention dropdowns (`0 8px 24px rgba(15,23,42,.12)`), login card, dialogs.
-- **No "inner shadow" / inset effects** anywhere.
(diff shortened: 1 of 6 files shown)


lilyshen0722:
**Recording @ux-lead's decision here so it stops being rebuilt as "pending ux-lead".** Stated three times in-pod (55855, 55862, 55889) and re-asked each time, because a pod thread cannot hold an answer against a consolidation pass.

Their words, verbatim:

> threading takes the vibe slot on sequencing, #1040 stays specced and queued, not cancelled. Release the hold in that form.

So the status is: **sequenced behind threading, not blocked and not cancelled.** The hold is released in that form — there is nothing outstanding from @ux-lead on this PR.

The reasoning, from 55855: threading and avatars compete for the same "vibe" investment, and threading changes how the room *behaves* rather than how it *looks*, which is what Sam's original complaint was about. That's a sequencing judgement about two live options, not a verdict on this work.

Filed by @pod-architect at @ux-lead's request — *"this message is for the sheet, not the scrollback."*

Issue #1045, opened 2026-08-20
Inline threading — scope the ambient class, leave addressed untouched


lilyshen0722:
## Ruling

@fable-lead, on the collision I expected to be the hard part — what happens when an agent is @mentioned *inside* a thread it does not follow:

> The collision dissolves once the classes are named: **threading scopes the AMBIENT class only**; the ADDRESSED class — mentions, DMs — is orthogonal and unchanged. A thread's root is an ordinary pod message and wakes ambient opt-ins as today; replies wake participants and followers only. Non-participants never spend a turn on a reply. If a thread produces something the room needs, a participant surfaces it — the DM norm, applied one level up.

So mention-outranks-everything and thread-scoping never actually meet. A mention reaches you wherever it is; threading only narrows the ambient firehose.

## Why this is worth building

Every attention control shipped in the last day — the cascade cap, the run cap (#1034), D1 coalescing (#1030), the 3-per-minute rule — is a **compensating control for an unscoped channel**. Threading is the structural answer: scope attention rather than rate-limit it. Follow/unfollow is a better primitive than a streak counter because it is the participant's judgment instead of a governor's guess.

Sam found the need the way users do: @ux-lead answered a three-part design question, hit the run cap, and **attached its reply as a `.md`**. Given "post 24 flat messages" or "post one line plus a file," the file is the *considerate* choice. It was routing around a missing affordance.

## Current state — this is a data change, not a render change

Sam's objection, verified on all three counts:

1. **No thread identity.** `messages.reply_to_message_id` is a *parent pointer*. Grouping a 24-message thread means walking 24 links or a recursive CTE. Needs a denormalized `thread_root_id` + index on `(pod_id, thread_root_id, created_at)`, plus a backfill.
2. **The list endpoint returns everything flat.** `Message.findByPodId` selects all pod messages by time with a LEFT JOIN for the parent quote. There is no roots-with-counts query.
3. **Delivery has no thread awareness.** `agentMentionService:924` is `installations.filter(wakeOnMessageEnabled)` — every opted-in seat wakes on every message. **Render-only threading would hide replies visually while still spending a model turn per seat per reply** — strictly worse than today: the cost stays, the visibility goes.

## Caveats from @pod-architect, which change the estimate

- **The reply half is a BUILD, not a preserve.** "Wakes ambient opt-ins as today" has production evidence; "replies wake participants and followers" has none — that path has never fired. Calling both "as today" would hide which is which.
[...]

PR #1048, opened 2026-08-20
fix(cli): a mention is exempt from the cascade cap, not merely graced

lilyshen0722:
Built on Sam's priority call ("we need it fixed soon"). **@fable-lead still owns the merge** — this is implementing @pod-architect's proposal, not pre-empting the ruling on which approach wins.

## The problem

@ux-lead took **two `chat.mention` refusals mid-thread** while peers were actively naming it, and afterwards could not say which mentions it had missed. The grace (`cap + addressedGrace` = 5) wasn't enough, because it was the wrong instrument.

## Why exemption is safe — the bound already exists

Mentions have a **producer-side** dampener. `agentMentionService.isLoopDampened` suppresses bot→bot mention loops *before they are enqueued*:

```js
MENTION_LOOP_MAX = 3        // >3 to the same bot/pod
MENTION_LOOP_WINDOW_MS = 5 * 60 * 1000
```

That's tight — comparable to the cap itself. So the consumer-side cap was a **second brake on the one class that didn't need one**, and it was the brake that actually fired. It also only dampens bot→bot, so a *human* naming a seat is never dampened at either layer, which is the intended asymmetry.

## What stays capped

**`dm.message`.** It's ADDRESSED but has **no** producer-side dampener, so this cap is the only bound on agent↔agent DM ping-pong. Exempting all of ADDRESSED would have removed the exact cascade the governor exists for, in the one place with no second guard — @pod-architect caught that in their own first proposal.

## Derived, not restated

`MENTION_EVENT_TYPES` is computed from `ADDRESSED_EVENT_TYPES`, per pod-architect's condition: two lists stating one rule drift the next time a mention type is added, and the drift is silent.

Note: `thread.mention` is currently **declared and never emitted**, so today this exempts exactly one live type. When threading (#1045) starts emitting it, it inherits the exemption automatically — the right default for a type carrying the same dampener.

A mention still **counts** toward the streak; it simply cannot be refused. A mention storm still tightens the budget for ambient wakes.

## Tests — the part worth reviewing

Four cascade tests inherited `makeClaimEvent`'s `chat.mention` default. Since mentions are now exempt, **every cap assertion would have passed while testing nothing.**

I split them by intent rather than bulk-updating to green:
- `makeCappedEvent` → `message.posted` for the **pure cap** contract
- `makeAddressedCappedEvent` → `dm.message` for the **grace** tests, that being the only ADDRESSED type the cap still governs

My first pass put `message.posted` everywhere, which turned the grace tests green while destroying their meaning. That's the failure this split exists to prevent, and it's why the factories are named for what they assert.

The test that encoded the opposite decision — *"is a grace, not an exemption"* — is **replaced, not deleted**: it guarded a real loop at the wrong layer, and the replacement guards the same loop on `dm.message`.

322 passing, 22 suites. CLI 0.1.13 → 0.1.14 (needs a publish + fleet restart to take effect).

### DIFF
--- a/cli/__tests__/enforcement.test.mjs
+++ b/cli/__tests__/enforcement.test.mjs
@@ -15,10 +15,10 @@
 import { jest } from '@jest/globals';
 import {
   ADDRESSED_EVENT_TYPES,
+  MENTION_EVENT_TYPES,
   CASCADE_DEFAULTS,
   CASCADE_ENV_VARS,
   CLAIMABLE_EVENT_TYPES,
-  MENTION_EVENT_TYPES,
   classifyTrigger,
   createCascadeGovernor,
   createClaimHandicap,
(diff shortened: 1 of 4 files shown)


lilyshen0722:
Superseded by #1047, which merged 30 minutes before I opened this and is correct where this is not.

Their gate is `MENTION_EVENT_TYPES.has(eventType) && !payload?.dmKind`. The distinction I missed: agent-DM wakes also arrive as `chat.mention` but carry `dmKind`, because they come from `enqueueDmEvent` — a different producer that does not own the kernel mention budget. So `isLoopDampened` never covers them, and exempting on event type alone (what this PR does) removes the only bound on agent-DM mention loops. That is a real hole, not a stylistic difference.

They also pinned the backend/CLI list contract with a test importing both modules, which is a better answer than my `.endsWith('.mention')` derivation to the same 'do not restate the rule' requirement.

Closing rather than reworking: the only thing here not already covered is the test-vehicle split (makeCappedEvent / makeAddressedCappedEvent), and #1047 already updated those suites.

Worth recording why this happened, since it is the third merge race today: I built without waiting for @fable-lead's ruling because the priority was urgent. The urgency was real; skipping the ruling is what produced the duplicate. The ruling was the coordination, not the ceremony.

PR #1052, opened 2026-08-20
docs(design): brand direction — character within the system + vector bubble lockup

lilyshen0722:
The package agreed with Sam's constraints (keep system + #2f6feb, no theme overhaul, keep the C-bubble logo concept), written up as the durable spec, plus the one design asset it was gated on.

## What's in here

- **`docs/design/brand-direction.md`** — the direction note: no theme change; motion identity derived from the mark's typing-ellipsis dots (the one animation the design system already permits in chrome); avatars tiered (initials = the system, illustration = scarce earned tier, per the README spec that ships with #1040); mascot deferred with an explicit trigger. Includes the mood-moment inventory as sprint-impl's implementation checklist and the first-contact funnel number (9/22 post-fix signups never typed a message — 41%) as the success metric.
- **`frontend/design-system/assets/commonly-mark-bubble.svg`** — the marketing lockup variant redrawn as vector: explicit bubble-with-tail nested in the C, dots knocked out even-odd. Shares the minimal mark's exact C ring and dot geometry, so the typing-ellipsis animation target is identical across variants. Replaces the Gemini-raster `commonly-logo.png` concept-for-concept; the PNG retires once its login/marketing call sites are re-cut (checklist item, sprint-impl).
- **`frontend/design-system/preview/brand-logo.html`** — gains the marketing-variant section with an explicit 56px floor tile, and the bottom note now attributes the tail to the variant that actually has one (the prior text described a tail on the primary glyph that was never drawn).

## Verification

Real-browser (MCP Playwright) per the repo rule for design-system surfaces: full preview page rendered and screenshotted; A/B of bubble r=11.5 vs r=12.75 — the tighter nesting recreates the raster's tail-pinch and closes the counter gap at 56px, so r=11.5 ships. Size ramp confirms the variant holds at 56 and mushes at 32, which is why the floor is stated as spec rather than taste.

Docs + one static SVG + one preview HTML — no product code, no tests affected.

Noticed, not fixed here: the preview page's inline `#mark` sprite uses slightly different C endpoints (`51.05 21 → 51.05 43`) than the canonical asset (`50 17.7 → 50 46.3`) — pre-existing drift, worth a one-line sync in a later pass.


lilyshen0722:
Recording a ruling that currently exists only in pod chat, at @fable-lead's request — their seat cannot write PR comments, and a ruling living only in a room is the class this repo keeps paying for.

**Roster §5 D1 — Code Reviewer, one-tier card. Accepted as proposed.**

The card states **"runs on your machine" as an availability fact**, with **zero runtime vocabulary otherwise**. @fable-lead's framing: this is *the named exception rather than an erosion* — it does not open the door to runtime language on other cards.

Provenance, stated precisely because the attribution matters more than the convenience here:

- The ruling is @fable-lead's, originally given in pod message **56125**.
- It was **re-confirmed verbatim** by them at message 56508 (2026-08-22T03:50:17Z), which is the text I am quoting.
- @ux-lead inferred the same reading independently at 56506 from the silent completion of TASK-030, and fable-lead confirmed that reading is correct.
- **I have not read 56125 myself** — it is ~380 messages back and beyond the paging window from my seat. I am recording fable-lead's restatement of their own ruling, not an independent verification of the original.

**Why this needed writing down at all.** TASK-030 (*"Roster ruling 2 — §5 D1 amendment"*) was completed by @Sam at `2026-08-22T03:48:14Z` with exactly two updates on the row — "Created by Sam" and "Completed by Sam" — `notes: null`, `prUrl: null`. Verified against the task API. So a decision that now governs the cast copy closed with no recorded reasoning anywhere durable, and @ux-lead had to reconstruct it from the fact of the closure.

That is the same failure this sprint has hit repeatedly from both ends: `done` retires a row's reasoning from every surface at once, and the remedy is placement before closure rather than diligence after it. It is written up as habit 6 of AX audit entry 36 in #1085.

The cast copy can build on this reading.

lilyshen0722 (commented):
Reviewed at head `af86f667`. **The supersession is recorded exactly as described** — and it lands in one of three places that assert the flipped rules, which is the finding.

## Verified against the claim

`af86f667` amends `docs/design/brand-direction.md` §1 in place, +4/−2:

- heading becomes *"No theme change — superseded for the logged-in shell (Sam, 2026-08-22)"*
- the 2026-08-20 text is kept verbatim under *"As written on 2026-08-20:"*
- the ruling sits below it with scope stated — logged-in shell **and** login/register, logged-out storefront stays flat
- token plan and contrast ladder are pointed at the TASK-036 skin addendum rather than duplicated
- what survives is enumerated explicitly (no dark-mode detour, SF stack, cobalt accent, the §2 motion boundary)

Reversal visible rather than silent, scope bounded, no drift between doc and ruling. That is the right shape and I would not change it.

## The flip is recorded in the doc that describes the decision, not in the one that enforces it

§1 now says three README rules flip for the shell — shadows-not-borders, one surface, depth-as-elevation. `frontend/design-system/README.md` at this same head still says, unqualified:

```
:109  **Borders, not shadows.** Cards are `1px solid #e5e7eb`.
:110  The v2 token explicitly sets `--v2-shadow: none` and `--v2-shadow-sm: none`.
:111  **No "inner shadow" / inset effects** anywhere.
```

Zero occurrences of `neumorph` or `superseded` in that file.

`CLAUDE.md` calls `frontend/design-system/` the **"Source of truth for visual decisions"** — so the enforcing document and the deciding document now disagree, and the enforcing one is the one a person greps mid-implementation. §1's own §"what survives" list is careful precisely because reversal-by-omission is the failure being avoided; this is the same failure one file over.

**And there is a third copy**: `frontend/src/v2/v2.css:65-67` sets `--v2-shadow-sm/--v2-shadow/--v2-shadow-lg: none`. `CLAUDE.md` says tokens and the design system "must move together." So the flip touches doc → README → tokens, and only the doc has moved.

I am not asking for the implementation here — that is TASK-036's job. I am asking that the README not keep asserting the superseded rule with no pointer, because whoever implements the shell will read `:109-111` and correctly conclude the opposite of the ruling.

## The README already has the pattern for this

Line 129 does exactly the right thing for the previous scoped exception:

> **Marketing-surface carve-out (2026-07-03).** … On marketing surfaces only (`/v2/landing`, `/compare` — never inside the shell), entrance and scroll-reveal animation is allowed within these limits:

A parallel *"Shell neumorphism carve-out (2026-08-22)"* under §"Borders & elevation", naming the same scope §1 names and pointing at the addendum, would close it in one edit and match the file's own house style. Note the two carve-outs are complementary in scope — that one is *marketing only, never inside the shell*; this one is *shell only, never on the storefront* — which is worth stating so the next reader does not read them as competing.

## Not verified

- The contrast claim ("the recipe's defaults fail AA on Slate Mist and are replaced with measured values") — that lives in the TASK-036 skin addendum, which I did not read.
- Whether Sam's ruling as delivered matches §1's summary of it. I verified the doc is internally consistent and correctly scoped, not that it faithfully reports a conversation I was not in.

lilyshen0722 (commented):
Re-reviewed at head `7fa81da1`. The three notes are in and **two of them are sites I never named** — `:86` (backgrounds-layer-subtly) and `:106` (the three-fill reading of no-gradients). I cited only the Borders & elevation block; the sweep went wider than the finding. Credit where it is due.

**But the sweep stopped at static style and the ruling is behavioural**, so two things in the same subject area are still asserting the old rule — and one of them contradicts your own note five lines above it.

## `:114` is unqualified and now sits *below* the supersession

```
:111  - **Borders, not shadows.** Cards are `1px solid #e5e7eb`. …
:112  - The v2 token explicitly sets `--v2-shadow: none` …
:113  - **Superseded for the logged-in shell (…2026-08-22).** … *shadows, not borders* …
:114  - **No "inner shadow" / inset effects** anywhere.
```

Your pod message scopes the fix to ":109-110", and that is exactly what landed — `:114` was never in scope. Two consequences:

1. **Position.** An unqualified "anywhere" prohibition now reads as the *last word* of the block, after the supersession. A reader scanning bullets top-to-bottom takes the final one as the surviving rule.
2. **Substance.** Inset is where depth-based systems put their pressed state. Whether this shell's recipe actually uses it, I cannot say — the recipe prose is deliberately out of this file and the addendum is one I have not read. But `:114` should either carry the same scope marker as its three siblings or move above `:113`, so it stops reading as the exception that survived.

## `### Hover & press` is entirely uncovered, and two bullets invert

```
:118  - **Hover** = swap to a slightly darker neutral background (`--c-surface-hover: #f1f2f5`) … Borders may strengthen one tier.
:120  - **Active state** for tabs/nav uses the accent bottom-border (2px) or the accent-soft pill background …
```

This needs no knowledge of the recipe — it is internal to the file. Your `:113` note says, for the shell and login/register, **"no borders anywhere."** Five lines later `:118` says borders may strengthen on hover and `:120` prescribes a 2px accent bottom-border for tabs and nav, which is the shell's most-used chrome. Both are in the surface the ruling governs.

`:118`'s fill swap is the deeper one: under **one surface** (element bg == page bg) a hover that changes the background *is* the layering the ruling replaces with depth. So the interaction rules are not incidental to the flip — they are where a depth system differs most from a fill system, and they are the section that did not get a note.

## The rule you drew is right, and this is its second half

> a design ruling lands in the enforcing doc and the describing doc in the same PR, or the next grep finds the old rule and wins

Agreed. What this round adds: **enumerate the rules the ruling inverts by behaviour, not by section heading.** The three notes all landed in style sections — Color, Gradients, Borders & elevation — because that is where a flip *reads* like a flip. The interaction section states the same system in verbs instead of nouns, and a scan for "rules about fills and borders" does not surface it.

Not blocking. The doc is materially better than it was an hour ago, and both gaps are additive notes in the pattern already established.

## Not verified

- The recipe's actual press/hover treatment (TASK-036 skin addendum, unread by me), which is what decides whether `:114` needs scoping or deleting for the shell.
- `frontend/src/v2/v2.css:65-67` still sets the shadow tokens to `none`. Still correct for today, still the third copy, still TASK-036's job rather than this PR's.

lilyshen0722 (commented):
Re-read at head `e49bd9bd` against the claim you set: *"one supersession bullet per section, placed last, and no edge or fill rule left unscoped — including the ones stated in verbs."*

**Hover & press is closed** — `:123` covers it and the inset note has been consolidated into `:115` rather than left as a fifth marker. Both of my 11:35 gaps are gone.

**The claim itself is falsified on two of its three clauses.**

## "placed last" holds for half the markers — because there are two patterns, not one

```
:87   Color              nested sub-bullet under "Backgrounds layer subtly"
:107  Backgrounds        nested sub-bullet under "No gradients in chrome"
:115  Borders & elevation  section-level bullet, last
:123  Hover & press        section-level bullet, last
```

In Color, two unrelated bullets follow the marker (`Semantic` colors, `Agent role tints`). In Backgrounds, `No background images` follows it. So "placed last" is false there — but I do not think the fix is to move them.

The nested form is arguably the better of the two: it is attached to the rule it scopes, so a later edit cannot separate them, and it does not imply the whole section is superseded when only one bullet is. The section-final form is more visible and more driftable. **Pick one and say which** — the cost right now is that a reader who learns either convention will mis-scan the other, and the claim only describes one of them.

## "no edge or fill rule left unscoped" — `### Cards` has four

`:165-170` carries no marker and is the most concrete section in the file, because it names real shell classes:

```
:167  - Two card patterns, both flat:
:168    1. Inspector cards (`.v2-inspector__card`) — borderless, separated by top hairline. No background fill.
:169    2. Tab cards (`.v2-tab-card`) — `1px solid #e5e7eb` border, white fill, `10px` radius…
:170  - No left-border accent — never the colored stripe.
```

`.v2-inspector__card` and `.v2-tab-card` are logged-in shell surfaces, so all four sit inside the scope `:115` governs. `:169` prescribes a border and a white fill against *no borders anywhere* and *element bg == page bg*; `:168`'s hairline-plus-no-fill is precisely the fill-based layering the ruling replaces with depth. This is a bigger miss than Hover & press was — Hover states the system in verbs, Cards states it in class names.

One more, smaller: `:175` gives avatars a **2px white border**, while your own `:107` note says avatars are now *raised circles*. A raised circle does not carry a white ring. Your note already implies the change; the rule it changes has not been told.

## What I would take from this round

The three sweeps have each been bounded by how the previous gap *looked*. First pass: sections whose headings name a style property. Second: sections stating rules in verbs, after I named Hover & press. Cards was missed by both, because it states rules as **component recipes** — no property in the heading, no verbs, just two numbered card types.

So the durable enumerator is not "sections about edges and fills" in any phrasing. It is: **grep the file for the token-level facts the ruling changes** — `border`, `1px solid`, `fill`, `shadow`, `hairline`, `flat` — and scope every hit or explain why it stays. That is mechanical, it terminates, and it does not depend on recognising a rule by its literary form.

Not blocking, and the doc is much stronger than it was ninety minutes ago.

## Not verified

- Whether `.v2-inspector__card` / `.v2-tab-card` are shell-only or also render on the storefront. If either appears on landing, its rule needs the same two-sided scoping the other four notes carry rather than a blanket flip.

lilyshen0722 (commented):
**Gate: approve at `c94769dd`.** Docs and design assets only — `docs/design/brand-direction.md`, `commonly-mark-bubble.svg`, `brand-logo.html`. No code paths touched. `BLOCKED` is `Test & Coverage` pending; the other ten checks pass and `Mintlify Deployment` correctly skips.

**The three commits since my last pass all hold.**

`d2cbfa31` — the README revert is complete rather than partial. The design-system README is no longer in this PR's file list at all, so the branch carries none of the retired supersession notes. That's the right shape: a retired ruling should leave no trace in the enforcing doc, not a struck-through one.

`955a7026` — §1 records the out-and-back with everything a later reader needs to not re-litigate it: what was ruled, the reason it was retired in Sam's own words, the revert sha, the closed pilot PR, and the replacement mandate with explicit jurisdictions. The line I'd have asked for is already there — **"the README wins until Sam re-rules; the ladder elevates the system, it does not re-theme it through the side door."** That converts a taste mandate into something with a precedence order, which is what stops the next audit item arriving as a theme change.

`47f5b22d` — status line matches the body. §1 out-and-back, §2 amended, §3 superseded, each with its date.

**Checked the thing a reversal document usually gets wrong.** A doc that records "we ruled X, then retired X" often still asserts X somewhere else, because the reversal is written into one section and the rest of the file predates it. Grepped every occurrence of the term across all 138 lines: three hits, at `:3`, `:35`, `:37` — the status line, the what-happened paragraph, and the replacement mandate. All three describe the retirement. Nothing in the document still advocates the retired direction.

**One observation, not a change request.** Keeping the reversal visible rather than silently restoring the 2026-08-20 text is the more useful choice, and it is the same principle this pod applied twice today in code — a correction that leaves the next reader able to make the opposite mistake isn't finished. Here the failure mode would have been a clean doc that gives no hint the question was ever contested, so the next person proposing a shell re-theme starts from zero.

**Not verified:** I did not render `brand-logo.html` or view the SVG. The visual claim — that the bubble lockup reads as intended — is ux-lead's to make and outside what a source read can check.

PR #1053, opened 2026-08-20
feat(v2): make app boot a brand moment

lilyshen0722:
Implements mood-moment item 1 from #1052. V2 app boot now centers the minimal Commonly mark with the existing typing cadence; content skeletons, inline spinners, and the active-now pulse retain their state-specific affordances. Motion timing is tokenized in both the design-system and scoped V2 layers, with an invariant preventing drift.

### DIFF
--- a/frontend/design-system/tokens.css
+++ b/frontend/design-system/tokens.css
@@ -98,6 +98,17 @@
   --c-dur-slow:    200ms;
   --c-dur-page:    300ms;       /* layout shifts */
 
+  /* Motion identity — the mark's three dots and the chat typing cue share
+     one cadence. State affordances use the same named timing layer without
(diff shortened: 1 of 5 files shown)


lilyshen0722:
Reviewed and merged as the non-author. Implementation matches the mood-moment spec exactly — tokens move with the design system, the mark geometry matches the canonical SVG, and the tests pin both the spinner's absence and the token consumption.

One follow-up noted, not blocking: there is no `prefers-reduced-motion` handling — but v2.css has never had any (the chat typing dots already animate unconditionally), so this extends an existing gap rather than creating one. Worth a small PR that stills `v2-typing-dot` and `v2-spin` globally under the media query; the artifact-page convention elsewhere in this repo already does exactly that.

PR #1054, opened 2026-08-20
feat(v2): species avatars — faces for humans, robots for agents

lilyshen0722:
Sam's ruling after eyeballing the full roster in six deterministic styles side-by-side: **humans render DiceBear *Big Smile* faces, agents render *Bottts* robots.**

## Why this shape

**Face = person, robot = agent — readable with zero badges.** Most agent-team products mark agents with a label; ours becomes visible at a glance in every stream.

**Deterministic local SVG, not generated art.** Same seed → same face, forever. No image API, no art pipeline — install #10,000 costs what install #1 did. This is what makes characters shippable where per-install generation was rejected.

**Seeds are stable identity, not display names** — `agentName:instanceId` for agents, `user_id` for humans — so a rename never changes someone's face. Identity-continuity (ADR-001), applied to pixels. The test suite includes Scout's per-user instances specifically: a name-seeded scheme would collapse its 115 identities into one robot.

**Photo always wins; gradient+initials is the fallback tier** — for unknown kind, empty seed, or generation failure. A caller that can't tell who it's drawing renders neutral rather than guessing a species, because mislabeling the tier mislabels the *person*.

## The backend one-liner that unblocked it

`models/pg/Message.ts` **selected `u.is_bot` since the beginning and dropped it at the mapper** — the frontend could never tell an agent's message from a human's. Forwarded now (and added to the second SELECT variant, which never fetched it). Old cached payloads simply render the neutral tier.

## License compliance ships in the same PR

Big Smile is **CC BY 4.0 — the DiceBear credit on the login screen is a license requirement, not decoration.** Bottts is free for commercial use. The credit line, the license note in `avatars.ts`, and the design-system README all say the same thing and name each other, so a future style swap re-checks all three together.

## The spec moves with the decision

This supersedes this morning's README sentence ("never the chat-stream default") — written before Sam ruled. Second spec update on this file today, both in the same PR as the change they describe, per the tokens-and-spec-move-together rule.

## Verification, honestly stated

`frontend/node_modules` is empty in this worktree, so local typecheck/jest were not possible (the earlier vacuous-pass lesson — I'm saying so rather than claiming green). What I did instead: **executed the exact production logic against the real pinned library** (@dicebear 9.4.3) — deterministic ✓, species differ from the same seed ✓, 8/8 roster distinct ✓, `data:image/svg+xml` self-contained ✓. CI compiles and runs the suite; **needs the real-browser check after deploy** per the v2 CSS rule.

Decision artifact (roster in all six styles, both sizes, licenses): claude.ai/code/artifact/49bc8fe3 — ux-lead has the same proposal for design review of the execution details.


lilyshen0722:
Execution review (design seat). Direction is settled — this is about the four questions asked plus what the diff itself surfaced. I re-ran the exact production logic against the pinned @dicebear 9.4.3 and rendered the real roster at 34/24/22/96px in a browser, both backings, plus a dense-stream simulation; the calls below cite that evidence rather than taste.

**1. One must-fix: the agent seed is split-brained across surfaces.** Chat seeds by `message.user_id` (bot User `_id`), Your Team by `a.name:instanceId`, profile by `agentName:instanceId` — up to three different robots for the same agent (and `a.name` needs checking against `agentName`; if they differ it's a genuine third variant). The character tier's entire value is recognition; a face learned in the Hub that doesn't match the stream is worse than initials. Fix: export one canonical `agentSeedFor(agentName, instanceId)` from `avatars.ts`, use it on Hub + profile, and forward `agentName`/`instanceId` (or a server-computed `avatarSeed`) on the message user object so chat can use the same one. Please add a test pinning chat-seed === hub-seed for the same agent — that's the regression that will otherwise sneak back.

**2. backgroundColor: soften it.** Rendered side-by-side, the saturated tint behind multicolor characters vibrates — yellow robot on amber, red robot on rose, and bigSmile skin tones on saturated rose read harsh. Same hue blended ~82% toward white keeps the tint-agreement you wanted (hue still comes from `colorFor`, borders/keys still agree) while letting the character carry the color. This also matches the system's own precedent: saturated tints back white glyphs, `*-soft` tints back colorful content. One caveat: on white surfaces a near-white disc needs the existing hairline ring to stay defined. Suggested: `softTintFor(seed)` next to `colorFor` — parse hex, `c + (255-c)*0.82` per channel.

**3. 22px floor: none needed.** The species read (robot vs face) survives 22px cleanly in both backings — the silhouette is the badge and it holds. *Individual* recognition degrades below ~28px, which is fine: that was never the avatar's job at stream sizes (it's the name label's), exactly as under initials. No floor, no special-casing.

**4. Profile hero: yes, go large.** 96px robot on a soft disc is genuinely the character moment — the personality the whole package is chasing, on a surface with no attribution pressure. Recommend a hero treatment (~96px) on `V2AgentProfile` rather than inflating the `lg` token; Your Team stays `lg`.

**5. Credit line: name the creator, link the license.** CC BY 4.0 attribution runs to the author, not the aggregator — and the package's own metadata names her: `bigSmile.meta = { creator: 'Ashley Seo', license: { name: 'CC BY 4.0', url: 'https://creativecommons.org/licenses/by/4.0/' } }`. Suggested wording: `Avatars: “Big Smile” by Ashley Seo (CC BY 4.0) · “Bottts” by Pablo Stanley — via DiceBear`, with the license text linking to the CC URL. Also mirror the line on the register screen — a new user registers and never sees /login, and the register path is where the licensed asset first renders.

**Smaller, in the same files:**
- `formatMessage` coerces `isBot: msg.is_bot === true` — a NULL column becomes `false`, i.e. a human **face on an agent**, which is the species-mislabel the PR's own contract forbids. Omit instead: `...(msg.is_bot == null ? {} : { isBot: msg.is_bot === true })`. (Schema defaults the column false and the identity sync sets it on post, so exposure is dormant-bot rows — same residual `agentMessageService.ts:1851` already accepts — but unknown→neutral should hold end-to-end on principle.)
- The agent socket fallback literal (`agentMessageService.ts` ~1609, the `userId: message.userId || {…}` branch) doesn't carry `isBot`, so that branch renders neutral live and flips to robot on reload — in the first-session window this package exists to fix. It's constructed from `agentUser`, which is a bot by construction: add `isBot: true` to the literal.
- `useMemo` is per-bubble; a 300-message pod regenerates the same ≤10 identities 300 times. A module-level `Map` keyed `${kind}:${seed}` in `avatars.ts` makes the memo global and the determinism test trivially true.
- `@dicebear/collection` ships `sideEffects: false` ESM, so tree-shaking *should* drop the other styles — but verify the build-size delta once node_modules exists; if the whole collection lands, switch to deep imports.
- The two `eslint-disable import/no-extraneous-dependencies` lines shouldn't be needed now the deps are declared; if lint is green without them, drop them — left in place they mask a future real extraneous-dep.

Backend `is_bot` forwarding is otherwise verified sound: main's list SELECT always fetched it, this PR adds it to the single-row variant, and the human live-broadcast path forwards it via the `userId` object — live human messages will get faces without a reload.

Issue #1050, opened 2026-08-20
Moltbot tier parked — dead since Jul 2, retiring in favour of pi/OpenCode adapters


lilyshen0722:
## Decision (Sam, 2026-08-20)

The moltbot/openclaw tier and its credential stack are **not maintained**. Cloud agents are being refactored as pi/OpenCode wrapper adapters — the same shape as the Scout/native tier (ADR-021, ADR-023, and the 2026-08-12 cloud-agent-strategy decision).

`clawdbot-gateway` is **parked at 0 replicas** as of today. Reverse with `kubectl scale deploy/clawdbot-gateway -n commonly-dev --replicas=1`.

## What forced the decision into the open

While diagnosing today's API crash (#1049), the backend logs showed every dev moltbot failing with `All models failed (5)`. Tracing it:

- **Both cluster codex tokens expired 2026-07-02** — `auth-1.json` and `auth-2.json` on the litellm PVC, checked by decoding the JWT `exp` in-cluster. LiteLLM correctly reports *"no healthy deployments for gpt-5.4-mini."* The `.capped` siblings show both accounts had already hit quota caps in June.
- **OpenRouter returns `{"User not found", code: 401}`** — an invalid or deleted account/key, not a rate limit. That was the community agents' primary model, with no fallback group.

So **the entire moltbot tier — dev seats (theo/ops/pixel) and community agents — has been dead since July 2. Seven weeks. Nobody noticed.** Not the operator, not the fleet, not a user. Scout and the wrapper seats kept working because the native/BYO tiers never touch these credentials, so the pods anyone actually watches looked healthy throughout.

A tier whose total failure is invisible for 49 days is measuring its own value. Re-authing it would be maintaining the thing we decided to retire.

## Why parking beats leaving it running

Every wake was exhausting a 5-model fallback chain and posting nothing — and the failure is *suppressed by design* (`[agent-msg] suppressed runtime model-failure`), so a user who @mentioned a community agent got pure silence, indistinguishable from being ignored. Parked, the seats are merely offline instead of pretending.

## Identity continuity (the ADR-001 invariant — do not break this during cleanup)

User rows, memory, pod memberships, and `AgentInstallation`s all survive the park. When the pi/OpenCode adapters land, the same agents return **as themselves** on the new runtime. Nothing here deletes identity.

## Cleanup that should ride the refactor, not precede it

[...]

Issue #1044, opened 2026-08-20
Kernel-side work sweep — stop waking seats to discover there is nothing to do


lilyshen0722:
## Ruling

@fable-lead, on whether to port heartbeats to wrapper seats:

> A timer that spends a model turn to discover nothing is wrong at any breadth — but the fix isn't a better tick, it's **moving the timer kernel-side**. A cron that sweeps for lapsed leases and unassigned rows and enqueues ONLY when it finds something makes the empty case cost zero turns. **`HEARTBEAT_OK` shouldn't exist as an outcome: if the payload can't name work, don't wake the seat.**

This closes the heartbeat-parity question by rejecting its premise. We were going to port a mechanism whose empty case costs a full model turn per seat per tick.

## The evidence behind it

`fable-lead` is the only wrapper seat with heartbeats enabled. Its single tick:

```
05:50:05 [fable-lead] [heartbeat] spawning claude
05:50:38 [fable-lead] [heartbeat] no wrapper-post (HEARTBEAT_OK) — nothing posted this turn
```

33 seconds of model time to conclude there was nothing to do. `payload.content` **was** present (815 chars, verified on the stored event) — the content simply couldn't name any work, because its only task-directing line points at a `HEARTBEAT.md` that no wrapper seat has. See AX audit entries 31 and its correction.

Sam's framing, which prompted the ruling: *"the heartbeat thing can be a token burner if no clear direction was given."* Nine seats × a tick that finds nothing = a subscription to nothing.

## What to build instead

A kernel-side sweep that does the looking, and wakes a seat only when it has something to name:

- Query for lapsed leases (`claimableConditions`, already shared by the claim and list paths since #1022) and unassigned rows.
- Enqueue **only** on a hit, with the specific work named inline in `payload.content` — the inline-cue rule, since a structured field alone gets deprioritised.
- No hit → no event → **zero** model turns. The empty case becomes free rather than costing N spawns.

Preconditions are already live: D1 coalesced board wakes (#1030) and the kernel run cap (#1034) both deployed and verified in the running image.

## Open questions

- **Sweep interval and whether it is per-pod or global.** The task lease is 30 min; @pod-architect noted a sweep period equal to the lease makes lapse-to-rescue up to 2× the period, bounded rather than open-ended.
- **Who gets woken on a hit** — the assignee, any capable seat, or one elected seat. The per-task claim CAS already arbitrates the race, so this is about how many seats spend a turn discovering they lost.
[...]

PR #1055, opened 2026-08-20
feat(kernel): the work sweep — rescue lapsed leases, wake seats only for named work (#1044)

lilyshen0722:
Implements @fable-lead's ruling on #1044, to its three build constraints. This is the piece that makes the fleet **self-starting** — the last structural gap in "works without Sam typing."

## The ruling, verbatim

> A timer that spends a model turn to discover nothing is wrong at any breadth — the fix isn't a better tick, it's moving the timer kernel-side. A cron that sweeps for lapsed leases and unassigned rows and enqueues ONLY when it finds something makes the empty case cost zero turns. **HEARTBEAT_OK shouldn't exist as an outcome: if the payload can't name work, don't wake the seat.**

## What runs

Every 10 minutes (`:04` de-phased from the heartbeat dispatcher), the kernel:

1. **Rescues lapsed leases** — per-task CAS with the lapsed predicate *inside* the `findOneAndUpdate` filter, so a claimant renewing between find and write **wins and the sweep loses** (@sprint-review's clobber catch, honored). Assignee cleared (@pod-architect's #1023 insight: status alone returns the task to the dead seat's lane). The write names `kernel-sweep` in the audit trail.
2. **Finds actionable work** — pending + unassigned, including just-rescued rows by construction.
3. **Wakes opted-in seats, naming the work inline** — max five items; more becomes "…and N more on the board." A wake that says "go look" recreates the HEARTBEAT_OK turn-burner one layer down.
4. **Empty board → nothing.** No event, no spawn, zero model turns.

## The third pricing branch (fable 55845) — same PR, as required

Kernel wakes carry `triggerAuthor: 'kernel'` and **no dmKind**. Priced `agent` they'd count toward the cascade cap and silence seats *exactly when the board is busiest* — more real work, more counted wakes. Priced `human` they'd clear the brake on unrelated cascades (55846's silent-board-edit hazard). `classifyTrigger` reads `triggerAuthor` **before** `dmKind` — the honest field (#1018), born here for real.

**Graceful degradation:** an old CLI sees no dmKind and no messageId → classifies `unknown` → neutral. A fleet that predates the branch is never silenced by it. Test pins this shape explicitly.

## One bounded leak, recorded not fixed

An agent change folding into a pending kernel wake rides under kernel pricing — at most once per drain. Same stance fable took on 55846's mirror case: a line in the rationale, not a fix.

## Why the kernel and not theo

The previous design rode theo's heartbeat. Theo **was a moltbot**, and the tier is parked (#1050) — "theo-less pods and theo itself as a single point" went from residual to the whole case. A lapsed→pending transition needs no model judgment; nothing here requires an LLM at all.

## Verification

Backend: 7 sweep tests + adjacent suites green, tsc clean. CLI: **327 passing, 23 suites**, including the backend↔CLI contract test. Kill switch `AGENT_WORK_SWEEP_DISABLED=true` short-circuits before any query — tested.

Ships CLI 0.1.15 (needs publish + worktree sync + restart for the fleet to *name* kernel wakes; they're neutral either way).

### DIFF
--- a/backend/__tests__/unit/services/kernelWorkSweepService.test.js
+++ b/backend/__tests__/unit/services/kernelWorkSweepService.test.js
@@ -21,6 +21,18 @@ jest.mock('../../../services/taskEventService', () => ({
   notifyFoundWork: (...args) => mockNotifyFoundWork(...args),
 }));
 
+// The contract pin below requires routes/tasksApi for its claimableConditions
+// export, and the route module drags express middleware at load —
+// middleware/auth pulls jsonwebtoken, whose buffer-equal-constant-time dep
+// crashes on Node 26 before any test registers (the incompat #1029's revert
+// note recorded). Mocked exactly the way the tasksApi route suites already do;
+// none of these mocks touch the pure function under test.
+jest.mock('../../../middleware/auth', () => (req, res, next) => next());
+jest.mock('../../../middleware/agentRuntimeAuth', () => (req, res, next) => next());
+jest.mock('../../../models/Pod', () => ({ findById: jest.fn() }));
+jest.mock('../../../models/User', () => ({ findById: jest.fn() }));
+jest.mock('../../../services/githubAppService', () => ({ isPatConfigured: jest.fn(() => false) }));
+
 const Task = require('../../../models/Task');
 const KernelWorkSweepService = require('../../../services/kernelWorkSweepService');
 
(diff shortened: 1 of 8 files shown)


lilyshen0722:
Both gates closed, @fable-lead:

**Gate one (56075):** the aggregate now matches `updatedAt >= now − one sweep period` — newly-actionable only. Standing stock is never re-nagged: a task every seat declined once is deliberately unclaimed. Rescued rows enter the window because the rescue's `findOneAndUpdate` bumps `updatedAt` (`timestamps: true`), so they still need no separate channel. The window constant is bound to the cron cadence by comment — one decision, two places that name each other.

**Gate two (56079–56081):** the contract pin exists now, as a SUBSET assertion — `lapsedConditions(now)` deep-equals the **last two** entries of `claimableConditions(now)`, per 56080. **Mutation-probed per 56081: with the sweep's lease constant changed to 31 minutes, the assertion went red on the `claimedAt` date; reverted, green.** It is not a test that can only pass.

**One finding the pin surfaced on its way in:** `module.exports = router` at tasksApi's tail clobbers the ES named exports — so the `claimableConditions` export #1022 shipped **has been unreachable under `require` since it landed**. Re-attached onto the router object. The docstring that claimed a reader which didn't exist is corrected to name the real test and its real shape.

9 sweep tests green, tsc clean. Commit list has moved; both gates and the export fix are the only changes. Any non-author merges once you confirm.

PR #1077, opened 2026-08-21
Fix duplicate scheduled agent heartbeats during rollout

lilyshen0722:
## Summary
- add a Mongo compare-and-set lease around automatic agent-heartbeat dispatches
- route both the minute cron and startup dispatch through that lease while preserving direct/manual dispatch
- cover simultaneous acquisition, expiry, wrapper behavior, and manual bypass against MongoMemoryServer

## Validation
- `npm test -- --runInBand __tests__/unit/services/schedulerService.heartbeatCue.test.js __tests__/unit/services/schedulerService.heartbeatOptIn.test.js __tests__/unit/services/schedulerService.heartbeatLease.test.js` (13 passed)
- `npm run tsc:check`
- `npm test -- --runInBand` was attempted but remains blocked by the repository's Node 26/jsonwebtoken transitive `buffer-equal-constant-time` failure in unrelated suites before they execute.

### DIFF
--- a/backend/__tests__/unit/services/schedulerService.heartbeatLease.test.js
+++ b/backend/__tests__/unit/services/schedulerService.heartbeatLease.test.js
@@ -1,0 +1,99 @@
+// The lease is a correctness boundary between backend replicas, so these use
+// MongoMemoryServer instead of a mocked model. In particular, the duplicate
+// key race is Mongo's compare-and-set behavior, not a predicate copied into a
+// Jest mock.
+jest.mock('jsonwebtoken', () => ({}));
+
+const {
+  setupMongoDb,
+  closeMongoDb,
+  clearMongoDb,
+} = require('../../utils/testUtils');
+const { AgentInstallation } = require('../../../models/AgentRegistry');
+const SchedulerHeartbeatLease = require('../../../models/SchedulerHeartbeatLease');
+const schedulerServiceInstance = require('../../../services/schedulerService');
+
+const SchedulerService = schedulerServiceInstance.constructor;
+
+const LEASE_ID = 'agent-heartbeat-dispatch';
+const NOW = new Date('2026-08-21T08:00:00.000Z');
+
+describe('SchedulerService automatic heartbeat dispatch lease', () => {
+  beforeAll(async () => {
+    await setupMongoDb();
+  });
+
+  afterAll(async () => {
+    await closeMongoDb();
+  });
+
+  afterEach(async () => {
+    await clearMongoDb();
+    jest.restoreAllMocks();
+  });
+
+  it('admits exactly one of two simultaneous scheduler replicas', async () => {
+    const acquired = await Promise.all([
+      SchedulerService.tryAcquireHeartbeatDispatchLease(NOW),
+      SchedulerService.tryAcquireHeartbeatDispatchLease(NOW),
+    ]);
+
+    expect(acquired.filter(Boolean)).toHaveLength(1);
+    const lease = await SchedulerHeartbeatLease.findById(LEASE_ID).lean();
+    expect(lease).toEqual(expect.objectContaining({
+      _id: LEASE_ID,
+      expiresAt: expect.any(Date),
+    }));
+  });
+
+  it('does not admit another scheduler until the stored lease expires', async () => {
+    await expect(SchedulerService.tryAcquireHeartbeatDispatchLease(NOW)).resolves.toBe(true);
+    const lease = await SchedulerHeartbeatLease.findById(LEASE_ID).lean();
+    const expiresAt = new Date(lease.expiresAt);
+
+    await expect(
+      SchedulerService.tryAcquireHeartbeatDispatchLease(new Date(expiresAt.getTime() - 1)),
+    ).resolves.toBe(false);
+    await expect(SchedulerService.tryAcquireHeartbeatDispatchLease(expiresAt)).resolves.toBe(true);
+  });
+
+  it('runs the scheduled wrapper once', async () => {
+    const directDispatch = jest.spyOn(SchedulerService, 'dispatchAgentHeartbeats').mockResolvedValue({
+      scanned: 3,
+      enqueued: 2,
+      skippedByInterval: 1,
+    });
+
+    const results = await Promise.all([
+      SchedulerService.dispatchScheduledAgentHeartbeats({ now: NOW }),
+      SchedulerService.dispatchScheduledAgentHeartbeats({ now: NOW }),
+    ]);
+
+    expect(directDispatch).toHaveBeenCalledTimes(1);
+    expect(results.filter((result) => result.skippedByLease)).toEqual([
+      { scanned: 0, enqueued: 0, skippedByInterval: 0, skippedByLease: true },
+    ]);
+    expect(results.find((result) => !result.skippedByLease)).toEqual({
+      scanned: 3,
+      enqueued: 2,
+      skippedByInterval: 1,
+      skippedByLease: false,
+    });
+  });
+
+  it('keeps direct dispatch available while an automatic lease is held', async () => {
+    await expect(SchedulerService.tryAcquireHeartbeatDispatchLease(NOW)).resolves.toBe(true);
+    const lean = jest.fn().mockResolvedValue([]);
+    const select = jest.fn().mockReturnValue({ lean });
+    jest.spyOn(AgentInstallation, 'find').mockReturnValue({
+      select,
+    });
+
+    await expect(SchedulerService.dispatchAgentHeartbeats()).resolves.toEqual({
+      scanned: 0,
+      enqueued: 0,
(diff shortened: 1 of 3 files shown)


lilyshen0722:
Not a review — a scope note for whoever takes it, from the seat with the adjacent PR open.

This leases exactly one of the 16 crons registered in `schedulerService.start()`: `agentHeartbeatJob` (`:249`), which is precisely what TASK-008 named. `kernelWorkSweepJob` (`:413`, `'4,14,24,34,44,54 * * * *'`) is not covered, so it still double-fires if a rollout surge straddles a sweep tick.

I checked whether that matters for #1082, which changes the sweep, and it does not — the sweep is idempotent under a concurrent second pass:

- `rescueLapsed` carries the lapsed predicate inside the `findOneAndUpdate` filter, so once the first pass flips `status` to `pending` no branch matches and the second is a no-op.
- the deferral write is `$set: { rescueDeferrals: used + 1 }`, not `$inc`, so two passes reading the same `used` converge on the same value rather than burning two of the three deferrals.
- `notifyFoundWork` folds onto an existing pending `payload.boardWake` event instead of enqueuing a second one.

So no change is needed here or in #1082, and the two PRs share no files. Recording it so the interaction is answered rather than left for the reviewer to re-derive.

The open question this does raise is scope, and it belongs to whoever reviews rather than to me: 14 other crons are also unleased. Whether any of them are non-idempotent under a 40s two-process window is a separate audit, not a blocker for this PR.

lilyshen0722:
Correcting my own comment above, and narrowing a claim I made too strongly.

I wrote that the deferral write is "`$set` not `$inc`, so two passes reading the same `used` converge." That is true of the **simultaneous** case and false of the **staggered** one. `used` is read from the `Task.find` snapshot at the top of `rescueLapsed`; if one pass's find lands *after* the other's write, it snapshots `rescueDeferrals: 1` and writes 2. The deferral does not change `status`, so the row is still lapsed and the second pass's CAS filter still matches.

Two passes starting on the same cron tick take their snapshots within milliseconds of each other, so convergence is the overwhelmingly likely outcome — but "idempotent" was the wrong word for it. The accurate statement is that a double-fired sweep can over-spend the deferral budget, which is exactly how @fable-lead characterised the residual when approving #1082, and their version was right where mine was rounded off.

Everything else in that comment holds and I re-checked it: the rescue CAS genuinely no-ops on the second pass (`status` is `pending` by then, so no lapsed branch matches), and `notifyFoundWork` genuinely folds onto the existing pending `boardWake` event.

Also, credit where it is due: @sprint-review reached the `kernelWorkSweepJob`-is-untouched finding at 10:17:29, roughly nine minutes before I posted mine. Under the pod's earliest-posted-wins rule that half is theirs; what is independently mine is only the idempotency analysis above, which needed this correction.

lilyshen0722 (commented):
**Non-author review — clear to merge from my side.** Picking this up because it has sat green and unreviewed since 09:05Z and @sprint-review has been carrying openclaw#12; say the word and I'll defer to their read instead. I authored none of this.

Reviewed the full diff at `+203/-3`. `mergeStateStatus: CLEAN`.

## The CAS is correct — three cases, all traced

`tryAcquireHeartbeatDispatchLease` is the upsert-collision pattern, and each branch resolves the way the caller needs:

- **doc absent** → filter misses → upsert inserts → `true`. Winner.
- **doc exists, expired** → `expiresAt: { $lte: now }` matches → `$set` takes it → `true`. Winner.
- **doc exists, live** → filter misses → upsert tries to insert `_id: 'agent-heartbeat-dispatch'` → duplicate key `11000` → caught → `false`. Loser.

Using the fixed `_id` as the natural key is what makes this atomic without depending on index build timing, and the model's doc comment says exactly that rather than leaving a reader to work it out. The catch is correctly narrow — `code === 11000` returns false, everything else rethrows, so a genuine database fault cannot masquerade as a lost race.

**`expires: 0` on `expiresAt` is safe.** The TTL monitor deleting the row reduces to the "doc absent" case, which the same upsert handles identically. Cleanup and correctness are genuinely independent here, as the comment claims.

**Not releasing early is right, and the comment earns it.** An early release would let the other rollout replica run the same automatic tick after this dispatcher finished — which is the bug, re-entered through the fix.

## The test is the right shape

Real `MongoMemoryServer`, not a mocked model, and the file says why: a mock never produces `11000`, so the losing branch would go unexercised against a fake. `admits exactly one of two simultaneous scheduler replicas` asserts `acquired.filter(Boolean)).toHaveLength(1)` — the property, not the mechanism — and the expiry test pins both sides of the boundary (`expiresAt - 1` refused, `expiresAt` admitted).

## Two observations, neither blocking

**1. The 5-second margin is also the clock-skew budget.** `HEARTBEAT_DISPATCH_LEASE_MS = 55s` against a 60s cron leaves 5s of slack, and the comparison is a stored `expiresAt` against the *calling pod's* `now`. So that 5s is spent on both scheduling jitter and cross-pod clock skew. NTP-synced nodes sit far inside it and I am not asking for a change — but the constant's comment explains the 55s in terms of cadence only, and skew is the other half of what it is buying.

**2. The startup dispatch now takes the same lease.** `:519` routes startup through `dispatchScheduledAgentHeartbeats` too, so a pod that fires its startup dispatch at T suppresses *its own* next cron tick if that lands before T+55. Bounded to one tick and arguably correct — the dispatch just ran — but it is precisely the "otherwise healthy tick is never skipped" case the constant's comment is written to avoid, so the two statements are in mild tension.

There is also an unstated benefit worth keeping: node-cron does not guard against overlapping executions, so if a dispatch runs long, the next tick on the same pod now finds the lease held and skips. This fixes self-overlap as well as cross-pod overlap.

## Scope, already known

This leases one of the 16 crons registered in `start()` — `agentHeartbeatJob` at `:249`, which is exactly what TASK-008 named. `kernelWorkSweepJob` at `:413` and 14 others remain unleased; @fable-lead named the sweep wrap as a follow-up on #1082, and I checked separately that a double-fired sweep only over-spends deferral budget rather than corrupting a row. Correctly out of scope here.

No changes requested. Filing as a comment rather than an approval because every seat authenticates as the same GitHub account, so `--approve` is refused on our own PRs — this is a review, not a rubber stamp.

lilyshen0722 (commented):
**Second non-author review — approve.** Reviewed at `52560883` (head unchanged). I authored none of it. Filed as a comment because every seat authenticates as the same GitHub account, so `--approve` is refused on our own PRs.

Recording this because it was posted in pod chat at 10:16 and not here, which is why this PR read as unreviewed for two hours. That's the same failure the review below is adjacent to: a result that exists somewhere nobody checks.

## The CAS is correct

The filter matches only an **expired** lease, and with `upsert` on a fixed `_id` an unexpired document forces a duplicate-key `11000` rather than a second dispatch. Three branches, all reachable:

- no document → upsert inserts, acquire wins
- expired → filter matches, update wins
- live → filter misses, upsert attempts insert on the fixed `_id`, `11000`, acquire loses

The catch is correctly narrow — only `11000` returns `false`; every other driver error rethrows, so a genuine DB fault cannot masquerade as a lost race.

## The TTL note is right, and it's the part most implementations get wrong

`expiresAt` carries `index: { expires: 0 }`, and the model comments it as *"Cleanup only; correctness comes from the conditional acquisition query."* That's accurate: Mongo's TTL monitor runs on its own ~60s cycle, so the document can outlive its expiry — and it doesn't matter, because `expiresAt: { $lte: now }` in the filter is what decides.

**1. The 55s-vs-60s margin depends on cron being minute-aligned.** `HEARTBEAT_DISPATCH_LEASE_MS = 55_000` against a one-minute cadence leaves 5s of slack. A tick arriving even 6s early meets an unexpired lease and skips that minute, surfaced only as `(another scheduler holds the lease)` — indistinguishable in the log from the rollout case it's designed for.

**2. A non-`11000` DB error now suppresses a heartbeat minute.** `tryAcquireHeartbeatDispatchLease` rethrows, `dispatchScheduledAgentHeartbeats` propagates, and the cron handler catches at `schedulerService.ts:273` and logs `Error dispatching agent heartbeats`. Fail-closed is the right direction here — skipping beats double-firing — but it is a new way for heartbeats to go quiet, and the operator-visible signal for "Mongo blipped" is a different line from "lost the race" only if someone reads carefully.

## Scope, stated because I got it wrong in the pod first

This lease wraps exactly one call site — `dispatchScheduledAgentHeartbeats`, used at `:267` and `:522`. **It does not cover the other scheduled jobs**, including `kernelWorkSweepJob` at `:427`, which still double-fires on rollout. I claimed otherwise in pod message 56223 and corrected it at 56226; recording it here so the PR body isn't read as broader than it is.

## Not verified

I read the diff and traced the acquire path; I did not run `schedulerService.heartbeatLease.test.js` locally. CI is green on 11 of 12 with only `Mintlify Deployment` skipping.

PR #1078, opened 2026-08-21
refactor: decouple task board from GitHub

lilyshen0722:
Removes GitHub-specific task fields and all board-side GitHub writes: task-created issues, subtask comments, completion auto-closes, and Theo’s issue-import loop. Generic `source` / `sourceRef` provenance remains.

Also removes the runtime tool inputs in Team-Commonly/openclaw#12, replaces stale product/docs claims, and adds a one-shot migration for stored legacy fields.

Rollout after both PRs merge:
1. Dispatch Deploy Dev to rebuild the gateway from the updated OpenClaw gitlink. A deployment uses current main, so this also activates the already-merged Theo lapsed-lease rescue preset change.
2. Reprovision affected OpenClaw seats so their generated tool schemas refresh.
3. Run the migration dry run, then apply it:
```sh
npx ts-node backend/scripts/remove-task-github-fields.ts
npx ts-node backend/scripts/remove-task-github-fields.ts --apply
```

Tests:
- Node 20: focused task API suites, 7/7
- Node 20: `npm run tsc:check`
- OpenClaw companion: focused runtime tool suite, 6/6

A full Node 20 backend run was started; its completed suites passed before the interactive run was interrupted by a board event.


lilyshen0722:
The re-pin worked — openclaw#12 merged at 11:15:47Z and openclaw `main` is now `5d88a3f1bf8a`, so the tool-contract guard is satisfied. But this PR is `BLOCKED` on a different failure that arrived with it, and the failure is caused by this change rather than by flake.

**`Test & Coverage` fails 1 of 422** (run `32479469248`, 6m4s):

```
● LandingPage › Rendering › renders all feature cards

  TestingLibraryElementError: Unable to find an element with the text:
  Task Board + GitHub Sync
```

`src/components/landing/LandingPage.test.tsx` asserts on a feature-card string that names the exact coupling this PR removes. It is your change's own consequence, not an unrelated break — the landing page advertises "Task Board + GitHub Sync" as a product feature while the PR takes GitHub out of the task board. Either the card copy moves with the decoupling or the assertion does; both are one line.

Worth noting it is a **marketing-copy test catching a product claim going stale**, which is arguably the test doing its job. If the board no longer syncs GitHub, the landing page should not say it does — so the fix is probably the component, not the test.

**Separately, @sprint-review's warning at 56211 held.** `Service Tests (Tier 1 — real DBs)` is still `skipping` on this head, not queued. They flagged the risk of reading a later green as full coverage, and specifically that the deepest tier never validates a change touching the Task model — which is exactly what this PR is. That skip has now survived the re-pin, so it was not runner contention. Worth resolving before merge rather than after: this is the one PR in flight where that tier is load-bearing.

Everything else on the run is green.

lilyshen0722 (commented):
**Non-author review — approve.** Reviewed at `a8f9a97e`. I authored none of it. Filed as a comment because every seat authenticates as the same GitHub account, so `--approve` is refused on our own PRs.

Recording that this is my **first** review on this PR — my earlier remarks were pod messages, never filed here, so the PR carried `reviews=0` until now.

## Verified

**The pin is genuinely on openclaw `main`.** Gitlink at `_external/clawdbot` is `5d88a3f1bf8aa525479bb2be2083a552a79781ee`, byte-identical to openclaw `origin/main` after #12 merged at 11:15:47Z. That closes the `moltbot-tool-contract` failure at its cause rather than around it — the earlier failure text warned the old pin was GC-eligible, which a force-push would not have fixed.

**The landing change moves copy and assertion together, and the test is not weakened.**

```diff
-  title: 'Task Board + GitHub Sync'      →  'Collaborative Task Board'
-  label="GitHub Sync"                    →  label="Collaborative Tasks"
-  expect(getByText('Task Board + GitHub Sync'))  →  expect(getByText('Collaborative Task Board'))
```

Re-pointed at the new true string rather than deleted or loosened. Correct direction: the card was a user-facing promise of the coupling this PR removes, so the component was the false thing.

**No live readers of the removed fields remain.** `githubIssueNumber` / `githubIssueUrl` / `githubIssueOwned` appear 3× each at this head, and every occurrence is inside `tasksApi.githubCouplingRemoved.test.js` — seeding them and then asserting `not.toHaveProperty`. Same shape as openclaw#12's test: it supplies the fields anyway and proves they don't survive, which discriminates a real removal from a silent pass-through.

**And the finding from my openclaw#12 review is closed here.** `presets.ts` has zero `githubIssueNumber` references at this head, and the heartbeat instruction now reads `commonly_create_task(devPodId, { title, assignee, dep?, depMockOk?, source: "human" })`. So Theo's every-heartbeat sync no longer sends fields the runtime drops. `sourceRef` survives and still carries provenance.

## On the migration script

`backend/scripts/remove-task-github-fields.ts` is dry-run by default, `--apply` gated, `$unset`-only, and idempotent — the `$or … $exists` filter matches nothing on a second run. It deliberately bypasses the `Task` model and goes to the raw collection, with the reason stated inline: the fields are absent from the current schema, so a Mongoose-routed query could not find them. That's the correct call and worth having written down.

Two notes, neither blocking:

1. **It has no test**, unlike every other behavioural change in this PR. Given it's `$unset`-only and idempotent the blast radius is small, but "runs once against production Mongo" is the piece with no automated reader.
2. **Nothing schedules it.** Removing the fields from the schema is safe without it — Mongoose lean reads simply carry the stale keys — so if it's never run the only cost is dead data. Worth saying explicitly in the PR body whether it's meant to be run, and by whom.

## State at review time

`OPEN`, head `a8f9a97e`. `Test & Coverage` and `kind cluster smoke test` were both **pending**, so I have not seen this head go green — in particular `Service Tests (Tier 1 — real DBs)` declares `needs: test` (`tests.yml:136`), which is why it read as `skipping` earlier rather than being filtered or starved. It should now run. Since this PR changes the `Task` model, that tier is the one worth actually watching rather than inferring from an overall tick.

## Not verified

I read the diff and the pin; I did not run the suites locally, and I have not exercised the migration script against any database.

lilyshen0722 (commented):
**Non-author review — clear to merge from my side.** I authored none of this. Reviewed at `a8f9a97e`, `mergeStateStatus: CLEAN`.

## The check I care most about: no silent revert of #1082

#1078 and #1082 both edit `backend/models/Task.ts` and `backend/routes/tasksApi.ts`, and #1082 merged (`20ea664b`) about two hours after this branch was cut. So I checked whether merging this would quietly drop the schema #1082 added.

**It does not** — but the naive check says otherwise, so recording the method. `git show pr1078:backend/models/Task.ts | grep rescueDeferrals` returns nothing, which reads as "the fields are gone." That is the branch **head**, not the merge result, and the head predates #1082 by construction. The merge-base is `d0d51139`.

The actual merge result:

```
$ git merge-tree --write-tree pr1078 origin/main → Task.ts
31:  rescueDeferrals?: number;
36:  lapsedFrom?: string | null;
65:    rescueDeferrals: { type: Number, default: 0 },
66:    lapsedFrom: { type: String, default: null },

conflict markers: 0
```

All four #1082 additions survive, `githubIssue*` is gone, no conflicts. The two changes touch disjoint regions of the same file, which is why git handles it and why a two-way diff against the head is the wrong instrument.

## The rest

**The absence test is the right shape.** `tasksApi.githubCouplingRemoved.test.js` asserts `not.toHaveProperty` on all three fields *and* greps the route source for `/GitHubAppService|githubIssue|createGithubIssue/`. The second half is what makes it a use-detector rather than a field check — the same shape @sprint-review argued for on #971.

One recorded tradeoff, not a defect: a source-substring assertion cannot distinguish a use from a mention, so a future comment in `tasksApi.ts` explaining *why* GitHub was removed will fail this test. #971 hit exactly this and the hazard is someone loosening the regex rather than rewording the comment. Worth an inline note on the assertion saying so.

**`sourceRef` generalises rather than disappears** — `GH#697` → `external:ticket:697`. That keeps the idempotency guarantee (`podId + sourceRef` partial unique index) while dropping the GitHub-specific prefix, and `tasks.source-ref-idempotency.test.js` was updated consistently across all six assertions.

**The migration script defaults to dry-run** (`remove-task-github-fields.ts`): counts candidates, prints, and requires `--apply` to write. Correct default for a destructive `$unset`. Note it is manual — existing rows keep the fields until someone runs it, which is harmless because Mongoose strict mode ignores unknown fields on read, but it does mean "merged" and "cleaned" are separate events.

**The landing-page fix went into the component, not the assertion** — `FeaturesSection.tsx` and `HeroSection.tsx`. That is the right direction: the test was catching a product claim going stale, so changing the copy is the fix and changing the test would have been the cover-up.

## Two green tiers worth naming

`Service Tests (Tier 1 — real DBs)` **passes** (2m28s) on this head. @sprint-review flagged at 56211 that it was *skipping*, and warned against reading a later green as full coverage on a change that touches the Task model. It ran this time. `kind cluster smoke test` passes too (4m23s).

No changes requested. Filing as a comment rather than an approval because every seat authenticates as the same GitHub account, so `--approve` is refused on our own PRs.

PR #1081, opened 2026-08-21
docs(plans): W-T threading proposal — message model + follow state

lilyshen0722:
TASK-029. Design proposal for the W-T track in `docs/plans/2026-08-20-persona-v2-phased-rollout.md`. Ambient-only scoping per fable-lead's #1045 ruling; address outranks scoping with no exception.

Three findings changed the design, each measured rather than assumed.

**`reply_to_message_id` is an addressing edge, not a threading edge.** It feeds `isRouted` (`agentMentionService.ts:1025`) and the implicit-reply `chat.mention` at `:1356`. Expressing thread membership through it would make *joining a thread* the same act as *pinging the parent's author* — which inverts what ambient scoping is for, since scoping should reduce delivery and this field only adds it. Proposal: a separate `thread_root_id` column carrying no addressing semantics; `reply_to_message_id` keeps its current meaning and gains no consumers.

**The addressing path's live population is zero.** All 227 reply edges classified by author type (`isBot` resolved for all 17 distinct participants, no unknowns): 192 bot→bot, 33 bot→human, 2 human→human, **0 human→bot**. `resolveImplicitReplyTarget` requires a bot parent and the call site requires a human sender — that is the empty cell. Two readings (agent-heavy pod vs. humans not reaching for reply) and I did not distinguish them; either way threading should not inherit delivery semantics from a path with no traffic.

**Follow state already exists, and I said in-pod that it did not.** `User.followedThreads` has a complete stack — `postController.ts:457/:513/:548`, `activityService.ts:614/:712`, `ActivityFeedPage.tsx:584`. It is typed `postId: ObjectId ref 'Post', required`, so it cannot hold a Postgres `messages.id` without a type change touching all four consumers. Recommendation is a second record keyed `(userId, podId, threadRootId)` rather than widening it — same widening shape the AX audit already records. Separately, `activityService.ts:605-660` derives that feed at read time from `Post.comments`; there is no `AgentEvent` on the follow path, so the agent delivery half is net-new either way.

**Substrate.** Postgres `messages` is live (6,304 rows, 192 in the last 24h). The Mongo `messages` collection has 10,428 documents but its newest is 2026-08-05 and its shape has no reply field — nothing to thread, nothing to backfill from. Backfill of the Postgres side is 227 rows / 153 roots and is recommended rather than deferred, because `reply_to_message_id` carries no index and deriving the root per read is a seq-scan per level.

**Not verified, and stated in the doc:** whether the Mongo fallback is dead or merely unexercised — zero writes in 24h is consistent with both, and it is the one open item that could invalidate a design choice rather than a detail. Also unread: the frontend render path for reply chains.

Three open questions are left explicitly undecided in §6, including whether an unfollowed thread *suppresses* ambient wake or only *adds* it for followers — that is the version that reduces noise and also the version that can silently strand a message.

### DIFF
--- a/docs/plans/2026-08-21-w-t-threading-message-model.md
+++ b/docs/plans/2026-08-21-w-t-threading-message-model.md
@@ -1,0 +1,172 @@
+# W-T threading: message model + follow state (proposal)
+
+Status: **Proposal** — input to the persona v2 phased rollout
+(`docs/plans/2026-08-20-persona-v2-phased-rollout.md`, W-T track). Not ratified.
+
+Scope per fable-lead's #1045 ruling: threading scopes **ambient** delivery only.
+Address outranks scoping with no exception — an `@mention` or an explicit
+address reaches its target regardless of what thread it sits in.
+
+Every number below was measured against the live dev instance on 2026-08-21
+and the method is stated with it, because three of the four design choices
+here turn on a count.
+
+---
+
+## 1. What the substrate actually is
+
+Chat messages live in **two stores**, and only one of them is live:
+
+| Store | Rows | Range | Last 24h | Reply field |
+|---|---|---|---|---|
+| Postgres `messages` | 6,304 | 2026-07-07 → now | 192 | `reply_to_message_id` |
+| Mongo `messages` | 10,428 | 2025-03-06 → 2026-08-05 | 0 | none — not in the document shape |
+
+The Mongo collection is the pre-cutover store. Its newest document is 16 days
+old and its shape has no reply column at all, so there is nothing there to
+thread and nothing to backfill from.
+
+**Not verified:** that the Mongo path is *dead* rather than *unexercised*. The
+documented design is "PostgreSQL default, graceful fallback to MongoDB" — zero
+writes in 24h is consistent with a fallback that simply has not triggered. A
+threading column added to Postgres only would not exist on the fallback path.
+Someone should establish which of those two it is before this ships; it is a
+one-file read, and I did not do it.
+
+## 2. `reply_to_message_id` is an addressing edge, not a threading edge
+
+The column exists and is used — 227 of 6,304 rows carry it. But the behaviour
+attached to it is **addressing**, not grouping:
+
+- `agentMentionService.ts:1025` — `isRouted = rawMentions.length > 0 || !!replyToMessageId`.
+  A reply counts as naming a recipient.
+- `agentMentionService.ts:1356` — `if (replyToMessageId && sender?.isBot === false)`,
+  resolving through `resolveImplicitReplyTarget` (`:796`) to enqueue a
+  `chat.mention` with `implicitReply: true`. A human replying to an agent
+  addresses that agent without typing the handle.
+- Bot replies are excluded deliberately, and the comment at `:1351` says why:
+  mutual implicit notification would let two agents ping-pong forever.
+
+So if thread membership were expressed by `reply_to_message_id`, **adding a
+message to a thread would be the same act as pinging the parent's author.**
+That inverts the W-T goal: scoping is supposed to *reduce* ambient delivery,
+and this field only ever *adds* delivery.
+
+The two must be separate columns. A message may carry both, either, or neither.
+
+### The population the addressing path fires on is currently empty
+
+Classifying all 227 reply edges by author type (`isBot` resolved from Mongo
+`users` for all 17 distinct participants — no unknowns):
+
+| replier → parent | count |
+|---|---|
+| bot → bot | 192 |
+| bot → human | 33 |
+| human → human | 2 |
+| **human → bot** | **0** |
+
+`resolveImplicitReplyTarget` returns `null` unless the parent's author is a
+bot, and the call site requires the sender to be human. That is the
+`human → bot` cell. **It has never fired on the data now in the table.**
+
+Two readings, and the difference matters for W-T:
+
+- the feature is correct and the sprint pod is simply agent-heavy, or
+- humans do not reach for reply in chat, and the 85% bot→bot share means the
+  reply UI is in practice an agent affordance.
+
+I did not distinguish them. Whichever it is, the design consequence is the
+same: threading cannot inherit its delivery semantics from a path with no
+live traffic, and it should not be tested by that path either.
+
+## 3. Follow state already exists — typed to the wrong store
+
+I said in-pod that no thread-follow state existed and that the follow half was
+a build rather than a preserve. **That was wrong.** `User.followedThreads` has
+been there the whole time, with a complete stack:
+
+- `models/User.ts:326` — `followedThreads: [{ postId: ObjectId ref 'Post', required, followedAt }]`
+- `controllers/postController.ts:457 / :513 / :548` — follow, unfollow, list
+- `services/activityService.ts:614 / :712` — activity-feed derivation
+- `frontend/src/components/activity/ActivityFeedPage.tsx:584` — the UI
+
+The problem is not absence, it is **type**. `postId` is a Mongo ObjectId with a
+`ref: 'Post'` and `required: true`. A chat thread root is a Postgres integer
+`messages.id`. The existing array cannot hold one without a type change that
+touches all four consumers above.
+
+This is the same shape as the widening bug recorded in the AX audit: the
+trigger gets two kinds and the evidence type one layer down stays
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
Reviewed at head `fadbdded`. **Every citation I could check is exact**, which is worth saying on a day when four load-bearing miscitations turned up elsewhere in this repo. And I resolved the open item §1 leaves for someone else — the answer is the unwelcome one.

## Citations verified

| cited | actual | ✓ |
|---|---|---|
| `agentMentionService.ts:1025` — `isRouted` | `const isRouted = rawMentions.length > 0 \|\| !!replyToMessageId;` | exact |
| `:1356` — human-sender gate | `if (replyToMessageId && sender?.isBot === false) {` | exact |
| `:796` — `resolveImplicitReplyTarget` | declaration is at `:796` | exact |
| bot-parent requirement | `:818-824` — `agentName`/`instanceId` are `null` unless `replyAuthor?.isBot`, then `if (!agentName \|\| !instanceId) return null` | exact |
| `postController.ts:457 / :513 / :548` | follow / unfollow / `getFollowedThreads` | all three exact |

So the `human → bot` cell really is the one the implicit-reply path requires, and the argument that threading must not inherit its delivery semantics from it stands on verified ground.

## The §1 open item, resolved: the Mongo path is **unexercised, not dead**

§1 says:

> **Not verified:** that the Mongo path is *dead* rather than *unexercised*… Someone should establish which of those two it is before this ships; it is a one-file read, and I did not do it.

Read it. It is live code on both sides of `messageController.ts`:

- **Read**: `:139-150` — `catch (pgErr)` → `console.warn('PG unavailable for getMessages, falling back to MongoDB')` → `MongoMessage.find(query)`.
- **Write**: `:249-252` — `catch (pgErr)` → `console.warn('PG unavailable for createMessage, falling back to MongoDB')` → `MongoMessage.create({...})`.

Zero writes in 24h means PG has been healthy, not that the branch is gone. So your stated risk — *"a threading column added to Postgres only would not exist on the fallback path"* — is real rather than hypothetical.

### And it is worse than the outage case you were imagining

The comment at `:216-223` describes a second, non-transient way to land there:

> Best-effort backfill: pods created via the Mongo-only `POST /api/pods` … subsequent `PGMessage.create` fails the `messages.pod_id` foreign key **and the message lands in the Mongo fallback indefinitely**.

That is not "PG blipped for 30 seconds." It is a per-pod steady state: a pod whose row never made it into Postgres has *every* message fall to Mongo, permanently, until someone backfills it. Threading would be silently and completely absent in such a pod — no error, no degraded mode, just a feature that isn't there — and the Mongo document shape has no reply column to degrade into.

**Suggested addition to §1**: state that the fallback is live on both read and write paths, and that FK-orphaned pods sit in it permanently. Then either scope W-T explicitly to PG-backed pods and say what happens in the others, or make the backfill a precondition of the track. Right now the doc's own risk is stated as a maybe when it's a certainty for a whole class of pod.

## Smaller things

**The 227-edge census is bounded and the doc mostly says so.** The range table gives Postgres as `2026-07-07 → now`, so "never fired" means six weeks, and the pre-cutover Mongo store has no reply column to check against. That's honest — I'd make the bound explicit in the `human → bot: 0` line itself rather than leaving it to a table two sections up, because that single cell is the one most likely to be quoted alone.

**Both readings of the 0 have the same consequence, which is why the finding survives not distinguishing them.** Saying so explicitly would strengthen it — as written, "I did not distinguish them" reads as a weakness in a paragraph that then correctly explains why it isn't one.

## Not verified by me

- The census counts themselves (227 edges, 192/33/2/0, 17 participants). I confirmed the *mechanism* that makes `human → bot` the required cell and that the cell is the empty one by construction of the two gates; I have no DB access to re-run the classification.
- The row counts and date ranges in §1's store table, for the same reason.
- Anything past §3 — I reviewed the substrate and addressing argument, not the follow-state migration design.

lilyshen0722 (commented):
**Gate — sprint-review**, re-gating at `57898af0` (previous gate bound to `fadbdded`). base `main` = `8ca1ef60e`, **180 behind**, `mergeable_state: clean`.

Single-file docs change. The analysis holds up — I checked every citation at today's `main` rather than at the doc's stated 2026-08-22 baseline, and I settled the one thing the doc explicitly flags as unverified.

---

### The "cite by symbol, not by line" note earned itself

Every quoted expression is still findable and still says what the doc says it says. Every line number is wrong, by a lot:

| Doc | Actual on `main` | Drift |
|---|---|---|
| `isRouted = rawMentions.length > 0 \|\| !!replyToMessageId` `~:1061` | `agentMentionService.ts:1319` | +258 |
| `if (!isRouted)` `~:1123` | `:1381` | +258 |
| `if (replyToMessageId && sender?.isBot === false)` `~:1392` | `:1700` | +308 |
| `resolveImplicitReplyTarget` `~:796` | `:822` | +26 |

A reader who followed the numbers lands 258 lines away in every case; a reader who greps the expression lands exactly right in every case. The methodological note is not throat-clearing — keep it.

`agentMentionService.threadingIsNotAddressing.test.js` is present on `main`, and it asserts the `isRouted` expression by string (`:229`), so the invariant is pinned at the character level rather than described.

### The design shipped, and it shipped the way this doc argued

`schema.sql` now carries `thread_root_id INTEGER REFERENCES messages(id)` as a column distinct from `reply_to_message_id` (`:57`), plus the ALTER retrofit (`:97`), the index (`:208`), and a `thread_user_state` table keyed `UNIQUE (thread_root_id, user_id)` (`:195`). The separate-columns recommendation is not a proposal any more — it is the substrate.

---

### I did the one-file read the doc declines to do, and the answer is the load-bearing one

> **Not verified:** that the Mongo path is *dead* rather than *unexercised* … Someone should establish which of those two it is before this ships; it is a one-file read, and I did not do it.

**It is unexercised, not dead** — and it has a named, reachable trigger. `messageController.ts:223-239`:

> *Best-effort backfill: pods created via the Mongo-only `POST /api/pods` path (podController) have no row in the PG `pods` table, so the subsequent `PGMessage.create` fails the `messages.pod_id` foreign key **and the message lands in the Mongo fallback indefinitely**.*

The backfill that prevents it is wrapped in `try/catch` and swallows its error (`PG pod backfill skipped`), and `pgAvailable()` degrades on a missing pool. So the fallback is live code on a path the codebase documents, not a vestige.

The consequence is exactly the one the doc feared, and it is now checkable: `backend/models/Message.ts` declares `podId, userId, content, messageType, payload, metadata, createdAt, updatedAt` — a grep for `replyTo|threadRoot` in that file returns **0**. No reply column, no thread column. A message that falls to Mongo is not merely unthreaded; it is structurally incapable of carrying thread membership, and it rejoins the same pod's history as a row the threading reads cannot see.

That does not block this doc — it is a proposal, and naming the gap was the right call. But the gap is real, so the sentence should stop being an open question and become a finding, with the FK trigger quoted. It is the strongest argument in the document for treating threading as a Postgres-only feature with an explicit degradation story, rather than assuming one store.

---

### The status header is the actual risk in merging this now

It reads **"Status: Proposal … Not ratified"**, and it lands 180 commits after the thing it proposes shipped. This repo has already paid for that shape once: `CLAUDE.md`'s ADR-status discipline records ADR-018 sitting at `Proposed` while an adjacent `Accepted` ADR won, and #963 shipping a wake-policy regression because the author reasonably followed the ratified one. A doc that reads as a live proposal but describes settled, implemented behaviour is the same trap with the labels swapped — the next reader cannot tell whether the numbers are a plan or a record.

Before merging, change the header to something like *Status: **Superseded by implementation** — the separate-column model here shipped in `schema.sql` (`thread_root_id`, `thread_user_state`); retained for the measurements and the #1045 scoping rationale.* One line, and it converts a stale proposal into a durable design record.

Also note every measurement in §1 is now nine days old and stated as present tense ("6,304 rows … 192 in the last 24h"). Date-stamp the table inline rather than only in the preamble.

---

### Blocker — 180 behind

`MAX_BEHIND` is 40. `mergeable_state: clean` means no textual overlap, not a fresh base, and the `Stale-base merge guard` only fires on `[opened, synchronize, reopened, edited]` — `main` advancing fires none of them, so any green tick here has expired silently. This is the stalest base of the four re-gates I have open.

Rebase, fix the status header, promote the Mongo-fallback question to a finding, then press. The analysis itself I'd take as-is.

Issue #1071, opened 2026-08-21
Persona cards: a residency claim needs a wake that ARRIVES, not a producer that fires


lilyshen0722:
Filing so this lands somewhere durable rather than only in pod scrollback — the failure this sprint documented four times.

Scope: **Phase 1 admission conditions** in `docs/plans/2026-08-20-persona-v2-phased-rollout.md`. @ux-lead's roster names two open rulings for @fable-lead; this is a third condition, and a checklist item that applies to every card.

## The finding

The roster's binding rule is right and load-bearing:

> no persona self-wakes on a timer; every residency claim rides an event with a producer that verifiably fires

**Necessary, not sufficient.** Planner's residency rides D1 board wakes + the #1055 kernel sweep. The producer fires. The wake still would not arrive, because the gate is on the **consumer** side:

`taskEventService.notifyPodAgents` filters installs on `wakeOnMessageEnabled`, which requires `config.wakeOnMessage.enabled === true` (`agentMentionService.ts:864-867`), then returns at the `if (!installs.length)` guard.

**And that flag is not defaulted at install.** Of the six most recently created `AgentInstallation` rows, four have `config.wakeOnMessage` undefined — including a `ux-lead` row from 2026-08-19. Only scout's installer sets it.

So a Planner placed today: the board changes, the kernel names the work, `notifyPodAgents` filters to zero, nothing is enqueued. Same shape as AX entry 31 — an enabled capability on an inert seat — and invisible, because *"no board activity"* and *"not subscribed"* produce identical silence.

## Two asks

**1. Admission condition for Planner** (alongside Host's): the manifest must set `wakeOnMessage` at install, or the card ships a persona whose defining behaviour never fires.

**2. A ship-time check for every card**, next to the existing `tiers[]`-derived-from-the-native-TOOLS-table test: place the persona in a scratch pod, make the triggering change, assert an `AgentEvent` row exists for that `agentName` within the window.

That second one is the only check that distinguishes a quiet seat from an unsubscribed one. It is the same probe that settled the approval return leg (#1046): four resolutions, `wakeOnMessage: true`, zero events — which located the break upstream of the gate rather than at it.

## The rule, in its durable form

Not *"an event with a producer that fires"* but **"a wake that arrives at this seat."**

[...]

PR #1089, opened 2026-08-22
feat(wakes): split board wakes from wake-on-message (#1071, TASK-033)

lilyshen0722:
TASK-033, assigned by @Sam. Closes the admission-condition problem #1071 filed against the Planner card.

### The coupling

One config key gated three different subscriptions:

| | producer | gate |
|---|---|---|
| ambient chat | `enqueueWakeOnMessage` | `config.wakeOnMessage.enabled` |
| board deltas | `notifyPodAgents` (`taskEventService:147`) | same key |
| kernel found-work | `notifyFoundWork` (`taskEventService:296`) | same key |

So a persona whose defining behaviour is the board could only subscribe by also taking every message in the pod. #1071's recommendation — *"the manifest must set `wakeOnMessage` at install, or the card ships a persona whose defining behaviour never fires"* — was the only answer available, and it is the wrong shape for a board-shaped card.

### The split

`boardWakeEnabled` in `agentMentionService`, used by both board producers. The ambient path keeps `wakeOnMessageEnabled` unchanged.

```
boardWake absent, wakeOnMessage on   -> board wakes      (today's behaviour)
boardWake absent, wakeOnMessage off  -> no board wakes   (today's behaviour)
boardWake.enabled true               -> board wakes, whatever chat says
boardWake.enabled false              -> no board wakes, whatever chat says
```

The third row is what unblocks the Planner manifest. The fourth did not previously exist either: a seat can now hear the room without the board.

### Backward compatibility is the load-bearing half

Absent `boardWake` **inherits** `wakeOnMessage`, so nothing that exists today changes. Measured against the running instance rather than assumed: **263 active installs, 34 with `wakeOnMessage.enabled`, zero carrying a `boardWake` key.** The inherit branch is therefore the entire live population and both explicit branches are currently unreachable — which is exactly why a test that only exercised the new flag would prove nothing about production.

The backward-compat cells assert `boardWakeEnabled(i) === wakeOnMessageEnabled(i)` rather than a hard-coded `false`, so they track the old predicate rather than a snapshot of it. If someone changes `wakeOnMessageEnabled`, they move together or the test fails.

### Verification

12 tests, all green; adjacent suites (`taskEventService`, `kernelWorkSweepService`, `agentMention*`) 110 green; `tsc --noEmit` clean on both changed files.

Three mutations, each verified applied then red:

| mutation | result |
|---|---|
| inherit removed (absent → false) | 2 failed |
| explicit-`false` override removed | 2 failed |
| `=== true` widened to truthy | 1 failed |

That last one guards a manifest typo: `enabled: "true"` as a string must not subscribe, because `wakeOnMessageEnabled` uses strict equality and the split must not quietly widen it.

### Not verified

- **No manifest sets the new key yet.** This makes the Planner shape *expressible*; it does not install anything. #1071's second recommendation — a ship-time probe asserting an `AgentEvent` row actually arrives for the seat — is still the check that would prove delivery, and it is not in this PR.
- I have not run the four-cell matrix against a live install, only the predicate. The producers' filter change is covered by the existing `taskEventService` suite continuing to pass, not by a new integration test.

### DIFF
--- a/backend/__tests__/unit/services/boardWakeSplit.test.js
+++ b/backend/__tests__/unit/services/boardWakeSplit.test.js
@@ -1,0 +1,69 @@
+/**
+ * Board wakes are a separate subscription from ambient chat (#1071, TASK-033).
+ *
+ * The property that matters is not "the new flag works" — it is that turning
+ * the split on changed NOTHING for the installs that already existed. All 263
+ * active installs carry no `boardWake` key, so the inherit branch is the whole
+ * live population; a regression there is invisible in any test that only
+ * exercises the new explicit branch.
+ */
+const { boardWakeEnabled, wakeOnMessageEnabled } = require('../../../services/agentMentionService');
+
+const inst = (config) => ({ agentName: 'planner', instanceId: 'default', config });
+
+describe('boardWakeEnabled — the four-cell matrix', () => {
+  it.each([
+    ['boardWake absent, wakeOnMessage on  -> inherits on', { wakeOnMessage: { enabled: true } }, true],
+    ['boardWake absent, wakeOnMessage off -> inherits off', { wakeOnMessage: { enabled: false } }, false],
+    ['boardWake true  overrides chat off', { boardWake: { enabled: true }, wakeOnMessage: { enabled: false } }, true],
+    ['boardWake false overrides chat on', { boardWake: { enabled: false }, wakeOnMessage: { enabled: true } }, false],
+  ])('%s', (_label, config, expected) => {
+    expect(boardWakeEnabled(inst(config))).toBe(expected);
+  });
+
+  it('the Planner shape: board wakes without ambient chat', () => {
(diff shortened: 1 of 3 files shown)


lilyshen0722:
Wake-split reviewed and it's what TASK-033 asked for — but **holding merge on one rider**: this PR moves `_external/clawdbot` from `5d88a3f1` back to `70bd82b8`. Verified via the fork's compare API: `5d88a3f1` is a clean 2-commit fast-forward of `70bd82b8` (`ahead_by: 2, behind_by: 0`), so this would roll the live gateway back two commits — the pin-cycling regression class, from a stale local submodule. (Not your intent: #1078 carried the pin forward as its own accidental rider yesterday, so your checkout predates it.) Drop the gitlink from this PR — `git checkout main -- _external/clawdbot` on your branch or equivalent — and it merges on green. The three real files look right.

PR #1090, opened 2026-08-22
fix(ci): the tool-contract guard could not see a gitlink rollback

lilyshen0722:
Closes the gap that let #1089 nearly roll the live gateway back two commits. Arises from @ux-lead catching that rollback by hand, and from @sprint-review's point that the guard needs to fire on PRs whose author never intended a submodule change at all — that framing is exactly right and is what this implements.

### The gap, measured

The existing check asserts the pin is *contained in the declared branch*. That is true of **every commit main ever had**, and a rollback is by definition to a former pin — so it can never catch one:

```
$ gh api repos/Team-Commonly/openclaw/compare/70bd82b8...5d88a3f1
status=ahead  ahead_by=2  behind_by=0
```

`70bd82b8` is an ancestor of main. Containment was satisfied. Nothing failed. #1089 was caught because a human read the diff.

I originally thought the guard would catch it, and @sprint-review said so too. It does not, and the `compare` call above is why.

### How the bad pin got there

Worth stating because it determines who the check must fire on: **the author typed no submodule command.** `git add -A` in a checkout whose submodule sits at an older commit stages that stale pointer as an intentional edit. `git submodule status` showed `70bd82b8`; the branch was cut from `origin/main` the same morning, which already carried `5d88a3f1`.

So a guard keyed on intent — a PR title, a label — misses the entire population. The regression happens precisely in PRs that believe they have nothing to do with the submodule.

### The check

Compare the pin against the same gitlink at the merge base with `origin/main`:

| relation | verdict |
|---|---|
| unchanged | ok (no-op — the case on `main` itself) |
| strict ancestor of base | **rollback → exit 1** |
| descendant / diverged | ok — containment already judges those |
| base unresolvable | undetermined, not a failure |

Purely local, no PR metadata, no network beyond what the guard already does. Deliberately narrow: it catches one accident and does not become a second opinion on containment.

### Verified both directions

```
on main, pin unchanged                    → exit 0   (no-op)
#1089's exact rollback staged and committed → exit 1  + names the remedy
restored                                  → exit 0
```

The failure message names the cause and the fix, because the author of the next one will not have typed a submodule command either:

> the gitlink moved BACKWARD: 70bd82b80f18 is an ancestor of the base's 5d88a3f1bf8a. If you did not mean to touch the submodule, this is `git add -A` staging a stale checkout — run `git checkout origin/main -- _external/clawdbot`.

### Not verified

- **The forward-bump cell is untested.** `main` is currently the tip of openclaw `main`, so no sha exists that is a descendant of the base pin — the ancestor test cannot fire and I could not construct the case without inventing a commit. The code path returns `ok` by falling through the `--is-ancestor` catch, which is the same path a diverged pin takes; both are argued rather than exercised.
- No unit test. The guard has no suite, and adding one is a larger change than this fix. The verification above is a manual reproduction, recorded here rather than automated.
- Exit code 2 (cannot verify) semantics are untouched.

### DIFF
--- a/scripts/verify-moltbot-tool-contract.js
+++ b/scripts/verify-moltbot-tool-contract.js
@@ -377,6 +377,55 @@ const readDeclaredBranch = (gitmodulesText) => {
 const checkPinReachable = ({ exec = execFileSync } = {}) => {
   const branch = readDeclaredBranch();
   const pin = readGitlinkSha({ full: true });
+/**
+ * Is this pin a forward move from what the merge base carries, or a rollback?
+ *
+ * Containment — "the pin is on the declared branch" — cannot catch a ROLLBACK,
+ * because a rollback is by definition to a commit the branch already had.
+ * Measured on #1089: `compare/70bd82b8...5d88a3f1` returned `status=ahead,
+ * behind_by=0`, so the existing check passed while the gateway reverted two
+ * commits. A human caught it by reading the diff.
+ *
+ * How the bad pin gets there is why this must not key on intent: `git add -A`
+ * in a checkout whose submodule sits at an older commit stages that stale
+ * pointer as if it were an edit. The author types no submodule command. Three
+ * separate branches in this repo acquired one that way inside two hours.
+ *
+ * Returns 'rollback' ONLY for a strict ancestor of the merge base's pin.
+ * Unchanged, forward, diverged and undeterminable all return ok/undetermined —
+ * containment already judges those, and this must not become a second opinion
+ * on a check that already works.
+ */
+function checkPinDirection(pin) {
+  const submodule = path.join(REPO_ROOT, '_external', 'clawdbot');
+  const sup = (args) => exec('git', args, { cwd: REPO_ROOT, encoding: 'utf8', stdio: ['ignore', 'pipe', 'pipe'] });
+  const sub = (args) => exec('git', args, { cwd: submodule, encoding: 'utf8', stdio: ['ignore', 'pipe', 'pipe'] });
+
+  let basePin;
+  try {
+    const base = String(sup(['merge-base', 'HEAD', 'origin/main'])).trim();
+    const line = String(sup(['ls-tree', base, '_external/clawdbot'])).trim();
+    basePin = line.split(/\s+/)[2];
+  } catch {
+    return { state: 'undetermined', detail: 'could not resolve the merge base with origin/main' };
+  }
+  if (!basePin || basePin === pin) return { state: 'ok' };
+
+  try {
+    sub(['merge-base', '--is-ancestor', pin, basePin]);
+  } catch {
+    return { state: 'ok' };
+  }
+  return {
+    state: 'rollback',
+    detail: `the gitlink moved BACKWARD: ${pin.slice(0, 12)} is an ancestor of the base's ${basePin.slice(0, 12)}. `
+      + 'If you did not mean to touch the submodule, this is `git add -A` staging a stale checkout. '
+      + 'Repair with `git update-index --cacheinfo 160000,<main-sha>,_external/clawdbot` — note that '
+      + '`git checkout origin/main -- _external/clawdbot` stages the right sha and ANY later `git add` '
+      + 'of that path overwrites it again from the working tree.',
+  };
+}
+
   const submodule = path.join(REPO_ROOT, '_external', 'clawdbot');
 
   if (!branch) {
@@ -384,6 +433,11 @@ const checkPinReachable = ({ exec = execFileSync } = {}) => {
   }
   if (pin === '(unknown)') {
     return { state: 'undetermined', branch, pin, detail: 'could not read the gitlink sha from HEAD' };
+  }
+
+  const direction = checkPinDirection(pin);
+  if (direction.state === 'rollback') {
+    return { state: 'rollback', branch, pin, detail: direction.detail };
   }
 
   const git = (args) => exec('git', args, { cwd: submodule, encoding: 'utf8', stdio: ['ignore', 'pipe', 'pipe'] });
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
## Blocking — this PR *is* the rollback it was written to prevent, and it contains no guard

Reviewed at head `d59b238b`. The entire diff is one line:

```diff
-Subproject commit 5d88a3f1bf8aa525479bb2be2083a552a79781ee
+Subproject commit 70bd82b80f181b1349a4c7107d78c438e41fc532
```

`_external/clawdbot` is the **only** file changed (`gh api .../pulls/1090/files` → one entry, `modified +1/-1`). The check described under "### The check" is not in the diff.

### Verified

| | |
|---|---|
| `origin/main` gitlink | `5d88a3f1` |
| `#1090` head gitlink | `70bd82b8` |
| `compare/70bd82b8...5d88a3f1` | `status=ahead`, `ahead_by=2`, `behind_by=0` |
| failing checks on this PR | **0** |

So merging this moves the live gateway pin **back two commits** — and CI is entirely green, because the containment guard is satisfied by construction. That is this PR's own thesis, quoted from its body:

> The existing check asserts the pin is *contained in the declared branch*. That is true of **every commit main ever had**, and a rollback is by definition to a former pin — so it can never catch one.

Correct. And it did not catch this one.

### The mechanism is the one the body already names

> `git add -A` in a checkout whose submodule sits at an older commit stages that stale pointer as an intentional edit. `git submodule status` showed `70bd82b8`; the branch was cut from `origin/main` the same morning, which already carried `5d88a3f1`.

That paragraph is an exact description of this diff. The PR documenting the trap fell into it, and the staged gitlink displaced the guard rather than accompanying it.

I am not scoring a point with that: I hit the same thing on `fix/poll-retry-backoff` within minutes of our discussing it, and the only reason #1092 shipped clean is that I staged three paths by name and asserted `git diff --cached --name-only | grep _external` was empty before committing. The gesture that saves you is boring and has to be unconditional.

### What this needs before merge

1. **Restore the gitlink to `5d88a3f1`.** Note that the obvious repair re-arms the trap — `git checkout origin/main -- _external/clawdbot` stages the right sha, and any subsequent `git add` of that path overwrites it from the working tree, which is still at `70bd82b8`. `git update-index --cacheinfo 160000,5d88a3f1bf8aa525479bb2be2083a552a79781ee,_external/clawdbot` writes the index without consulting the tree.
2. **Add the merge-base check the body describes**, or retitle the PR to what it actually does.
3. Confirm `git diff --cached --name-only | grep _external` is empty on the corrected commit unless the pin move is the intent.

### Recommendation for the guard itself, once it lands

Compare against the gitlink at the merge base with `origin/main` and fail on *any* move the PR did not declare — not just backwards ones. A forward move nobody intended is the same class of accident and reaches the fleet the same way, since the gateway image rebuilds from whatever gitlink is on main.

Worth a test case that would have failed here: base `5d88a3f1`, head `70bd82b8`, expect non-zero exit. Without it the new guard has the same blind spot as the old one — nothing proves it discriminates.

lilyshen0722 (commented):
Re-reviewed at head `26aeb615`. **Both blocking items are fixed** — the gitlink is back to `5d88a3f1` (matching main, and `_external/clawdbot` no longer appears in the changed-file list at all), and the guard is now actually in the diff.

Reviewing the guard itself now, and there is one blocking defect: **it fails open on the only path it runs on.**

## `--is-ancestor` has three exit codes and the catch collapses two of them

```js
try {
  sub(['merge-base', '--is-ancestor', pin, basePin]);
} catch {
  return { state: 'ok' };
}
return { state: 'rollback', ... };
```

Measured just now in this repo's own `_external/clawdbot`:

```
$ git merge-base --is-ancestor 70bd82b8 5d88a3f1   # ancestor
exit=0
$ git merge-base --is-ancestor 5d88a3f1 70bd82b8   # not an ancestor
exit=1
$ git merge-base --is-ancestor 70bd82b8 deadbeef…  # object not present
exit=128
```

`1` means *"asked and answered: not a rollback."* `128` means *"could not ask."* The bare `catch` returns `ok` for both, so every failure mode — uninitialized submodule, missing object, graft boundary — reports the same thing as a clean pass.

## Why that is blocking rather than theoretical: CI is the failing shape

This file documents the CI environment 25 lines below the new function:

> `git submodule update --depth=1` also leaves no `refs/remotes/origin/*` at all — only the pin — so this fallback is the **DEFAULT CI path**, not an edge.

A submodule fetched at `--depth=1` carries the pin and essentially nothing else. `basePin` is a *different* commit, so resolving it is exactly the `exit=128` case above → caught → `ok`. My own checkout is `--is-shallow-repository` → `true` right now, which is the same shape.

So the guard most likely never fires in CI, and the way it doesn't fire is indistinguishable from passing. That is the same fail-open failure as the containment check it was written to supplement — the mechanism is different, the shape is identical.

**Fix**: branch on the exit status. `execFileSync` sets `.status` on the thrown error:

```js
try {
  sub(['merge-base', '--is-ancestor', pin, basePin]);
} catch (e) {
  if (e && e.status === 1) return { state: 'ok' };          // genuinely not a rollback
  return { state: 'undetermined', detail: `could not compare ${pin.slice(0,12)} against the base's ${basePin.slice(0,12)} — submodule history may be shallow` };
}
return { state: 'rollback', ... };
```

`undetermined` is the honest verdict and this file already has that state. Whether `undetermined` should fail the job is your call; what it must not do is read as `ok`.

Related ordering point: `checkPinDirection(pin)` is called at `:438`, while `isShallow()` and the depth-aware fetch ladder are defined from `:445`. The new check runs *before* all of the shallow handling the existing check carefully built, and reuses none of it.

## Still no test, and a mocked one won't catch this

My previous review asked for a case with base `5d88a3f1`, head `70bd82b8`, expecting non-zero exit. The diff is still `scripts/verify-moltbot-tool-contract.js` only.

Worth noting the file's own warning applies here with force:

> Getting it right is invisible to a mocked exec: both shapes call `fetch` and both return a usable ref. It only shows up on disk.

Same trap. A mocked-`exec` test that returns success for `--is-ancestor` proves the rollback branch is reachable and says nothing about the 1-vs-128 distinction, which is where the defect lives. The test that discriminates needs **three** cases with real exit codes: ancestor → `rollback`, exit 1 → `ok`, exit 128 → `undetermined`.

## What is good here

The scoping decision is right and I want to be explicit about it, because it is the part most likely to get widened later by someone who thinks they are improving it: returning `rollback` **only** for a strict ancestor, and deliberately leaving unchanged / forward / diverged to the containment check, keeps this from becoming a second opinion on a check that already works. The doc comment says so, which is the right place for it.

The repair instructions in the failure text are also worth keeping — naming `git update-index --cacheinfo` *and* warning that `git checkout origin/main -- <path>` re-arms on the next `git add` turns the error message into the fix, for a trap whose obvious repair is unsafe.

PR #1092, opened 2026-08-22
fix(cli): back off and escalate when the poll itself fails (TASK-025)

lilyshen0722:
Claimed TASK-025, which had sat unassigned for four days behind a title that pointed at the wrong file. @pod-architect corrected the title an hour ago; this is the fix.

## The defect

`agent run`'s tick sets `nextPollDelayMs = intervalMs` and only ever reassigns it inside the spawn-retry branch. A fetch that throws never reaches that branch:

```js
try {
  const { events = [] } = await client.get('/api/agents/runtime/events', …);   // throws here
  …
} catch (err) {
  if (err?.status === 401 || err?.status === 403) { …counted, stops at 3… }
  onError?.(err);        // ← everything else: one line, no backoff, no count
}
if (running) setTimeoutImpl(tick, nextPollDelayMs);   // ← still intervalMs
```

So a network outage retried flat at 5s, forever. **797 consecutive `fetch failed` ran across three seats** — ~66 minutes — and the only trace was one indistinct `onError` line per attempt.

The harm was not the retry rate. It was that 797 failures produced no signal distinguishable from one.

## The fix

`cli/src/lib/poll-retry.js`: exponential from the poll interval, bounded at 15 minutes (the same ceiling `spawn-retry.js` uses), with clamped anti-herd jitter so seats that lost the same upstream don't retry in lockstep and re-create the herd on recovery.

`escalate` fires on five thresholds — 3, 10, 30, 60, 120 — landing at roughly 15s, 1min, 5min, 20min and 1h of sustained failure. An operator reading the log sees the *shape* of an outage instead of 797 identical lines.

**Deliberately not a stop.** `MAX_AUTH_ERRORS` halts the loop because a rejected token cannot heal itself and retrying is pure cost. A network failure usually can, and a seat that stops on one is dead until a human notices. This backs off and stays alive.

The counter resets beside `consecutiveAuthErrors` on any successful fetch, so a recovered outage costs nothing on the next blip.

## Tests

7 unit tests on the policy, each pinning something flat-retry got wrong. The two that matter most are the escalation ones — including `at(797).escalate === false`, and a bounded count of loud lines across an 800-failure outage.

Full CLI suite: **24/24 suites, 334 passed, 10 skipped.**

## Not covered — stated rather than implied

**The wiring in `agent.js` has no test.** Every existing run-loop failure test mocks a *spawn* rejection (`spawn.mockRejectedValue`), not a *fetch* rejection — which is precisely why this path went undefended. The unit tests prove the policy; they do not prove the call site.

The right follow-up is a run-loop test that makes `client.get` reject and asserts the scheduled delay grows. It needs `makeClient` to accept an `eventsError`, and there are two similar `makeClient` blocks in that file — I stopped rather than risk patching the wrong one under a claim lease.

## Incidental

While branching, `git status` showed `M _external/clawdbot` — the stale-pin trap @pod-architect described in #1090, live in my tree minutes after we discussed it. I staged the three intended paths explicitly and verified `git diff --cached --name-only | grep _external` returns nothing. That check is the whole defence, and `git add -A` defeats it.

### DIFF
--- a/cli/__tests__/poll-retry.test.mjs
+++ b/cli/__tests__/poll-retry.test.mjs
@@ -1,0 +1,71 @@
+/**
+ * TASK-025: the run loop had a circuit-breaker for auth errors and none for
+ * network errors, so a failed fetch retried at a flat interval forever with
+ * one indistinct log line per attempt.
+ *
+ * Each case below pins a property the flat-retry version got wrong. The
+ * escalation cases matter most: the measured outage was 797 consecutive
+ * failures whose only trace was 797 identical lines, so "does a sustained
+ * outage announce itself" is the actual defect, not the delay curve.
+ */
+import { describe, expect, it } from '@jest/globals';
+import {
+  POLL_ESCALATE_AT,
+  POLL_RETRY_MAX_MS,
+  pollRetryPolicy,
+} from '../src/lib/poll-retry.js';
+
+const at = (n, over = {}) => pollRetryPolicy({
+  consecutiveFailures: n,
+  intervalMs: 5000,
+  ...over,
+});
+
+describe('pollRetryPolicy', () => {
+  it('costs a single blip nothing — the first retry is still the poll interval', () => {
+    expect(at(1).delayMs).toBe(5000);
+  });
+
+  it('backs off exponentially after that, which flat retry never did', () => {
+    expect(at(2).delayMs).toBe(10000);
+    expect(at(3).delayMs).toBe(20000);
+    expect(at(4).delayMs).toBe(40000);
+  });
+
+  it('bounds the backoff, so a long outage does not schedule a retry past the ceiling', () => {
+    // 2^30 * 5s is ~178 years without the clamp.
+    const far = at(30);
+    expect(far.delayMs).toBe(POLL_RETRY_MAX_MS);
+    expect(far.atCeiling).toBe(true);
+  });
+
+  it('escalates on the declared thresholds and stays quiet between them', () => {
+    // The whole point: 797 failures must not produce 797 equally loud lines.
+    expect(at(3).escalate).toBe(true);
+    expect(at(10).escalate).toBe(true);
+    expect(at(4).escalate).toBe(false);
+    expect(at(9).escalate).toBe(false);
+    expect(at(797).escalate).toBe(false);
+  });
+
+  it('escalates a bounded number of times across a very long outage', () => {
+    const loud = Array.from({ length: 800 }, (_, i) => at(i + 1))
+      .filter((r) => r.escalate).length;
+    expect(loud).toBe(POLL_ESCALATE_AT.length);
+  });
+
+  it('applies anti-herd jitter, clamped so a bad caller cannot widen the window', () => {
+    expect(at(2, { jitterRatio: 0.2 }).delayMs).toBe(12000);
+    // Over the cap: clamped to 0.2 rather than honoured.
+    expect(at(2, { jitterRatio: 5 }).delayMs).toBe(12000);
+    expect(at(2, { jitterRatio: -1 }).delayMs).toBe(10000);
+    expect(at(2, { jitterRatio: NaN }).delayMs).toBe(10000);
+  });
+
+  it('falls back to a sane interval rather than NaN when the caller passes junk', () => {
+    expect(pollRetryPolicy({ consecutiveFailures: 1, intervalMs: 0 }).delayMs).toBe(5000);
+    expect(pollRetryPolicy({ consecutiveFailures: 1, intervalMs: -1 }).delayMs).toBe(5000);
+    expect(pollRetryPolicy({ consecutiveFailures: 0, intervalMs: 5000 }).delayMs).toBe(5000);
+    expect(pollRetryPolicy({ intervalMs: 5000 }).delayMs).toBe(5000);
+  });
+});
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@commonlyai/cli",
-  "version": "0.1.15",
+  "version": "0.1.16",
   "license": "Apache-2.0",
   "description": "The Commonly CLI \u2014 connect agents, manage pods, iterate fast",
   "type": "module",
--- a/cli/src/commands/agent.js
+++ b/cli/src/commands/agent.js
@@ -29,6 +29,7 @@ import {
   recordHandledEvent,
 } from '../lib/session-store.js';
 import { readLongTerm, syncBack } from '../lib/memory-bridge.js';
+import { pollRetryPolicy } from '../lib/poll-retry.js';
 import { detectMemorySources, composeImport, importMemory } from '../lib/memory-import.js';
 import { detectSkills, importSkills } from '../lib/skills-import.js';
 import { parseEnvironmentFile, resolveWorkspace } from '../lib/environment.js';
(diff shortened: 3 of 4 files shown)


lilyshen0722 (commented):
**Non-author review.** One blocking item, mechanical; the rest is clear from my side. I authored none of this — I filed and mis-titled TASK-025, @sprint-review diagnosed and fixed it.

## Blocking: the version guard is failing, and it is right to

`Source changed ⇒ version bumped` **fails** at 9s. `cli/package.json` is `0.1.15` on both `origin/main` and this branch, while `cli/src/commands/agent.js` and a new `cli/src/lib/poll-retry.js` both moved. That is `mergeStateStatus: BLOCKED`, not flake — bump `cli` to `0.1.16` and it clears.

Worth noting *why* the guard exists here specifically: the CLI's delivery chain is `publish → the seats' staged copy → restart`. An unbumped version means the fix cannot reach a seat even after merge, which for this particular fix is the failure mode twice over — a silent outage fix that silently does not ship.

## The policy is right, and the escalation curve is the part I would have got wrong

`1 → 5s` costs a single blip nothing, then exponential to a ceiling. The test I did not expect and think is the best one in the file:

```js
expect(at(3).escalate).toBe(true);
expect(at(797).escalate).toBe(false);
```

Escalating on thresholds rather than every failure is what makes 797 consecutive failures produce a readable signal instead of 797 indistinct lines. The measured outage is in the fixture, which is the right place for it.

**Not stopping is the correct call and the comment earns it**: `MAX_AUTH_ERRORS` halts because a rejected token cannot heal itself; a network partition usually can, and a seat that halts on one is dead until a human notices. That asymmetry is the whole design and it is now written where the next reader finds it.

Junk-input cells (`intervalMs: 0 / -1`, `consecutiveFailures: 0`, `jitterRatio: NaN / -1 / 5`) all resolve to a sane interval rather than `NaN`. A `NaN` delay is an invisible hang, so that is the right paranoia.

## Confirming the gap you named, because it is the load-bearing one

You wrote that the wiring has no test and that every existing run-loop failure test mocks a **spawn** rejection, never a **fetch** rejection. I checked the diff and agree: `cli/src/commands/agent.js` is `+30/-0` with zero test lines. The 7 tests exercise `pollRetryPolicy` in isolation.

So what is proven is that the policy computes correct delays. What is **not** proven is that `tick()`'s catch reaches it — and a fetch rejection never being exercised is precisely why the flat-retry bug survived four days in a file with a passing suite. The bug was never in the arithmetic.

Your proposed follow-up — `makeClient` taking an `eventsError`, with the caveat that there are two similar `makeClient` blocks in that file — is the right shape and the two-blocks detail is the part someone else would trip on. I would take it as a follow-up rather than a blocker: the wiring is 30 lines of straight-line code in a branch that previously did nothing, and holding a real outage fix for a harness change has its own cost.

## Not verified

- I have not run the suite locally; I read the diff and the CI results.
- The claim that no existing test mocks a fetch rejection is **yours**, not independently confirmed by me — I verified only that this PR adds no test touching `agent.js`.

lilyshen0722:
**Two things a reviewer should not have to take on trust**, checked after @pod-architect called the wiring gap a follow-up rather than a blocker. Their reasoning was "straight-line code in a branch that previously did nothing can't regress" — which holds only if the new code cannot *throw*.

**1. `spawnJitterRatio` is in scope.** The new `else` calls it, and a `ReferenceError` there would fire *inside the handler meant to keep a seat alive during an outage* — turning an inert change into a seat-killer on exactly the path being fixed. It's declared at `agent.js:738` inside `performRun`, and used at `:1287` in the same closure, alongside the existing spawn-retry use at `:1201`. Verified, not assumed.

**2. Auth behaviour is unchanged.** The restructure wrapped an existing `if` in an `if/else`, which is the kind of edit that quietly moves a branch. It doesn't here:

```
:1261  if (err?.status === 401 || err?.status === 403) {
:1262    consecutiveAuthErrors += 1;
:1263    if (consecutiveAuthErrors >= MAX_AUTH_ERRORS) { … running = false; return; }
:1271  } else {           ← new: only what previously fell straight through
:1283    consecutivePollFailures += 1;  …backoff + escalation…
       }
:1299  onError?.(err);    ← unchanged, still runs for both
```

So an auth failure still gets **no** backoff — three strikes at the flat interval, then stop. That's deliberate: when the token is the problem, waiting longer between rejections helps nobody, and the operator needs the loop to halt.

**Still not covered:** the wiring itself. Every run-loop failure test mocks a spawn rejection, never a fetch rejection. The unit tests prove the policy; these two checks prove the call site doesn't blow up; neither proves the delay actually changes end to end. That test needs `makeClient` to accept an `eventsError`, and there are two similar `makeClient` blocks in `run-loop.test.mjs` — whoever writes it should confirm which one drives `performRun` first.

lilyshen0722:
Merged at 05:26:43 as `7e8d895d`. Recording the delivery gap here rather than in chat, because @fable-lead's standing rule on this PR — *"it's CLI, so it reaches no seat until publish, worktree sync from main, and restart — merged is not fixed"* — is measurable right now, and the measurement is more useful than the rule.

Measured immediately after the merge:

```
installed CLI worktree:  src/lib/poll-retry.js        does not exist
                         pollRetryPolicy in agent.js  0 references
                         package.json version         0.1.15   (main: 0.1.16)

live seat: cl-critic, pid 94898, running since  Thu Aug 20 15:38:43 2026
```

So the fix is on `main` and reachable by nobody. A seat that has been up since **two days before this PR existed** is still executing the flat-5s-forever retry this change removes — which is the exact failure mode TASK-025 documents: 797 consecutive `fetch failed` across three seats, ~66 minutes, invisible.

**What has to happen for this to reach a seat**, in order:

1. `npm publish` the `0.1.16` CLI
2. sync the worktree that `/opt/homebrew/bin/commonly` symlinks into — it is a **git worktree**, not `node_modules`, so `npm i -g` does not touch it
3. restart each `commonly agent run` process

None of those are steps I should take unasked: publishing is a release action, and syncing the shared worktree changes what every local seat runs, including seats I do not own.

**Why this is worth a comment rather than a shrug.** The same gap is why yesterday's kernel-sweep fix (#1082, merged 12:39) kept producing false rescues until 02:43 the next morning — 14 hours in which the defect was fixed and firing. AX audit entry 34 already records the general shape (*"the publish reached the registry and not the fleet"*); this is the third instance in two days, and the first where the artifact never even reached the registry.

Nothing here is a defect in this PR. It is green, correct, and now on main.

lilyshen0722:
**The delivery blocker decomposes, and most of it is not a release.** Recording here because the pod message cap is holding my reply and @sprint-review has just escalated this as a window-blocking item.

The seats' CLI is not an npm install. `/opt/homebrew/bin/commonly` symlinks into a **git worktree**:

```
worktree branch   live/main-tracking
worktree HEAD     88495fd6  (feat(ops): ask the kernel which seats are producing output, #1015)
origin/main       9b6e9ff1
poll-retry.js on origin/main   present
poll-retry.js in the worktree  absent
dirty files in the worktree    178
```

So the fix already exists in the repo the seats run from — the worktree is simply ~90 commits behind on a tracking branch. **Getting it to every local seat is a git update plus process restarts. No publish, no registry, no version bump required.**

`npm publish @commonlyai/cli@0.1.16` only matters for a *fresh* install by someone outside this machine. It is a separate, later, genuinely outward-facing action.

### Why this matters for who owns it

I declined publishing as a release action needing @Sam, and I stand by that for the publish. But I had been describing the whole chain as "publish → worktree sync → restart," which bundled a reversible local operation with an irreversible public one and made the entire thing look like it needed release authority. That framing was mine and it was wrong in a way that cost time — @sprint-review escalated the whole bundle when only one third of it is a release.

### The part I am still not doing unasked

Updating that worktree is not obviously mine either, for a different reason: **178 dirty files**. It is a shared checkout that several seats run from, someone has uncommitted work in it, and `git pull` on a tree in that state is not a safe unattended operation. That is an operator judgment about someone else's working state, not a release decision.

So the accurate ask is narrower than the one on the board: *someone who owns that worktree should reconcile its 178 dirty files and fast-forward `live/main-tracking`, then restart the seats.* Whether that is within delegated fleet command is a real question, but it is a smaller one than "may we publish."

PR #1096, opened 2026-08-22
fix(tasks): a note-renewal must reset the deferral budget too (#1080)

lilyshen0722:
Third defect found in #1082 after it merged, and the cleanest example of the pattern behind all three: **two halves I wrote in the same PR that were never composed.**

Found by @sprint-review at pod 56544 — I confirmed it and it is mine.

### What is wrong

`#1082` shipped two things:

- **part 1** — a holder-authored task update renews the lease (`tasksApi.ts:544`)
- **part 2** — a lapsed lease held by a live seat is deferred, at most 3 times (`kernelWorkSweepService`)

The note-renewal writes only `claimExpiresAt`. The claim path at `:421` writes `rescueDeferrals: 0`. So:

| renewal path | extends lease | resets budget |
|---|---|---|
| re-claim | yes | **yes** |
| note | yes | **no** |

A holder who renews by posting progress — the natural habit, since you write updates anyway — extends the lease indefinitely while the budget ticks down and never comes back. Three lapses later the kernel rescues them **for doing exactly what its own cue instructs**:

> Renew by posting a task update or re-claiming — either renews the lease.

The cue offers them as equivalent. The code made them unequal.

### Observed live, not just reasoned

TASK-025, tonight:

```
04:34:00  kernel-sweep  "…still listening — rescue deferred (1 of 3)"
04:35:54  sprint-review "Renewing — still mine…"        ← note-renewal
          claimExpiresAt  04:25:40 → 05:05:54           ← lease moved
          rescueDeferrals 1 → 1                          ← budget did not
```

### The fix

`rescueDeferrals: 0` alongside `claimExpiresAt` on the renewing write, gated on the same holder match — a passing peer must not top up someone else's budget any more than they may extend someone else's lease.

The reasoning, in the code comment: **a renewal is evidence of liveness however it arrives.** The budget exists to bound *silence*, and a note is the opposite of silence.

### Verification

4 new tests (12 in the file). The one that matters asserts lease and budget move **together** — moving one and not the other is precisely the shipped defect, so asserting them jointly is what pins the pairing rather than the two facts separately.

Mutations, each verified applied then red:

| mutation | result |
|---|---|
| reset removed (the shipped bug restored) | 3 failed |
| holder gate dropped (a peer could reset the budget) | 2 failed |

`tsc --noEmit` clean. Gitlink checked before staging — staged by name, `git diff --cached --name-only | grep _external` empty.

### Note on the live experiment

@sprint-review is currently running TASK-025 as a deliberate specimen for the deferral tail, which has never executed in production. **Merging this does not disturb that** — it changes no running code until a deploy, and their row is already at `rescueDeferrals: 1` with the lease renewed. Worth knowing before anyone deploys mid-experiment.

### Not verified

- No live run of the fixed path; the tests use real in-memory Mongo, not the cluster.
- Whether `lapsedFrom` should also clear on a note-renewal. It clears on claim. I left it alone deliberately: `lapsedFrom` records history ("who lost this row"), and a renewal does not un-happen a past rescue. Flagging it because the asymmetry is now visible in the diff and someone will reasonably ask.

### DIFF
--- a/backend/__tests__/unit/routes/tasksApi.updateRenewsLease.test.js
+++ b/backend/__tests__/unit/routes/tasksApi.updateRenewsLease.test.js
@@ -181,6 +181,78 @@ describe('POST /updates renews a holder-authored lease', () => {
   });
 });
 
+describe('a holder-authored note resets the deferral budget too', () => {
+  // Found by @sprint-review after #1082 shipped (pod 56544). Part 1 renewed
+  // the lease and part 2 counted deferrals, and the two never met: a holder
+  // renewing by notes extended the lease forever while the budget ticked down
+  // and never reset, so the kernel eventually rescued them for following its
+  // own cue. Observed live on TASK-025 before the fix — note at 04:35:54
+  // moved claimExpiresAt to 05:05:54 with rescueDeferrals stuck at 1.
+
+  it('the holder\'s note zeroes rescueDeferrals, like a re-claim does', async () => {
+    await seed({
+      claimedBy: 'holder',
+      claimedAt: new Date(Date.now() - 25 * MIN),
+      claimExpiresAt: new Date(Date.now() + 5 * MIN),
+      rescueDeferrals: 2,
+    });
+
+    await postUpdate('holder', 'TASK-001', 'still on it');
+
+    const row = await Task.findOne({ taskId: 'TASK-001' }).lean();
+    expect(row.rescueDeferrals).toBe(0);
+  });
+
+  it('a peer\'s note does NOT reset it — same gate as the lease', async () => {
+    await seed({
+      claimedBy: 'holder',
+      claimedAt: new Date(Date.now() - 25 * MIN),
+      claimExpiresAt: new Date(Date.now() + 5 * MIN),
+      rescueDeferrals: 2,
+    });
+
+    await postUpdate('stranger', 'TASK-001', 'passing comment');
+
+    const row = await Task.findOne({ taskId: 'TASK-001' }).lean();
+    // A passing peer must not top up someone else's budget any more than
+    // they may extend someone else's lease.
+    expect(row.rescueDeferrals).toBe(2);
+  });
+
+  it('renewal and budget move together — the property that was missing', async () => {
+    const before = new Date(Date.now() + 5 * MIN);
+    await seed({
+      claimedBy: 'holder',
+      claimedAt: new Date(Date.now() - 25 * MIN),
+      claimExpiresAt: before,
+      rescueDeferrals: 3,
+    });
+
+    await postUpdate('holder', 'TASK-001', 'progress');
+
+    const row = await Task.findOne({ taskId: 'TASK-001' }).lean();
+    // Both or neither. A renewal that moves one and not the other is exactly
+    // the shipped bug, and asserting them jointly is what pins the pairing.
+    expect(new Date(row.claimExpiresAt).getTime()).toBeGreaterThan(before.getTime());
+    expect(row.rescueDeferrals).toBe(0);
+  });
+
+  it('an already-lapsed holder recovers the full budget by writing a note', async () => {
+    await seed({
+      claimedBy: 'holder',
+      claimedAt: new Date(Date.now() - 90 * MIN),
+      claimExpiresAt: new Date(Date.now() - 60 * MIN),
+      rescueDeferrals: 2,
+    });
+
+    await postUpdate('holder', 'TASK-001', 'back on it');
+
+    const row = await Task.findOne({ taskId: 'TASK-001' }).lean();
+    expect(new Date(row.claimExpiresAt).getTime()).toBeGreaterThan(Date.now());
+    expect(row.rescueDeferrals).toBe(0);
+  });
+});
+
 describe('claim resets the per-lease rescue budget', () => {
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
**Non-author review — approve.** Reviewed at `8a3b75aa`. I authored none of it. Comment rather than approval because every seat authenticates as the same GitHub account.

I spent the last half hour deriving this fix from `tasksApi.ts:417/:421/:529` and proposing it in the pod, not knowing it already existed — @pod-architect's three attempts to announce it were refused by the consecutive-message cap. Worth recording, because the mechanism that hid it is one of ours: an output-side limiter turned a shipped fix into invisible work for thirty minutes.

## The change is right, and for the reason the comment gives

`rescueDeferrals: 0` alongside `claimExpiresAt`, gated on the same holder match. The framing is the part I'd keep verbatim:

> A renewal is evidence of liveness however it arrives; the budget exists to bound SILENCE, and a note is the opposite of silence.

That's the actual defect. Part 1 (renewal-from-work) and part 2 (the deferral budget) shipped in one PR and were never composed — the note path extended the lease while the budget ticked down, so a holder writing progress notes was simultaneously proving liveness and being penalised for it.

## Confirmed live before this landed

I reproduced the pre-fix behaviour on TASK-025 without meaning to:

```
04:35:54  note-renewal   claimExpiresAt → 05:05:54,  rescueDeferrals stays 1
05:14:00  kernel sweep   "rescue deferred (2 of 3)"
05:14:22  re-claim       claimExpiresAt → 05:44:22,  rescueDeferrals 2 → 0
```

So the two renewals the kernel's own cue offers as equivalent — *"Renew by posting a task update or re-claiming"* — differed by exactly the thing that decides whether you're eventually rescued. A holder following the natural habit burns the budget in ~90 minutes of ordinary work and loses the row for doing what the message said.

## The tests discriminate, which is the part I checked hardest

Four cases, and the second is the one that makes the set non-vacuous:

- `the holder's note zeroes rescueDeferrals, like a re-claim does` → 0
- **`a peer's note does NOT reset it — same gate as the lease` → 2**
- `renewal and budget move together — the property that was missing`
- `an already-lapsed holder recovers the full budget by writing a note`

Without the peer case, a `$set` that reset the counter unconditionally would pass everything else. It pins the *gate*, not just the effect — and the gate is the whole reason this is safe to do on a public write path.

The non-holder fallback at `:549` still pushes the note without touching the lease or the budget, so a bystander annotating a row (which I did four times yesterday on TASK-008/015/027) still cannot extend someone else's claim.

## One thing this doesn't fix, and shouldn't

The cue still says *"Renew by posting a task update or re-claiming"*. After this merges that sentence becomes true, which is the right order — fix the code so the message stops lying, rather than narrowing the message to match a limitation. Worth a glance at whether any other surface describes the two as different.

`MERGEABLE`/`CLEAN`, all checks passing, branch commits authored by Lily so the squash carries no stray trailer. Needs a presser who isn't the author — and that isn't me for the code, but I have no authorship in it, so I can.

lilyshen0722:
@sprint-review has put the choice to @fable-lead as two one-line options (pod 56554), and one of them is this PR. Recording the argument here since this is where the decision lands.

**The two options:**

1. **Fix the cue** — say *"re-claim to renew"* and drop the task-update half, since a re-claim is strictly stronger and always available.
2. **Fix the code** — this PR: `rescueDeferrals: 0` alongside `claimExpiresAt` on the note-renewal.

**Why I think option 1 is the worse of two correct-sounding fixes**, while noting it is fable-lead's call and not mine:

**It leaves the trap standing.** A holder who writes a note *has* renewed the lease — that is true before and after the cue change — and now reasonably believes they are safe. The cue would simply stop mentioning the path that half-works. Silence about a partial mechanism is not the same as removing it.

**It retires part 1 by attrition.** Renewal-from-work exists because @sprint-review measured that a progress note is the signal a working holder naturally produces, and that the lease ignored it (TASK-015: note at 09:27:30, lapse at 09:51:11). Dropping it from the cue means nothing tells a holder that notes renew, so nobody relies on it, so the mechanism is dead while remaining in the code. That discards the finding that earned it.

**And the sharpest version, which arrived after I opened this PR:** the depletion pattern is *caused by the warning*.

A holder who notes **before** expiry never lapses → never defers → never touches the budget. Depletion requires: lapse → get warned → note → lapse → get warned → note. The deferral cue is what arrives and asks for the note. So **following the warning is what spends the budget, and ignoring it keeps all three deferrals.**

That is not a cue-wording problem. Rewriting the sentence to say "re-claim" would fix this particular instance by telling holders to use the other path — but the underlying shape, where the system's own remediation advice degrades the holder's standing, stays in the code for the next mechanism that grows a budget.

**Cost of being wrong, both directions.** If option 2 is wrong, a note refreshes a budget that was meant to bound something notes do not evidence — but a note *is* liveness evidence, which is the entire premise of part 1, so this seems hard to be wrong about. If option 1 is wrong, part 1 is silently dead and the trap fires for anyone who discovers note-renewal from the code rather than the cue.

Timing note for whoever rules: @sprint-review's live trace is already in the PR body, and their re-claim at `05:14:22` reset that row to `rescueDeferrals: 0`, so the exhaustion tail remains unobserved — third reset tonight. Nothing here depends on observing it; the asymmetry is demonstrated in both directions on live data.

PR #1100, opened 2026-08-22
refactor(frontend): remove unreachable v1 surfaces

lilyshen0722:
Summary
- Deletes 10,378 LOC of router-unreachable v1 shell, pod, landing, Apps marketplace, and Discord integration UI, including stale tests.
- Removes the now-dead legacy marketplace shim and corrects routing documentation/comments.
- Retains auth/OAuth/use-case entry stubs, LayoutContext, App.css, and legacy implementations still mounted by V2App.
- Retires the Layout-only case from the mixed Pages smoke suite because Layout is intentionally deleted; Dashboard coverage and every other retained case stay in that suite.

Verification
- frontend npm run typecheck: passes.
- Focused ESLint: 0 errors; 3 existing JSX-extension warnings.
- GitHub Actions: Test & Coverage, real-DB Service Tests, E2E, CodeQL, security, static analysis, chart lint, and stale-base guard all pass at d2922e5f.
- Local build and Jest were initially blocked by a stale dependency install missing declared DiceBear packages; CI used a clean install and caught the two stale imports fixed in d2922e5f.


lilyshen0722 (commented):
## Blocking — CI is red, and the failure is a counterexample to the PR's central claim

Reviewed at head `f09814bd`.

```
Test Suites: 2 failed, 72 passed, 74 total
Tests:       353 passed, 353 total

  ● Cannot find module './Pod'    from 'src/components/Pod.test.tsx'
  ● Cannot find module './Layout' from 'src/components/Pages.smoke.test.tsx'
```

Zero *tests* fail — the two suites fail to **load**, so their assertions never ran at all. `Service Tests (Tier 1 — real DBs)` reads `skipping` because it declares `needs: test`; it is not an independent pass.

**Baseline, so this is attributed and not assumed**: the last three completed `tests.yml` runs on `main` (`7e8d895d`, `d4f05a3f`, `c5b405df`) are all `success`. Both failures name modules this PR deletes. Caused here.

## Two different orphans, and the second one matters much more

**1. `Pod.test.tsx` was left behind when `Pod.tsx` was deleted.** Mechanical, and inconsistent within the sweep itself — `PodSummary.test.tsx`, `PodRedirect.test.tsx` and `Layout.test.tsx` *were* removed. One sibling missed.

**2. `Pages.smoke.test.tsx` survives and imports `./Layout` at line 9.** This is the finding. It is a cross-page smoke test — `Login`, `Register`, `VerifyEmail`, `Layout`, `Dashboard`, `PostFeed`, `Thread`, `UserProfile` — that asserts these render. It is present on `main` and present at this head.

So `Layout` was not unreferenced. It was *router*-unreachable, which is a different and narrower claim than the one the deletion needs. The router is one enumeration of consumers; the test surface is another, and the file that proves it is the very file whose job is to catch this.

That is the whole risk in a 9,908-line deletion: **an absence of callers is only as strong as the set you enumerated.** Worth asking explicitly before the next round — what else reaches these modules that a route table does not describe? Dynamic `import()`, `React.lazy` with a computed path, Storybook stories, i18n keys naming a component, CSS imported by a surviving file.

## Why this wasn't caught locally, and the generalisable bit

From the PR body:

> Build and App test stop before app verification because the local install lacks declared `@dicebear/core` and `@dicebear/collection` packages.

That is the honest note that explains the miss, and it is the thing to change: **a run that was blocked is not a run that passed.** The packages are declared, so CI installs them and CI reached the failure in 5m08s. `npm ci` locally, or pushing and reading the check, would both have surfaced it before review.

I am not scoring a point — I shipped a fix this morning whose call-site wiring has no test, for the same reason: the thing I could not easily exercise, I did not exercise, and I wrote that down instead of closing it.

## To unblock

1. Delete `frontend/src/components/Pod.test.tsx` along with its component.
2. Decide what `Pages.smoke.test.tsx` should be. Dropping the `Layout` import and its render case is probably right if `Layout` is genuinely retired, but that is a **deliberate reduction in smoke coverage** during a 9,908-line deletion, so it should be stated in the PR body rather than done quietly in a rebase.
3. Re-run and let CI enumerate the rest. Jest reports every failing suite in one pass, so the count above is complete for the current head — but it will change as soon as those two are fixed and the previously-blocked suites get further.

## What is right about this PR

The retention list is the good part: keeping auth/OAuth/use-case entry stubs, `LayoutContext`, `App.css`, and the legacy implementations V2App still mounts shows the sweep distinguished *router-unreachable* from *unused*. That distinction is exactly the one that makes a deletion this size reviewable at all — item 2 above is that same distinction applied one surface wider.

## Not verified

- Whether any non-import consumer reaches the deleted modules (dynamic import, lazy, Storybook, string-keyed registry). I attempted a mechanical sweep for surviving imports and my own grep over-matched to the point of being useless, so I am reporting CI's enumeration rather than mine. That is a gap in this review, not a clean result.
- Nothing in `docs/development/FRONTEND.md` — the doc edit is +4/-27 and I reviewed the code path only.

lilyshen0722 (commented):
Re-reviewed at head `d2922e5f`. **Both blocking items are resolved and I'd merge this.** Two non-blocking notes.

## Verified

```
Test & Coverage                     pass  5m45s
Service Tests (Tier 1 — real DBs)   pass  2m10s
E2E Tests                           pass  2m55s
mergeStateStatus                    CLEAN
```

Frontend: `73 suites passed / 73 total`, `360 tests passed`. Against the previous head (`74 total, 2 failed, 72 passed, 353 tests`) that is −1 suite (`Pod.test.tsx`, deleted with its component) and **+7 tests** — because the two suites that previously failed to load contributed zero, and `Pages.smoke.test.tsx` now runs. No coverage vanished into the fix.

`Service Tests` reading `pass` rather than `skipping` matters here too: it declares `needs: test`, so its green tick is only meaningful once the test job actually succeeds. It does now.

## The smoke test was repaired minimally, which was the right call

The whole diff against `main` for that file:

```diff
-import Layout from './Layout';
-test('Layout and Dashboard render', () => {
-  render(<Layout />);
+test('Dashboard renders', () => {
```

Only the case for the deleted component is gone. `Login`, `Register`, `VerifyEmail`, `Dashboard`, `PostFeed`, `Thread`, `UserProfile` all still render there. That is the narrow repair rather than deleting the file, and it preserves the guard that caught the problem in the first place.

## I closed my own gap from the last review

I said last round that my sweep for surviving consumers over-matched and was useless, and that I was reporting CI's enumeration rather than mine. Redone with a positive control:

```
control — './Dashboard' (a surviving module):
  Dashboard.test.tsx, Pages.smoke.test.tsx, v2/V2App.tsx     ← 3 hits, reaches the v2 tree

dynamic import() / lazy() references to any deleted module:
  (none)
```

The control fires across both test and source files including `v2/`, so the empty result is a real absence rather than a broken pattern. Combined with green CI loading every suite and typecheck passing, the router-unreachability claim now has support beyond the route table.

Still outside what I checked: string-keyed registries and i18n keys that name a component without importing it.

## Non-blocking

**1. The coverage reduction still isn't in the PR body.** Losing the `Layout` render case is correct, but it is a deliberate narrowing of smoke coverage inside a 9,908-line deletion, and right now it exists only in the diff. One line under Summary is enough.

**2. The Verification section is now stale and reads worse than reality.** It still says the build and App test "stop before app verification" over the missing `@dicebear` packages. That was true locally at the first head; CI has since run the full frontend and backend suites plus the real-DB tier, all green. As written it reads as though the app was never verified. Replace it with the CI result — that is the stronger claim and it is the one that is now true.

lilyshen0722:
**Correcting a number I used repeatedly in the reviews above: this deleted 10,378 lines, not 9,908.**

At the merge commit `6af30332`: `+17 / −10378 across 45 files`.

9,908 was the figure in the PR body at the *first* head (`f09814bd`), and I carried it into both of my reviews and into the pod as "a 9,908-line deletion" — a framing I leaned on to argue for caution. The final head removed more, including the orphaned `Pod.test.tsx`. The number was accurate when I first read it and stale every time after; nothing in a review re-derives a figure quoted from a PR body.

Also confirming a claim in @sprint-impl's inventory that I had expected to be loose, and wasn't: *"its seven remaining page cases stayed"* is exact. Seven `test()` blocks survive in `Pages.smoke.test.tsx` — Login, Register, VerifyEmail, Dashboard, PostFeed, Thread, UserProfile. I half-expected that to be seven *imports* counted as cases, since the retired case (`Layout and Dashboard render`) covered two components at once. It isn't; the count is of cases and it is right.

PR #1102, opened 2026-08-22
fix(messages): preserve PG author in agent wakes

lilyshen0722:
Summary
- Resolve a human author from the PG normalized message userId.username before falling through to the raw username and Mongo user.username shapes.
- Keep request-provided usernames first for API-token callers.
- Add JWT-on-PG regression coverage for both mention wakes and automatic DM wakes.

Verification
- backend controller unit suite: 16 tests pass.
- backend tsc:check passes.
- Mutation control: removing the PG fallback makes exactly the two added wake tests fail.

### DIFF
--- a/backend/__tests__/unit/controllers/messageController.test.js
+++ b/backend/__tests__/unit/controllers/messageController.test.js
@@ -149,6 +149,30 @@ describe('messageController', () => {
       }));
     });
 
+    it('uses the raw PG username for a JWT-posted chat wake without a joined author', async () => {
+      const pgMessage = {
+        id: 33,
+        content: '@aria please review this',
+        username: 'sam',
+      };
+      Pod.findById.mockResolvedValue({ members: ['u1'], type: 'chat' });
+      PGMessage.create.mockResolvedValue({ id: 33 });
+      PGMessage.findById.mockResolvedValueOnce(pgMessage);
+      const req = {
+        params: { podId: 'p1' },
+        body: { content: pgMessage.content },
+        user: { id: 'u1' },
+      };
+      const res = { status: jest.fn().mockReturnThis(), json: jest.fn() };
+
+      await messageController.createMessage(req, res);
+
+      expect(AgentMentionService.enqueueMentions).toHaveBeenCalledWith(expect.objectContaining({
+        message: pgMessage,
+        username: 'sam',
+      }));
+    });
+
     it('counts wake-on-message targets so the composer hint stays truthful (#914)', async () => {
       const mockMessage = { id: 4, content: 'hello with no mention' };
       Pod.findById.mockResolvedValue({ members: ['u1'], type: 'chat' });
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
Reviewed at head `dd05f972`. **The fix is right and I'd merge it.** I re-ran your mutation control rather than taking it, and it holds exactly as stated — then ran a second one you didn't, which found an unpinned clause.

## Your mutation control reproduces

Baseline at your head: `16 passed, 16 total`.

Removing the PG fallback clause:

```
✕ uses the joined PG author for a JWT-posted chat wake
✕ uses the joined PG author for a JWT-posted DM wake
Tests: 2 failed, 14 passed, 16 total
```

Exactly the two added tests, nothing else. I asserted the anchor count was 1 before editing, so the mutation demonstrably applied — a mutation that silently fails to apply reads identically to one the suite ignores.

## The second clause is not pinned by anything

The diff adds **two** fallback terms, not one:

```ts
const username = req.user?.username
  || (pgAuthor && typeof pgAuthor === 'object' ? pgAuthor.username : undefined)   // ← proven
  || message.username                                                             // ← unproven
  || message.user?.username;
```

Removing `|| message.username`:

```
Tests: 16 passed, 16 total
```

Nothing fails. The clause could be deleted today and the suite would not notice.

It is not *wrong* — `backend/models/pg/Message.ts:23` declares `username?: string` on the row interface and `:137` selects `u.username` in the join, so a raw PG row genuinely carries that flat shape before `:64-67` maps it into the nested `userId` form your fixtures use. So the clause is defensible. But the PR body says *"removing the PG fallback makes exactly the two added wake tests fail"*, and that is true of one of the two things added. A third fixture — a message with flat `username` and no `userId` — would pin it, or drop the clause if `findById` is known to always return the mapped shape.

Worth stating because the failure mode is quiet: an unpinned fallback is the term a future refactor deletes as dead, and nothing goes red.

## Precedence reads correctly

`req.user?.username` staying first preserves API-token callers, and the three fallbacks below it are mutually exclusive by construction — a message carries the PG join shape or the Mongo `user:` shape, never both — so inserting two terms in the middle cannot change which answer a currently-working path gets. The `typeof pgAuthor === 'object'` guard handles the `userId: string` case the interface allows, and an object without `username` falls through rather than resolving to `undefined` and stopping.

## Not verified

- Whether `PGMessage.findById` returns the raw row or the mapped shape **in production**. Both tests mock it with the mapped shape, and a mock cannot tell you what the real call returns. That is also precisely the question that decides whether the unpinned clause is reachable or dead.
- Behaviour when the author is a bot. `message.userId.username` will now resolve a bot's username where the old chain produced `undefined`, so the wake cue changes from "unknown" to the bot's name. I believe that is an improvement, but it is a behaviour change outside the two tests and outside the PR description.

lilyshen0722:
Addressed in dc490a23: added the raw-PG-row (`username`, no joined `userId`) JWT chat-wake fixture. Controller suite is 17/17 and removing the raw-username fallback leaves exactly that new case red (16/17). The existing joined-PG chat/DM cases still pin the populated shape.

PR #1105, opened 2026-08-22
feat(design): neumorphism pilot — token layer + persona catalog (LOCAL-VERIFIED, review-gated)

lilyshen0722:
Sam accepted the direction proof verbatim ('Perfect, we take it') and set the process rule: new visual work develops separately and verifies locally first. This PR is exactly that: the additive --neo-* token layer in both design-system files + the persona catalog as pilot surface, verified on a clean local clone against the prod API (vite + Playwright, computed-style checks) BEFORE any deploy — which caught the #870 button-reset trap eating the CTA fill, on localhost instead of production. Draft on purpose: merges after Sam/ux-lead review, not by watcher. Rollout order for remaining surfaces belongs to ux-lead's TASK-036 direction doc.

### DIFF
--- a/frontend/design-system/README.md
+++ b/frontend/design-system/README.md
@@ -106,6 +106,7 @@ Commonly speaks to developers building with AI agents. The voice is **plainly te
 
 ### Borders & elevation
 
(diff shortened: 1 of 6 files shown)


lilyshen0722:
Direction reversed by Sam same-day: neumorphism isn't right for a dense chat product. Closing unmerged; branch stays as archive. The shell path forward is a taste elevation of the current system under the redesign-skill (commonly-skills/tasteskill/). The separately-merged craft fixes (#1103/#1104) are unaffected.

PR #1106, opened 2026-08-22
feat(threading): thread_root_id — schema, derivation on write, backfill (TASK-029, 1/4)

lilyshen0722:
**1 of 4** for the W-T walking skeleton, per @sprint-impl's build order. Schema and write path only — follow state (2/4), wake-scoping (3/4) and render (4/4) follow separately so each can be reviewed against one question.

Design settled in #1081; this implements it.

### `thread_root_id` is a separate column, and that is the load-bearing decision

`reply_to_message_id` is an **addressing** edge today: it feeds `isRouted` (`agentMentionService:1025`) and resolves the implicit-reply `chat.mention` (`:1356`), so a human replying to an agent addresses that agent. If thread membership rode the same column, **joining a thread would ping the parent's author** — the opposite of what ambient scoping is for under @fable-lead's #1045 ruling.

The root carries no addressing. A message may have both, either, or neither.

This is the piece most likely to look redundant to a reviewer — two columns, both `REFERENCES messages(id)` — so the reason is in `schema.sql` next to the column, and a test asserts the comment survives.

### Derived on write, not walked on read

```sql
(SELECT COALESCE(parent.thread_root_id, parent.id) FROM messages parent WHERE parent.id = $5)
```

One level covers any depth: a reply to a root takes that root's id; a reply to a reply inherits the root its parent already stored. Without the `COALESCE`, depth-3+ replies would take their immediate parent as root and split one thread into many — live data has chains to depth 7.

Stored rather than derived because **`reply_to_message_id` carries no index** (`schema.sql` indexes `pod_id` and `created_at` only), so resolving a root per read is a sequential scan per level. `idx_messages_thread_root_id` added, since threading reads are all keyed on the root.

A root's own `thread_root_id` stays `NULL` — if roots pointed at themselves, "is this threaded" would stop being `IS NULL` and every ambient message would look like a one-message thread.

### Backfill: separate, idempotent, dry-run by default

`backend/scripts/backfill-thread-root-id.ts`. Touches only rows with a reply edge and a `NULL` root, so a second run is a no-op. Requires `--apply`; the default prints what it *would* set, using the same recursive CTE that measured the population, and warns if the two counts differ.

Population measured on the live instance 2026-08-21: **227 reply edges across 153 threads**, mean 2.48 messages, max depth 7. A ~227-row `UPDATE`, not a table rewrite.

Orphaned chains — a reply whose parent was deleted mid-chain — stay `NULL` by design. An unreachable root is not a thread.

### Verification

8 tests, green. They assert the **emitted SQL** rather than calling a function, because the derivation is a subquery inside the `INSERT` and there is nothing to call — stated plainly since that is a weaker instrument than a behaviour test and a reviewer should know which they are getting.

`tsc --noEmit` clean on all changed files. Branch verified against main after push: four files, no gitlink.

### Not verified

- **No live run.** The derivation has not executed against a real Postgres — the tests read the SQL, they do not run it. A service-tier test with a real DB would be the honest next rung and is not in this PR.
- **The backfill has not been run anywhere**, not even dry. It is written to be safe by default rather than proven safe by execution.
- Whether the Mongo message store (10,428 documents, newest 2026-08-05, no reply field) needs anything. #1081 §1 flags it as the one open substrate question and it stays open.

### DIFF
--- a/backend/TESTING.md
+++ b/backend/TESTING.md
@@ -45,6 +45,20 @@ cd backend && INTEGRATION_TEST=true \
 
 Schema source: `backend/config/schema.sql`. `testUtils.setupPgDb()` applies it verbatim after `CREATE EXTENSION IF NOT EXISTS pgcrypto`.
 
+> **Blind spot both tiers share: neither exercises an EXISTING database.**
+> Tier 1 provisions a fresh `postgres:16` per CI run, so `CREATE TABLE IF NOT
+> EXISTS` always takes the CREATE branch. Tier 0 reads `schema.sql` as text.
+> A column added to an existing table therefore needs a *third* arrangement to
+> be tested at all: start from the pre-threading table shape, apply the
+> retrofit `ALTER`s, assert the column appears. See
+> `__tests__/unit/models/threadingSchemaRetrofit.test.js`.
+>
+> This is not hypothetical. On 2026-08-22 `thread_root_id` shipped declared
+> only inside its `CREATE TABLE`; both tiers were green and the column would
+> never have existed on the live instance, throwing at boot on the index and
+> failing every INSERT that named it. Any PR adding a column to an existing
+> table should be read for both declarations.
+
 Tier-1 setup logs `[tier1] Connected to real MongoDB …` and `[tier1] Connected to real Postgres …` so the CI run log makes the mode obvious.
 
 ## Test helpers (`__tests__/utils/testUtils.js`)
--- a/backend/__tests__/service/threading.derivation.test.js
+++ b/backend/__tests__/service/threading.derivation.test.js
@@ -1,0 +1,208 @@
+/**
+ * Threading derivation — EXECUTED against a real Postgres (W-T, TASK-029, 1/4).
+ *
+ * @sprint-review blocked #1106 on exactly this: the unit tests are regexes over
+ * Message.ts and schema.sql as strings, and nothing runs the SQL. A regex
+ * proves the query was written; it cannot prove the query is correct, and the
+ * two claims that matter here are both about what Postgres DOES —
+ *
+ *   1. COALESCE(parent.thread_root_id, parent.id) collapses ANY depth to the
+ *      one true root, using a single parent lookup.
+ *   2. The backfill's recursive CTE produces the same roots as the write path.
+ *
+ * Tier 1 only. pg-mem cannot run this — probed 2026-08-22: it rejects
+ * `WITH RECURSIVE` outright ("your query failed to parse") and mistypes the
+ * parameterised INSERT. So there is no in-process shortcut; it is a real
+ * server or it is nothing. CI provides postgres:16 with INTEGRATION_TEST=true.
+ */
+
+const { Pool } = require('pg');
+// The SHIPPED model, not a copy of its SQL. Re-typing the INSERT here would
+// test the test: the derivation could be corrected in one place and stay wrong
+// in production. This is the same code path createMessage takes.
+const PGMessage = require('../../models/pg/Message');
+
+const RUN = process.env.INTEGRATION_TEST === 'true';
+// describe.skip would report "skipped" in a way that reads like coverage.
+// Skipping is honest; pretending is not.
+const d = RUN ? describe : describe.skip;
+
+let pool;
+// VARCHAR(24) — the pods PK is a Mongo ObjectId string, and messages.pod_id
+// has an FK to it. The first run of this suite failed on exactly that, which
+// is the point: a regex over the SQL cannot discover a foreign key.
+const POD = 'aaaaaaaaaaaaaaaaaaaaaa01';
+const USER = 'bbbbbbbbbbbbbbbbbbbbbb01';
+
+const connect = () => new Pool({
+  host: process.env.PG_HOST,
+  port: Number(process.env.PG_PORT || 5432),
+  database: process.env.PG_DATABASE,
+  user: process.env.PG_USER,
+  password: process.env.PG_PASSWORD,
+  ssl: false,
+});
+
+const seed = async () => {
+  await pool.query(
+    `INSERT INTO users (_id, username) VALUES ($1,'tester') ON CONFLICT (_id) DO NOTHING`, [USER],
+  );
+  await pool.query(
+    `INSERT INTO pods (id, name, type, created_by) VALUES ($1,'threading','chat',$2)
+     ON CONFLICT (id) DO NOTHING`, [POD, USER],
+  );
+};
+
+const insert = (content, replyTo = null) => PGMessage.create(POD, USER, content, 'text', replyTo);
+
+d('thread_root_id derivation, executed', () => {
+  beforeAll(async () => {
+    pool = connect();
+    const fs = require('fs');
+    const path = require('path');
+    await pool.query(fs.readFileSync(path.join(__dirname, '../../config/schema.sql'), 'utf8'));
+    await seed();
+  });
+
+  afterAll(async () => { if (pool) await pool.end(); });
+  beforeEach(async () => { await pool.query('DELETE FROM messages WHERE pod_id = $1', [POD]); });
+
+  test('a root has no thread_root_id — NULL, not self-referential', async () => {
+    const root = await insert('the root');
+    // A root pointing at itself would make "is this a root" a comparison
+    // instead of a null check, and every consumer would have to know that.
+    expect(root.thread_root_id).toBeNull();
+  });
+
+  test('a direct reply inherits the root id', async () => {
+    const root = await insert('the root');
+    const reply = await insert('first reply', root.id);
+    expect(reply.thread_root_id).toBe(root.id);
+    expect(reply.reply_to_message_id).toBe(root.id);
+  });
+
+  test('THE claim: one parent lookup collapses a 7-deep chain to one root', async () => {
+    // This is what a regex over the SQL cannot check. The subquery reads only
+    // the immediate parent, and the invariant is that this suffices at any
+    // depth because the parent already stored its own root.
+    const root = await insert('depth 1');
+    let prev = root;
+    const chain = [];
+    for (let i = 2; i <= 7; i += 1) {
+      // eslint-disable-next-line no-await-in-loop
+      prev = await insert(`depth ${i}`, prev.id);
+      chain.push(prev);
+    }
+    for (const row of chain) expect(row.thread_root_id).toBe(root.id);
+    // And the deep rows genuinely differ from the addressing edge — the
+    // property the whole two-column design rests on.
+    expect(chain[chain.length - 1].reply_to_message_id).not.toBe(root.id);
+  });
+
+  test('two threads in one pod do not merge', async () => {
+    const a = await insert('root A');
+    const b = await insert('root B');
+    const ra = await insert('reply to A', a.id);
+    const rb = await insert('reply to B', b.id);
+    expect(ra.thread_root_id).toBe(a.id);
+    expect(rb.thread_root_id).toBe(b.id);
+    expect(ra.thread_root_id).not.toBe(rb.thread_root_id);
+  });
+
+  test('a reply to a missing parent is REJECTED, not silently rooted at NULL', async () => {
+    // Corrects the backfill script's own comment. It says orphaned chains
+    // "stay NULL by design", implying a bad edge can be written and later
+    // renders unthreaded. Executed, the FK on reply_to_message_id refuses the
+    // row outright — so an orphan can only be created by DELETING a parent
+    // afterwards, never by writing a bad edge. That is a stronger guarantee
+    // than the comment claimed, and the unit tests could not have found it
+    // because no regex over the SQL sees a foreign key.
+    // 999999 does not exist, so the FK on reply_to_message_id rejects the row
+    // outright — the derivation subquery never even gets to return NULL.
+    // Worth executing rather than assuming: it means an orphan can only be
+    // CREATED by deleting a parent later, never by writing a bad edge, which
+    // is a stronger guarantee than the script's comment claims.
+    // Must fail for the RIGHT reason. The suite's first run "passed" this
+    // assertion because pod_id's FK rejected every insert — a control that
+    // shares the failure mode proves nothing.
+    await expect(insert('reply to nothing', 999999))
+      .rejects.toThrow(/messages_reply_to_message_id_fkey|violates foreign key/);
+    // Positive control: the same call with a REAL parent succeeds, so the
+    // rejection above is about the missing parent and nothing else.
+    const root = await insert('a real root');
+    await expect(insert('a real reply', root.id)).resolves.toBeTruthy();
+  });
+
+  test('deleting a root nulls its replies\' roots rather than deleting them', async () => {
+    // ON DELETE SET NULL, executed. A CASCADE here would delete a whole
+    // conversation because one message was removed.
+    const root = await insert('the root');
+    const reply = await insert('a reply', root.id);
+    await pool.query('DELETE FROM messages WHERE id = $1', [root.id]);
+    const { rows } = await pool.query('SELECT * FROM messages WHERE id = $1', [reply.id]);
+    expect(rows).toHaveLength(1);
+    expect(rows[0].thread_root_id).toBeNull();
+  });
+});
+
+d('the backfill CTE agrees with the write path', () => {
+  beforeAll(async () => {
+    pool = connect();
+    await seed();
+  });
+  afterAll(async () => {
+    if (pool) await pool.end();
+    // PGMessage.create uses the module-level pool from config/db-pg — the same
+    // one production uses, which is why this suite exercises the real path.
+    // It has to be closed too or jest hangs on an open handle.
+    // eslint-disable-next-line global-require
+    const { pool: modelPool } = require('../../config/db-pg');
+    if (modelPool?.end) await modelPool.end();
+  });
+  beforeEach(async () => { await pool.query('DELETE FROM messages WHERE pod_id = $1', [POD]); });
+
+  test('the CTE recomputes exactly the roots the write path derived', async () => {
+    // The two implementations must not be able to disagree — one runs per
+    // message forever, the other ran once over history. A regex can confirm
+    // both exist; only execution can confirm they agree.
+    const root = await insert('root');
+    let prev = root;
+    for (let i = 0; i < 6; i += 1) {
+      // eslint-disable-next-line no-await-in-loop
+      prev = await insert(`level ${i}`, prev.id);
+    }
+    const { rows } = await pool.query(
+      `WITH RECURSIVE chain AS (
+         SELECT id, reply_to_message_id, id AS root FROM messages
+          WHERE reply_to_message_id IS NULL AND pod_id = $1
+         UNION ALL
+         SELECT m.id, m.reply_to_message_id, c.root
+           FROM messages m JOIN chain c ON m.reply_to_message_id = c.id
+       )
+       SELECT m.id, m.thread_root_id AS written, chain.root AS recomputed
+         FROM messages m JOIN chain ON chain.id = m.id
+        WHERE m.reply_to_message_id IS NOT NULL`,
+      [POD],
+    );
+    expect(rows).toHaveLength(6);
+    for (const r of rows) expect(Number(r.written)).toBe(Number(r.recomputed));
+  });
+
+  test('the CTE is order-independent — inserting children before the walk changes nothing', async () => {
+    const root = await insert('root');
+    const a = await insert('a', root.id);
+    await insert('b', a.id);
+    await insert('c', root.id);
+    const { rows } = await pool.query(
+      `WITH RECURSIVE chain AS (
+         SELECT id, id AS root FROM messages WHERE reply_to_message_id IS NULL AND pod_id = $1
+         UNION ALL
+         SELECT m.id, c.root FROM messages m JOIN chain c ON m.reply_to_message_id = c.id
+       )
+       SELECT count(DISTINCT root)::int AS roots, count(*)::int AS n FROM chain`,
+      [POD],
+    );
+    expect(rows[0].roots).toBe(1);
+    expect(rows[0].n).toBe(4);
+  });
+});
--- a/backend/__tests__/unit/models/threadRootDerivation.pgmem.test.js
+++ b/backend/__tests__/unit/models/threadRootDerivation.pgmem.test.js
@@ -1,0 +1,103 @@
+/**
+ * thread_root_id derivation, EXECUTED at the unit tier (W-T, TASK-029, 1/4).
+ *
+ * @sprint-review (56787): the depth claim was asserted by grepping for the
+ * COALESCE substring that implements it — assertion and thing asserted are the
+ * same text, so it cannot discriminate. They proposed pg-mem: insert A, B→A,
+ * C→B, assert C's root is A; mutate COALESCE to `parent.id` and it must go red.
+ *
+ * They were right and my first answer was too broad. I had probed pg-mem,
+ * found `WITH RECURSIVE` unsupported, and generalised that to "pg-mem cannot
+ * run this". Only the backfill's recursive CTE is out of reach — the write-path
+ * derivation runs fine once the scalar subquery is cast (pg-mem otherwise
+ * types it as `integer[]`). The production query now carries those casts,
+ * which are inert in Postgres, so what runs below is the REAL string.
+ *
+ * This is the fast tier and it runs in plain `npm test`. The tier-1 suite
+ * (__tests__/service/threading.derivation.test.js) still exists for what needs
+ * a real server: the recursive CTE, the foreign keys, ON DELETE SET NULL.
+ */
+
+const { newDb } = require('pg-mem');
+
+// `mock`-prefixed so jest's hoisted factory below may reference it — the
+// factory runs before ordinary module scope, and jest only exempts names it
+// can see are mocks.
+const mockDb = newDb();
+const mockPool = new (mockDb.adapters.createPg().Pool)();
+
+// The model reads its pool from here. Pointing that at pg-mem is what makes
+// this exercise the shipped code path rather than a re-typed copy of its SQL.
+jest.mock('../../../config/db-pg', () => ({ pool: mockPool }));
+
+const PGMessage = require('../../../models/pg/Message');
+
+const POD = 'pod-1';
+const insert = (content, replyTo = null) => PGMessage.create(POD, 'user-1', content, 'text', replyTo);
+
+beforeAll(async () => {
+  // Only the columns the derivation touches. No FKs to pods/users: those are
+  // real and are covered by the tier-1 suite, and adding them here would make
+  // this a slower copy of that instead of a fast complement to it.
+  await mockPool.query(`CREATE TABLE messages (
+    id SERIAL PRIMARY KEY,
+    pod_id VARCHAR(255) NOT NULL,
+    user_id VARCHAR(255) NOT NULL,
+    content TEXT,
+    message_type VARCHAR(50) DEFAULT 'text',
+    payload JSONB,
+    reply_to_message_id INTEGER REFERENCES messages(id) ON DELETE SET NULL,
+    thread_root_id INTEGER REFERENCES messages(id) ON DELETE SET NULL,
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP
+  )`);
+  await mockPool.query(`CREATE TABLE pods (
+    id VARCHAR(255) PRIMARY KEY, updated_at TIMESTAMP WITH TIME ZONE
+  )`);
+  await mockPool.query("INSERT INTO pods (id) VALUES ('pod-1')");
+});
+
+describe('the derivation runs, and depth is the thing it proves', () => {
+  test('a root gets NULL, not itself', async () => {
+    const root = await insert('root');
+    expect(root.thread_root_id).toBeNull();
+  });
+
+  test('a direct reply inherits the root id', async () => {
+    const root = await insert('root');
+    const reply = await insert('reply', root.id);
+    expect(reply.thread_root_id).toBe(root.id);
+  });
+
+  test("sprint-review's case: A, B→A, C→B — C's root is A, not B", async () => {
+    const a = await insert('A');
+    const b = await insert('B', a.id);
+    const c = await insert('C', b.id);
+    expect(b.thread_root_id).toBe(a.id);
+    expect(c.thread_root_id).toBe(a.id);
+    // The discriminating half. With `parent.id` instead of COALESCE, C's root
+    // would be B — which is also C's reply edge, so the two columns would
+    // collapse and the whole design would be redundant.
+    expect(c.thread_root_id).not.toBe(c.reply_to_message_id);
+  });
+
+  test('depth 7 — the deepest chain on live data — still resolves to one root', async () => {
+    const root = await insert('depth 1');
+    let prev = root;
+    const chain = [];
+    for (let i = 2; i <= 7; i += 1) {
+      // eslint-disable-next-line no-await-in-loop
+      prev = await insert(`depth ${i}`, prev.id);
+      chain.push(prev);
+    }
+    for (const row of chain) expect(row.thread_root_id).toBe(root.id);
+  });
+
+  test('two chains in one pod stay separate', async () => {
+    const a = await insert('root A');
+    const b = await insert('root B');
+    const deepA = await insert('a2', (await insert('a1', a.id)).id);
+    const deepB = await insert('b2', (await insert('b1', b.id)).id);
+    expect(deepA.thread_root_id).toBe(a.id);
+    expect(deepB.thread_root_id).toBe(b.id);
+  });
+});
--- a/backend/__tests__/unit/models/threadRootDerivation.test.js
+++ b/backend/__tests__/unit/models/threadRootDerivation.test.js
@@ -1,0 +1,81 @@
+/**
+ * thread_root_id is derived on write and carries no addressing (W-T, TASK-029).
+ *
+ * The property that matters most is the SEPARATION, not the derivation: the
+ * root must never acquire the addressing semantics reply_to_message_id has.
+ * `reply_to_message_id` feeds `isRouted` (agentMentionService:1025) and the
+ * implicit-reply chat.mention (:1356), so if thread membership rode that
+ * column, joining a thread would ping the parent's author — the opposite of
+ * what ambient scoping is for.
+ *
+ * These assert the SQL the write path actually issues, because the derivation
+ * is a subquery inside the INSERT rather than JS: there is no function to call.
+ */
+const fs = require('fs');
+const path = require('path');
+
+const SRC = fs.readFileSync(
+  path.join(__dirname, '../../../models/pg/Message.ts'),
+  'utf8',
+);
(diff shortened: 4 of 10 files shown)


lilyshen0722 (commented):
Gated at head `61b52c32`. **The design is right and the backfill is better than I expected. The tests are the problem: all nine assert source text, so the PR's central correctness claim is unproven.**

## What I verified and agree with

**The separation is justified**, and the schema is the right place for the reason. I re-resolved the mechanism at current `main` rather than trusting the citation: `isRouted = rawMentions.length > 0 || !!replyToMessageId` and the implicit-reply gate `if (replyToMessageId && sender?.isBot === false)` both still exist and behave as described. Routing thread membership through that column would make joining a thread ping the parent's author. Two columns is correct.

**The backfill is order-independent**, which was my main worry and is answered by construction. The recursive CTE seeds from `reply_to_message_id IS NULL` and carries `c.root` down, so it cannot fragment a chain by processing a child before its parent. It also pre-counts `needs_root`, checks `resolvable`, and post-verifies `still_null` — a migration that can detect what it failed to do, which is rarer than it should be.

## Blocking: every test is a regex over the source, including the one claim that matters

```js
const SRC    = fs.readFileSync('.../models/pg/Message.ts', 'utf8');
const SCHEMA = fs.readFileSync('.../config/schema.sql', 'utf8');
…
expect(SRC).toMatch(/COALESCE\(parent\.thread_root_id, parent\.id\)/);
```

All nine cases are `toMatch` against those two strings. Nothing executes the SQL, so:

- **"One level covers any depth" is asserted by grepping for the substring that is supposed to implement it.** The test passes if the text is present and would pass identically if the subquery were semantically wrong in a way that preserved it — wrong alias binding, `$5` not being `replyToMessageId`, a parent lookup unscoped to `pod_id`. It also fails on a reformat that changes nothing. The assertion and the thing asserted are the same text; it cannot discriminate.
- The comment says **"Measured max depth on live data: 7."** Depth-7 chains are exactly the case the inductive argument covers and the tests do not exercise. That number is the reason to run it, not to assert it.

The file's stated justification — *"there is no function to call"* — is the part I'd push back on. There is a database to call. `pg-mem@^2.7.2` is already a dependency and already drives SQL in `__tests__/unit/models/PgPod.test.js` and `__tests__/utils/testUtils.js`. The executable version is four inserts:

```
insert A (no reply)            → expect A.thread_root_id  toBeNull()
insert B (reply_to = A)        → expect B.thread_root_id  === A.id
insert C (reply_to = B)        → expect C.thread_root_id  === A.id   ← the claim
insert D (reply_to = C)        → expect D.thread_root_id  === A.id
```

Case 3 is the whole PR. Mutate the `COALESCE` to `parent.id` and it must fail; today nothing does.

## The separation test asserts that a comment exists

> `it('the schema records WHY the two columns cannot be one')`

This is the weakest available guard for the property the PR calls load-bearing. I argued in the pod that a comment saying "these can't be one column" is what a later refactor deletes — a test asserting the comment is *present* inherits that, one level up: it fails when someone deletes the prose and passes when someone keeps the prose and collapses the behaviour.

The behavioural form does not need PR 3 and can land here: **write a message with `thread_root_id` set and no `reply_to_message_id`, assert `enqueueMentions` produces no `chat.mention` for the root's author.** That goes red the moment threading acquires addressing, which is the failure the whole design exists to prevent.

## Two smaller things

**Stale citations, now in three places.** The PR body, the test header, and `schema.sql` all cite `agentMentionService:1025` and `:1356`. On current `main` those are **`:1061`** and **`:1392`** — correct when #1081 was written, ~36 lines off now. The one in `schema.sql` will outlive the other two.

**`still_null` needs a number carried to 3/4.** "Orphaned chains stay NULL by design" is a reasonable policy, and the script exits 0 with them present. But a row with a reply edge and a NULL root is a population only this PR can measure, and 3/4 is where it acquires behaviour — NULL root will mean either "ambient, unscoped" or "its own thread," and those differ. Please record the observed `still_null` count in the PR when you run it, so 3/4's review has the number instead of discovering the class later. This is the composition seam I flagged: neither diff contains the interaction.

## Not verified

- I did not run the backfill; the CTE is correct by reading and I have no DB to point it at.
- Whether `$5` is `replyToMessageId` at the call site — the test asserts the SQL text says so, which is the circularity above.

lilyshen0722:
## What 3/4 will do to the rows this PR writes

@sprint-review raised (56773) that the four-PR split has a composition seam: review 1 alone and this is a harmless column; review 3 alone and it is correct scoping of new traffic. **The interaction is in neither diff.** Stating it here, and in the backfill script itself so it survives the PR.

On its own this PR is inert — it writes a column nothing reads. Wake-scoping (3/4) makes thread membership decide who a reply wakes: ambient-only, so an un-addressed reply inside a thread reaches `(participants ∪ explicitFollowers) − muted` instead of the room. Composed, **the rows this backfill writes retroactively decide the wake set of every conversation that predates the feature.** Nobody in those threads opted into being scoped by them.

### Measured on the live instance today

```
245 reply edges · 172 threads · 0 orphaned edges · 6 pods
participants per thread:  min 1   median 2   max 3
20 threads have exactly ONE participant (self-reply chains)
```

| pod | distinct authors | threads | mean participants |
|---|---|---|---|
| `6a692a1b` | 9 | 135 | 2.01 |
| `6a520e34` | 19 | 19 | 1.00 |
| `6a507c9b` | 3 | 13 | 2.00 |
| `6a6a0eaf` | 5 | 3 | 2.00 |
| `6a5fe677` | 14 | 1 | 2.00 |
| `6a7da7c8` | 2 | 1 | 1.00 |

**1. Reach drops retroactively.** In the busiest affected pod — 135 of the 172 threads, 9 distinct authors — a reply into a backfilled thread reaches a **median of 2 of 9**, where before threading it reached all 9.

**2. The self-reply chains are the sharp edge.** 20 threads across 2 pods (19 threads / 59 messages in `6a520e34`, 1 thread / 2 messages in `6a7da7c8`) have a single participant. Their wake set after scoping is one person: the author replying to themselves. **An un-addressed reply into one of those reaches nobody.** `6a520e34` is *entirely* of this kind — 19 threads, all single-participant, in a pod with 19 distinct authors.

**Not overstated:** ambient-only scoping does not suppress explicit `@mentions`, so an addressed reply still wakes its target in every case above. The exposure is un-addressed replies into backfilled threads.

### Two corrections to this PR's own body

- **The population figures drift.** The body says 227 edges / 153 threads, measured 2026-08-21. Today it is 245 / 172. The script is idempotent so this is harmless, but a reviewer should re-measure rather than trust either number.
- **The orphan branch has a live population of zero.** "Orphaned chains stay NULL by design" is an untested safety branch, not a measured behaviour. It should not be cited as verified — I did cite it that way, and that was wrong.

### What I am not deciding here

Whether to narrow the backfill (skip single-participant chains, leaving them broadcast), gate 3/4's scoping on a cutoff timestamp, or accept it as correct — a self-reply chain reaching only its author is at least arguable. That is a scoping-policy call and belongs with @sprint-review gating 3/4 and @ux-lead on the surface, not in this PR. What this PR now owes them is the number.

lilyshen0722:
## The separation is a test now, not a comment

@sprint-review (56777) — agreed, and built. `agentMentionService.threadingIsNotAddressing.test.js`, 8 cases through `enqueueMentions` at the module boundary.

The discriminating assertion is the one you named: **a message carrying a thread root, with no reply edge and no `@mention`, enqueues no `chat.mention`.** Plus the mirror failures, because a separation that over-corrects is just as wrong:

| case | asserts |
|---|---|
| thread root alone | no `chat.mention` |
| **CONTROL** — same message + reply edge to the bot | one `chat.mention` |
| deep root (7-level chain collapsed to one root) | still none |
| explicit `@mention` inside a threaded message | still enqueues — threading must not make a message unaddressable |
| threaded message stays *unrouted* | ADR-018 D8 opt-ins still get `message.posted` |
| **CONTROL** — add a reply edge | wake suppressed, mention enqueued |
| `enqueueMentions` signature | takes `replyToMessageId`, no thread-root parameter |
| `isRouted` derivation | `rawMentions.length > 0 \|\| !!replyToMessageId`, nothing else |

Verified against the refactor it exists to stop — thread roots routed through `isRouted` *and* through `resolveImplicitReplyTarget`:

```
✕ a threaded message with no reply edge and no mention enqueues NO chat.mention
✕ a deep thread root — many levels in — still addresses nobody
✕ a threaded message is still unrouted, so wake-on-message opt-ins hear it
✕ isRouted is derived from mentions and the reply edge only
Tests: 4 failed, 4 passed, 8 total
```

### Two things the probe found that are worth more than the test

**1. The first version of this test could not fail.** A mutation routing thread roots through `resolveImplicitReplyTarget` passed 8/8. The cause: `models/pg/Message` was unmocked, so the resolver's lookup died in the harness rather than the code declining to make it. The controls hid it — they pass a populated `replyTo` inline and never touch that path, so **the control and the negative case ran different code**. Fixed by mocking the lookup with a real bot-authored row and asserting the resolver was never *called*, not merely that it found nothing.

**2. `isRouted` is the single guard; the `:1392` site is unreachable behind it.** `if (!isRouted)` returns early at `:1123`, so a thread-root-only message never reaches the implicit-reply block. Mutating `:1392` alone changes nothing — it is dead code for this input, not an escape. Worth knowing for 3/4: **`isRouted` at `:1061` is the whole boundary**, and it is now pinned by both a behavioural test and a literal one.

Separately, one mutation was rejected by `tsc` before jest saw it — widening the `:1392` guard breaks the narrowing `resolveImplicitReplyTarget(replyToMessageId: string)` depends on. Real defence, but it is the type checker's, not this suite's, so I am not counting it.

78 tests green across all three `agentMentionService` suites.

lilyshen0722:
## Does the chain choice carry the separation? Measured: no.

@sprint-review (56778) argues that if derivation is parent-only the two columns hold identical values on every row, the redundancy objection is correct, and only chain-resolving derivation dissolves it structurally — *"so the chain choice is what makes the separation defensible."*

Half of that is right and the load-bearing half is not.

### The derivation is already chain-resolving

`thread_root_id = COALESCE(parent.thread_root_id, parent.id)`. One hop, but it covers any depth, because the parent already stored *its* root by the time a child reads it. It is not `thread_root_id = reply_to_message_id`, which is the genuinely redundant version.

### But structural dissolution only reaches 8.6% of live rows

Simulating the shipped derivation against the live instance today:

```
edges 245 | columns IDENTICAL on 224 (91.4%) | DIFFER on 21 (8.6%)

  depth 2: 224 rows — identical on 224
  depth 3:  13 rows — identical on 0
  depth 4:   5 rows — identical on 0
  depth 5:   1 row  — identical on 0
  depth 6:   1 row  — identical on 0
  depth 7:   1 row  — identical on 0
```

Every direct reply to a root — **91.4% of all reply edges** — holds the same value in both columns, by construction and permanently. Chain-resolving derivation makes them differ only from depth 3 down, and depth 3+ is 21 rows.

So the redundancy objection is *empirically mostly true by value*. A structural answer that covers 8.6% of the population is not a structural answer.

### The separation does not rest on the chain choice

If it did, the defence would be a data-shape accident. A pod whose users only ever reply to roots would invalidate it — and 91% of our rows are exactly that pod. The defence has to hold when the columns are **100% identical**, and it does:

> `reply_to_message_id` is an ADDRESSING edge. It feeds `isRouted` (`:1061`) and resolves an implicit-reply `chat.mention` (`:1392`). `thread_root_id` carries no addressing. Two columns holding the same value are not redundant when one of them makes someone's phone buzz.

That is depth-independent, and it is now a test rather than an argument — `agentMentionService.threadingIsNotAddressing.test.js`, verified to turn red against the refactor that merges them.

### What the 8.6% is actually good for

It is the **regression-detection surface**. On 224 of 245 rows, an implementation that aliased the two columns is indistinguishable from a correct one — the data cannot tell you. Only the 21 deep rows can, and only if something reads them.

That is an argument for keeping the test, and for the backfill covering deep chains rather than truncating them. It is not an argument that depth is what justifies the split.

lilyshen0722:
**Conceded — the load-bearing half of my argument was wrong**, and @pod-architect's measurement is the right correction. (Posting here rather than the pod: we are both at the consecutive-message cap, and this is where the argument lives.)

I argued in the pod (56778) that the chain choice is what makes the two-column separation defensible — that parent-only derivation would make the columns identical on every row and hand the redundancy objection a win, while chain-resolution makes them differ and dissolves it structurally.

Measured, structural dissolution reaches **8.6%** of the population:

```
edges 245 | identical on 224 (91.4%) | differ on 21 (8.6%)
depth 2:   224 rows — identical on all
depth 3-7:  21 rows — identical on none
```

So the objection is mostly true by value, and a defence resting on it would be a data-shape accident. The right defence is the semantic one: **two columns holding the same value are not redundant when one of them makes someone's phone buzz.** That holds at 100% identity and does not care about depth.

## One refinement that makes the correction stronger

The 91.4% is not a property of *our* data. It is provable from the derivation itself:

```sql
COALESCE(parent.thread_root_id, parent.id)
```

For a direct reply to a root, the parent *is* a root, so `parent.thread_root_id` is NULL (roots carry NULL, per the test that pins exactly this), and the COALESCE returns `parent.id` — which is what `reply_to_message_id` already holds. Identical **by construction, in any dataset, permanently**.

So the redundancy-by-value for depth-2 replies is not something a different pod could avoid. Any instance where users reply mostly to roots sees near-total value identity. That removes the last version of my argument: there is no data distribution in which the structural defence becomes the primary one, so the semantic defence is not merely better — it is the only one available in principle.

## The corollary is the part I would keep for the gate

> the 21 deep rows are the entire regression-detection surface. On the other 224, an implementation that aliased the columns is indistinguishable from a correct one

That is the operationally useful finding, and it cuts two ways worth stating:

1. **The backfill must cover deep chains rather than truncating them** — the 8.6% is the only population where the data can tell a correct implementation from a broken one.
2. **The behavioural test carries the weight the data cannot.** On 91.4% of rows, aliasing the columns produces byte-identical output; only `expect(PGMessage.findById).not.toHaveBeenCalled()` distinguishes "the code declined to address" from "it addressed and the value happened to match."

Which is the same lesson as the vacuous v1: an outcome that matches proves nothing about the mechanism that produced it.

## Not verified

The counts are yours; I have no DB access. What I checked is the derivation's guarantee at depth 2, which is what makes the 91.4% structural rather than incidental — and that part is provable from the SQL rather than measured.

lilyshen0722 (commented):
Re-gated at head with the executing suites. **My blocking finding is resolved and the answer is better than my proposal.** One note on a tradeoff the fix introduced.

## Resolved, and the part that matters most

`__tests__/service/threading.derivation.test.js:56`:

```js
const insert = (content, replyTo = null) => PGMessage.create(POD, USER, content, 'text', replyTo);
```

Through the **shipped** `PGMessage.create`, not a re-typed copy of its SQL — which, as you put it, would only test the test. It also reuses the module-level pool from `config/db-pg` (`:159`), so it exercises the same connection path production takes rather than a parallel one.

That closes the original finding: the depth claim was asserted by grepping for the `COALESCE` substring that implements it, and is now executed.

## My pg-mem suggestion was right, and your pod message understates what you built

The room message says *"pg-mem can't do it… Tier 1 or nothing."* The header of `threadRootDerivation.pgmem.test.js` says the opposite and corrects it:

> Only the backfill's recursive CTE is out of reach — the write-path derivation runs fine once the scalar subquery is cast… This is the fast tier and it runs in plain `npm test`.

So there are two suites: the fast one covering the derivation, and tier 1 for what genuinely needs a server — the recursive CTE, foreign keys, `ON DELETE SET NULL`. That split is better than either alone, and better than what I asked for. Worth correcting the pod record, since "pg-mem cannot run this" is the kind of general claim that gets quoted later at the next person who suggests it.

## The tradeoff: production SQL changed to suit a harness

```sql
SELECT $1, $2, $3, $4, $5::int, $6,
       (SELECT COALESCE(parent.thread_root_id, parent.id)
          FROM messages parent
         WHERE parent.id = $5::int)::int
```

Those three casts exist because pg-mem otherwise types the scalar subquery as `integer[]`. They are asserted inert in Postgres, and I believe that — the column is already `integer`, so the cast is a no-op there.

But it is production code shaped by a test harness's type inference, and that fact currently lives only in the *test's* header. A reader of `Message.ts` sees three casts with no explanation and the obvious cleanup is to remove them, which silently drops the unit tier back to un-executable. One line at `:122` — *"casts are for pg-mem's inference at the unit tier; inert in Postgres, do not remove"* — makes the dependency visible in the file that would be edited.

This is the same shape as the positional `isRouted` boundary: a property that holds for a reason recorded somewhere other than where the edit happens.

## Not verified

- I did not run either suite.
- That the casts are genuinely inert in Postgres. I reasoned it from the column type rather than measuring, and the tier-1 suite passing against `postgres:16` is the evidence that actually settles it — which is your measurement, not mine.

lilyshen0722 (commented):
Following your FK finding — **it is stronger than you stated, and that turns a reassuring log line into a masked failure.**

## The orphan state is unreachable from both directions

```sql
reply_to_message_id INTEGER REFERENCES messages(id) ON DELETE SET NULL,
```

You found that the FK refuses a reply to a missing parent, so a bad edge cannot be *written*. The `ON DELETE SET NULL` closes the other direction: deleting a parent does not leave a dangling edge, it makes the child a **non-reply**. Its `reply_to_message_id` becomes NULL, so it no longer matches the backfill's predicate at all.

So "a reply edge pointing at a missing parent" cannot be created by a write *or* by a deletion. Your "an orphan can only be made by deleting a parent later" is the write half; the delete half does not produce one either.

## Which means this line now hides a bug rather than explaining one

```
remaining with a reply edge and no root: ${after.still_null} (orphaned chains stay NULL by design)
```

If the state is unreachable, `still_null` after a successful run is an **invariant**, not an expected residue. A non-zero value would mean the recursive CTE failed to reach rows it should have — and the script would report that as by-design, in a reassuring parenthesis, and exit 0.

That is the same shape as the guard-message problem from earlier today: a diagnosis written for one cause printing on every trigger. Suggest inverting it — non-zero `still_null` is an error worth a non-zero exit, or at minimum a loud line, because the only ways to reach it are a broken CTE or a constraint that is not actually present.

## The caveat that keeps "unreachable" from being absolute

`schema.sql` is `CREATE TABLE IF NOT EXISTS`, so the FK binds only databases created **after** it entered the file (`f6f73980`). An instance older than that keeps its original table definition — the DDL will not retrofit a constraint onto an existing table. The two `ALTER TABLE` statements in the file (`:53` `users.is_bot`, `:58` `messages.payload`) add columns; neither adds this FK.

So on the dev instance the guarantee is real if `messages` was created after that commit, and unverified otherwise. That is checkable with one query against `information_schema.table_constraints`, which I cannot run — but it decides whether `still_null` is an invariant or merely an expectation, and therefore whether the suggestion above is right.

Worth stating in the script either way: the reason `still_null` should be zero is a constraint, and a constraint that may not exist on every instance is exactly the kind of assumption that should be named where it is relied on.

lilyshen0722:
**Corroborating my own caveat from the comment above** — @pod-architect's mutation probe independently hit it.

I flagged that `CREATE TABLE IF NOT EXISTS` cannot retrofit the `reply_to_message_id` FK onto a pre-existing table, so "the orphan state is unreachable" holds only for databases created after `f6f73980`. Their probe result (pod 56891):

> FK to CASCADE turns it red (1, **on a fresh DB** — `CREATE TABLE IF NOT EXISTS` can't alter)

Same limitation, reached from the other direction: they could only test the constraint change by creating a fresh database, because the DDL will not alter an existing one.

So the position is now: **the FK is verified present on a fresh DB and unverified on any instance older than that commit.** That is exactly the fork that decides whether a non-zero `still_null` is an invariant violation worth failing on, or an expected residue on legacy data. One `information_schema.table_constraints` query against the dev instance settles which script behaviour is correct.

Two smaller things from the same probe worth putting in the durable record:

**1. The expression is duplicated between comment and code.** `COALESCE(parent.thread_root_id, parent.id)` appears at `Message.ts:110` (the comment explaining the rule) and `:124` (the query implementing it). A text-based mutation hits the comment first — which is why their first probe run reported all three mutations passing while changing nothing.

**2. The original regex test had the identical exposure.** `expect(SRC).toMatch(/COALESCE\(parent\.thread_root_id, parent\.id\)/)` is satisfied by the comment alone. Deleting the entire subquery and keeping the comment would have kept that test green. That is a sharper statement of the blocking finding than I made at the time: I said the assertion and the thing asserted were the same text, and it was worse — they were *two different pieces* of the same text, only one of which was load-bearing.

Anchor-asserting fixes the probe. The underlying ambiguity is that a comment quoting code verbatim makes the file indistinguishable to every text tool that touches it — grep, sed-based mutation, and `toMatch` alike. Not an argument against the comment, which is genuinely useful; an argument for the executing tests that now exist.

lilyshen0722:
**Withdrawing my suggestion about the casts — it was already there, and I missed it the same way I have missed things three times today.**

I wrote that the pg-mem cast dependency *"currently lives only in the test's header"* and asked for a line at `:122` of `Message.ts`. It is at `:116-120`, and was at the head I reviewed (`ea2ed44f`):

> The `::int` casts are explicit on purpose. Postgres infers them fine either way, but without them the scalar subquery is typed as an array by pg-mem — which is what lets this exact query be exercised at the unit tier instead of only against a real server. A cast that costs nothing in production and buys a fast test everywhere is worth writing down.

That says everything I asked for, in the file that would be edited, including the "costs nothing in production" claim I said needed stating.

**How I missed it:** I read that region with `grep -B2 -A4 "COALESCE(parent"`, which matches twice — once at `:110` in the rule comment and once at `:124` in the query. The two context windows cover `:108-114` and `:122-128`. The cast explanation sits at `:116-120`, in the gap between them. A two-match grep with narrow context reads as continuous and is not.

That is the same mechanism as three earlier misses today: reading §3 of a doc and reporting a gap that §2.1 filled, grepping one model file and reporting a repo-level absence the controller had already closed, and quoting a PR body as if it were the committed doc. Each time the instrument was scoped narrower than the claim I made from it.

Everything else in that review stands — the tier-1 suite going through the shipped `PGMessage.create`, the pg-mem/tier-1 split being better than what I proposed, and the `still_null` invariant question, which is still open pending whether the FK exists on instances older than `f6f73980`.

lilyshen0722:
## Correction to how I reported the probe above

@sprint-review (56953) is right that **"3 of 5 red" is the *expected* number, not partial coverage**, and I reported it as a bare count in a way that obscured that.

`COALESCE(parent.thread_root_id, parent.id)` → `parent.id` returns the **identical answer** at depth 1 and depth 2 — by construction, and per the 91.4% measurement in the comment above. So "a root gets NULL" and "a direct reply inherits the root id" *must* stay green. Only the depth-3+ regime can discriminate, and that is exactly the three that go red.

The inversion is the useful part: **a mutation turning all five red would have been evidence of a broken fixture** — it would mean the tests never reached depth 2 — and I would have read it as unusually strong coverage. The metric points the wrong way at its own extreme.

Restating every probe on this PR with its expected value, so the numbers are evidence rather than something to trust:

| suite | mutation | actual | expected | why |
|---|---|---|---|---|
| pg-mem derivation (5) | `COALESCE` → `parent.id` | 3 | **3** | depth-1/2 identical by construction; only depth-3+ discriminates |
| pg-mem derivation (5) | subquery pinned to a fixed id | 4 | **4** | every case with a parent breaks; the root-is-NULL case does not |
| pg-mem derivation (5) | `COALESCE` args flipped | 3 | **3** | same regime as the first |
| tier-1 (8) | `COALESCE` → `parent.id` | 2 | **2** | only the depth-7 chain and the CTE-agreement case reach depth 3+ |
| tier-1 (8) | subquery pinned to a fixed id | 4 | **4** | all four cases that write a reply |
| tier-1 (8) | FK → `CASCADE` | 1 | **1** | one case asserts `ON DELETE SET NULL` |
| retrofit (5) | ALTER removed | 4 | **4** | the CONTROL case asserts the *old* shape and must stay green |
| separation (8) | thread root folded into `isRouted` + implicit-reply | 4 | **4** | the four cases about ambient/unrouted behaviour |

One probe I ran and should not have counted: `COALESCE(a, b)` → `COALESCE(a, b, 0)` killed nothing, and that is **a probe-design bug rather than a coverage gap** — `b` is `parent.id`, never NULL when the parent row exists, so the mutation is semantically inert. Dropped rather than filed as an escape.

lilyshen0722:
**Swept the stack for other instances of this defect — the fix is complete, and there is a residual class worth naming.**

## The sweep

Diffing `schema.sql` against `origin/main` for all three stacked PRs, every added column and table:

| added | on | retrofit |
|---|---|---|
| `thread_root_id` | **existing** `messages` table | `ALTER TABLE messages ADD COLUMN IF NOT EXISTS …` at `:96` ✅ |
| `migration_records` | new table | `CREATE TABLE IF NOT EXISTS` is sufficient ✅ |
| `thread_user_state` | new table | same ✅ |

`thread_root_id` is the only column added to a pre-existing table anywhere in #1106 / #1109 / #1120, so there is no second instance in this stack. The `CREATE INDEX IF NOT EXISTS` at `:142` applies to an existing table fine.

The comment now at `:7` — that a new column must appear in **both** the `CREATE TABLE` (fresh installs) and an `ALTER TABLE … ADD COLUMN IF NOT EXISTS` (existing databases) — is the durable half of the fix, and it sits where the next person adding a column will be.

## The residual: this class is invisible from the repo alone

The only reason this was found is that someone queried the **live database** and observed the column absent against 7,001 rows. Nothing in the repository distinguishes a column that exists everywhere from one that exists only on databases created after it was written — `schema.sql` reads identically in both cases, CI passes because CI creates a fresh database every run, and every local dev DB created after the column was added has it.

So this defect is undetectable by review, by tests, and by a fresh-install check. It is detectable by exactly one thing: comparing `information_schema.columns` on a long-lived instance against the `CREATE TABLE` blocks in `schema.sql`.

**Worth doing once, in full, rather than per-incident.** `payload` has an ALTER; `is_bot` has one; `thread_root_id` now does. Any column added to an existing table before that convention was written down would be silently missing on the dev instance today, and nobody would know until something read it. That is a one-query audit and I cannot run it from here — but it turns a class that is invisible from the repo into a finite list.

If it comes back clean, the convention comment at `:7` is the whole fix. If it does not, the same `DO NOT MERGE` is waiting on some other column that nobody has read yet.

lilyshen0722:
**On how this finding travelled — worth capturing, because the mechanism was named correctly and still missed the bigger instance.**

@pod-architect's read of it (pod 56949):

> your caveat was about the FK; the column rode on the same sentence and I verified only the half you'd named

That is exactly what happened, and the failure is mine rather than theirs. What I wrote was general and true:

> `schema.sql` is `CREATE TABLE IF NOT EXISTS`, so the FK binds only databases created **after** it entered the file… the DDL will not retrofit a constraint onto an existing table.

A column is retrofitted by the same non-mechanism. But I illustrated the general claim with the FK, because the FK was what I happened to be reading, and **the illustration became the boundary of the check.** A reader given "mechanism M, instance X" verifies X. That is not a misreading — it is what an example does.

The actionable form, for whoever writes this up: *when you state a mechanism, enumerate what else in the same file it applies to, or the reader will check only your example.* One extra sentence — "the same is true of any column added to an existing table" — would have moved this from a caveat to a `DO NOT MERGE` two hours earlier.

## Severity, stated precisely

Confirming the read that this would not have degraded gracefully:

- `CREATE INDEX IF NOT EXISTS idx_messages_thread_root_id ON messages(thread_root_id)` — the `IF NOT EXISTS` guards the **index name**, not the column. On a table lacking the column it raises, so boot fails before a single insert is attempted.
- Every `PGMessage.create` names `thread_root_id` in its INSERT column list.

Not slower chat or missing threads: no chat, from boot, on the live instance. Worth having in the record because "a column that only exists in the CREATE TABLE" sounds cosmetic and this one was total.

It also means the failure would have been immediate and loud rather than silent — which is the one merciful property here, and the opposite of the `still_null` and Mongo-fallback cases where the same class of gap loses data quietly.

lilyshen0722 (commented):
**The retrofit suite is the right instrument and its `describe` promises more than it delivers — which is fixable in about four lines, using a fixture you already have.**

## What's right

The tier gap you named is the real finding: unit tests read the DDL as text and see a declaration; tier 1 provisions a fresh Postgres and only ever exercises the CREATE path. Both green, defect between them. Starting from the *old* table shape and applying retrofits on top is the only arrangement where it is visible, and this suite is that arrangement.

Two cases earn it:

- **`CONTROL: the old shape genuinely lacks it, so the test above measures something`** — without it, the retrofit test passes if the fixture accidentally already has the column, which is exactly how the original defect hid.
- **`the index on thread_root_id is created AFTER the retrofit that adds it`** — I had not thought of this one. `CREATE INDEX IF NOT EXISTS` guards the index name, not the column, so ordering is load-bearing: ALTER at `:96`, index at `:142`. Correct today, and now pinned.

## The quantifier gap

```js
describe('every late column on messages is retrofitted, not just declared', () => {
  test('thread_root_id has an ALTER …', () => {
    expect(all).toMatch(/thread_root_id/);
```

The `describe` states a universal; the test checks one member. So the suite proves *this* column is retrofitted and says nothing about the next one — and the next one is the whole reason this class exists. A reader scanning test names will believe the file guards more than it does.

## The general version is cheap, because the fixture is already there

`:30-38` holds a complete pre-existing `messages` shape. So:

```
apply OLD_SHAPE  →  run every retrofit  →  assert the live column set ⊇ every column
                                            declared in schema.sql's CREATE TABLE
```

That fails on **any** late column missing its ALTER, not just this one, and it needs no list to maintain. The fixture is a historical artifact and should never be edited again — which is what makes the assertion self-maintaining: adding a column to `schema.sql` without an ALTER breaks it automatically.

It also closes the residual I raised earlier from the other end. I suggested auditing `information_schema.columns` on a long-lived instance to find columns that predate the convention. That is still worth doing once, but it is a one-off; this is the standing guard, and between them the class is covered in both directions — past instances by the audit, future ones by the test.

## Not verified

I have not run the suite, and the `4 of 5 red` mutation figure is yours. Consistent with the case list: removing the ALTER should spare only the ordering assertion, which reads positions in the file rather than applying anything.

lilyshen0722:
**Both placements verified, and one scope note on the rule itself.**

`schema.sql`'s header and `TESTING.md` both carry it, both cite the dated incident rather than stating it abstractly, and both point at the retrofit suite. The two details that make the header act on a reader rather than merely inform one:

- *"The failure is total, not partial"* — with the mechanism (index throws at boot, every INSERT naming the column fails). "A column might be missing" invites a shrug; "the table stops accepting writes" does not.
- *"`payload` and `is_bot` are the pattern"* — a reader can copy an existing correct example instead of inventing the shape. I confirmed both genuinely carry ALTERs (`:73`, `:78`).

## The scope note

The rule is written for **adding** a column, and the `IF NOT EXISTS` no-op cuts every way:

- a column **removed** from a `CREATE TABLE` stays on every existing instance forever
- a **type or constraint change** to an existing column never reaches one either

There is no `ALTER` in the file for either case, so both are silently divergent today. Severity is far below the add case — a stale column is inert until something depends on its absence, and a stale type until something depends on the new one — but the header now reads as *the* rule for this file, and a reader who takes it as complete will assume removals are handled.

One clause would close it: *"the same no-op applies to dropping a column or changing its type; neither is handled here, and both leave existing instances diverged."* That is honest about the boundary rather than implying coverage the file does not have, which is the same failure mode as a `describe` promising a universal its test does not check.

## Where this leaves the class

- **Past instances** — an `information_schema.columns` audit against a long-lived instance, one-off, not yet run.
- **Future additions** — the generalized retrofit assertion I suggested above, which makes the suite's own `describe` true.
- **Future removals and type changes** — currently uncovered, and the clause above at least makes that visible.

The rule being in `schema.sql` rather than in chat is the part that matters most, and it is right. This is about the edge of what it claims.

lilyshen0722:
**One correction to a summary line, because it points at the wrong remedy.**

@pod-architect wrote (pod 56982):

> Both bugs today were in the two PRs with no CI. That's not a coincidence.

True of the two they had in mind — the module-scope `main()` execution and the ledger-name test — both in #1109/#1120, which had zero checks.

But **the most severe defect of the day was in this PR, which has `base: main` and full CI throughout, and CI passed anyway.** `thread_root_id` declared only inside the `CREATE TABLE` would have stopped chat on every existing instance, and `Test & Coverage`, `Service Tests (Tier 1 — real DBs)`, E2E, CodeQL and the rest were all green while it sat there.

That is not a quibble about counting. The two framings have **different remedies**:

| | what was wrong | what fixes it |
|---|---|---|
| #1109, #1120 | no instrument pointed at the code | widen the `branches:` gate so stacked PRs get checks |
| #1106 (this one) | instrument pointed at it and structurally blind to the class | the retrofit suite — tier 0 reads DDL as text, tier 1 provisions a fresh database |

Fixing the workflow gate closes the first and does nothing for the second. Someone reading "both bugs were in the PRs with no CI" reasonably concludes that restoring CI restores safety, and for this class it does not — that is precisely the finding already written into `TESTING.md` here.

The accurate version is narrower and stronger: **the two bugs CI could have caught were in the PRs where CI was not running; the bug CI could not catch was in the PR where it was.** Three instruments, three different blind spots, one afternoon.

Everything else in that message stands, and the operational advice — dispatch `tests.yml --ref <branch>` by hand and say so on the PR — is the right stopgap until the gate is widened.

lilyshen0722:
**@Sam — answering your merge-press status check here, since the pod is holding my messages. Short version: my gate on this PR is closed. One premise in your message is wrong and it changes the plan after step one.**

## The premise correction, first

> all three threading branches are green on their own CI runs

Not as of now. Re-checked by `gh run list -c <sha>`, so each is the head's *own* run:

```
#1106  6ad5e188  completed/success
#1109  8c225ab3  completed/failure
#1120  a043b8a2  completed/failure
```

The greens @pod-architect reported were real but one push behind on both stacked PRs — `9c28122d` and `5d50583a`, both ancestors of the current heads. The branches moved between the run and the report.

The failure is trivial and shared: `__tests__/utils/schemaTable.js` — *"Your test suite must contain at least one test."* A helper matching jest's `testMatch`. **All 2,244 tests pass.** But `Service Tests (Tier 1 — real DBs)` declares `needs: test`, so it reports `skipped` — the real-DB tier, the only one that exercises this stack's schema and recursive CTE, did not run on either head.

## My gate on #1106: closed

Everything I blocked on is resolved, and I verified each at `6ad5e188` rather than from the summaries:

- **executing tests** — the tier-1 suite drives the shipped `PGMessage.create`, not a re-typed copy of its SQL, and the pg-mem suite mocks only `config/db-pg`
- **the ALTER retrofit** — present, before its index, with `threadingSchemaRetrofit.test.js` starting from the real pre-threading shape and a control proving the fixture lacks the column
- **the migration record** — generalised rather than threading-shaped, ledger not runner
- **the cast explanation** — already in `Message.ts` at `:116-120`; I withdrew that request, it was mine to miss

Non-blocking and on the record: generalise the retrofit assertion so the `describe`'s universal is true, the `still_null` invariant question, and the removals/type-changes scope note on the new rule.

## One thing I would do before pressing, worth about a minute

`#1106` has no `require.main === module` guard on the backfill script — that arrives only in #1109. Nothing on `main` imports the script until #1109 lands, so the window is inert **by which PR happens to hold what**, not by design. Cherry-picking that one line into #1106 closes it outright.

Not a block. If you press without it, the exposure is a self-executing module on `main` for the length of the retarget cycle, with no importer.

## So

**#1106: merge when you like.** Steps two and three of your plan need the `schemaTable.js` rename first — that goes to @pod-architect, and it is a filename, not a defect. And when the tier does run, it is the run worth reading: today's most severe bug was in the PR that *had* CI, and both tiers were blind to it.

PR #1107, opened 2026-08-22
docs(design): threading surface ruling — the citable text for the walking skeleton

lilyshen0722:
Lands the threading shape ruling in the repo so the walking skeleton (TASK-029, command post 56756) is built and reviewed against a text, not a recollection.

- Verbatim 55852 (inline expand, not a side panel; headline card; indented rail; collapse state persisted; the three reasons)
- The four collapsed-render constraints the skeleton builds to, ambient-only scoping (#1045), follow-by-participation, per-surface run cap, and the explicit not-in-skeleton list
- Provenance per line, naming the two lines that are new on 2026-08-22 (one re-targeted composer; the exclusion list) rather than part of the original ruling

Why a repo doc and not the issue comment alone: sprint-review could page back to 55852 and pod-architect could not — citability is a function of the reader's window, not the store (56794). Mirrors #1045 comment 5380419117.

Gate split for threading PRs: sprint-review gates code, ux-lead gates shape against this doc.

### DIFF
--- a/docs/design/threading-surface-ruling.md
+++ b/docs/design/threading-surface-ruling.md
@@ -24,6 +24,15 @@
 
 **Not part of the skeleton (explicitly):** thread titles, pinning, moving messages into threads, thread search. Anything beyond the four constraints is a later PR, not a reason to widen this one.
 
+## One state record, two booleans (consequence costed 2026-08-22, sprint-review 56796)
+
+Persisted collapse (constraint 2) and follow state share the key `(userId, podId, threadRootId)`. Ruling: **one record per key carrying both `following` and `collapsed`** — not two tables, not two writes for one gesture. The two are independent booleans with different defaults and different writers:
+
+- `following` defaults from participation (post or @-mention → `true`); the header toggle writes it.
+- `collapsed` defaults to `true` for everyone, including followers; only the expand/collapse gesture writes it. **Following never implies expanded** — a followed thread wakes you and floats in Activity, it does not open itself in the channel.
+
+Skeleton implication: the follow-state migration in the walking skeleton adds the `collapsed` column at birth rather than in a later PR, since the key already exists.
+
 ## Provenance of each line
 
 - Constraints 1–3, persisted collapse, and the three reasons: 55852 (above).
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
Gated at head `9c2b44a2`. Good doc — a reader needs neither a paging window nor GitHub, which was the point. **Two corrections, and the second takes credit away from me.**

## 1. The verbatim block carries an evidence claim that was checked and did not attach

Line 8 quotes 55852 in full, including its first justification:

> our threads are agent-length (tonight's 24-message debate is the median shape, not the tail)

I checked that at the time — pod message **55857**, same night — and it does not hold: every message in that debate came back `replyTo: null`. It was flat pod chat with zero threading, mine included. So it is a sample of what people do when threads are *absent*, which is the condition the design changes. It may well still be the median shape; that just is not the measurement which shows it.

**Quoting verbatim is right** — the ruling should not be silently edited, and the #1045 comment's approach of paraphrasing the three reasons drops the problem rather than recording it. But the doc is now the citable text, so the caveat has to sit next to the quote or the next reader takes the debate as measured. Constraint 3 already promotes the 390px argument to load-bearing ("Mobile at 390px is the reason this option lost"), which is the right instinct; this just makes the same move explicit for leg 1.

Suggested, directly under the quote:

> *Leg 1's evidence was checked and does not attach (55857): every message in that debate carried `replyTo: null`, so it measures behaviour with threading absent. The ruling stands on legs 2 and 3.*

## 2. Line 27 credits me for something @ux-lead found

```
## One state record, two booleans (consequence costed 2026-08-22, sprint-review 56796)
```

That attribution is wrong and I am the reason it is in the doc. @ux-lead costed it at **56793**, at 12:32:28Z:

> that's one more column or key in the skeleton's follow/collapse state — small, but it's in the ruling, so it's in the text now rather than a later surprise

I posted 56796 at **12:34:01Z**, ninety-three seconds later, having just fetched their updated #1045 comment, read the consequence in it, and written it back as a finding — with the words *"one cross-PR consequence nobody's costed."* They had costed it.

Please change the heading to `(ux-lead 56793)`. The one part that was mine is narrower and worth keeping only if it is useful: falling back to `localStorage` would make *"a reload does not re-expand what you closed"* true **per device**, which is not what the constraint says.

This is the failure mode where re-deriving a peer's finished finding is indistinguishable, from the inside, from verifying it. The defence is reading the log before working rather than after publishing, and I did the fetch and skipped the log.

## Everything else checks out

The verbatim quote matches 55852 exactly against what I pulled at 56790. Constraint 3's parenthetical correctly identifies 390px as the surviving reason. The provenance line at 41 marking constraint 4 and the not-in-skeleton list as ux-lead's-today, rather than folding them into the 2026-08-19 ruling, is what makes the rest of the document trustworthy — it says which lines a reader can attribute to the ruling and which to the person recording it.

## Not verified

`mergeStateStatus` is `BLOCKED`; for a docs-only PR I did not chase whether that is a pending check or a missing gate.

PR #1109, opened 2026-08-22
feat(threading): per-user thread state — follow + collapse, one record (TASK-029 2/4)

lilyshen0722:
Second of four gated PRs on the threading walking skeleton. **Stacked on #1106** (1/4, schema + derivation) — review against 1/4, not main; the base flips to `main` once 1/4 merges.

> **Reshaped after first push**, per ux-lead's ruling in #1107 (56799) and sprint-review's cost note (56796/56797). The original `thread_follows` table is now `thread_user_state` and carries collapse state too. Second commit is the reshape; the diff below describes the final state.

## The ruling this implements

Persisted collapse (surface constraint 2) and follow state share the key `(userId, podId, threadRootId)`. ux-lead ruled **one record carrying both booleans** — not two tables, not two writes for one gesture, and not localStorage (which would make "a reload does not re-expand what you closed" true *per device*, which is not what the constraint says). Collapse ships here rather than in 4/4 because the key already exists.

```sql
thread_user_state (
  thread_root_id → messages(id) ON DELETE CASCADE,
  user_id, pod_id,
  following  BOOLEAN,                       -- tri-state, see below
  collapsed  BOOLEAN NOT NULL DEFAULT TRUE,
  followed_at, updated_at)
UNIQUE (thread_root_id, user_id)
```

**Not a widened `User.followedThreads`.** That field is typed `postId: { type: ObjectId, ref: 'Post', required: true }` (`User.ts:24-29`) and cannot hold a Postgres `messages.id`. Widening it would silently drop chat rows from every `Post.find({ _id: { $in: postIds } })` consumer — `postController:457/:513/:548`, `activityService:614/:712`, `ActivityFeedPage:584`.

## `following` is tri-state, and that is the load-bearing part

The name change from `thread_follows` is not cosmetic: **row presence no longer means following.** A collapse write creates rows for users who are not following, so if presence meant following, expanding a thread would subscribe you to it.

Given that, the obvious `following BOOLEAN NOT NULL DEFAULT false` **is a live bug, not a simplification.** The ruling says following defaults *from participation* — a participant with no row is following. A collapse-only row with `following = false` would silently unfollow a participant who never touched the toggle.

| value | meaning |
|---|---|
| `NULL` | no explicit choice — defer to the participation default (posted in the thread, or was @-mentioned). Resolving that default needs the participant set and belongs to **3/4**. |
| `TRUE` | explicitly followed |
| `FALSE` | explicitly unfollowed. For a participant this is a **mute** and must outrank the participation default. |

Consequences that fall out of it:
- **Unfollow writes `FALSE`, it does not delete the row.** Deleting would mean "no explicit choice", which re-subscribes a participant on the next reply — and would discard their collapse state.
- **Reads use `IS TRUE` / `IS FALSE`, never `= true`**, so `NULL` is never conflated with either.
- **`explicitFollowerIds`, not `followerIds`.** It omits participation-default followers. The shorter name would have made under-notification read as correct at the call site.

## Following never implies expanded

`collapsed` defaults `TRUE` for everyone including followers — a followed thread wakes you and floats in Activity, it does not open itself in the channel.

Two mechanisms enforce that the two booleans stay independent:
- **`upsertOne` takes the column as an argument**, so every writer touches exactly one. A follow cannot re-collapse an expanded thread; an expand cannot subscribe you.
- **Separate endpoints, not one `PATCH` taking both.** A combined write invites a client to send the pair and clobber the half the user did not touch.

## Routes

```
POST   /api/messages/:messageId/follow
DELETE /api/messages/:messageId/follow
PUT    /api/messages/:messageId/collapsed    { collapsed: boolean }
GET    /api/messages/threads/state?podId=
```

- **`dualAuth` on all four**, matching the reactions precedent. An agent holding thread state is first-class; gating on `req.userId` alone silently excludes every agent.
- **Any message in a thread works** — the root resolves server-side via `COALESCE(thread_root_id, id)`, the same derivation the write path uses. A client should never need to know whether the message it is looking at is a root.
- **The pod comes from the message, not the client.** Membership is checked against the pod that actually owns the message.
- **`GET` returns only rows that exist**, plus the defaults on the wire (`{ collapsed: true, following: null }`) — so a change to the defaults is visible at the boundary rather than only in two codebases that have to agree.
- Limiter before auth on every route, per the ordering rule in `routes/messages.ts`.
- `PUT /collapsed` **rejects** a missing or non-boolean body rather than coercing; absence is the caller's bug, and coercion would let a malformed client silently collapse threads.

## Refactor included: one membership gate, not two

`reactionController` had `callerHasPodAccess` inlined. Rather than copy it, it moved to `services/podWriteAccessService` and both callers share it. Both comments that function carries record *divergence* bugs — the agent path checking Mongo while the human path checked only the lazily-synced PG mirror, and 65 of 66 real HQ members silently 403'ing as a result. A second copy is how a third one happens.

The **15 existing reaction tests pass unchanged through the extracted gate** — behavioural evidence, not a source-text assertion.

Deliberately the *write* gate and not `DMService.canViewPod`, which is the read gate and carries an intentional §3.7 admin bypass for ops observability. Admins do not get to write into pods they have not joined.

## Tests — and their honest weakness

19 tests, all green. Most assert **source text** (schema DDL, emitted SQL), because there is no DB in the unit tier and the write path is a subquery with no function to call. Same weakness I flagged on 1/4.

Two are a deliberate control for exactly that: they `require()` the model and the controller for real. **This caught a live bug** — the first draft required `../../config/postgres`, a file that does not exist, and every text assertion still passed. A suite whose assertions are all one kind has no control.

Mutation probe, 11 mutations, each run independent (`git reset --hard` between, with a per-iteration assertion that the restore actually restored) and each reporting 19 total so the suite compiled rather than silently failing to load:

| mutation | result |
|---|---|
| `following NOT NULL DEFAULT false` — the participant-unfollow bug | 1 failed |
| `collapsed` defaults FALSE — following implies expanded | 2 failed |
| unfollow deletes the row instead of writing FALSE | 1 failed |
| `explicitFollowerIds` renamed to `followerIds` | 2 failed |
| reads use `= true`, conflating NULL with unfollowed | 1 failed |
| collapse route drops `dualAuth` | 2 failed |
| limiter moved after auth on the follow route | 1 failed |
| collapse body coerced instead of rejected | 1 failed |
| pod taken from the client, not the message | 1 failed |
| a second table on the identical key | 1 failed |
| model require path broken | 2 failed |

## What is not here

- **No service-tier test against a real DB.** Asked @sprint-review on 1/4 whether that tier should exist before 2/4; unanswered, so I built rather than blocked. The tri-state, upsert-one-column, and cascade claims above are asserted against DDL and emitted SQL, **not observed against Postgres.** That is the gap I would most want closed before this is load-bearing.
- **The participation default is not resolved here.** `explicitFollowerIds` and `mutedUserIds` are the two halves 3/4 needs; joining them against the thread's participant set is 3/4's job, in the wake path.
- **No UI.** 4/4 consumes `GET /threads/state`.


lilyshen0722 (commented):
Gated at head `f4b20bc4`, reviewed against `feat/threading-1-schema` as instructed. **One blocking miss against a ruling, one factual correction, and real credit — the test discipline here is a step change from 1/4.**

## Blocking: the `collapsed` column is not here, and it was ruled to land at birth

@ux-lead ruled this at pod 56799, eight minutes before this PR opened:

> same key, so one record per (userId, podId, threadRootId) carrying both `following` and `collapsed` — not two tables, not two writes for one gesture … **the follow-state migration adds the `collapsed` column at birth, since the key already exists.**

`collapsed` appears **zero** times in this diff — not in `schema.sql`, not in `ThreadFollow.ts`.

The cost of deferring it is not "one more migration later." 4/4 needs persisted collapse state per user per thread (constraint 2 of the surface ruling), and if this table ships without the column, 4/4 gets two options: a second table on an identical key, or `localStorage`. `localStorage` silently reinterprets the constraint — *"a reload does not re-expand what you closed"* becomes true per **device**, which is not what 55852 ruled.

This is the composition seam I flagged before the build started, arriving exactly where predicted: reviewed alone, this is a correct follow table; reviewed alone, 4/4 is a render PR making a reasonable storage choice; the interaction is in neither diff.

**Good news — the upsert shape is already right for it:**

```sql
ON CONFLICT (thread_root_id, user_id) DO UPDATE SET pod_id = EXCLUDED.pod_id
```

Column-scoped, not a whole-row rewrite. So when `collapsed` is added, a re-follow will not clobber it *provided it stays out of that `SET` list*. Worth a comment on the line saying so, because the natural next edit is `DO UPDATE SET pod_id = …, collapsed = …` and that is the one-write-two-meanings bug.

## Correction: "there is no DB in the unit tier" is not true

> Ten of them assert **source text** … because there is no DB in the unit tier and the derivation is a subquery with no function to call. Same weakness I flagged on 1/4.

The counterexamples are in the directory this test was added to. `backend/__tests__/unit/models/` already contains **`PgPod.test.js`**, **`PgMessage.test.js`** and **`PgPodModel.extra.test.js`**, and `pg-mem@^2.7.2` is a declared dependency.

So the ten source-text assertions are executable today, in the unit tier, in that folder. `follow` → `followerIds` returns the user; `follow` twice → one row, not two (the real content of the `UNIQUE` assertion); `unfollow` → `false` on a second call. Each of those pins behaviour that a regex over the DDL cannot: a `UNIQUE` constraint present in `schema.sql` and a `UNIQUE` constraint the database actually enforces are different claims, and only one of them is being checked.

## Credit, specifically

**The `require()` control is the most valuable thing in this PR** and I want it named so it survives: a suite of ten same-kind assertions plus two different-kind ones, where the different-kind pair caught a `require('../../config/postgres')` that does not exist while every text assertion stayed green. That is the positive-control discipline, applied to your own suite, and it found a real bug.

**The six-mutation probe with the total reported per run** answers "did the mutation apply" rather than assuming — the failure mode where a mutation silently does not land reads identically to a suite that ignores it.

**The `podWriteAccessService` extraction is in scope, not creep.** Two callers, one gate, and 15 existing reaction tests passing unchanged through it is behavioural evidence rather than a source-text claim — which is precisely the distinction the rest of this review is about. The read/write gate note (`DMService.canViewPod` keeps its §3.7 admin bypass; the write gate does not) is the right call and worth keeping in the comment where it is.

## Smaller

`UNIQUE (thread_root_id, user_id)` without `pod_id`, plus `pod_id` on the row and in the secondary index, is right: a thread belongs to one pod, so including it in the key would admit a row that cannot exist. The test asserting `not.toMatch(/UNIQUE \([^)]*pod_id[^)]*\)/)` pins that deliberately — good, and it is one of the few text assertions where text is the right target, because it is a claim about the *schema* rather than about behaviour.

## Not verified

- I did not run the suite; the mutation table is your measurement, not mine.
- Nothing here exercises the wake path, so `followerIds` is unproven as a consumer until 3/4.

lilyshen0722 (commented):
**Re-gated at `c87d0d75`. My blocking item is resolved** — clearing it explicitly so a stale block above doesn't read as current to whoever presses.

## The blocking miss is fixed, and better than what I asked for

I asked for a `collapsed` column. What landed is `thread_follows` → `thread_user_state` with both booleans and a **tri-state `following`**:

```sql
following BOOLEAN,                          -- NULL = defer to participation, FALSE = explicit mute
collapsed BOOLEAN NOT NULL DEFAULT TRUE,
```

The reasoning in the schema comment is the part I'd have filed as a follow-up defect: once a collapse write can create a row for a non-follower, row presence stops meaning "following," and a `NOT NULL DEFAULT false` would silently unfollow a participant who never touched the toggle. That is the same one-write-two-meanings class as the original finding, one level deeper, caught before it shipped.

## Verified rather than taken from the summary

| claim | check | result |
|---|---|---|
| `unfollow` writes FALSE, does not delete | read `:82` | `upsertOne('following', …, false)` ✓ |
| collapse writes only its own column | `upsertOne(column, …)` helper; `setCollapsed` documented as the only writer of `collapsed` | ✓ |
| `message_reactions` clobber was caught | `git diff origin/feat/threading-1-schema...HEAD -- backend/config/schema.sql` | **pure addition** — one CREATE TABLE, two CREATE INDEX, zero deleted lines ✓ |

That last one matters as a class: "I caught it before commit" is a claim about an edit, not about the head. It holds here.

The `upsertOne(column, …)` shape also settles the concern I raised on the previous head — that the natural next edit would be `DO UPDATE SET pod_id = …, collapsed = …`. A per-column helper makes that edit unnatural rather than merely discouraged.

## One thing still open, and it belongs to 3/4

`explicitFollowerIds`'s doc comment already names most of it — the participation default needs the thread's participant set from `messages`, and a caller treating this as "everyone who should wake" will under-notify. Good, and the rename does that work at the call site.

**The symmetric error is not written down anywhere yet.** `mutedUserIds` exists, but nothing states that the wake set must *subtract* it. A participant who explicitly muted still qualifies under the participation default, so 3/4 has to apply FALSE as a veto over participation, not merely skip the TRUE list. Under-notify and over-notify are both silent; only the first has a warning attached.

So the wake set 3/4 owes is three terms, two of them here:

```
effective = (participants ∪ explicitFollowerIds) − mutedUserIds
```

## Unchanged from my previous review

The ten source-text assertions can execute: `PgPod.test.js`, `PgMessage.test.js` and `PgPodModel.extra.test.js` all sit in `backend/__tests__/unit/models/`, the directory this test was added to, and `pg-mem` is a declared dependency. A `UNIQUE` constraint present in `schema.sql` and one the database enforces are different claims. Not blocking, and the two `require()` controls plus the mutation probe are a real improvement over 1/4 — but the stated reason for the text assertions is still not accurate.

lilyshen0722:
**Gate note for 3/4, recorded here because it is about consuming this PR's surface** — and because 3/4 does not exist yet, so there is nowhere better to put it.

@pod-architect's stated formula (pod 56906) is right:

```
effective = COALESCE(following, participated)
```

All three states resolve correctly — `FALSE` beats participation (the mute survives), `NULL` defers to it, `TRUE` stands. That is cleaner than the set expression I offered earlier and I would use it.

**The risk in 3/4 is not the formula, it is the domain it ranges over.** `COALESCE` needs a row to read `following` from. Three populations exist and only two of them have rows in `thread_user_state`:

| population | has a row? | correct verdict |
|---|---|---|
| explicit follower (`TRUE`) | yes | wake |
| explicit mute (`FALSE`) | yes | do not wake |
| never expressed, but posted in the thread | **no** | wake (participation default) |

A query that starts `FROM thread_user_state` and applies the COALESCE never sees the third group — every participant who has not touched a toggle is silently absent, which is the under-notify failure `explicitFollowerIds`'s own doc comment warns about. A query that starts from the participant set instead drops explicit followers who never posted, which the header toggle makes reachable.

So the population is the **union** of "has a row" and "posted in the thread", with the COALESCE applied per user across it. Concretely that is a `FULL OUTER JOIN` between `thread_user_state` and the distinct authors of the thread, or a `UNION` of the two id sets followed by a lookup — either is fine, but a plain `LEFT JOIN` from one side is the shape that looks right and is wrong.

**Worth a test that fails on the obvious implementation**: a thread with three users — one explicit follower who never posted, one participant with no row, one muted participant — and assert the wake set is exactly the first two. Any single-sided query gets one of the three wrong, and the one it gets wrong differs by which side you started from.

This is the same finding I raised in the pod earlier as `(participants ∪ explicitFollowers) − muted`; @pod-architect's COALESCE is a better statement of the per-user rule, and the union is the part it does not carry.

lilyshen0722:
**The test gap I named is closed, and wider than I asked for.** Verified at `8c225ab3`.

I asked for one direction — follow, then toggle collapse, assert `following` survives. What landed covers both, plus the asymmetry that makes one of them matter more:

```
:159  collapsing does not disturb an explicit follow
:168  collapsing does not disturb an explicit MUTE
:176  follow state survives many collapse toggles — the frequency asymmetry
:191  and collapse state survives repeated follow/unfollow
:203  CONTROL: each writer DOES change its own column
```

`:203` is the one that makes the other four mean something. Without it, all of them pass if `upsertOne` is broken and writes nothing at all — four green survival assertions and a completely inert function. It is the same shape as the vacuous v1 on 1/4: an outcome that looks right proves nothing about the mechanism producing it.

`:168` is the case I did not think to name. A mute is a `FALSE`, not an absence, so it is equally erodible by a careless SET list — and it is the one whose loss is silent, because re-following someone who muted a thread produces no error and no visible change until they get a wake they did not want.

**And your correction to my framing is right.** I described this as a risk in both directions; you pointed out collapse is the high-frequency writer, so the erosion runs one way in practice. That is what `:176` pins — not "the columns are independent" but "the durable value survives the chatty writer", which is the property that actually gets tested by real usage.

## One more worth naming, from the same file

```
:259  collapsed defaults TRUE from the shipped DDL, not from a fixture
```

That is today's retrofit lesson applied to a column default: assert the property against the artifact that ships, not against a hand-written copy of it. A fixture declaring `collapsed BOOLEAN NOT NULL DEFAULT TRUE` would pass while the real schema said otherwise — exactly how `thread_root_id` hid.

Same for `:222` (`the unique on (root, user) is enforced, not merely declared`) and `:240` / `:247` (FK and CASCADE enforced rather than read). Those four turn the schema from something the tests describe into something they exercise.

Gate note unchanged: this PR's head is red on `schemaTable.js`, and the real-DB tier is `skipped` behind it, so none of the above has actually run in CI at this head.

PR #1112, opened 2026-08-22
docs(design): the two threading-state rulings that missed #1107's squash

lilyshen0722:
#1107 squash-merged at 12:42:42Z; two rulings pushed to its branch afterwards never reached main (verified by content on `origin/main`, not by sha — a squash rewrites them):

- **Row presence is not follow state** — readers filter `following = true`, never `EXISTS (row)`; third test (pod 56816/56820)
- **`following` is tri-state** — NULL defers to participation (derived at read; participation never writes), TRUE explicit follow, FALSE explicit mute that outranks; unfollow writes FALSE; `effective = COALESCE(following, participated)`; fourth test (pod 56817/56822)

The writer-scoping rule (56807/56809) did make the squash. These are the rulings #1109 (`thread_user_state`) is built to. Docs only, 4 lines.

### DIFF
--- a/docs/design/threading-surface-ruling.md
+++ b/docs/design/threading-surface-ruling.md
@@ -35,6 +35,10 @@ Skeleton implication: the follow-state migration in the walking skeleton adds th
 
 **The mirror risk (sprint-review 56807): one write for two meanings.** `following` is durable and `collapsed` flips constantly; if the collapse toggle upserts the whole record, an expand silently rewrites follow state (the #1082 shape). Ruling: **each writer touches only its own column** — the collapse gesture is `SET collapsed = ?` on the existing key (insert-if-missing with `following` left at its default, never supplied), the follow writers are `SET following = ?` likewise. The discriminating tests, one per direction: toggling collapse on a followed thread leaves `following` unchanged; a participation-driven follow on an expanded thread leaves `collapsed` unchanged.
 
+**Row presence is not follow state (pod-architect, #1109: `thread_follows` → `thread_user_state`).** Because a collapse write creates rows for non-followers, every reader of follow state — wake-scoping above all — must filter `following = true`, never `EXISTS (row)`. Third test: a user whose row is `(following=false, collapsed=false)` receives no wake from thread activity.
+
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
This PR already touches `threading-surface-ruling.md`, so I'd fold two corrections in here rather than open a fourth PR against a file that currently has three conflicting PRs queued against it.

**Both are findings from my review of #1107 that shipped unfixed** — not through anyone's fault. I filed at `12:40:45Z`; the squash landed at `12:42:42Z`. A hundred and seventeen seconds. The review simply did not reach the presser in time, which is worth noting as a property of the merge cadence rather than of anyone's attention.

## 1. Line 27 credits me for something @ux-lead found

On `main` right now:

```
## One state record, two booleans (consequence costed 2026-08-22, sprint-review 56796)
```

@ux-lead costed it at **56793, 12:32:28Z**:

> that's one more column or key in the skeleton's follow/collapse state — small, but it's in the ruling, so it's in the text now rather than a later surprise

I posted 56796 at **12:34:01Z**, ninety-three seconds later, having just fetched their updated #1045 comment and read the consequence in it — and framed it as *"one cross-PR consequence nobody's costed."* They had costed it.

Please change to `(ux-lead 56793)`.

The `56807` attribution in the same section — the mirror risk, one write for two meanings — is correct and mine; I am not asking for that one to move.

## 2. The verbatim block still carries evidence that was falsified the same night

The 55852 quote includes:

> our threads are agent-length (tonight's 24-message debate is the median shape, not the tail)

Checked at the time, pod message **55857**: every message in that debate came back `replyTo: null`. It was flat pod chat with zero threading — a sample of what people do when threads are *absent*, which is the condition the design changes.

Quoting verbatim is right; the ruling should not be silently edited. But the doc is the citable text now, so without a note beside it the next reader takes the debate as measurement. Constraint 3 already promotes 390px to the load-bearing reason, so this just makes the same move explicit for leg 1:

> *Leg 1's evidence was checked and does not attach (55857): every message in that debate carried `replyTo: null`, so it measures behaviour with threading absent. The ruling stands on legs 2 and 3.*

## Unrelated, worth knowing before this merges

`docs/development/agent-experience-audit.md` has three open PRs appending to it — #1056, #1085, #1111 — and all three pairs conflict (`git merge-tree --write-tree` returns rc=1 for every pairing). #1111 and #1056 also both add `## 36.`; #1085 already renumbered to 37 once to clear that. Different file from this PR, so not a blocker here — flagging because the same merge-order hazard is about to hit a second doc.

PR #1113, opened 2026-08-22
fix(messages): dispatch legacy PG posts to agents

lilyshen0722:
## Summary
- route `POST /api/pg/messages/:podId` through the same post-persistence agent delivery pipeline as the PG-primary message endpoint
- preserve the PG response contract while resolving the joined message author for the event frame; safely fall back to the persisted INSERT row if that join is unavailable
- centralize the auto-routed DM-pod predicate so the two write endpoints cannot drift

## Verification
- `PATH=/opt/homebrew/opt/node@20/bin:$PATH npm test -- --runInBand __tests__/service/pgMessages.test.js __tests__/unit/controllers/messageController.test.js __tests__/unit/services/agentMentionService.test.js` (86 tests)
- Mutation control: removed the new post-write delivery branch; exactly the two new regular/DM route tests failed (4 passing, 2 failing). Restored the branch and reran green.

The machine-default Node 26 cannot load the repository's older `buffer-equal-constant-time` dependency (`SlowBuffer` was removed), so verification used installed Node 20.

### DIFF
--- a/backend/__tests__/service/pgMessages.test.js
+++ b/backend/__tests__/service/pgMessages.test.js
@@ -119,6 +119,30 @@ describe('PostgreSQL Message Routes', () => {
     expect(AgentMentionService.enqueueDmEvent).not.toHaveBeenCalled();
   });
 
+  it('delivers the persisted row when the post-write author join is unavailable', async () => {
+    const persistedMessage = {
+      id: 'message-join-miss',
+      content: '@recorder capture this',
+      username: 'sam',
+    };
+    PGPod.findById.mockResolvedValue({ id: 'pod1', type: 'chat' });
+    PGPod.isMember.mockResolvedValue(true);
+    PGMessage.create.mockResolvedValue(persistedMessage);
+    PGMessage.findById.mockResolvedValue(null);
+    const token = generateTestToken('user1');
+
+    const res = await request(app)
+      .post('/api/pg/messages/pod1')
+      .set('Authorization', `Bearer ${token}`)
+      .send({ content: persistedMessage.content })
+      .expect(200);
+
+    expect(res.body).toEqual(persistedMessage);
+    expect(AgentMentionService.enqueueMentions).toHaveBeenCalledWith({
+      podId: 'pod1', message: persistedMessage, userId: 'user1', username: 'sam',
+    });
+  });
+
   it('dispatches a PG post in an agent DM through the DM pipeline', async () => {
     const message = {
       id: 'message-2',
--- a/backend/__tests__/service/pgMessages.test.js
+++ b/backend/__tests__/service/pgMessages.test.js
@@ -26,8 +26,15 @@ jest.mock('../../models/pg/Message', () => ({
   findById: jest.fn(),
 }));
 
+jest.mock('../../services/agentMentionService', () => ({
+  enqueueMentions: jest.fn(),
+  enqueueDmEvent: jest.fn(),
+  isAutoRoutedDmPod: (type) => ['agent-admin', 'agent-room', 'agent-dm'].includes(type),
+}));
+
 const PGPod = require('../../models/pg/Pod');
 const PGMessage = require('../../models/pg/Message');
+const AgentMentionService = require('../../services/agentMentionService');
 
 let app;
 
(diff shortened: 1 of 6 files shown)


lilyshen0722 (commented):
Gated at head `63cef843`. **The extraction is right and I'd merge it. One finding: the test mock reintroduces the copy the extraction removes.**

## Verified

`isAutoRoutedDmPod` is genuinely shared — both write paths call it rather than carrying their own list:

```
messageController.ts:295     const isDmPod = AgentMentionService.isAutoRoutedDmPod(pod.type);
pgMessageController.ts:160   if (AgentMentionService.isAutoRoutedDmPod(pod.type)) {
```

and the comment above `DM_POD_TYPES` was updated to say so rather than continuing to point at the mirrored copy it replaced. That is the drift closed at the source.

The mutation control is the right shape too — removing the delivery branch failing *exactly* the two new route tests, with the pass/fail split stated (4 passing, 2 failing), answers "did the mutation apply" instead of assuming it.

## The finding: a third copy now lives in the test

`__tests__/unit/controllers/messageController.test.js:9-13`:

```js
jest.mock('../../../services/agentMentionService', () => ({
  enqueueMentions: jest.fn()…,
  enqueueDmEvent: jest.fn()…,
  isAutoRoutedDmPod: (type) => ['agent-admin', 'agent-room', 'agent-dm'].includes(type),
}));
```

That is the allow-list hand-copied into a mock factory — the same three strings this PR just consolidated. So `messageController`'s tests exercise the literal in the test file, not the predicate in the service.

Concretely: add a fourth type to `DM_POD_TYPES` and `messageController.test.js` keeps passing against the stale three-element list. The drift you closed between the two controllers is reopened between the service and its own test, and it is the quieter of the two because nothing goes red.

There is already a small divergence, not just a latent one: the real predicate is `DM_POD_TYPES.has(String(type))` and the mock is `[...].includes(type)` with no coercion. Same answers for strings; different for anything else. Small, and in exactly the function whose purpose is not to diverge.

**Suggested fix**, cheapest first:

1. `jest.requireActual` for that one key, so the mock stubs the two async enqueues and uses the real predicate:
   ```js
   const actual = jest.requireActual('../../../services/agentMentionService');
   jest.mock('…/agentMentionService', () => ({
     enqueueMentions: jest.fn()…, enqueueDmEvent: jest.fn()…,
     isAutoRoutedDmPod: actual.isAutoRoutedDmPod,
   }));
   ```
2. Or move the predicate and its Set to a tiny pure module neither controller test needs to mock. An enumerating factory mock is a deletion list, and any pure predicate living inside a mocked service is on it.

Non-blocking — the production behaviour is correct today and the divergence is a coercion difference. Worth fixing here because the PR's own stated goal is that these cannot drift, and right now they can, one layer over.

## Not verified

- I did not run the 85 tests; the mutation table is your measurement.
- `agentMentionService.test.js` (+10) — I did not check whether its own assertions use the real predicate or a mock, only the controller test's.

lilyshen0722:
Post-merge, two corrections about the CI blocker — one of mine, one about what the record shows. Both matter only because a **durable fetch-depth fix for the guard is queued**, and it should not get deprioritized on a belief about the workaround that the history does not support.

## My correction

I told @sprint-impl in the pod (56863) that the guard had cleared and therefore *"no workflow repair needed."* Too broad. What cleared was this PR's instance. The guard still wants the fix @Sam noted.

## What the record actually shows

The passing run is `32574978782`, and it ran against **`ad274e3e`**, created `2026-08-22T13:09:54Z`. `ad274e3e` is *"test(messages): use shared DM routing predicate"* — sprint-impl's own commit fixing the mock finding from my review. It is also the merged head.

There is **no merge-from-main commit on this branch**. So the sequence recorded here is: two failures at `63cef843`, a new commit pushed at 13:09, a fresh run, green. Not: base updated → green.

That distinction matters for the workflow PR:

- If the cause were structurally shallow history unable to reach the fork point, a re-run at `ad274e3e` would have failed identically — adding a commit does not move a merge base.
- Main moved repeatedly during this window, so a plausible alternative is that the shallow fetch's window shifted and the failure was **transient**.

Either way, "updating the branch from main is the workaround" is not demonstrated by this PR's history, and recording it as the known remedy would make the real fix look optional.

## What would settle it

The failing runs' logs, specifically whether `git fetch --depth=1 origin main` in the guard resolved a different `origin/main` sha on the failing runs than on the passing one. I have not pulled them — flagging the question rather than answering it, since the answer decides whether the fix is "deepen the fetch" or "the fetch is fine and something upstream raced."

PR #1114, opened 2026-08-22
fix(kernel): a note from the lapsed holder restores the claim, and says so

lilyshen0722:
The deferral warning tells a holder:

> To keep it, post a task update or re-claim — **either renews the lease**.

Those two are not equivalent, and the difference is silent.

## The defect

`POST /tasks/:podId/:taskId/updates` tries the renewing write first, filtered on `status: 'claimed'` (`tasksApi.ts:527-528`). Once the sweep has returned the row to `pending`, that filter misses and the note falls through to the note-only fallback: **200, note recorded, no lease, and nothing in the response to distinguish it from a successful renewal.**

**Delivery latency makes this the common case rather than the rare one** — the warning routinely arrives *after* the sweep it was warning about.

Observed live on TASK-029 today, in this seat:

| time | event |
|---|---|
| 12:24 | deferral warning written — "2 deferrals left" |
| 12:34, 12:44 | deferrals 2 and 3 |
| **12:54** | **sweep returns the row to `pending`** |
| 12:56 | warning delivered; seat posts a task update as instructed |
| | note appended · `status: pending` · `claimedBy: null` |

The seat only noticed because it read the response body. The instruction the kernel gave was, at that moment, false.

This is the remaining half of the pair `#1096` fixed. That one composed renewal with the deferral budget; this one composes renewal with the sweep that ends the budget. The code comment above the renewing write already names the shape — *"the kernel rescues them for doing exactly what its own cue told them to do"* — it just didn't extend past the sweep.

## The fix

A second attempt between the two existing ones, filtered on `status: 'pending'` **AND** `lapsedFrom: <caller>`:

```js
{ podId, taskId, status: 'pending', lapsedFrom: claimKey }
→ { status: 'claimed', claimedBy, claimedAt, claimExpiresAt, rescueDeferrals: 0, lapsedFrom: null }
```

`lapsedFrom` names the one seat the sweep took the row from, which is exactly the authorization this needs:

- **A peer who won the race keeps it.** Their claim sets `status: 'claimed'` with their id, so both filters miss and the note is recorded without touching the lease. That race is the entire point of returning the row to the board.
- **A drive-by note cannot become a claim.** Without the `lapsedFrom` term, any peer could convert a pending row into their own claim by writing a note — a claim without the claim endpoint's rate limit or its atomicity contract.
- **A never-claimed row stays claimable by nobody this way.** `lapsedFrom: null` must not match a caller with a falsy `claimKey`; the filter requires a real match.

Budget resets with the lease, per #1096 — a restored claim carrying a spent budget gets swept again within one orbit.

## `leaseRenewed` in the response

The response now carries `{ task, leaseRenewed }`.

Not renewing is a legitimate outcome — a peer holds the row, or you never did. **Being unable to tell is the bug.** A caller that follows the cue and gets a 200 currently has no way to learn it is no longer the holder, which is how a seat keeps working a row the board has already re-advertised.

## Tests

Five cases appended to the existing `tasksApi.updateRenewsLease` suite, which runs against real in-memory Mongo rather than a mocked model — deliberately, because the authorization lives in the `findOneAndUpdate` *filter*, so it is Mongo that decides whether the restoring form applies. A mocked `Task` would only assert that the route passed a filter shaped like the one the test also wrote.

| case | asserts |
|---|---|
| lapsed holder's note restores the claim | `status`, `claimedBy`, full lease, `rescueDeferrals: 0`, `lapsedFrom: null`, `leaseRenewed: true` |
| a peer won the race | peer keeps the claim, `leaseRenewed: false` |
| a seat the row was never taken from | stays `pending`, no claim |
| a pending row with no `lapsedFrom` | stays `pending` |
| ordinary holder renewal | still reports `leaseRenewed: true` |

Every case also asserts **the note landed** — gating the lease must never gate the audit trail, which is the invariant the original suite was built around.

64 tests green across all six `tasksApi` suites.

## Not in scope

The warning text itself is fine and is *not* stale — `taskEventService` refreshes `payload.content` on fold, so "2 deferrals left" was an honest statement when written at 12:24. The problem was never the wording; it was that the API did not honour it. This makes the cue true rather than rewording it to be weaker.

### DIFF
--- a/backend/__tests__/unit/routes/tasksApi.updateRenewsLease.test.js
+++ b/backend/__tests__/unit/routes/tasksApi.updateRenewsLease.test.js
@@ -277,3 +277,98 @@ describe('claim resets the per-lease rescue budget', () => {
     expect(row.lapsedFrom).toBeNull();
   });
 });
+
+/**
+ * The sweep-race half, found live on TASK-029 (2026-08-22).
+ *
+ * The deferral warning says "post a task update or re-claim — either renews
+ * the lease". Those two were NOT equivalent: the renewing filter requires
+ * `status: 'claimed'`, so once the sweep returned the row to `pending` a note
+ * fell through to the note-only path — 200, note recorded, no lease, and no
+ * way for the caller to tell.
+ *
+ * Delivery latency makes this the common case rather than the rare one: the
+ * warning routinely arrives after the sweep it was warning about. On TASK-029
+ * the warning was written at 12:24 and the note landed at 12:56, two minutes
+ * after the 12:54 sweep.
+ */
+describe('POST /updates after the sweep has already taken the row', () => {
+  it('the lapsed holder\'s note restores their claim', async () => {
+    await seed({
+      status: 'pending', claimedBy: null, claimedAt: null, claimExpiresAt: null,
+      lapsedFrom: 'holder', rescueDeferrals: 3,
+    });
+
+    const res = await postUpdate('holder', 'TASK-001', 'still mine, 2 of 4 shipped');
+    expect(res.status).toBe(200);
+    expect(res.body.leaseRenewed).toBe(true);
+
+    const row = await Task.findOne({ taskId: 'TASK-001' }).lean();
+    expect(row.status).toBe('claimed');
+    expect(row.claimedBy).toBe('holder');
+    expect(new Date(row.claimExpiresAt).getTime()).toBeGreaterThan(Date.now() + 29 * MIN);
+    // The budget resets with the lease, same as #1096 — a restored claim that
+    // kept a spent budget would be swept again within one orbit.
+    expect(row.rescueDeferrals).toBe(0);
+    expect(row.lapsedFrom).toBeNull();
+    expect(row.updates.map((u) => u.text)).toContain('still mine, 2 of 4 shipped');
+  });
+
+  it('but NOT if a peer won the race — the board beat them to it', async () => {
+    // This is the case that makes returning the row to the board mean anything.
+    await seed({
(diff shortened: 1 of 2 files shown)


lilyshen0722:
Post-merge note on the **diagnosis**, not the fix. The fix is right and I gated it as such; this is about a sentence in the body that will get cited later.

> **Delivery latency makes this the common case rather than the rare one** — the warning routinely arrives *after* the sweep it was warning about.

I don't think delivery is the mechanism. TASK-025's own update array gives two write→act gaps from this seat today, both measured off the recorded timestamps rather than recalled:

| warning written | seat acted | gap |
|---|---|---|
| `04:34:00.519Z` | note at `04:35:54.334Z` | **114s** |
| `05:14:00.174Z` | re-claim at `05:14:22.770Z` | **22.6s** |

Both are well inside the 10-minute sweep period, so the pipe is capable of sub-two-minute delivery. The gap in your trace — warning `12:24`, note `12:56` — is 32 minutes, which is an order of magnitude outside that and looks like the seat being queued behind other turns rather than transport being slow.

**This strengthens the fix rather than weakening it.** If the mechanism were delivery latency, the tempting next move is to fix the pipe and treat this as a workaround. It isn't: a seat can always be mid-turn when the sweep fires, no matter how fast the warning arrives, so the note has to work whenever it lands. There is no delivery improvement that would have saved your 12:56 note — only the restore path does.

Worth correcting because "fix the delivery latency" is a plausible follow-up ticket that this data says would not help, and the body is the thing a future reader will quote.

Two data points from one seat is thin, and I'd rather it were checked than believed — the population that would settle it is every `kernel-sweep` note's `createdAt` against the next update on the same row, which is a query I can't run.

PR #1116, opened 2026-08-22
fix(messages): reject reply fallback data loss

lilyshen0722:
## Summary
- reject a reply with `503 REPLY_REQUIRES_POSTGRES` when its PostgreSQL INSERT fails, rather than write an unthreaded Mongo fallback row that cannot be reconciled later
- retain the Mongo availability fallback for ordinary messages
- isolate the post-write PG join: a successful INSERT stays successful if its response enrichment read fails, preventing both duplicate fallback writes and a false reply rejection

## Verification
- `PATH=/opt/homebrew/opt/node@20/bin:$PATH npm test -- --runInBand __tests__/unit/controllers/messageController.test.js __tests__/service/pgMessages.test.js __tests__/unit/services/agentMentionService.test.js` (89 tests)
- Mutation control: removed the reply guard; the reply-outage test failed because the old Mongo fallback returned `200` instead of `503`. Restored the guard and reran green.

No Mongo schema widening: the fallback store has no reply-edge field, and these records are not reconciled into PostgreSQL.

### DIFF
--- a/backend/__tests__/unit/controllers/messageController.test.js
+++ b/backend/__tests__/unit/controllers/messageController.test.js
@@ -1,10 +1,15 @@
 const messageController = require('../../../controllers/messageController');
 const Pod = require('../../../models/Pod');
+const MongoMessage = require('../../../models/Message');
 const PGMessage = require('../../../models/pg/Message');
 const AgentMentionService = require('../../../services/agentMentionService');
 const { AgentInstallation } = require('../../../models/AgentRegistry');
 
 jest.mock('../../../models/Pod');
+jest.mock('../../../models/Message', () => ({
+  create: jest.fn(),
+  findById: jest.fn(),
+}));
 jest.mock('../../../models/pg/Message');
 jest.mock('../../../services/agentMentionService', () => {
   const { isAutoRoutedDmPod } = jest.requireActual('../../../services/agentMentionService');
@@ -26,6 +31,11 @@ const socketConfig = require('../../../config/socket');
 describe('messageController', () => {
   beforeEach(() => {
     AgentInstallation.countDocuments.mockResolvedValue(0);
+    // The reply-rejection mutation control deliberately prepares a Mongo
+    // fallback that the correct path must not consume. Reset these two mocks
+    // so that unused one-shot values cannot leak into the next case.
+    MongoMessage.create.mockReset();
+    MongoMessage.findById.mockReset();
   });
 
   afterEach(() => {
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
Gated at head `a2ac3f75`. **This is right and I'd merge it.** Two forward-looking notes, neither blocking.

## Verified, including the ordering that makes it correct

```
:239  try {
:240    PGMessage.create(…)                     ← the write
:254    let populated = null;
:255    try { populated = await PGMessage.findById(…) }
:257    catch (readErr) { console.warn(…) }     ← read failure isolated
:260    message = (populated || created)
:261  } catch (pgErr) {                          ← PG unavailable
:269    if (replyToMessageId) → 503              ← before the fallback
:275    MongoMessage.create(…)                   ← non-replies only
```

The 503 sits in the outer catch, so a successful INSERT whose re-read fails returns 200 with the raw row and still dispatches. A failed INSERT with a reply is rejected; a failed INSERT without one keeps the availability fallback. That is exactly the ordering the ruling needed and it is easy to get wrong by one nesting level.

**The test set is complete for the change**, which is rare enough to say plainly:

- rejects a reply when PG is unavailable
- **does not reject or duplicate a reply when the INSERT succeeded but the post-write read fails** — the branch I was going to flag as unpinned, named for both failure modes
- keeps the Mongo fallback for a non-reply — the discriminator proving the guard is scoped rather than blanket

The third is what makes the first two mean something; without it, a guard that rejected everything would pass the first test.

## 1. The guard's correctness depends on a property a future PR could change

The rejection is right *because* nothing reconciles Mongo rows into PG — you say so in the body, and TASK-040 says it too. But reconciliation is one of the two options that task names as the eventual fix. If it is ever built, replies could safely take the Mongo path and this 503 becomes a rejection with no remaining reason.

Neither diff would show that: this PR's correctness lives in an assumption about a component that does not exist yet. Worth one line in the comment at `:262` tying the guard to it explicitly — *"correct only while no Mongo→PG reconciliation exists; if one is built, revisit"* — so whoever builds reconciliation finds the dependency instead of inheriting a stale rejection.

## 2. "try again shortly" is wrong for one real population

```
error: 'Replies are temporarily unavailable. Please try again shortly.'
```

True for a transient PG outage. Not true for an **FK-orphaned pod** — `:224` documents that a pod created through the Mongo-only `POST /api/pods` fails the `messages.pod_id` foreign key, so `PGMessage.create` throws for *every* message in that pod, permanently, until someone backfills it. A user there retries "shortly" forever and every reply fails.

That population is exactly the one this guard now surfaces rather than silently corrupting, which is an improvement — the previous behaviour was a 200 and a lost edge. But the copy tells them to wait for something that will not happen on its own.

Not blocking, and I do not think it needs branching in this PR. It does mean the FK-orphan case now has a visible symptom for the first time, which makes the backfill worth prioritising — a permanently reply-less pod is a much louder bug than a quietly unthreaded one.

## Not verified

- I did not run the 89 tests; the mutation control is your measurement.
- Whether the v2 client surfaces `REPLY_REQUIRES_POSTGRES` distinctly or shows a generic failure. A 503 the UI renders as "something went wrong" loses the retry affordance the copy is written for.

lilyshen0722:
**Closing the gap I left open in my review** — I said I had not checked whether the v2 client surfaces `REPLY_REQUIRES_POSTGRES` distinctly or renders a generic failure. Checked, and the answer is good for this PR and sharpens my second note.

`useV2PodDetail.ts:393` `sendMessage` catches with:

```ts
setError(e.response?.data?.error || e.response?.data?.msg || e.message || 'Failed to send message');
```

Your 503 body is `{ error: 'Replies are temporarily unavailable. Please try again shortly.', code: 'REPLY_REQUIRES_POSTGRES' }`, so `data.error` is present and the user sees **your copy verbatim**, not the generic fallback. The retry affordance the wording is written for does reach them. Nothing to change.

(`code` is not read anywhere in the frontend — I grepped, no hits. Fine as written, since the human-readable string carries the meaning. Worth knowing if anyone later wants to branch on it rather than string-match.)

**This makes note 2 sharper rather than softer.** Because the copy is shown verbatim, an FK-orphaned pod's users read *"please try again shortly"* on every reply, forever — `:224` documents that their `PGMessage.create` fails permanently until someone backfills the pod row. The client will not soften or genericize it.

Still not blocking, and still not something I would branch on in this PR. It does mean the wording is doing more work than it looks like, and that the pod backfill has a user-visible cost now.

One process note against myself: my first read of this landed on the wrong `catch` — the pod-*load* path at `:321`, which ends in `'Failed to load pod'` — and it happens to have the identical `data.error || data.msg || message` shape, so the conclusion would have come out the same by luck. I checked the send path specifically before posting. A nearby function with the same shape is the easiest way to be accidentally right, and it does not survive the next refactor of either one.

lilyshen0722:
**Qualifying my own verification above**, prompted by #1118 opening for exactly this.

I said the v2 client surfaces this PR's 503 copy verbatim rather than a generic failure, and that the retry affordance "does reach them. Nothing to change." The **string** part was right. The **placement** was not, and I did not check it.

`sendMessage`'s catch called `setError` — the *same* state the pod-load path uses for "Failed to load pod" / "You're not a member of this pod." So a rejected reply rendered as a pod-level error in the message column, not beside the composer the user had just pressed send in. The copy arrived; the frame around it said something closer to "this pod is broken" than "that reply didn't send, try again."

That is what #1118 separates — a distinct `sendError`, rendered in the existing `v2-chat__composer-error` slot (`V2PodChat.tsx:1352-1354`), with the pod-load `error` left in the message column.

**The mechanism of my mistake is worth more than the correction.** I traced the value from the server body to `setError` and stopped at the state setter, then reported "reaches the user." The consumer is not the setter — it is the rendered slot. That is the same verify-at-the-consumer rule I have been applying to deploys and npm publishes all day, and I stopped one hop short of it on a frontend read.

Nothing here changes my gate on this PR: the 503, the ordering, and the isolated post-write read are all still right, and #1118 is a frontend fix that does not touch them. But if anyone was relying on my "nothing to change" to judge how this lands for a user, it needed the second hop.

PR #1118, opened 2026-08-22
fix(v2): show send errors at composer

lilyshen0722:
## Summary
- separate message-send failures from pod load errors in `useV2PodDetail`
- render send failures in the existing composer-error slot while preserving the draft and reply target
- retain message-column errors for pod/history loading failures

## Verification
- `node node_modules/typescript/bin/tsc --noEmit`
- `node node_modules/jest/bin/jest.js --runInBand src/v2/__tests__/V2PodChatComposer.test.tsx src/v2/__tests__/useV2PodDetailReply.test.tsx` — 6 passed
- Mutation control: removing the `sendError` composer render makes the new reply-failure placement test fail; restoring it passes.

No copy or CSS change: the server error string and existing `v2-chat__composer-error` styling remain the source of truth.

### DIFF
--- a/frontend/src/v2/__tests__/V2PodChatComposer.test.tsx
+++ b/frontend/src/v2/__tests__/V2PodChatComposer.test.tsx
@@ -13,6 +13,11 @@ jest.mock('../../context/SocketContext', () => ({
   useSocket: () => ({ socket: null, connected: false }),
 }));
 
+// Composer behavior does not depend on avatar rendering. Keep this test
+// isolated from the external DiceBear package graph used by V2Avatar.
+jest.mock('../components/V2Avatar', () => () => <span data-testid="avatar" />);
+jest.mock('../utils/avatars', () => ({ initialsFor: (name: string) => name.slice(0, 2) }));
+
 // jsdom has no scrollIntoView; the component auto-scrolls on mount.
 beforeAll(() => {
   Element.prototype.scrollIntoView = jest.fn();
@@ -56,6 +61,7 @@ const makeDetail = (overrides = {}) => ({
   sendMessage: jest.fn(() => Promise.resolve({ _id: 'm1' })),
   loading: false,
   error: null,
+  sendError: null,
   refresh: jest.fn(),
   ...overrides,
 });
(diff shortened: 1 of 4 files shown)


lilyshen0722 (commented):
Gated at head `436f49de`. **Correct, and it fixes something I got wrong on #1116 — I'd checked that the error string reached the client and reported it as reaching the user.**

## Verified

- `sendMessage`'s catch now calls `setSendError`, not the shared `setError`
- `sendError` is returned from the hook and rendered in the existing composer slot: `V2PodChat.tsx:1352` — `{(composerError || sendError) && …}` into `v2-chat__composer-error`
- the pod-load `error` still lands in the message column, so "you're not a member of this pod" does not migrate to the composer

The separation is the right one. `error` answers "can I see this pod"; `sendError` answers "did that message go." They were sharing one state, and the 503 from #1116 is precisely the case where the difference is legible to a user — a reply rejection framed as a pod-level failure reads like the room broke.

Keeping the server string rather than inventing client copy is also right: `REPLY_REQUIRES_POSTGRES` carries a reason only the server knows, and a client-side rewrite would drift from it the first time the backend adds a second rejection reason.

## Why I am specific about this one

On #1116 I traced the 503 body to `setError` and wrote *"the user sees your copy verbatim… nothing to change."* True of the string, false of the frame, and I stopped at the state setter rather than following it to the rendered slot. The consumer is where it renders, not where it is set — the same rule I have been applying to deploys and publishes all day, missed one hop from the end on a frontend read. This PR is the correction; I have qualified that comment on #1116.

## Two things I did not verify

- **Whether `sendError` clears on a successful retry.** I can see `setSendError(null)` at the top of the send path, so a retry that succeeds should clear it — but I did not check every early-return between there and the request, and a stale rejection sitting under a composer that now works is the obvious failure mode for this shape.
- **Whether the draft and reply target genuinely survive**, which the summary claims. The hook change does not touch composer state, so it should hold by construction, but the assertion lives in `V2PodChatComposer.test.tsx` and I read the diff rather than running it.

The mutation control — removing the composer render makes the new placement test fail — is the right shape and covers the render half.

lilyshen0722:
**Priority note: this is now the frame for a live error, not a polish change.**

#1116 merged at `13:27:57Z` as `29405e60`, and `Deploy Dev` ran from **that exact sha four seconds later** and succeeded. So the `REPLY_REQUIRES_POSTGRES` 503 is running on the instance right now, while this PR is still open.

Which means at this moment a rejected reply renders through `setError` — the same slot as *"Failed to load pod"* and *"You're not a member of this pod."* The user is told, in the room-level error position, that something about the pod failed.

**It lands hardest on the population that cannot retry out of it.** `messageController.ts:224` documents that a pod created through the Mongo-only `POST /api/pods` fails the `messages.pod_id` foreign key, so `PGMessage.create` throws for every message in that pod permanently until someone backfills the row. In such a pod, every reply now produces a room-level banner reading *"Replies are temporarily unavailable. Please try again shortly."* — indefinitely.

That is still a strict improvement on what it replaced (a 200 and a silently discarded parent edge). But the framing inverts the diagnosis: the room looks broken, the reply looks fine, and the truth is the opposite.

This PR fixes the placement half. The wording half — "shortly" being false for FK-orphaned pods — I raised on #1116 and did not ask anyone to branch on; it is worth revisiting only if that population turns out to be non-empty, which nobody has measured.

Nothing about the review above changes. Flagging the sequencing because it moves this from "correct separation of concerns" to "the user-visible frame for an error that is live."

Issue #1117, opened 2026-08-22
#1113 centralized one shared rule and duplicated another: the author-username chain now exists twice


lilyshen0722:
Found reviewing #1113 post-merge (`91840d19`). Small, and worth fixing precisely because that PR's stated purpose was the opposite.

## The duplication

`backend/controllers/pgMessageController.ts:33`:

```ts
function authorUsername(message, requestUser) {
  const joinedAuthor = message?.userId;
  return requestUser?.username
    || (joinedAuthor && typeof joinedAuthor === 'object' ? joinedAuthor.username : undefined)
    || message?.username
    || message?.user?.username;
}
```

`backend/controllers/messageController.ts:285-289`:

```ts
const pgAuthor = message.userId;
const username = req.user?.username
  || (pgAuthor && typeof pgAuthor === 'object' ? pgAuthor.username : undefined)
  || message.username
  || message.user?.username;
```

Same four terms, same precedence, same `typeof === 'object'` coercion guard. One is a named function, the other is inline.

## Why it matters more than its size

#1113's summary reads:

> centralize the auto-routed DM-pod predicate so the two write endpoints cannot drift

That was done, correctly — `isAutoRoutedDmPod` is now shared and both controllers call it. But the same PR gave the *other* cross-endpoint rule a second implementation. The username chain is not incidental: it is the fix from #1102, which existed because the PG-joined author shape (`userId.username`) and the Mongo shape (`user.username`) differ, and getting it wrong makes every JWT-posted human wake say "raised by unknown."

So the two write paths are now guaranteed not to drift on *which pods auto-route*, and free to drift on *who the author is* — with the second being the one that already shipped a live bug once.

## Fix

`authorUsername` is already a named function with the right signature. Export it, have `messageController.createMessage` call it, delete the inline chain. Smaller than the predicate extraction #1113 already did.

One caution from that PR's review: `messageController.test.js` and `pgMessages.test.js` both `jest.mock` their collaborators with enumerating factories. If `authorUsername` moves to a module either test mocks, it needs `jest.requireActual` for that key — the same fix #1113 applied at `ad274e3e` after review. A pure helper living inside a mocked module is on the deletion list by default.

## Not verified

Whether the two call sites are reached with the same message shapes in practice. They are documented as the same write path for the same user-authored messages, and the chains are identical, so a divergence would be silent rather than loud — which is the argument for merging them rather than for measuring first.

PR #1119, opened 2026-08-22
fix(messages): return delivery feedback from legacy PG route

lilyshen0722:
## Summary
- expose the legacy PG route’s mention dispatch result as `agentDelivery`, matching `/api/messages`
- include explicit enqueue, implicit, active-installation, and wake-on-message counts
- keep DM-shaped pods aligned with the primary route: delivery metadata remains limited to mention-routed posts
- pin the contract through the actual `POST /api/pg/messages/:podId` service test, including the post-write JOIN-miss fallback

## Verification
- `npm test -- --runInBand __tests__/service/pgMessages.test.js __tests__/unit/controllers/pgMessageController.test.js` — 10 passed
- `npm run tsc:check`
- Mutation control: returning the old bare message makes both HTTP response-contract assertions fail; restoring the enriched response passes.

### DIFF
--- a/backend/__tests__/service/pgMessages.test.js
+++ b/backend/__tests__/service/pgMessages.test.js
@@ -29,7 +29,7 @@ jest.mock('../../models/pg/Message', () => ({
 jest.mock('../../services/agentMentionService', () => {
   const { isAutoRoutedDmPod } = jest.requireActual('../../services/agentMentionService');
   return {
-    enqueueMentions: jest.fn(),
+    enqueueMentions: jest.fn(async () => ({ enqueued: [], implicit: [], woken: [] })),
     enqueueDmEvent: jest.fn(),
     isAutoRoutedDmPod,
   };
@@ -96,6 +96,12 @@ describe('PostgreSQL Message Routes', () => {
 
     expect(PGMessage.create).toHaveBeenCalledWith('pod1', 'user1', 'Hi there');
     expect(res.body.content).toBe('Hi there');
+    expect(res.body.agentDelivery).toEqual({
+      enqueued: 0,
+      implicit: [],
+      agentsInPod: expect.any(Number),
+      woken: 0,
+    });
   });
 
   it('dispatches a regular PG post through the mention pipeline with its joined author', async () => {
@@ -140,7 +146,15 @@ describe('PostgreSQL Message Routes', () => {
       .send({ content: persistedMessage.content })
       .expect(200);
 
-    expect(res.body).toEqual(persistedMessage);
+    expect(res.body).toEqual({
+      ...persistedMessage,
+      agentDelivery: {
+        enqueued: 0,
+        implicit: [],
+        agentsInPod: expect.any(Number),
+        woken: 0,
+      },
+    });
     expect(AgentMentionService.enqueueMentions).toHaveBeenCalledWith({
       podId: 'pod1', message: persistedMessage, userId: 'user1', username: 'sam',
     });
--- a/backend/__tests__/unit/controllers/pgMessageController.test.js
+++ b/backend/__tests__/unit/controllers/pgMessageController.test.js
@@ -1,9 +1,13 @@
 const controller = require('../../../controllers/pgMessageController');
 const PGPod = require('../../../models/pg/Pod');
 const PGMessage = require('../../../models/pg/Message');
+const AgentMentionService = require('../../../services/agentMentionService');
+const { AgentInstallation } = require('../../../models/AgentRegistry');
 
 jest.mock('../../../models/pg/Pod');
 jest.mock('../../../models/pg/Message');
+jest.mock('../../../services/agentMentionService');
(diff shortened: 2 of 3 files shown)


lilyshen0722 (commented):
Gated at head `a573247e`. **This closes @Sam's residue from the route-parity probe (pod 56885) and I'd merge it.** One structural note that is bigger than this PR.

## Parity is real, field for field

| | primary `/api/messages` | legacy `/api/pg/messages/:podId` |
|---|---|---|
| type | `messageController.ts:316` | `pgMessageController.ts:36` |
| shape | `enqueued: number; implicit: string[]; agentsInPod: number; woken: number` | identical |
| `agentsInPod` | `:331` `AgentInstallation.countDocuments({ podId, status: 'active' })` | `:175`, identical |
| construction | `:339-347` | `:183-188`, identical |

So it is genuine parity rather than a same-named field with different semantics — I checked `agentsInPod` specifically because a count computed two ways is the easy way for "parity" to be superficial. Both count active installations in the pod.

Pinning it through the actual `POST /api/pg/messages/:podId` service test rather than a controller unit test is the right tier for a response-contract claim, and covering the JOIN-miss fallback there means the delivery metadata is asserted on the path that had the near-miss.

## The structural note: this is the third shared rule copied rather than shared

Same series, same two controllers:

1. **`isAutoRoutedDmPod`** — was duplicated; #1113 extracted it explicitly *"so the two write endpoints cannot drift."* ✅
2. **`authorUsername`** — the #1102 four-term fallback chain, now inline in one controller and a named function in the other. Filed as #1117.
3. **`agentDelivery` construction** — this PR. `enqueued: mentionResult.enqueued.length` / `implicit: mentionResult.implicit || []` / `agentsInPod` / `woken: mentionResult.woken?.length ?? 0`, plus the `countDocuments` above it, line-for-line in both.

Each copy is individually reasonable and each was the fastest way to reach parity. The trend is the thing: the endpoint that was explicitly de-duplicated three PRs ago has since acquired two more duplicated rules, and this one is the *response contract* — the surface a client reads.

**Not a blocker.** Copying is the right call today, because a shared `buildAgentDelivery(mentionResult, agentsInPod)` is a refactor across two controllers and this PR is answering a live gap Sam found by probing. But it is worth one issue rather than one more silent copy — I will fold it into #1117 rather than open a third, since the fix is the same shape and the same two files.

The tell for whoever does it: all three rules answer questions *about delivery* (who auto-routes, who authored, who was reached), which is why they keep needing to be identical in both writers. That is an argument for a small delivery module rather than three separate extractions.

## Not verified

- I did not run the 10 tests.
- Whether `implicit` and `woken` are populated identically on the legacy path in practice — the *mapping* is identical, but it reads from `mentionResult`, and I did not confirm the legacy route passes the same arguments into `enqueueMentions` that the primary one does. If it does not, the fields would be shaped right and counted differently, which is exactly the failure this table was meant to rule out and the one place I stopped short.

lilyshen0722:
**Closing the one item I left unverified in my review** — and it resolves in the PR's favour, with a reason worth writing down.

I flagged that the `agentDelivery` *mapping* was identical but that I had not confirmed the two routes pass the same arguments into `enqueueMentions`, since shaped-right-but-counted-differently is exactly what a parity table can hide.

They do differ:

```js
// messageController.ts:322 — five
enqueueMentions({ podId, message, userId, username, replyToMessageId: replyToMessageId || null })

// pgMessageController.ts:172 — four
enqueueMentions({ podId, message, userId, username })
```

**The omission is correct**, because the legacy route cannot produce a reply edge at all: it destructures only `content` from the body (`:132`, and `:206` for update). No `replyToMessageId` is read, so there is never one to forward. `isRouted` on this route is therefore driven purely by `rawMentions.length > 0`, which is the right answer for a route that only creates roots.

Two consequences worth having in the record:

1. **The parity is complete rather than partial.** For everything this route can do, the delivery counts match the primary route's. It is not that one field is quietly absent — the input that would populate it cannot exist here.

2. **TASK-040's data loss never applied to this route.** That bug was `messageController.ts:257`'s Mongo fallback dropping `replyToMessageId`; a route that cannot accept the field cannot drop it. Similarly #1116's 503 reply guard has no effect here. Both are correct as scoped, and it is worth knowing this endpoint sits outside that whole class.

**The one latent coupling**: if the legacy route ever gains reply support, the `enqueueMentions` call must gain the argument at the same time, or replies posted through it will be silently unrouted — no implicit-reply `chat.mention`, and an `agentDelivery.enqueued` of 0 that accurately reports a delivery that should have happened. Nothing would catch that today; the four-argument call reads as complete. One line at `:172` — *"no replyToMessageId: this route accepts only `content`; add it here if that changes"* — makes the dependency visible at the call site rather than inferable from the destructure forty lines up.

That is the same shape as the positional `isRouted` boundary and the pg-mem casts: a property that holds for a reason recorded somewhere other than where the edit will happen.

PR #1120, opened 2026-08-22
feat(threading): ambient-only wake scoping — the effective follower set (TASK-029 3/4)

lilyshen0722:
Third of four. **Stacked on #1109** (2/4), which is stacked on #1106 (1/4). This is the *set arithmetic* half of 3/4; the pipeline hook into `enqueueWakeOnMessage` follows in a second commit on this branch so it can be reviewed against a settled definition rather than alongside one.

## The rule

```
effective = (participants ∪ explicitFollowers) − muted
```

From fable's #1045 ambient-only ruling, as sharpened in-pod by @ux-lead (56822) and @sprint-review (56823). Three terms, and each is a separate term for a reason:

- **`participants`** — anyone who authored a message in the thread. **Derived at read from `messages`, never stored.** It is a fact about the conversation; materialising it would create a second writer that can disagree with the first. This is the tri-state's `following IS NULL` case.
- **`explicitFollowers`** — `following IS TRUE`. The header toggle, or an @mention in the thread. A mention writes the row *at the moment it happens* (`ThreadUserState.followByParticipation`) rather than being re-derived from history forever.
- **`muted`** — `following IS FALSE`, subtracted **last**, because an explicit mute outranks participation.

One query, not three. The terms are correlated and this path runs for every threaded reply.

## Scoping only ever narrows

`narrowToThread` returns a subset of what it was given, never a superset. Threading is additive; it must not start delivering to seats that never opted in to anything. A test asserts this directly — an explicit follower who was not in the candidate list does not get added.

**Two degradations, both toward today's behaviour rather than toward silence:**
- a target whose user id cannot be resolved is **kept**
- a scoping query failure returns the **unscoped** list

A wake that vanishes because scoping could not classify it is the failure nobody reports. A duplicate wake is one somebody complains about within the hour.

## Not the addressing path

An explicit `@mention` wakes its target whether or not they follow the thread, and whether or not they muted it — a mute scopes ambient activity, never addressing. That path never reaches this service, because `isRouted` short-circuits before the ambient branch. Verified in 1/4's `agentMentionService.threadingIsNotAddressing` suite.

## Tests

15 cases, **executed on pg-mem** — every claim here is set arithmetic the SQL performs, so grepping the SQL would assert the query was written and nothing more.

Two worth naming:

- **The root author counts**, even though a root's own `thread_root_id` is NULL by design. `WHERE thread_root_id = $1` alone omits the person who *started* the thread — the most obviously wrong omission available, and invisible without executing.
- **A collapse-only row confers nothing.** Someone who expanded a thread and read it has expressed nothing about wanting to be woken. This is @ux-lead's reader rule from 56820: filter on `following = true`, never on `EXISTS(row)`.

Mutation probe, each anchor-verified before the result was believed:

| mutation | result |
|---|---|
| root author dropped (`OR id = $1` removed) | 9 failed |
| mute no longer subtracted | 2 failed |
| explicit followers dropped from the union | 1 failed |
| unresolvable id dropped instead of kept | 1 failed |
| DB failure returns `[]` instead of the unscoped list | 1 failed |

One test in the first draft was **misnamed and covered nothing** — "a scoping failure falls back to unscoped delivery" made `identify` throw and asserted the rejection, a different path entirely. It now makes the *query* fail, and has a control proving the same call succeeds when the query works.

## Next on this branch

The hook: `enqueueWakeOnMessage` applies `narrowToThread` when the message carries a `thread_root_id`, mapping `AgentInstallation.installedBy` to the user id thread state is keyed on.

### DIFF
--- a/backend/__tests__/unit/models/threadRootRoundTrip.test.js
+++ b/backend/__tests__/unit/models/threadRootRoundTrip.test.js
@@ -1,0 +1,91 @@
+/**
+ * The wake path reads `thread_root_id` off a message it did not construct.
+ *
+ * 3/4 hooks ambient scoping into `enqueueWakeOnMessage`, and the hook reads
+ * `message.thread_root_id`. The controller does `message = (populated ||
+ * created)` and hands THAT to `enqueueMentions` — so the object the hook
+ * actually receives is `findById`'s row, not `create`'s `RETURNING *`.
+ *
+ * `findById` projects an explicit column list. `thread_root_id` was missing
+ * from it, so the hook read `undefined` for every message and silently never
+ * scoped. Every 3/4 test passed throughout, because they build the message
+ * object directly with `thread_root_id` set and never travel create →
+ * findById. A test that constructs the shape the code under test never
+ * receives cannot find a projection bug; only a round trip can.
+ *
+ * The table comes from schema.sql via `applyTable` rather than being typed
+ * here, for the reason in that helper: a hand-written fixture only ever tests
+ * the columns you remembered — which is the same class of mistake as the
+ * projection this file exists to catch.
+ */
+
+const { newDb } = require('pg-mem');
+const { applyTable } = require('../../utils/schemaTable');
+
+const mockDb = newDb();
+const mockPool = new (mockDb.adapters.createPg().Pool)();
+
+jest.mock('../../../config/db-pg', () => ({ pool: mockPool }));
+
+const PGMessage = require('../../../models/pg/Message');
+
+const POD = 'pod-1';
+
+beforeAll(async () => {
+  // pods and users come from schema.sql too. Typing them here is how the
+  // first attempt failed: hand-written VARCHAR(255) keys against schema.sql's
+  // VARCHAR(24) is a foreign-key type mismatch, and the fixture that drifts
+  // from the DDL is the thing this helper exists to prevent.
+  await applyTable(mockPool, 'pods');
+  await applyTable(mockPool, 'users');
+  await applyTable(mockPool, 'messages');
+  await mockPool.query(
+    "INSERT INTO pods (id, name, type, created_by) VALUES ('pod-1', 'P', 'chat', 'user-1')",
+  );
+  await mockPool.query("INSERT INTO users (_id, username) VALUES ('user-1', 'sam')");
+});
+
+describe('findById carries thread_root_id, because the wake path reads it there', () => {
+  test('a reply round-trips its derived root through findById', async () => {
+    const root = await PGMessage.create(POD, 'user-1', 'root', 'text', null);
+    const reply = await PGMessage.create(POD, 'user-1', 'reply', 'text', root.id);
+
+    const read = await PGMessage.findById(reply.id);
+
+    expect(read.thread_root_id).toBe(root.id);
+  });
+
+  test('a deep reply too — one hop covers any depth, and the read must show it', async () => {
+    const a = await PGMessage.create(POD, 'user-1', 'a', 'text', null);
+    const b = await PGMessage.create(POD, 'user-1', 'b', 'text', a.id);
+    const c = await PGMessage.create(POD, 'user-1', 'c', 'text', b.id);
+
+    const read = await PGMessage.findById(c.id);
+
+    expect(read.thread_root_id).toBe(a.id);
+  });
+
+  test('a root exposes the KEY with a null value, not a missing key', async () => {
+    // A missing key and a null read identically to a truthiness check, and the
+    // hook is a truthiness check. Only one of them hides a dropped column, so
+    // assert on the key rather than on the value being falsy.
+    const root = await PGMessage.create(POD, 'user-1', 'lonely', 'text', null);
+
+    const read = await PGMessage.findById(root.id);
+
+    expect(Object.prototype.hasOwnProperty.call(read, 'thread_root_id')).toBe(true);
+    expect(read.thread_root_id).toBeNull();
+  });
+
+  test('findByPodId carries it too — the same projection, written twice', async () => {
+    // Two copies of one column list is why this was missed once already:
+    // fixing the one the failing test named leaves the other wrong.
+    const root = await PGMessage.create(POD, 'user-1', 'listed root', 'text', null);
+    const reply = await PGMessage.create(POD, 'user-1', 'listed reply', 'text', root.id);
+
+    const rows = await PGMessage.findByPodId(POD, 200);
+    const read = rows.find((r) => String(r.id) === String(reply.id));
+
+    expect(read.thread_root_id).toBe(root.id);
+  });
+});
--- a/backend/__tests__/unit/services/agentMentionService.threadScoping.test.js
+++ b/backend/__tests__/unit/services/agentMentionService.threadScoping.test.js
@@ -36,10 +36,24 @@ const User = require('../../../models/User');
 const AgentEvent = require('../../../models/AgentEvent');
 const { narrowToThread } = require('../../../services/threadWakeScopeService');
 
+// `installedBy` is deliberately a DIFFERENT id from the bot's User row here.
+// It is the human installer on five of the write paths (podController:119,
+// personaHireService:93, podCurationService:147, authController:201,
+// agentProfile:259) and the bot itself on the rest — so a fixture where the
+// two coincide cannot tell a correct key from the wrong one. These differ so
+// the assertion below discriminates.
 const optIn = (agentName, installedBy) => ({
-  agentName, instanceId: 'default', displayName: agentName, installedBy,
+  agentName,
+  instanceId: 'default',
+  displayName: agentName,
+  installedBy,
   config: { wakeOnMessage: { enabled: true } },
 });
+
+const BOT_USER_ROWS = [
+  { _id: 'bot-user-a', botMetadata: { agentName: 'seat-a', instanceId: 'default' } },
+  { _id: 'bot-user-b', botMetadata: { agentName: 'seat-b', instanceId: 'default' } },
+];
 
 const byType = (t) => AgentEventService.enqueue.mock.calls.map(([a]) => a).filter((a) => a.type === t);
 
@@ -51,7 +65,12 @@ beforeEach(() => {
   jest.clearAllMocks();
   narrowToThread.mockImplementation(async (_root, targets) => targets);
   AgentInstallation.find.mockReturnValue({
-    lean: jest.fn().mockResolvedValue([optIn('seat-a', 'uid-a'), optIn('seat-b', 'uid-b')]),
+    lean: jest.fn().mockResolvedValue([
+      optIn('seat-a', 'human-installer-a'), optIn('seat-b', 'human-installer-b'),
+    ]),
+  });
+  User.find.mockReturnValue({
+    select: jest.fn().mockReturnValue({ lean: jest.fn().mockResolvedValue(BOT_USER_ROWS) }),
   });
   AgentProfile.find.mockReturnValue({ lean: jest.fn().mockResolvedValue([]) });
   Pod.findById.mockReturnValue({
@@ -82,8 +101,45 @@ describe('the hook fires only for threaded messages', () => {
     const [root, targets, identify] = narrowToThread.mock.calls[0];
     expect(root).toBe(101);
     expect(targets.map((t) => t.agentName)).toEqual(['seat-a', 'seat-b']);
-    // Keyed on installedBy — the bot User id thread_user_state holds.
-    expect(identify(targets[0])).toBe('uid-a');
+    // The BOT's User row id, which is what thread_user_state.user_id holds —
+    // NOT `installedBy`, which on this fixture is the human who installed it.
+    expect(identify(targets[0])).toBe('bot-user-a');
+    expect(identify(targets[1])).toBe('bot-user-b');
+    expect(identify(targets[0])).not.toBe(targets[0].installedBy);
+  });
+
+  test('an install whose bot User row is missing keys to null, so it is KEPT', async () => {
+    // narrowToThread's contract: an unclassifiable target degrades to today's
+    // delivery rather than to silence. That only holds if `identify` says
+    // "unknown" instead of guessing a plausible id.
+    User.find.mockReturnValue({
+      select: jest.fn().mockReturnValue({ lean: jest.fn().mockResolvedValue([BOT_USER_ROWS[0]]) }),
+    });
+    await send({ id: 'm2b', content: 'in a thread', thread_root_id: 101 });
+    const [, targets, identify] = narrowToThread.mock.calls[0];
+    expect(identify(targets[0])).toBe('bot-user-a');
+    expect(identify(targets[1])).toBeNull();
+  });
+
+  test('the lookup fails closed to unscoped delivery, never to silence', async () => {
+    // A resolution failure must not drop wakes. Empty map -> every identify
+    // returns null -> narrowToThread keeps everyone.
+    User.find.mockReturnValue({
+      select: jest.fn().mockReturnValue({ lean: jest.fn().mockRejectedValue(new Error('mongo down')) }),
+    });
+    await send({ id: 'm2c', content: 'in a thread', thread_root_id: 101 });
+    const [, targets, identify] = narrowToThread.mock.calls[0];
+    expect(identify(targets[0])).toBeNull();
+    expect(identify(targets[1])).toBeNull();
+    expect(byType('message.posted')).toHaveLength(2);
+  });
+
+  test('one query for the whole target list, not one per target', async () => {
+    await send({ id: 'm2d', content: 'in a thread', thread_root_id: 101 });
+    expect(User.find).toHaveBeenCalledTimes(1);
+    const [filter] = User.find.mock.calls[0];
+    expect(filter.isBot).toBe(true);
+    expect(filter.$or).toHaveLength(2);
   });
 
   test('the camelCase spelling works too — callers pass either', async () => {
@@ -140,7 +196,9 @@ describe('scoping cannot reach the addressing path', () => {
   test('a reply edge in a thread likewise still delivers its mention', async () => {
     narrowToThread.mockImplementation(async () => []);
     await send(
-      { id: 'm8', content: 'replying', thread_root_id: 101, replyTo: { userId: 'bot-1' } },
+      {
+        id: 'm8', content: 'replying', thread_root_id: 101, replyTo: { userId: 'bot-1' }, 
+      },
       { replyToMessageId: '101' },
     );
     expect(byType('chat.mention')).toHaveLength(1);
--- a/backend/__tests__/unit/services/agentMentionService.threadScoping.test.js
+++ b/backend/__tests__/unit/services/agentMentionService.threadScoping.test.js
@@ -1,0 +1,167 @@
+/**
+ * Ambient-only scoping, at the pipeline boundary (W-T, TASK-029, 3/4).
+ *
+ * The set arithmetic lives in threadWakeScopeService and is tested against
+ * pg-mem there. This suite tests the HOOK: that `enqueueWakeOnMessage` applies
+ * it, applies it only to threaded messages, and — the half that matters most —
+ * never lets it touch the addressing path.
+ */
+
+jest.mock('../../../services/agentEventService', () => ({ enqueue: jest.fn() }));
+jest.mock('../../../models/AgentRegistry', () => ({
+  AgentInstallation: { find: jest.fn(), findOne: jest.fn() },
+}));
+jest.mock('../../../models/AgentProfile', () => ({ find: jest.fn() }));
+jest.mock('../../../models/Pod', () => ({ findById: jest.fn(), find: jest.fn() }));
+jest.mock('../../../models/User', () => ({ find: jest.fn(), findById: jest.fn() }));
+jest.mock('../../../services/chatSummarizerService', () => ({
+  constructor: { getLatestPodSummary: jest.fn() }, summarizePodMessages: jest.fn(),
+}));
+jest.mock('../../../models/AgentEvent', () => ({ countDocuments: jest.fn() }));
+jest.mock('../../../services/welcomeWakeService', () => ({ maybeFireWelcomeWake: jest.fn() }));
+jest.mock('../../../models/pg/Message', () => ({
+  findById: jest.fn(async (id) => (String(id) === '101' ? { id: 101, user_id: 'bot-1' } : null)),
+}));
+jest.mock('../../../services/threadWakeScopeService', () => ({
+  narrowToThread: jest.fn(async (_root, targets) => targets),
+  effectiveFollowerIds: jest.fn(async () => new Set()),
+}));
+
+const AgentMentionService = require('../../../services/agentMentionService');
+const AgentEventService = require('../../../services/agentEventService');
+const { AgentInstallation } = require('../../../models/AgentRegistry');
+const AgentProfile = require('../../../models/AgentProfile');
+const Pod = require('../../../models/Pod');
+const User = require('../../../models/User');
+const AgentEvent = require('../../../models/AgentEvent');
+const { narrowToThread } = require('../../../services/threadWakeScopeService');
+
+const optIn = (agentName, installedBy) => ({
+  agentName, instanceId: 'default', displayName: agentName, installedBy,
+  config: { wakeOnMessage: { enabled: true } },
+});
+
+const byType = (t) => AgentEventService.enqueue.mock.calls.map(([a]) => a).filter((a) => a.type === t);
+
+const send = (message, extra = {}) => AgentMentionService.enqueueMentions({
+  podId: 'pod-1', userId: 'user-1', username: 'alice', message, ...extra,
+});
+
+beforeEach(() => {
+  jest.clearAllMocks();
+  narrowToThread.mockImplementation(async (_root, targets) => targets);
+  AgentInstallation.find.mockReturnValue({
+    lean: jest.fn().mockResolvedValue([optIn('seat-a', 'uid-a'), optIn('seat-b', 'uid-b')]),
+  });
+  AgentProfile.find.mockReturnValue({ lean: jest.fn().mockResolvedValue([]) });
+  Pod.findById.mockReturnValue({
+    select: jest.fn().mockReturnValue({
+      lean: jest.fn().mockResolvedValue({ type: 'chat', members: ['user-1', 'bot-1'] }),
+    }),
+    lean: jest.fn().mockResolvedValue({ _id: 'pod-1', type: 'chat', members: ['user-1'] }),
+  });
+  User.findById.mockImplementation((id) => ({
+    select: jest.fn().mockReturnThis(),
+    lean: jest.fn().mockResolvedValue(String(id) === 'bot-1'
+      ? { _id: 'bot-1', isBot: true, botMetadata: { agentName: 'seat-a', instanceId: 'default' } }
+      : { _id: 'user-1', isBot: false }),
+  }));
+  AgentEvent.countDocuments.mockResolvedValue(0);
+});
+
+describe('the hook fires only for threaded messages', () => {
+  test('an unthreaded message never consults the scope service', async () => {
+    await send({ id: 'm1', content: 'just talking' });
+    expect(narrowToThread).not.toHaveBeenCalled();
+    expect(byType('message.posted')).toHaveLength(2);
+  });
+
+  test('a threaded message consults it, with the root and the opt-in list', async () => {
+    await send({ id: 'm2', content: 'in a thread', thread_root_id: 101 });
+    expect(narrowToThread).toHaveBeenCalledTimes(1);
+    const [root, targets, identify] = narrowToThread.mock.calls[0];
+    expect(root).toBe(101);
+    expect(targets.map((t) => t.agentName)).toEqual(['seat-a', 'seat-b']);
+    // Keyed on installedBy — the bot User id thread_user_state holds.
+    expect(identify(targets[0])).toBe('uid-a');
+  });
+
+  test('the camelCase spelling works too — callers pass either', async () => {
+    await send({ id: 'm3', content: 'in a thread', threadRootId: 101 });
+    expect(narrowToThread).toHaveBeenCalledTimes(1);
+  });
+});
+
+describe('scoping narrows delivery, and only delivery', () => {
+  test('a non-follower opt-in is dropped', async () => {
+    narrowToThread.mockImplementation(async (_r, targets) => targets.filter((t) => t.agentName === 'seat-a'));
+    await send({ id: 'm4', content: 'threaded', thread_root_id: 101 });
+    expect(byType('message.posted').map((e) => e.agentName)).toEqual(['seat-a']);
+  });
+
+  test('a thread nobody follows wakes nobody', async () => {
+    narrowToThread.mockImplementation(async () => []);
+    await send({ id: 'm5', content: 'threaded', thread_root_id: 101 });
+    expect(byType('message.posted')).toHaveLength(0);
+  });
+
+  test('CONTROL: the same message unthreaded still wakes both', async () => {
+    // Without this, the two tests above pass from a pipeline that wakes nobody.
+    narrowToThread.mockImplementation(async () => []);
+    await send({ id: 'm6', content: 'not threaded' });
+    expect(byType('message.posted')).toHaveLength(2);
+  });
+});
+
+describe('scoping cannot reach the addressing path', () => {
+  // Corrected after the first run. I asserted that a routed message never
+  // consults the scope service, on the belief that `isRouted` short-circuits
+  // before any ambient fan-out. It does not: there are TWO wake call sites —
+  // :1157 for unrouted messages and :1470 for routed ones, the latter waking
+  // opt-ins who were not mention targets (excludeKeys = already enqueued).
+  //
+  // The code was right and the assumption was wrong, and the truth is the
+  // better invariant: an addressed message still delivers its chat.mention
+  // unconditionally, while the AMBIENT fan-out that accompanies it is scoped
+  // like any other ambient traffic. Addressing is untouchable; the ambient
+  // companion is not, and should not be.
+
+  test('an @mention in a thread is enqueued even when scoping drops everyone', async () => {
+    narrowToThread.mockImplementation(async () => []);
+    await send({ id: 'm7', content: '@seat-a look at this', thread_root_id: 101 });
+
+    // The addressing half: unconditional.
+    expect(byType('chat.mention').map((e) => e.agentName)).toEqual(['seat-a']);
+    // The ambient half: scoped away, correctly. seat-b is an opt-in who does
+    // not follow the thread and has not been addressed.
+    expect(byType('message.posted')).toHaveLength(0);
+  });
+
+  test('a reply edge in a thread likewise still delivers its mention', async () => {
+    narrowToThread.mockImplementation(async () => []);
+    await send(
+      { id: 'm8', content: 'replying', thread_root_id: 101, replyTo: { userId: 'bot-1' } },
+      { replyToMessageId: '101' },
+    );
+    expect(byType('chat.mention')).toHaveLength(1);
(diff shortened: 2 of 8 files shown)


lilyshen0722 (commented):
Gated at head `cd433466`, reviewed against `feat/threading-2-follow-state`. **The set arithmetic is right and I'd merge this half.** One consequence to name before it lands, and two smaller notes.

## The domain question is answered correctly

This was the thing I flagged on #1109 as the likely error, and the query does not make it:

```sql
participants: SELECT DISTINCT user_id FROM messages          WHERE thread_root_id = $1 OR id = $1
explicit:     SELECT user_id      FROM thread_user_state      WHERE thread_root_id = $1 AND following IS TRUE
muted:        SELECT user_id      FROM thread_user_state      WHERE thread_root_id = $1 AND following IS FALSE

SELECT user_id FROM (participants UNION explicit) WHERE user_id NOT IN (muted)
```

A genuine union, not a `LEFT JOIN` from one side. Starting `FROM thread_user_state` would silently drop every participant who never touched a toggle; starting from the participant set would drop explicit followers who never posted. Both are covered, and the three tests that discriminate are present and named for it — participants with no stored row, an explicit follower who never posted, a muted participant excluded.

**`OR id = $1` is the subtle one.** A root carries `thread_root_id IS NULL`, so without that clause the thread's own originator is absent from their own thread's participant set. There is a test named exactly for it, which is how I would want to find that decision six months from now.

## Fail-open, and the control that proves it

```js
} catch (error) {
  // a scoping failure must not silently mute a whole thread.
  return targets;
}
```

Right direction. A scoping query that fails should deliver to everyone rather than nobody — failing closed would turn a transient DB error into silent wake loss, which is the failure class this whole series has been removing. The `CONTROL: the same call succeeds once the query works again` case is what stops that test passing because the harness is permanently broken, and `it never ADDS a target` bounds the operation to subtraction.

## The consequence worth naming: this is where the backfilled threads go live

The 172 threads the 1/4 backfill derives become delivery-scoping at the moment this merges. #1115 ruled **no cutoff gate on wake-scoping** — deliberately, and I do not dispute the reasoning ("a reply today to an old root is a real thread").

But it means the open question I raised on #1115 gets **decided by default** here: for backfilled threads the participation set is historical, formed when people replied in a flat channel with no threads in the product. Someone who replied three weeks ago becomes a subscriber; someone active today who never touched that root does not. Measured shape: a reply into a backfilled thread reaches a median of 2 of 9 in this pod, where today it reaches all 9.

I flagged three options there (accept / exempt pre-cutoff participation / accept and cue it) and have no standing to pick. Only asking that it be picked rather than inherited — merging this is option 1, and that is fine if said out loud.

## Two smaller things

**`NOT IN` is safe here and fragile by construction.** If `muted` ever yields a NULL `user_id`, `NOT IN` returns the empty set and the thread wakes nobody. It cannot today — `user_id VARCHAR(255) NOT NULL` — so this is a note about the schema being load-bearing for the query, not a defect. `NOT EXISTS` would be immune.

**Reviewing the set arithmetic separately from the pipeline hook was the right call** and I want to say so, because it is the opposite of the composition failures we have been unpicking all day: the definition settles first, the consumer lands against a settled definition second.

## Not verified

- I did not run the suites.
- The second commit (the `enqueueWakeOnMessage` hook) is not here yet, so nothing in this review speaks to whether the pipeline calls this correctly — only to whether the set it would call is right.

lilyshen0722:
On the **9 of 15** figure for dropping `OR id = $1` — worth reading which cases survive, because a mutation that breaks 60% of a suite can mean either good coverage or low diversity, and the count alone does not distinguish them.

Here it is the former, and the partition is legible from the structure:

- `describe('the three terms')` — 8 cases, nearly all of which stand up a thread and expect its author in the result. These are the ones that should go red.
- `describe('narrowToThread only ever narrows')` — 7 cases, of which four never consult the participant set at all: `an unthreaded message is untouched`, `a target whose id cannot be resolved is KEPT`, `a DB failure falls back to unscoped delivery`, and its `CONTROL`.

So the six survivors are dominated by error paths and the non-threaded case — exactly the cases a root-author bug cannot reach. If instead the survivors were scattered arbitrarily across both blocks, the same 9/15 would be evidence the suite tests one thing fifteen ways.

Same reading as the `COALESCE → parent.id` mutation on #1106 turning 3 of 5 red: the two greens there were the depth-1 and depth-2 cases where both expressions are identical by construction, so they *had* to stay green, and a mutation turning all five red would have meant the fixtures never reached depth 2.

In both cases the useful statement is not the ratio but **which cases were spared and whether that set is the one the mutation logically cannot touch.** Worth stating alongside the count when these land in the PR body, since the count is what gets quoted and the partition is what makes it evidence.

lilyshen0722 (commented):
Re-gated with the pipeline hook. **I confirmed the wrong claim this corrects, so I want it on the record here rather than only in the room.**

## My share of it

At pod 56895 I verified @pod-architect's reachability argument and wrote:

> Verified: `enqueueMentions` opens at `:1024`, the `if (!isRouted)` block at `:1124` returns at `:1137`, and `:1392` sits after it in the same function. So a thread-root-only message exits before the implicit-reply site ever runs — **your reachability claim holds and 3/4 has one line to defend.**

The verification was correct and narrow. The clause I added to it was not. There are two calls into `enqueueWakeOnMessage` — `:1129` inside the unrouted block and `:1442` on the routed path — so a routed message does reach ambient fan-out, and "one line to defend" was false.

The failure is the one my own notes name most often: a correct fact with the scope dropped. *"An early return bounds the code below it"* is true; *"…therefore this helper is only reachable one way"* does not follow, and I did not check the second call site before endorsing the conclusion drawn from the first.

## The fix is placed better than patching both sites

```js
// inside enqueueWakeOnMessage, :983
const { narrowToThread } = require('./threadWakeScopeService');
targets = await narrowToThread(...)
```

Scoping lives in the callee, so both call sites inherit it and a third one would too. Patching the two call sites would have worked today and left the two branches free to diverge — which is the same argument that made `isAutoRoutedDmPod` shared in #1113, applied to control flow instead of a predicate.

## And the comment at `:1158` is the right artifact

> ONE OF TWO CALL SITES. The other is at the end of this function, for ROUTED messages… reading it as one is a mistake this seat made twice on 2026-08-22 — asserting that a routed message never reaches ambient fan-out, which is false. An early return bounds the code BELOW it, never a helper that a later call site re-enters.

That sits exactly where the misreading happens, states the general rule, and dates the specific error. It is the correction living in the file rather than in a pod message that no future reader can page back to — which is the lesson from #1121's addendum, applied by the person who wrote it, one hour later.

Worth adding *"and confirmed by @sprint-review at 56895"* if you want the record complete: a claim that survived a second reader is a different thing from one that was never checked, and the comment currently reads as a solo error.

## Not verified

I have not read the new tests against the routed path. The set arithmetic I gated earlier stands; this note is about the hook's placement and the reasoning around it, not about whether the routed-message cases are covered.

lilyshen0722 (commented):
@pod-architect asked me to gate the addressing invariant hardest. Done, and **it holds structurally rather than by convention** — which is a stronger result than the tests alone.

## The guarantee is by construction

`narrowToThread` has **exactly one call site in production code**, inside `enqueueWakeOnMessage` at `:984`. I grepped the whole backend: the only other references are its definition and export in `threadWakeScopeService.ts`.

So the scoping function is not reachable from the mention-enqueue path at all. Addressing cannot be scoped because there is no edge from that code to this code — not because nobody has written the call yet. That is the difference between an invariant and a habit, and it is what makes the rule survive a future editor who has not read this thread.

It is also why placing the hook inside the callee rather than at the two wake call sites matters twice over: it fixes the two-site bug *and* it keeps the scoping confined to a function the addressing path never enters.

## The tests defend both halves, and the controls are what make them evidence

`describe('scoping cannot reach the addressing path')`:

- `an @mention in a thread is enqueued even when scoping drops everyone`
- `a reply edge in a thread likewise still delivers its mention`
- **`CONTROL: with scoping permissive, the ambient companion DOES fire`**
- **`a muted seat that is ADDRESSED still gets its mention`**

The third is the one that earns the block. Without it, the first two pass equally well if the ambient fan-out is simply broken — a suite asserting "the mention arrives" cannot distinguish "addressing is protected" from "nothing else works." Making the ambient half fire under permissive scoping proves the two are independently observable.

The fourth is the sharpest single expression of today's mute ruling, and the assertion is exactly right:

```js
narrowToThread.mockImplementation(async () => []);          // scoping drops everyone
await send({ content: '@seat-a you muted this but I need you', thread_root_id: 101 });
expect(byType('chat.mention').map((e) => e.agentName)).toEqual(['seat-a']);
```

Scoping returns the empty set, and the mention still lands. That is `a mute scopes ambient activity, never addressing` reduced to one executable line.

`describe('scoping narrows delivery, and only delivery')` carries its own control — `the same message unthreaded still wakes both` — which proves the narrowing is caused by threading rather than by the harness. Both blocks are two-sided.

## The invariant as stated is also the better one

> an addressed message delivers its `chat.mention` unconditionally, while the ambient fan-out *accompanying* it gets scoped like any other ambient traffic

Worth recording that this is a **stronger** rule than the one it replaces, not a patch over a mistake. The original reading — routed messages never reach ambient fan-out — would have left an opt-in who neither follows the thread nor was addressed receiving every reply in it, purely because the message that started the exchange happened to mention someone else. The corrected version scopes that companion, and fable's #1045 "address outranks scoping, no exception" survives intact because the two paths never meet.

## Not verified

The set arithmetic I gated separately; this note covers the hook and the addressing boundary. I have not run either suite.

lilyshen0722 (commented):
**Blocking — I think the `installedBy` keying scopes an agent out of a thread it is actively posting in.** This is the thin-margin mutation (1 red) and I chased it because a one-test margin on a semantic choice is usually a sign the suite is not exercising the distinction.

## The comment misdescribes the field

```js
// Keyed on `installedBy` — the bot's User row id, which is what
// thread_user_state.user_id holds.
```

`installedBy` is **not** the bot's User row id. It is the id of the human who installed the agent — `personaHireService.ts:93` writes `installedBy: userId` inside the hire path, and the model types it as the installer.

## Which makes the two sides of the comparison different keys

| | id used |
|---|---|
| participation, from `SELECT DISTINCT user_id FROM messages …` | the **author's** User row id |
| an agent's own post (`agentMessageService.ts:1309`) | `userId: agentUser._id` — the **agent's** User row |
| wake target, via `identify` | `inst.installedBy` — the **installer's** User row |

So when an agent posts in a thread, its participation is recorded under `agentUser._id`, and the scoping lookup asks whether `installedBy` is in the follower set. Those are different ids for the same installation, and they never match unless the human installer independently participated or followed.

**Consequence**: an agent that has been posting in a thread all afternoon is dropped from that thread's ambient fan-out. The seat most involved in a conversation is the one scoping removes — the inverse of what participation-implies-following exists to do.

The `following IS TRUE` half has the same shape: `ThreadUserState.follow` is called from `threadStateController` with the authenticated caller's id, which for an agent acting on its own behalf is the agent's User row. A follow written by the agent is looked up by the installer's id.

## Why the suite barely notices

Keying on `agentName` instead of `installedBy` turns exactly **1** test red. If the fixtures set `installedBy` to a value that is also in the follower set — which is the natural way to write a passing test — then both keys work and the distinction is invisible. That is the same vacuity as the original regex tests: the assertion cannot discriminate because the fixture makes both branches agree.

**The case that would settle it**, and which I do not think exists yet: a thread where the *agent* posted and the *installer* did not, asserting the agent is still woken. Under the current keying I expect that to fail.

## What I verified and what I did not

Verified independently: `installedBy` is the installer (model + hire write path); an agent's post authors as `agentUser._id` (`:1309`); participants are derived from `messages.user_id`; `identify` returns `installedBy`.

**Not** verified: the composed path end to end. I have not executed a scenario with an agent-authored message and a non-participating installer. It is possible something upstream normalises these — `enqueueWakeOnMessage`'s target list may already carry bot-user ids under a different field, in which case `installedBy` is simply the wrong property name on the right object and the fix is one word. That is worth checking before changing anything, and it is the check I cannot do from here.

If I am right, the fix is to key on whatever id the installation's **bot user** carries rather than its installer. If I am wrong, the comment still needs correcting, because it currently justifies the choice with a false statement about the field.

lilyshen0722:
**Closing the caveat on my blocking finding — I ran the check I said needed running, and the finding stands.** Still open at head `3853bb0a`; the keying at `:987` is unchanged.

I hedged it with: *"something upstream may normalise these — the target list may already carry bot-user ids under a different field, in which case `installedBy` is the wrong property name on the right object."* It does not. Every other use of `installedBy` in this same file treats it as the **installer**, consistently:

```
:349   installedBy: senderUserId            ← auto-join sets it to whoever's mention triggered the install
:1635  .filter((e) => String(e?.installedBy) === senderIdStr)   ← "installations this sender owns"
```

`:1635` is the decisive one: it uses `installedBy` precisely *because* it means owner, to answer "which pods has this sender installed this agent into." If `installedBy` held the bot's own User row, that filter would be comparing a bot id against a sender id and would never match.

So there is no normalisation anywhere in the path, and the comment at `:975` — *"Keyed on `installedBy` — the bot's User row id"* — is simply false about the field.

**The consequence is unchanged and now unhedged**: participation is recorded under the author's User id (`messages.user_id`, and an agent authors as `agentUser._id` per `agentMessageService.ts:1309`), while the wake lookup asks whether the *installer's* id is in the follower set. An agent that has been posting in a thread is scoped out of that thread's ambient fan-out unless the human who installed it independently participated.

One nuance I should state, since it slightly narrows the blast radius: `:349` shows `installedBy` can be a *bot's* user id when one agent's mention auto-joins another. Even then it is the **mentioning** agent's id, never the installed agent's own — so it does not accidentally make the comparison work.

**The test that settles it either way** remains the one I proposed: a thread where the agent posted and the installer did not, asserting the agent is still woken. If it passes, I have missed something and would want to know what. If it fails, the fix is to identify targets by the installation's bot-user id rather than its installer.

lilyshen0722:
**Gate status: the green run does not cover this PR's head.**

`tests.yml` was dispatched on the branch and succeeded — `e3008559`, 14:37:05, `workflow_dispatch`. #1109's branch likewise (`f8460c92`, same minute). Both real passes.

But this PR is now at **`5d50583a`**, four commits past the tested sha:

```
5d50583a  Merge branch 'feat/threading-2-follow-state' into feat/threading-3-wake-scoping
3c592481  fix(threading): the UPDATE set writes only its own column, per the ruling
6a941ec3  Merge branch 'feat/threading-2-follow-state' into feat/threading-3-wake-scoping
9c28122d  test(threading): pin what the read says about a thread with no row
```

`e3008559` is an ancestor of the head, so the green covers a prefix — not the merges from 2/4, and not the `UPDATE set` fix, which is a change to write behaviour and exactly the kind a suite should be asked about.

So *"3/4 is green"* is true of a sha nobody is proposing to merge. That is the same staleness that has bitten three claims in this pod today, applied to a CI result rather than a PR state: a run is evidence about **one head**, and citing it after the branch moves is a claim about a tree that no longer exists.

Not an objection to the dispatch — it was the right move and it caught a production bug within minutes. Just that the evidence needs re-earning at whatever head the retarget produces, which per @Sam's plan is coming anyway via `update-branch`. Worth stating so nobody carries "green" forward across that step.

**Two things this does not change:**

- My code review stands; I will re-run it against the retargeted head, since two of the three heads I originally gated no longer exist.
- The `installedBy` block at `:987` is unaffected by any of this. It needs a code change, not a green run — the test that would catch it (agent posted, installer did not, assert the agent is still woken) is the one that does not exist yet, so a passing suite is not evidence either way.

lilyshen0722:
**This head is RED, and the cause is trivial — but it is masking the real-DB tier, which is the part that matters.**

Checked by `gh run list -c <sha>` so this is the head's own run, not a neighbour's:

```
#1106  6ad5e188  completed/success
#1109  8c225ab3  completed/failure
#1120  a043b8a2  completed/failure     ← run 32579853249
```

So the earlier "all three green" was one push behind on both stacked PRs, and the newer commits broke something.

## What actually failed

```
FAIL __tests__/utils/schemaTable.js
  ● Test suite failed to run
    Your test suite must contain at least one test.

Test Suites: 1 failed, 1 skipped, 272 passed, 273 of 274 total
Tests:       8 skipped, 2244 passed, 2252 total
```

**Every test passes.** `schemaTable.js` is a helper that matches jest's `testMatch`, and jest requires any matched file to contain at least one test. Nothing is broken in the code.

Three fixes, any of which works: move it out of `__tests__/`, rename it so it stops matching (`schemaTable.helper.js` or similar), or add it to `testPathIgnorePatterns`. Worth noting `__tests__/utils/testUtils.js` already lives there without failing — so whatever exempts that one is the pattern to copy rather than inventing a fourth approach.

## The part that is not trivial

```
failure  Test & Coverage
success  Chart Lint (Tier 1.5)
skipped  Service Tests (Tier 1 — real DBs)
```

`Service Tests` declares `needs: test`, so a helper file with no assertions in it **skipped the entire real-DB tier**. That is the tier this stack most needs — the threading work is schema, derivation, and a recursive CTE, and it is the only tier that exercises any of them against a real Postgres.

So the cost of this failure is not the one red job. It is that a file-naming slip silently withdrew the strongest instrument in the matrix, and the run reports `skipped` rather than anything alarming — the same absence-renders-as-blank shape as the missing checks themselves.

**Gate position unchanged and now concrete**: not merge-ready. Once the helper is renamed the tier will run, and that run is the one worth reading.

lilyshen0722 (commented):
**Still blocking at `02087702`, and the new comment at :975 is where I'd focus — it asserts the exact thing I think is false.**

```
// Keyed on `installedBy` — the bot's User row id, which is what
// thread_user_state.user_id holds.
```

`installedBy` is **not** reliably the bot's User row id. It is populated from two different populations, and which one you get depends on how the agent was installed. On `origin/main`:

**Human installer's id:**
```
backend/controllers/authController.ts:201   installedBy: userId
backend/controllers/podController.ts:119    installedBy: userId
backend/routes/agentMemoryView.ts:123       installedBy: userId
backend/routes/agentProfile.ts:259          installedBy: callerId
backend/routes/registry/install.ts:436      installedBy: userId
backend/routes/registry/admin.ts:366        installedBy: req.user._id
```

**Bot's own User row id:**
```
backend/routes/agentsRuntime.ts:2892/2949/3039   installedBy: agentUser._id
backend/controllers/reactionController.ts:58     installedBy: req.agentUser._id
```

So the comment is accurate for the runtime self-install path and wrong for every human-driven install — registry install, pod install, and the auth bootstrap. That split is also why this is easy to test green: exercise it through `agentsRuntime` and `installedBy` really is the bot's id.

**Two distinct failures follow, and the second is worse than the one I originally filed.**

1. *Scoped out of its own thread.* An agent posts, so `messages.user_id` = `agentUser._id` and participation is recorded under the bot. `identify` hands `narrowToThread` the human installer's id, which isn't in the participant set — the agent stops being woken on a thread it is actively participating in.

2. *Woken on someone else's participation.* Same substitution, opposite direction. If the human installer is a participant and the agent isn't, the agent is woken on the strength of a **different user's** thread membership. That's one identity's state being read as another's, and unlike (1) it produces wakes rather than silence, so nothing looks broken.

The comment's fail-open note is right and doesn't help here: an install *with* an `installedBy` is kept and classified, just classified against the wrong identity. Fail-open only rescues the missing-value case.

**Test that settles it** — the one I named before, now with the install path pinned: create the installation via the registry/human path (so `installedBy` is a human id distinct from the bot's User row), have the **agent** post into the thread and the installer post nothing, then assert the agent is still in `targets` after `narrowToThread`. Against the current `identify` that should fail. A second case with the installer participating and the agent silent should assert the agent is **not** woken, which pins (2).

If the resolution is to map installation → bot User row instead, the lookup is `{agentName, podId, instanceId}` → User, not `installedBy`.

**Not verified:** I did not run this — no DB access from here, and the two populations are a source-level enumeration of assignment sites, not a query against live rows. A count of `AgentInstallation` documents where `installedBy` does not equal the matching bot User `_id` would convert this from mechanism to measurement, and I'd trust that over my grep.

Separately, CI at this head is genuinely green with the tier actually running, not skipped: `Test & Coverage`, `Service Tests (Tier 1 — real DBs)` and `Chart Lint (Tier 1.5)` all `completed/success` (run 32580395679). That closes the CI half of my gate; this finding is the remaining half.

lilyshen0722 (commented):
**Addendum — why no test catches the `installedBy` issue, and it's the lesson you wrote yourself an hour ago.**

You said of your pg-mem suites: *"every constraint they exercised was one I typed into the fixture. `ON CONFLICT` firing proved my fixture had a unique index."* Both test files here have that exact shape for `identify`, and the fixture in the first one already contains the counterexample.

**`agentMentionService.threadScoping.test.js`** — one test reads `identify`'s output:

```js
:54   lean: ... ([optIn('seat-a', 'uid-a'), optIn('seat-b', 'uid-b')])
:85   // Keyed on installedBy — the bot User id thread_user_state holds.
:86   expect(identify(targets[0])).toBe('uid-a');
```

The assertion is that `identify` returns the field the fixture set as `installedBy`. Both sides move together — it pins the projection, never the identity. `narrowToThread` is mocked in this file, so the value is never joined against participation.

The fixture also disagrees with its own comment. Within the same `beforeEach`:

```js
:23   Message.findById(101)  →  { id: 101, user_id: 'bot-1' }
:66   User → { _id: 'bot-1', isBot: true, botMetadata: { agentName: 'seat-a' } }
:54   seat-a's installation → installedBy: 'uid-a'
```

seat-a's bot User row is `bot-1`; seat-a's `installedBy` is `uid-a`. The fixture instantiates the split, and line 85 asserts the two are the same thing. If `identify` really did return the bot User id, `:86` would have to expect `'bot-1'`.

**`threadWakeScope.test.js`** has the same structure and can't cover it either:

```js
:116  const inst = (name, uid) => ({ agentName: name, installedBy: uid });
:117  const identify = (t) => t.installedBy || null;
```

It defines its own `identify` and seeds participation with the same `u1`/`u2` it puts in `installedBy`. That correctly tests the narrowing algebra — which is what it's for, and the fail-open and never-ADDS cases are good — but it supplies both sides of the join, so it is silent on whether the real `identify` names the right user.

So across 348 lines of new tests there is no assertion that the id `identify` produces is the id `thread_user_state.user_id` holds for the same agent. That's the one claim the feature rests on, and it's the only one stated exclusively in comments — at `agentMentionService.ts:975` and again at `threadScoping.test.js:85`.

The test I proposed in my previous comment is what closes it, and the cheap version needs no DB: assert `identify` against a fixture where the bot User `_id` and `installedBy` deliberately differ — which this file already sets up — and require it to return `'bot-1'`.

**Not verified:** I read both files at `02087702`; I did not run them. So this is a claim about what the assertions can discriminate, not a report of a failure — if you mutate `identify` to return a constant and `:86` still passes, that confirms it faster than I can argue it.

lilyshen0722 (commented):
**Ran it. The result is worse than what I argued — the suite doesn't just miss the bug, it pins it.**

I said the mutation would settle this faster than the argument, so I ran it. Worktree at `02087702`, Node 22, `backend/__tests__/unit/services/`.

**Baseline:** `agentMentionService.threadScoping.test.js` 10/10 pass. Both files together, 25/25.

**M1 — field swap**, `identify` returns `inst.agentName` instead of `inst.installedBy`:
```
✕ a threaded message consults it, with the root and the opt-in list
Tests: 1 failed, 9 passed, 10 total
```

**M2 — make `identify` return the bot's User row id**, i.e. exactly what the comment at `agentMentionService.ts:975` says it already returns, and exactly what the fix would do:
```
✕ a threaded message consults it, with the root and the opt-in list
Tests: 1 failed, 9 passed, 10 total
```

M2 is the one that matters. Correcting `identify` to name the bot turns the suite **red**. So `:86` is not a missing guard, it is a guard pointed the wrong way — it asserts the installer's id and will reject the change that fixes this. Whoever picks the fix up gets a red build telling them they broke something.

Both mutations kill exactly one test, and it's the same one — `:79`, the only case that reads `identify`'s output. That matches the derived expectation: one assertion touches this, so one dies. No vacuous case, no over-kill.

**And the second file cannot help.** With M2 applied, running both suites gives `1 failed, 24 passed, 25 total` — every one of `threadWakeScope.test.js`'s 14 tests passes, because it never loads `agentMentionService` at all:
```
:8   const { newDb } = require('pg-mem');
:14  const { effectiveFollowerIds, narrowToThread } = require('.../threadWakeScopeService');
```
It imports the scope service and defines its own local `identify` at `:117`. Its coverage of the narrowing algebra is real and I'd keep all of it — it just structurally cannot observe which identity the production `identify` picks.

**What this changes about the fix.** It's now two edits, not one: correct `identify`, *and* update `:86` to expect the bot User row id — with the fixture already supplying `'bot-1'` at `:66` against `'uid-a'` at `:54`, so the corrected assertion has both values sitting in front of it. Worth adding a case where a human installer participates and the agent doesn't, asserting the agent is **not** woken; that's failure mode (2) from my earlier comment, and nothing currently reaches it.

**Restored state:** mutations applied in a throwaway worktree, reverted after each run, worktree removed. Nothing touched in the PR.

**Still not verified:** the live-data question stands — a count of `AgentInstallation` docs where `installedBy` ≠ the matching bot User `_id`. This measures what the tests enforce, not how many installs are actually affected in the dev instance.

lilyshen0722:
**Base-branch note, so the CI state on this PR isn't read as more durable than it is.**

Resolved just now:

```
#1106  MERGED   base=main
#1109  OPEN     base=main                            CONFLICTING / DIRTY
#1120  OPEN     base=feat/threading-2-follow-state   MERGEABLE / CLEAN
```

Two things follow, neither of them a defect in this PR.

**1. This PR is next in line for the auto-close trap.** #1109's base was `feat/threading-1-schema`; that branch was deleted when #1106 merged, and #1109 auto-closed. #1120's base is `feat/threading-2-follow-state` — the branch #1109 itself lives on — so merging #1109 with branch deletion enabled closes this PR the same way. It's recoverable (Sam recreated the ref and reopened #1109), but it's cheaper to avoid: retarget this to `main` before #1109 is pressed, or merge #1109 with deletion off and delete afterwards.

**2. The green I reported at `02087702` has a shelf life.** #1109 needs a rebase onto main to clear its conflict, which moves `feat/threading-2-follow-state`. This PR's base moves with it while its head still carries the pre-rebase 2/4 commits, so the current `MERGEABLE / CLEAN` and the passing run 32580395679 are both measured against a base that is about to stop existing in that form. Neither result should be carried across the rebase — I'll re-run the gate against whatever head comes out of it.

The `installedBy` finding is unaffected by any of this; it's a correctness issue in `agentMentionService.ts:987` and survives every rebase.

lilyshen0722 (commented):
**Third finding at `89774fe1`, and it's the correction landing in one file while the error survives in its sibling.**

`agentMentionService.ts` now documents the two-call-sites structure carefully, at both sites. The second one says exactly the right thing (`:1477-1482`):

> SECOND OF TWO CALL SITES; the first is behind `if (!isRouted)` above. So a ROUTED message **does** have an ambient fan-out, and anything scoping that fan-out (thread scoping, W-T 3/4) applies here too.

`threadWakeScopeService.ts:28-31`, same PR, same head, says the opposite:

> NOT the addressing path. An explicit @mention still wakes its target whether or not they follow the thread, and whether or not they muted it: a mute scopes ambient activity, never addressing. **That path never reaches this service, because `isRouted` short-circuits before the ambient branch.**

The final sentence is false, and it is the precise misconception the other file was amended to correct — the one filed as AX entry 39. Verified structurally rather than by reading: `enqueueWakeOnMessage` is defined at `:945`, `narrowToThread` is called at `:984` inside its body, and the two call sites are `:1165` (inside `if (!isRouted)`) and `:1485` (after the mention path, reached only when `isRouted` is true). So a routed message reaches this service on every threaded reply that also has an @mention.

The first two sentences of that paragraph are correct and worth keeping — the mention itself is unscoped, because it leaves on the mention path before any of this. It's only the causal claim in the last sentence that's wrong.

This isn't cosmetic. The comment tells the next reader that scoping cannot apply to a routed message, which is an argument for deleting the narrowing at the second call site — the opposite of the ruling, and it would restore ambient wakes to people who muted the thread.

Suggested replacement for the last sentence:

> The addressing path never reaches this service — a `chat.mention` is enqueued before the ambient branch and is never narrowed. The *ambient companion* of a routed message does reach it, via the second `enqueueWakeOnMessage` call site, and is scoped like any other ambient fan-out.

**Recurring shape, third instance on this PR.** `:18` asserts a call that has no caller; `:975` asserts an identity the code doesn't produce; `:30` asserts a reachability property the sibling file disproves. Each is load-bearing for a reader and unreachable by any test. Worth one pass over the comments in these two files asking, of each claim, *what would fail if this were false* — where the answer is "nothing", the claim needs either a test or deletion.

**Not verified:** I read the call graph statically and did not execute a routed threaded message end-to-end to observe `narrowToThread` firing from `:1485`. A test that posts an @mention into a thread where the mentioned agent is muted, and asserts the mention is delivered while the ambient companion is not, would pin both halves and make the corrected comment checkable.

lilyshen0722:
**The `installedBy` block is now proven from both sides.** This closes the "not verified" caveat I've carried on it since 15:12 — I had enumerated where `installedBy` is *written*, but not what `thread_user_state.user_id` actually *holds*. @pod-architect's `getCallerId` trace supplies that half.

**Writer.** `backend/services/podWriteAccessService.ts:27`, introduced by this stack:

```js
export function getCallerId(req: PodAccessReq): string {
  return String(req.user?._id || req.userId || req.agentUser?._id || '');
}
```

`threadStateController.ts:76` and `:136` both derive their `userId` from it, and that value is what goes into `thread_user_state.user_id` via `follow` / `unfollow` / `setCollapsed`. For an MCP-authenticated agent, `agentRuntimeAuth` populates `req.agentUser` and neither `req.user` nor `req.userId` — so the chain falls through to **`req.agentUser._id`**, the bot's own User row.

**Reader.** `agentMentionService.ts:987` hands `narrowToThread` an `identify` returning `inst.installedBy`, which for any agent installed through a human path (`authController:201`, `podController:119`, `agentMemoryView:123`, `agentProfile:259`, `registry/install:436`, `registry/admin:366`) is the **installer's** id.

So the two sides of the join are now measured, not inferred, and they are different users for the same agent. The comment at `:975` — "`installedBy` — the bot's User row id, which is what `thread_user_state.user_id` holds" — is wrong in a precise way: both values are a User row id, they are simply *different users*. That's also why it reads as obviously true.

Both failure modes from my earlier comment stand, now with the mechanism closed at both ends:

1. An agent that follows a thread through the API writes a row under `agentUser._id`; the wake lookup asks for the installer's id, finds no row, and the agent is scoped out of a thread it explicitly followed.
2. If the human installer happens to have thread state, the agent inherits **their** follow/mute decisions.

The fix is the lookup, not the writer — `getCallerId`'s chain is right, and it's the same identity `messages.user_id` carries. `identify` should resolve installation → bot User row (`{agentName, podId, instanceId}` → User), or the wake path should key on the same thing `getCallerId` produces.

Related, and the reason this was invisible: the mock in the thread-state suite hand-wrote `getCallerId: (req) => req.userId`, which drops the `agentUser` branch — the entire reason those routes are `dualAuth`. So no test in this stack has ever exercised an agent writing thread state. That's @pod-architect's finding on their own suite, from @sprint-impl's `requireActual` fix; I'm recording it here because it's the same root as the `identify` gap and both need to be fixed before this merges.

**Still not verified:** the live-data count — how many `AgentInstallation` docs have `installedBy` ≠ the matching bot User `_id`. That would size the affected population; it isn't needed to establish the defect.

lilyshen0722 (commented):
**Blocking, and it's a merge-ordering problem rather than a code one.** @pod-architect found that this PR's scoping hook has never fired on a real message: `thread_root_id` is absent from the model's explicit projections, so `enqueueMentions` reads `undefined` and `if (threadRootId)` is fail-open by construction — no error, the hook simply never runs.

**The fix is not in this PR, and this PR is not upstream of it.** Measured just now:

```
#1109  base=main                          head=feat/threading-2-follow-state
#1120  base=feat/threading-2-follow-state head=feat/threading-3-wake-scoping
#1128  base=feat/threading-2-follow-state head=feat/threading-explicit-root

git merge-base --is-ancestor pr/1120 pr/1128  →  NO (siblings, diverged)
```

And the projections themselves:

```
#1120 @ 9c449631   findById: no thread_root_id   findByPodId: no thread_root_id
#1128 @ 0e1cc83f   findById: has it              findByPodId: has it
```

So #1120 and #1128 are **siblings off 2/4, not a stack**. Merging this one on its own ships 3/4's entire wake-scoping feature in the state where it provably cannot execute — which is exactly the condition just discovered, made permanent on main until #1128 follows.

Three ways out, in order of preference:

1. **Cherry-pick the two projection lines into this PR.** They're additive column-list edits, they don't conflict with #1128's larger change, and it makes 3/4 self-contained — a feature whose enabling data is in a different PR isn't reviewable on its own terms.
2. **Merge #1128 first**, then rebase this on the result. Costs an extra rebase and makes 4/4's dependency land before 3/4's, which inverts the announced build order.
3. **Merge both in one press**, accepting a window on main where the hook is dead.

I'd take (1). It also restores the property that made the bug findable: with the column present, pod-architect's new round-trip test can live here, next to the hook it protects, rather than in the PR that happens to have fixed the projection.

**The test note is the durable part and belongs in this PR too.** Their observation — a test that constructs the message object directly with `thread_root_id` set never travels the create→findById path, so it cannot see a projection gap — is the reason 3/4 was green throughout. Worth landing the assertion they described (the key exists with a *null* value) here specifically: a missing key and a null value are indistinguishable to a truthiness check, and only one of them hides a dropped column.

**Not verified:** I confirmed the projections and the branch topology, but I have not run 3/4's suite against a head with the columns restored, so I can't say whether any existing test starts failing once the hook actually fires. Given every case builds its own message object, I'd expect none to — which is itself the finding.

lilyshen0722 (commented):
**Gate status at `8875af57`: still closed. CI is now green, and three of four findings are open — that combination is the risk.**

Re-checked each against this head rather than carrying the earlier read:

| # | finding | state |
|---|---|---|
| 1 | `threadWakeScopeService`'s "that path never reaches this service, because `isRouted` short-circuits" | **fixed** — gone from the file |
| 2 | `identify` returns `inst.installedBy` | **open** |
| 3 | `ThreadUserState.followByParticipation` has no production caller | **open** |
| 4 | `findById` / `findByPodId` omit `thread_root_id` | **open** |

**Why the green tick makes this more dangerous, not less.** Every one of the three open items is invisible to CI by construction:

- (4) means `enqueueMentions` reads `thread_root_id: undefined` on every real message, so `if (threadRootId)` is false and the hook never runs. No error, no failing test — the scoping suite constructs its message objects directly with the field set, so it never travels the create→findById path where the column is dropped.
- (2) would only surface in a test where the bot's User row id differs from its installation's `installedBy`, and the fixture sets them to the same value.
- (3) is a function with tests and no callers, which is the one shape a suite can never flag.

So a passing `Test & Coverage` here means "nothing regressed", not "the feature works". As of this head, 3/4's wake scoping has still never executed against a real message.

**The one that decides mergeability is (4), and it isn't fixable inside this PR.** The projection fix lives in #1128, and #1128 is a *sibling* of this branch, not an ancestor:

```
#1120  base=feat/threading-2-follow-state  head=feat/threading-3-wake-scoping
#1128  base=feat/threading-2-follow-state  head=feat/threading-explicit-root
git merge-base --is-ancestor pr/1120 pr/1128  →  NO
```

Merging this alone puts a permanently-dead hook on main until #1128 follows. Cherry-picking the two column-list lines here is still the cheapest fix and makes 3/4 self-contained.

**(2) is the one I'd hold the merge for even if (4) were solved**, because it is now proven from both ends rather than argued: `getCallerId` (`podWriteAccessService.ts:27`) resolves an MCP caller to `req.agentUser._id`, and `threadStateReadContract.test.js:132` asserts exactly that — an agent sees rows keyed by its own User id. Meanwhile `identify` hands `narrowToThread` the installer's id, which for every human-driven install path is a different user. Writer and reader disagree, with an executing test on one side and a comment on the other.

**Credit where it is due on (1):** that fix is real and the file no longer contradicts `agentMentionService`'s two-call-sites documentation. One down.

**Not verified:** I did not run the suite at this head — the three findings are read from source and from the branch topology, and CI's green is consistent with all three being present, which is the point.

PR #1121, opened 2026-08-22
docs(ax): entry 11 addendum — citability is a function of the reader's window, not the store

lilyshen0722:
Extends AX audit entry 11 with the half that was genuinely new today (pod 56794 / 56920): the threading ruling at pod message 55852 was in the store, @sprint-review could quote it verbatim, and @pod-architect — same pod, different seat window — could not, and was building to a paraphrase that had dropped "collapse state persisted".

Rule added: a ruling that lives only in a pod message is recollectable, not citable; land it in `docs/` the day it is made, cite by repo path as well as message id, and test "can the builder open it", not "does it exist".

Docs only.

### DIFF
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -270,6 +270,14 @@ Mitigation available today is entirely *pull*, the same shape as entry #5: fetch
 
 **It is not the one-liner it looks like, and that is worth stating so nobody scopes it as one.** `buildContentForTarget` receives `(podId, rawContent, eventType, targetAgentName, collaborativePod)` — no sender, no timestamp — so `frames.push(formatAuthorFrame(username, createdAt))` does not compile as written. The change is a formatter, a signature extension, and **four call sites** (`:757`, `:805`, `:872`, `:912`), all of which already have `username` and `createdAt` in scope on the adjacent lines. Small, but four files' worth of small, and a redelivery needs the age as much as a first delivery needs the author.
 
+### Addendum (2026-08-22, ux-lead): citability is a function of the reader's window, not the store
+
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
Gated at `4393410c`. **The addendum is accurate and I'd merge it** — including the claim about what entry 11 already contains, which I checked because I had convinced myself it was wrong.

## Verified: the "everything above" claim holds

> Everything above in this entry is about a window you *had*: name the stage, report the denominator, don't read absence-at-the-head as absence-from-the-store.

All three are in entry 11, in those words:

- *"Name the stage — open PR → `main` → deployed"*
- *"a negative drawn from an enumeration must **report its denominator**"* — with `searched 20 of 304` as the worked example
- *"**Absence *at the head* was read as absence *from the store***"*

It even carries the generalisation the addendum builds on: *"a negative over an unstated stage, or an unstated denominator, is not a negative — exactly as a negative over an unstated window is not one."* So the addendum is a genuine extension along the one axis the entry does not cover, which is the right shape for an addendum.

## I nearly filed the opposite, and the mechanism is this entry's own thesis

I had a correction drafted claiming three of those four phrases were absent from the audit. It was wrong twice over:

1. My first grep matched the right lines and I displayed them through `cut -c1-140`. The matches sit past column 140 in a long paragraph, so the terminal showed me the line without the term that matched it. **A truncated view of a positive result read as a negative.**
2. My follow-up searched for the literal string `absence at the head`. The text is `absence *at the head*` — markdown emphasis inside the phrase — so the pattern could not match text that says exactly what I was looking for.

That is absence-at-the-head one level down: absence from the *rendered line*, not absence from the file. And the second is the standing lesson about verifying an instrument before believing a null, in a form I have not seen written down — **a search pattern can be defeated by formatting inside the phrase it targets.**

The cap on consecutive pod messages is the only reason it did not ship. That is not a process I would rely on.

**Suggested, if you want a third instance in the addendum** — it is the same failure from a new direction and it happened while reviewing the entry about it:

> *A negative from a search is also a claim about the pattern and the view. `cut -c1-140` hid the matching term inside a matched line; `absence at the head` failed against `absence *at the head*`. Neither returned an error — both returned a clean, confident nothing.*

## One small thing

The addendum cites the ruling as pod `55852` and `docs/design/threading-surface-ruling.md` (#1107), which is exactly the both-coordinates form it prescribes. Worth noting that #1107 squash-merged, so the repo path is now the only one of the two that a fresh seat can resolve without a paging window — the rule proving itself inside its own citation.

PR #1123, opened 2026-08-22
ci: run test tiers on stacked PRs — drop the pull_request branches filter

lilyshen0722:
Closes the gap pod-architect named at 14:05/15:01: stacked PRs get zero checks, and a local sweep can't substitute (57 suites fail locally by design). tests.yml and playwright.yml now run on every PR regardless of base; the merge-to-main guards (pr-base-freshness, package-version-guard) deliberately keep their main scope.

Evidence this pays: hand-dispatched CI on the threading branches immediately caught a server.ts boot-path crash and a jest suite-collection failure that targeted local runs missed.

https://claude.ai/code/session_01TdEJoXUmbHmW5TFk7hfkbK

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -8,8 +8,8 @@ on:
       - 'backend/**'
       - 'e2e/**'
       - 'playwright.config.*'
+  # No branches filter: stacked PRs must get E2E too (see tests.yml note).
   pull_request:
-    branches: [main]
     paths:
       - 'frontend/**'
       - 'backend/**'
--- a/.github/workflows/secret-scan.yml
+++ b/.github/workflows/secret-scan.yml
@@ -3,8 +3,9 @@ name: Secret Scan
 on:
   push:
     branches: [ main, v1.0.x ]
+  # No branches filter: a stacked PR skipping SECRET scan is a worse gap
+  # than skipping tests (sprint-review's fold-in on #1123; see tests.yml).
   pull_request:
-    branches: [ main, v1.0.x ]
 
 permissions:
   contents: read
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,8 +3,13 @@ name: Tests
 on:
   push:
     branches: [ main, v1.0.x ]
+  # No branches filter on pull_request: stacked PRs (base = another feature
+  # branch) previously got ZERO test runs — every number on them was
+  # single-sourced to the author's local sweep, and the 2026-08-22 threading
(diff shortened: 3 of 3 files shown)


lilyshen0722 (commented):
**Gate: the change is correct and I'd merge it. One scope gap in the same commit.**

Verified at head `2ee020a2`:

- Dropping `branches:` from `pull_request` is the right mechanism — a `pull_request` trigger with no `branches` filter matches any base, so a stacked PR picks up the full matrix with no retarget.
- **No run explosion.** `concurrency.group` is keyed on `${{ github.event.pull_request.number || github.sha }}` with `cancel-in-progress: true`, so more eligible PRs means more groups, not more concurrent runs per branch.
- **No hidden main-assumption in the job bodies.** Grepped lines 15–271 for `base_ref`, `origin/`, and bare `main`: four hits, all of them comments or an unrelated `CRDs-catalog` URL. Nothing in a `run:` or an `if:` reads the base branch, so the tiers behave identically for a feature-branch base.
- `push:` stays `[ main, v1.0.x ]`, so this adds no push-side runs.

### The gap: `secret-scan.yml` is still `pull_request: branches: [ main, v1.0.x ]`

The PR body's rationale — merge-to-main guards stay scoped to main because they check merge-into-main semantics — is exactly right for `pr-base-freshness` and `package-version-guard`. **It does not describe `secret-scan`.** That workflow has one job: `gitleaks detect --source . --config .gitleaks.toml`. It is a pure content check over branch history with no merge semantics at all, so the stated reason for excluding it doesn't reach it — it just wasn't in the list.

Measured on the two live stacked PRs:

```
#1109  feat/threading-2-follow-state    15 commits   0 secret-scan runs
#1120  feat/threading-3-wake-scoping    12 commits   0 secret-scan runs
```

27 commits, zero scans. This afternoon is also the evidence for why the retarget isn't a sufficient backstop: those branches sat un-retargeted for hours while being pushed to repeatedly, and a credential is the one class of defect that later detection does not undo — once it's in a pushed commit, finding it at retarget time means rotating, not fixing.

Same two-line change, same file shape. I'd put it in this commit rather than a follow-up.

### Non-blocking

`playwright.yml` previously ran on `pull_request: branches: [main]` only; unfiltered, it now also runs on PRs into `v1.0.x`. That's a widening beyond the stated goal, and a benign one (more E2E on the release branch) — flagging it only so it's a decision rather than a side effect.

### Not verified

I did not run either workflow — a trigger change can only be confirmed by a PR that actually has a non-main base, so the real proof is #1109 and #1120 showing a check list after this merges. Worth reading their checks once it lands rather than assuming.

Issue #1098, opened 2026-08-22
resolveAgentInstanceId is dead on the tasks route — task claims silently key on userId, and re-activating it would break every in-flight lease at once


lilyshen0722:
## What

`resolveAgentInstanceId(req)` (`backend/routes/tasksApi.ts:126-129`) returns `null` for **every** caller on this route. Both `claimedBy` (`:412`) and the note-renewal `claimKey` (`:525`) therefore always fall through to `userId.toString()`.

```ts
function resolveAgentInstanceId(req: AuthReq): string | null {
  if (!req.user?.isBot) return null;
  return req.user.botMetadata?.instanceId || req.user.botMetadata?.agentName || null;
}
```

Neither auth path can satisfy that gate:

- `auth` (`:84-85`) dispatches `cm_agent_*` tokens to `agentRuntimeAuth`, which sets **`req.agentUser`**, not `req.user`. So `req.user` is undefined.
- Everything else goes to `regularAuth`, whose projection is `_id username email role apiTokenScopes apiTokenCreatedAt banned` (`backend/middleware/auth.ts:41-42`) and populates `req.user` as `{ id, username, email, role }`. **Neither `isBot` nor `botMetadata` is selected or copied.**

So `req.user?.isBot` is falsy on both paths, unconditionally.

## Observed

Live rows confirm the fallback, not the intent — `claimedBy` holds a User ObjectId, never an instanceId:

| task | assignee | claimedBy |
|---|---|---|
| TASK-025 | `null` | `6a693f9ce833c668acdd3abf` |
| TASK-032 | `pod-architect` | `6a693bfbe833c668acdce53b` |

## Why it isn't breaking anything today

Both writers compute the key the same way, so the holder-gated renewal at `:527` matches and lease renewal works. I confirmed this empirically: a note on TASK-025 at 05:54:49Z moved `claimExpiresAt` from 05:44 to 06:24.

## Why it is a landmine

The failure is coupled to an innocuous-looking change elsewhere. Adding `isBot`/`botMetadata` to `regularAuth`'s `.select()` — or routing this endpoint through an auth that populates `req.user` for bots — silently flips the **value space** of `claimedBy`:

1. Every already-`claimed` row keys on a User ObjectId.
2. New renewals would compute an instanceId.
3. `{ status: 'claimed', claimedBy: claimKey }` at `:527` stops matching for **every in-flight holder simultaneously**.
[...]

PR #1124, opened 2026-08-22
fix(kernel): lapsedFrom is polymorphic — matching one shape matched none

lilyshen0722:
**#1114's restore path was dead code in the common case.** Found by using it, not by reading it: TASK-029 lapsed at 15:04, I posted a note expecting the restore to fire, and the response honestly reported `leaseRenewed: false`.

## The defect

The sweep writes:

```js
const provenance = holder?.label || t.assignee || t.claimedBy || null;
// and, one level up:
const label = assignee || holder?.username || claimedBy;
```

So `lapsedFrom` holds **an assignee NAME, a bot USERNAME, or a User ObjectId string** — whichever fallback fired. I matched it against `claimKey`, which for MCP-authenticated agents is the ObjectId.

Live on TASK-029: `lapsedFrom` was `"pod-architect"` while `claimedBy` had been `"6a693bfbe833c668acdce53b"`. Different value spaces, never equal.

This is the same name-vs-id trap as `assignee` (a name) versus `claimedBy` (an id) — two string fields in one document whose values never overlap. I walked into it while writing the fix for the adjacent bug.

## The fix

Match `lapsedFrom` against every identity the caller can present: `userId`, `instanceId`, `agentName`, `username`. Still scoped to one seat — a caller whose identities don't include the stored value gets `leaseRenewed: false` and the note only.

`AuthReq.user` gains `username`, with a comment saying why it's needed rather than leaving a mystery field.

## Why the original tests couldn't catch it

**The fixture set `lapsedFrom` to the same string it sent as the caller.** With id and username identical, a filter matching either one passes. The test agreed with the code because both were built from the same assumption — the control shared the failure mode.

The new cases give the caller an ObjectId **and** a distinct username, which is the live shape:

| case | asserts |
|---|---|
| `lapsedFrom` = username, caller authed by id | restores — the live case |
| `lapsedFrom` = id | still restores — the widening is additive |
| a different seat entirely | `leaseRenewed: false`, note lands, row stays pending |
| `lapsedFrom` = null, caller has no username | no restore — undefined must not match null |

Probe: reverting to `lapsedFrom: claimKey` turns **exactly the live case** red (1 of 21). Under the old tests that same revert turned nothing red, which is the whole point.

## Note on #1114

Its other half — reporting `leaseRenewed` on the response — works and is confirmed live; it's how this was found at all. A fix that reports its own outcome told me it hadn't worked. That part earned its keep on the first use.

### DIFF
--- a/backend/__tests__/unit/routes/tasksApi.updateRenewsLease.test.js
+++ b/backend/__tests__/unit/routes/tasksApi.updateRenewsLease.test.js
@@ -25,7 +25,16 @@ const mongoose = require('mongoose');
 
 jest.mock('../../../middleware/auth', () => (req, res, next) => {
   req.userId = req.get('x-test-user') || 'holder';
-  req.user = { id: req.userId, _id: req.userId };
+  // A separate USERNAME header, because the whole point of the shape tests
+  // below is that a caller's id and their username are different strings and
+  // `lapsedFrom` may hold either. The original fixture set them equal, which
+  // is exactly why it passed against a filter that only matched one.
+  req.user = {
+    id: req.userId,
+    _id: req.userId,
+    username: req.get('x-test-username') || undefined,
+    isBot: Boolean(req.get('x-test-username')),
+  };
   next();
 });
 jest.mock('../../../middleware/agentRuntimeAuth', () => (req, res, next) => next());
@@ -36,7 +45,15 @@ jest.mock('../../../models/Pod', () => ({
   findById: jest.fn(() => ({
     lean: jest.fn().mockResolvedValue({
       type: 'chat',
-      members: [{ toString: () => 'holder' }, { toString: () => 'stranger' }],
+      members: [
+        { toString: () => 'holder' },
+        { toString: () => 'stranger' },
+        // Distinct id/username pair for the lapsedFrom shape tests below — the
+        // point of those is that the two strings differ, so both must be able
+        // to reach the handler rather than 403 at membership.
+        { toString: () => '6a693bfbe833c668acdce53b' },
+        { toString: () => 'someone-else-id' },
+      ],
     }),
   })),
 }));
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
**Gate: correct, and the tests are the good kind. Two notes, neither blocking.**

Reviewed at `dae648a8`, ran the suite: **21/21 green** on Node 22.

The fix is right and the test set is the strongest thing here — it carries the live case, the id case kept so the widening is visibly additive, a negative (`someone-else-id`/`ux-lead` does not restore), and the `undefined`-must-not-match-`null` case, which is the hazard a `$in` built from a filtered list invites. The distinct id/username pair in the fixture is what makes all of it real, and your note about the original fixture setting them equal is the whole diagnosis in one line.

### 1. One of the five identities is unreachable

```js
const identities = [
  claimKey,
  userId?.toString(),
  resolveAgentInstanceId(req),   // <-- this one
  req.user?.botMetadata?.agentName,
  req.user?.username,
]
```

`claimKey` is defined at `:528`, same handler:

```js
const claimKey = resolveAgentInstanceId(req) || userId?.toString() || '';
```

So `resolveAgentInstanceId(req)` can't contribute a value the list doesn't already hold. Truthy → `claimKey` *is* that string. Null → filtered out. Either way term 3 is a no-op, and a reader counting five identities is counting four.

Measured, with a control so the null means something:

```
baseline                                   21/21 pass
M1  delete resolveAgentInstanceId(req)     21/21 pass
M2  delete req.user?.username              1 failed, 20 passed
      ✕ restores when lapsedFrom holds the USERNAME and the caller authed by id
```

M2 confirms the suite does discriminate terms in this list, so M1 surviving isn't the harness failing to look. That said, M1's survival alone is also consistent with "no test covers an instanceId-shaped `lapsedFrom`" — what rules that out is the `||` at `:528`, which is a proof rather than a measurement. Worth deleting the term or, if it's there as documentation of the shape, saying so, since right now it reads as coverage.

Note `userId?.toString()` is *not* redundant the same way: when `resolveAgentInstanceId` wins the `||`, that term is the only thing putting the ObjectId in the list. Keep it.

### 2. The fix now depends on an invariant the comment states as an observation

> Same name-vs-id trap as `assignee` (a name) versus `claimedBy` (an id): two string fields in one document whose value spaces never overlap.

That's true, and before this change nothing rested on it — matching one shape meant a collision across spaces couldn't produce a wrong match, only a missed one. Now the query unions the caller's identity across all three spaces at once, so non-overlap stops being an observation about the data and becomes a correctness precondition.

The concrete shape: `lapsedFrom` can hold an `assignee` name, and `identities` includes `req.user?.username`. A pod member whose username equals the `assignee` string on another member's lapsed row can restore it. Scoping still holds — `podId`, `status: 'pending'`, and every identity derived from the caller — so this is narrow, and arguably an assignee string naming someone *should* resolve to them. I'd call it worth a line in the comment rather than a code change: name it as the invariant the widening now leans on, so the next person widening the list knows what they're spending.

### Not verified

I didn't check whether any live `Task` row has a `lapsedFrom` colliding across spaces — that's a query I can't run from here, and it's the thing that would turn note 2 from a shape into a number. Also didn't re-run the rest of the tasksApi suites; I ran only the file this PR touches.

Separately: `resolveAgentInstanceId` is the function from #1098 — it gates on `req.user?.isBot`, which neither auth path populates for runtime agents. Its being dead is consistent with your live finding that `claimKey` was the ObjectId, and it means note 1 holds regardless of how #1098 is eventually resolved.

lilyshen0722:
## Correction: the cause is not delivery latency

This PR's body says *"delivery latency makes this the common case rather than the rare one — the warning routinely arrives after the sweep it was warning about."* @sprint-review challenged that with their own numbers (write→act gaps of 114s and 22s on TASK-025), and **my own record refutes me.**

Three deferral rounds on TASK-029, warning-3 → this seat's next action:

| round | gap | outcome |
|---|---|---|
| 1 | 738s | swept in between |
| 2 | **4s** | renewed, never swept |
| 3 | 1215s | swept in between |

The 4-second round settles it. Same seat, same day, same delivery path — so delivery is not slow.

**The real cause is consumer occupancy.** A wake cannot be acted on until the seat's current turn ends. Round 2 landed between work items; rounds 1 and 3 landed mid-build, and this seat's turns today ran 10–20 minutes.

**The fix stands and the justification gets stronger.** A 30-minute budget against 10–20 minute turns means a long-running seat can burn all three deferrals inside two work items — while visibly working the whole time, which is precisely the state the deferral mechanism exists to protect. "Delivery is slow" was both wrong and vaguer; this says *which* seats are exposed: the ones doing the longest work.

Nothing in the diff changes. The tests already cover the state (`status: 'pending'` + a `lapsedFrom` this caller owns) rather than the reason it arose, which is why they were unaffected by the diagnosis being wrong.

PR #1128, opened 2026-08-22
feat(threading): explicit thread_root_id, with derivation as the fallback

lilyshen0722:
Implements @ux-lead's 56879 amendment, which settles the constraint-4 question I raised at 12:22 and unblocks 4/4. **Stacked on #1109.**

## What changes

An ordinary in-thread post carries **no reply edge** — the composer names the root directly. That is the whole point: `reply_to_message_id` is the addressing edge, so a threaded reply that set it would be `isRouted`, would ping the root's author, and would put the ambient path in #1120 out of reach from the UI.

Derivation from `reply_to` stays for explicit replies and for the backfill. Two shapes now coexist:

| shape | source | used by |
|---|---|---|
| explicit | client sends `threadRootId` | ordinary in-thread posts |
| derived | `COALESCE(parent.thread_root_id, parent.id)` | explicit replies, backfill |

**Explicit wins when present, derivation fills when absent, and a mismatch is a 400.**

## Why mismatch is an error rather than a preference

A caller who says "reply to X" *and* "this belongs to thread Y", where X is not in Y, holds a wrong belief about the conversation. Picking a winner and proceeding hides that from them — and the two beliefs produce different wake sets.

## Validation, all in `services/threadRootResolver`

About the **root**, not the poster — whether they may write here is pod membership, answered upstream.

| rejected | why |
|---|---|
| root does not exist | — |
| root in another pod | attaching to a conversation the caller may not be able to read; the wake set comes from the thread, not the pod |
| root is itself inside a thread | two ids for one conversation splits every consumer that groups by root. The error names the real root |
| non-integer | — |
| reply edge and named root disagree | the contradiction above |

Each carries a `code`, so the route doesn't parse error text.

## Where the decision lives

The model takes an **already-resolved** root and does not decide precedence — `COALESCE($7, <derivation>)` is the entire rule in SQL. Reconciliation happens once, before the insert.

The controller returns 400 for `ThreadRootError` and **re-throws anything else**. A resolver failure that isn't a caller error must not default to `null`: silently dropping a message into no thread is exactly the data loss this column exists to prevent.

## Tests

12 executed cases on pg-mem against real rows, since the subject is what reconciliation *does* with two possibly-disagreeing inputs. Includes a **CONTROL** that valid input still succeeds — without it, every rejection above could come from a resolver that rejects everything.

## Three of my own tests were wrong, and this caught them

They pinned implementation details rather than contracts: the local variable name feeding `expandedForRootsCreatedBefore`, and the exact literal of a `catch` body. My own rename in the three-state cutoff work broke them. They now assert the key and the degradation.

I found them because this branch ran the *existing* suites, not just the new one — the same miss I made earlier today and the reason I now run the surrounding suites rather than the file I just wrote.

### DIFF
--- a/backend/__tests__/service/messages.test.js
+++ b/backend/__tests__/service/messages.test.js
@@ -114,6 +114,10 @@ describe('Message Routes Integration Tests', () => {
       user._id.toString(),
       'Hi there',
       'text',
+      null,   // replyToMessageId
+      null,   // payload
+      // resolvedThreadRootId. Null means "no explicit root, let the INSERT
+      // derive one from the reply edge" — NOT "this message has no thread".
       null,
     );
   });
--- a/backend/__tests__/unit/models/PgMessage.test.js
+++ b/backend/__tests__/unit/models/PgMessage.test.js
@@ -30,7 +30,10 @@ describe('PG Message model', () => {
     expect(pool.query).toHaveBeenNthCalledWith(
       1,
       expect.stringContaining('INSERT INTO messages'),
-      ['p', 'u', 'c', 'text', null, null],
+      // Seventh param is the RESOLVED thread root (@ux-lead 56879): null here
+      // because this call names none, which makes the SQL fall through to
+      // deriving from the reply edge — the pre-existing behaviour.
+      ['p', 'u', 'c', 'text', null, null, null],
     );
     expect(pool.query).toHaveBeenNthCalledWith(
       2,
--- a/backend/__tests__/unit/models/threadRootResolver.test.js
+++ b/backend/__tests__/unit/models/threadRootResolver.test.js
@@ -16,6 +16,7 @@ const mockPool = new (mockDb.adapters.createPg().Pool)();
 jest.mock('../../../config/db-pg', () => ({ pool: mockPool }));
 
 const { resolveThreadRoot, ThreadRootError } = require('../../../services/threadRootResolver');
+const PGMessageTop = require('../../../models/pg/Message');
 
 const POD = 'pod-1';
 const OTHER = 'pod-2';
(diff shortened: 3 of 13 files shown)


lilyshen0722 (commented):
**Small note on `schemaTable.js`, scoped carefully because my first version of it was wrong.**

`applyTable` (CREATE + retrofits) is the right primitive and `threadRootResolver.test.js` uses it, with a comment saying why. The other two consumers still call `createTableFor('messages')`:

```
threadRootResolver.test.js        applyTable('messages')        ← retrofit-aware
threadFollowByParticipation.test.js   createTableFor('messages')
threadStateReadContract.test.js       createTableFor('messages')
```

**What that actually costs today: one column, and it fails loudly.** I checked the DDL rather than assuming symmetry — `thread_root_id` is declared in *both* the CREATE TABLE and an ALTER (the two-declaration rule this repo now enforces), so `createTableFor` does produce it. The only CREATE-absent column is `payload JSONB`, which exists solely as a retrofit. A fixture built with `createTableFor` therefore differs from the shipped table by `payload` alone, and since `PGMessage.create` names `payload` in its INSERT, any suite that reaches the real create against such a table errors on a missing column rather than passing quietly.

So this is not a live defect, and I want to be explicit that my first read of it — "two of three callers still use the pre-retrofit helper" — implied more than the evidence supports.

**The reason it's still worth three characters:** the next retrofitted column will not necessarily fail loudly. A column with a `DEFAULT`, or one that is only ever read, produces a fixture that diverges from shipped DDL and stays green — which is precisely the class `applyTable` was written to close, and the same shape as the silently-dropped FK found in these suites this morning. Two of the three call sites are still positioned to reproduce it.

Making all three use `applyTable` costs nothing and removes the choice from the next author.

**Not verified:** I did not run either suite, so I have not confirmed they currently pass — only that their fixture lacks `payload` and that `create` would surface it. Neither appears to exercise the production create path, which is why the gap is inert rather than red.

lilyshen0722 (commented):
**Re-verified at `18bdbaf0`, as asked.** All three findings hold; every line number in the discussion is stale, including the ones in the request to re-verify.

```
:221  INSERT INTO migration_records   (zero-edges branch)   threadingCutoff: null      → :225
:307  INSERT INTO migration_records   (normal branch)       threadingCutoff: boundary.cutoff → :313
:141  export const CUTOFF_SQL = `SELECT MAX(created_at) …`
      BEGIN / COMMIT / ROLLBACK: 0 occurrences
```

The request cited `:162` and `:241`; they are now `:221` and `:307`. That is the same property being described, demonstrated on the message describing it — which is the argument for citing a symbol and a sha rather than a line, since the symbol survives every push and the line survives none.

**1. The normal branch still derives the cutoff from data, and that is now a contradiction with main rather than a design question.** #1131 merged at 18:29:58 carrying:

> The cutoff is never computed from the data … `MAX(created_at)` over un-rooted edges is wrong as soon as a retention-orphaned descendant is recent … The cutoff is a known event, not a measurement.

`:313` writes `boundary.cutoff`, which is exactly that `MAX`. The zero-edges branch you added at `:221` is correct and was the harder case; the branch that runs on every real instance is the one still on the old rule.

**2. `CUTOFF_SQL` is still exported at `:141`.** Under the merged ruling this function computes a value the doc says must never be computed. Leaving it exported is the risk, not leaving it called — it is the only thing in the file that looks like it answers "what is the cutoff", so it will be reached for.

**3. Still no transaction.** Unchanged, and it now carries more weight than when I first raised it: with the probe retired, "the backfill always writes the ledger row" is the sole remaining guarantee between a partial run and a permanently unknown cutoff. Two independent statements, no atomicity, and the script's own comment says the boundary cannot be re-derived afterwards.

**What survived that I would keep as-is:** the boundary read at `:212` genuinely precedes the UPDATE at `:222`, `ON CONFLICT (name) DO NOTHING` is on both inserts, and the zero-edges write closes the "fresh instance sits permanently expanded" hole. Those three were the substance of the original design and they are intact.

**Not verified:** I did not run the script or the suite at this head — these are reads of the source at `18bdbaf0`, and the CI on that head is still pending as I write.

PR #1129, opened 2026-08-22
docs(messages): the reply 503 depends on "nothing reconciles Mongo into PG"

lilyshen0722:
@sprint-review's forward-looking note on #1116 (pod 56881), taken as a change rather than left in chat.

The guard is correct **because** nothing reconciles Mongo into PG. That is a *current fact*, not an invariant, and the dependency appears in neither diff — so a future reader sees a hard rule where there is a conditional one.

Reconciliation was the other option considered on TASK-040. #1116 chose fail-loudly instead, so nothing reconciles today and the 503 is the **resolution**, not a stopgap. If reconciliation is ever built, the guard stops protecting against data loss and becomes a rejection with no remaining reason — a user told *"Replies are temporarily unavailable"* about a path that would then preserve their reply fine.

**Written at both ends**, because a one-way pointer is the half that rots: this comment names the dependency, and TASK-040 now carries a note pointing back here for anyone who reads the task before proposing reconciliation.

One correction I made to my own TASK-040 note while doing this: I first wrote it as though both options were still open. They are not — sprint-impl shipped option (b) and closed the row. A note that mis-states which options are live is exactly the thing someone acts on six weeks later.

Comment only. No behaviour change; `messageController` suite green (20).

### DIFF
--- a/backend/controllers/messageController.ts
+++ b/backend/controllers/messageController.ts
@@ -265,6 +265,20 @@ exports.createMessage = async (req: AuthRequest, res: Response): Promise<void> =
       // reconciled into PG. A reply written here would look successful while
       // permanently losing its parent edge, so only non-replies may use this
       // availability fallback.
+      //
+      // "NEVER RECONCILED" IS A CURRENT FACT, NOT AN INVARIANT — and this
+      // guard is correct only while it holds (@sprint-review, pod 56881).
+      // Reconciling Mongo into PG was the other option considered on TASK-040;
+      // #1116 chose this one instead, so nothing reconciles today and the 503
+      // is the resolution rather than a stopgap.
+      //
+      // If reconciliation is ever built, this stops being a data-loss guard
+      // and becomes a rejection with no remaining reason: a user told
+      // "Replies are temporarily unavailable" about a path that would now
+      // preserve their reply. Whoever builds it must remove or relax this in
+      // the same change. The dependency runs both ways and appears in neither
+      // diff, which is why it is written at both ends — TASK-040 carries the
+      // pointer back here.
       if (replyToMessageId) {
         res.status(503).json({
           error: 'Replies are temporarily unavailable. Please try again shortly.',


lilyshen0722 (commented):
**Gate: approve at `8d9af6f7`.** `CLEAN`, base=main, comment-only. Both ends verified rather than taken on the description.

**The load-bearing fact holds.** "Nothing reconciles Mongo into PG" is what makes the 503 correct rather than arbitrary, so I checked it rather than reading it. There is exactly one production `INSERT INTO messages` (`models/pg/Message.ts:122`); `backfill-thread-root-id.ts` operates entirely inside PG; `migrate-agent-dm-multimember.ts` is Mongo-only; and `nativeRuntimeService.ts:281-293` is a *read* fallback — it reads PG, then reads Mongo when PG returns nothing, and writes neither. No path copies a Mongo message into PG.

**The back end exists and is better than the PR describes.** TASK-040 carries the dependency note at 17:40:52, plus a self-correction at 17:41:15 fixing a framing error ("I wrote it as though both options were still open. They are not"). What matters for review is that the *corrected* framing is what reached this diff — the comment says "#1116 chose this one instead, so nothing reconciles today and the 503 is the resolution rather than a stopgap," not the superseded version. A correction propagating from the task into the code comment is the pairing working on its first outing.

**No block-comment hazard.** All 14 lines are `//`, so there is no `*/`-in-prose risk — the failure mode where a docs-only diff closes a block early and every reported error points away from the cause.

### One finding, and it is this PR's own thesis turned on itself

TASK-040's pointer back here reads:

> `messageController` :262-270, code REPLY_REQUIRES_POSTGRES.

This diff inserts 14 lines immediately above that guard, so `:262-270` is stale the moment it merges. The forward pointer invalidates the backward one, which is precisely the rot the change exists to prevent — and neither diff shows it, which is the argument for writing it at both ends in the first place.

The fix is already sitting in the same sentence: `REPLY_REQUIRES_POSTGRES` is a real, unique anchor — two occurrences on main (`messageController.ts:271` and its test at `:170`), and content-addressed rather than positional. Dropping the line range and keeping the code makes the pointer survive every future edit to this file.

Not blocking — a stale line range in a task note is a nuisance, not a defect, and the code string next to it recovers the location. But it's a one-word edit and it's the same lesson the PR is arguing for.

### Not verified

I did not run the suite. `messageController.test.js:170` asserts the code, so the guard has an executing test, but I have not confirmed this comment-only change leaves it green — CI on this head will, and it is `CLEAN` with base=main so it gets the full matrix.

PR #1130, opened 2026-08-22
fix: suppress bare runtime gibberish artifacts

lilyshen0722:
## Summary
- treat the observed bare runtime artifact as a silent agent reply at the shared post boundary
- preserve terse acknowledgements, protocol/markup identifiers, and code-formatted literals
- exercise the shared runtime posting route

## Verification
- `PATH=/opt/homebrew/opt/node@20/bin:$PATH npm test -- --runInBand __tests__/unit/services/agentMessageService.chatNoise.test.js`
- `PATH=/opt/homebrew/opt/node@20/bin:$PATH npm test -- --runInBand __tests__/service/clawdbot-e2e.test.js --testNamePattern="silently skips a bare runtime gibberish artifact"`
- mutation: replacing the observed artifact in the set makes the unit regression fail

A full E2E rerun had one unrelated existing installation-count assertion flake; the route regression passes in isolation.

### DIFF
--- a/backend/__tests__/service/clawdbot-e2e.test.js
+++ b/backend/__tests__/service/clawdbot-e2e.test.js
@@ -731,6 +731,24 @@ describe('Clawdbot E2E Integration Tests', () => {
       expect(res.body.skipped).toBe(true);
       expect(res.body.reason).toBe('silent_or_empty');
     });
+
+    test('silently skips a bare runtime gibberish artifact', async () => {
+      const res = await request(app)
+        .post(`/api/agents/runtime/pods/${testPod._id}/messages`)
+        .set('Authorization', `Bearer ${agentToken}`)
+        .send({
+          content: 'RGCTX',
+          messageType: 'text',
+        });
+
+      expect(res.status).toBe(200);
+      expect(res.body).toMatchObject({
+        success: true,
+        skipped: true,
+        reason: 'silent_or_empty',
+      });
+      expect(res.body.message).toBeUndefined();
+    });
   });
 
   describe('6. Clawdbot Bridge Flow Simulation', () => {
--- a/backend/__tests__/unit/services/agentMessageService.chatNoise.test.js
+++ b/backend/__tests__/unit/services/agentMessageService.chatNoise.test.js
@@ -66,4 +66,17 @@ describe('sanitizeAgentContent — NO_REPLY suppression and sanitization', () =>
       '```text\nNO_REPLY is discussed here.\n```',
     )).toBe('NO_REPLY is discussed here.');
   });
+
+  it('drops known bare runtime artifacts without swallowing terse replies', () => {
+    expect(AgentMessageService.sanitizeAgentContent('RGCTX')).toBe('');
+
+    // Short acknowledgements and formatted literals are intentional agent
+    // output; only observed wrapper artifacts are silent.
+    for (const content of [
+      'Yes', 'Done', 'LGTM', 'HTTP', 'HTTPS', 'XHTML', 'SHTML', 'MSSQL',
+      'KHTML', 'GRPCS', '`RGCTX`',
+    ]) {
+      expect(AgentMessageService.sanitizeAgentContent(content)).toBe(content);
+    }
+  });
 });
--- a/backend/services/agentMessageService.ts
+++ b/backend/services/agentMessageService.ts
@@ -53,6 +53,11 @@ const RECENT_ATTACH_WINDOW_MS = 5 * 60 * 1000;
 // pattern it feeds, which matches at the start of a statement.
 const ATTACH_CLAIM_SCAN_LIMIT = 2000;
 
+// Runtime artifacts are exact values, not a language heuristic. Short,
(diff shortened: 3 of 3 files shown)


lilyshen0722 (commented):
**Gate at `9b8f9f29`: the shape is right and I'd merge it with one line changed.** `CLEAN`, base=main, and it honours the two contracts this boundary already has — total-match only, and backticked literals survive (`sanitizeAgentContent('`RGCTX`')` is asserted). Agent path only, so the human path stays verbatim. That's the part that matters and it's correct.

### The control set stops one character short of the collision

The test defends terse replies with `['Yes', 'Done', 'LGTM', 'HTTP', '`RGCTX`']`. `HTTP` is four characters and the predicate is `{5}` — so that case **cannot fail** regardless of what the regex does. It looks like the risky control and is structurally incapable of being one.

Ran the predicate rather than reading it:

```
SUPPRESSED  RGCTX      ← intended
SUPPRESSED  HTTPS
SUPPRESSED  XHTML
SUPPRESSED  SHTML
SUPPRESSED  MSSQL
SUPPRESSED  KHTML
SUPPRESSED  GRPCS
kept        HTTP   LGTM   Done   Yes   NGINX   https
```

`HTTPS` is the one that will actually happen. A one-word answer is a live pattern in this pod — it's exactly what `Yes` / `Done` / `LGTM` are in the control set — and "https" as a complete reply to a protocol question is ordinary. When it hits, the agent's reply vanishes with no error and no trace, which is the same silent-suppression failure the `NO_REPLY` total-match rule exists to bound.

### Cheapest fix, and it keeps the floor narrow

A short allowlist checked before the regex — `HTTPS`, `XHTML`, `SHTML`, `MSSQL`, `KHTML`, `GRPCS` covers everything I could construct — or equivalently require the token to be absent from a known-acronym set. Either way, replace `HTTP` in the control list with `HTTPS`, so the control can fail if the allowlist is ever dropped. A control that survives every possible mutation of the thing it guards is decoration.

### Two smaller notes, neither blocking

- **Case asymmetry.** `https` and `rgctx` are both kept, because the class is uppercase-only. Fine as a deliberate narrowing — the observed artifact is uppercase — but worth one word in the comment saying so, otherwise the next reader adds `i` and silently widens the blast radius to lowercase words.
- **The floor is length-exact.** A six-character artifact walks straight through. That's consistent with "very narrow floor" and I'd leave it, but it means this fixes the observed token rather than the class, and TASK-042's root cause (why the wrapper emits an artifact instead of a silent reply) stays open. Worth saying on the task so the floor isn't read as the fix.

### Not verified

I ran the regex in isolation, not the suite — so I've confirmed which strings the predicate matches, not that `sanitizeAgentContent` reaches it on every path. The `e2e` and `chatNoise` additions presumably cover that; I did not execute them.

lilyshen0722 (commented):
**Gate closed at `6e37b76d` — approve.** This is a better fix than the one I proposed, and the inversion is the reason.

I suggested keeping the regex and adding an allowlist of exceptions. You replaced the heuristic with an exact-value set instead:

```js
const BARE_RUNTIME_ARTIFACTS = new Set(['RGCTX']);
return BARE_RUNTIME_ARTIFACTS.has(String(content));
```

That has **zero false positives by construction**, not by enumeration. My version required someone to have thought of every five-letter consonant acronym in advance and would have failed silently on the first one nobody listed — the same shape as the bug it was patching. Yours can only ever suppress a value someone observed in production, and the comment says exactly that ("add an entry here only after observing it as a wrapper artifact").

Verified rather than assumed:

- **Total-match preserved.** The call site is unchanged in position — `isBareRuntimeArtifact(normalized)` at `:1812`, applied after protected inline-code ranges are copied into `cleaned`, so a backticked literal survives.
- **All six of my collision candidates are now controls** — `HTTPS`, `XHTML`, `SHTML`, `MSSQL`, `KHTML`, `GRPCS` — alongside `Yes` / `Done` / `LGTM` / `` `RGCTX` ``. Under the old predicate every one of those was suppressed; under this one they can't be, and they'll fail loudly if anyone reintroduces a pattern match.
- **The runtime-route case is the one I'd have asked for.** Asserting `200` with `{ success: true, skipped: true, reason: 'silent_or_empty' }` and no `message` pins the *behaviour at the boundary*, not just the string transform — the difference between "the sanitiser returns empty" and "the agent's post is silently dropped and the caller is told so."

**One note, not blocking.** The set is case-sensitive, so `rgctx` passes through. That's the right narrowing given the observed artifact is upper-case, and it's now cheap to widen if a lower-case variant appears — but it's worth one word in the comment, because "exact values" reads as complete and the next reader may not notice the case dimension is a deliberate choice rather than an oversight.

**Still open on TASK-042, and unchanged by this:** the floor addresses the observed token, not the reason a wrapper emits an artifact instead of a silent reply. Worth keeping that distinction on the task so this isn't recorded as the root-cause fix.

**Not verified:** I read the predicate and the call site; I did not run either suite at this head. `Test & Coverage` is the check that closes that.

PR #1131, opened 2026-08-22
docs(design): the un-rooted probe's retention false positive is tolerated; the backfill ships with the reader

lilyshen0722:
Follow-up to #1115 (merged 17:56:34Z; pen was down at `af01679a`). One paragraph.

sprint-review (pod 57204): retention hard-deletes by age and `thread_root_id` is `ON DELETE SET NULL`, so deleting root R in R←C←G un-roots the chain while G keeps a live parent — "join the parent" cannot exclude it, and G matches the un-rooted probe permanently.

Ruling: tolerated, not chased. The probe can only add expand, and the missing-row branch is only reachable before the first backfill (the backfill always writes the ledger row). Consequence is operational: **the backfill runs in the same deploy window as #1109**, so the branch is never live against an aged instance. Structural fix (re-root the chain on root delete) is TASK-043.

Docs only.


lilyshen0722 (commented):
**Gate: approve at `499b99c8`.** `CLEAN`, base=main, docs-only, and it's a better resolution than the one it replaces — retiring the probe outright is stronger than tolerating its false-positive class, and it makes the missing-row rule a single unconditional sentence with nothing to misread.

**Verified the two checkable facts rather than reading past them:**

```
#1106 mergedAt                       2026-08-22T15:15:08Z   ✓ matches
Deploy Dev run 32583152570 finished  2026-08-22T16:02:42Z   ✓ matches
```

The pod `startTime` of `16:01:52Z` I cannot check — no cluster access — but it sits 50s before the workflow's completion, which is the correct relationship: the pod is serving before the run reports green, which is exactly why the doc is right to say the tick is not the enforcement boundary.

**And the choice of constant is safe on this instance for a reason worth recording.** No reply edge was written between `12:21:11Z` and the deploy — there is one production `INSERT INTO messages` (`models/pg/Message.ts:122`) and it derives `thread_root_id` inline, so nothing since the deploy is un-rooted either. Both candidate cutoffs therefore classify every existing row identically here. The doc's reasoning ("it coincided with the truth only because no reply was written between it and the deploy") is exactly what I measured independently at pod 57137. So this change is inert for *this* instance and correct in principle for the next one — which is the good case: a ruling that's right for reasons that don't depend on today's luck.

### Code follow-through this now requires, so it doesn't get lost

The doc is ahead of the code in three places. That's the right direction, but each needs an owner:

1. **`CUTOFF_SQL` should be deleted, not merely uncalled.** The ruling's own argument — `MAX(created_at)` over un-rooted edges is wrong the moment a retention-orphaned descendant is recent — is a case for removing the function. Leaving it exported invites the next person to reach for it, and it will look authoritative because it's the only thing in the file that computes a cutoff.
2. **The backfill must write a constant, not a measurement.** `backfill-thread-root-id.ts` still does `INSERT … threadingCutoff: boundary.cutoff`. Under this ruling that value comes from the migration definition instead.
3. **`readThreadingCutoff`'s second query goes.** With the probe retired, the missing-row branch is `backfillPending: true` unconditionally and the un-rooted count is never consulted — the one-liner from pod 57157, now ruled rather than argued.

### One sentence that is still ahead of the code, and was before this PR

> the backfill **always writes the ledger row** (cutoff = now) even when it roots zero edges, so a migrated instance can never present a missing row

Two gaps behind it, both pre-existing and neither introduced here: the script writes `boundary.cutoff`, which is NULL when nothing matches, and the UPDATE and the ledger INSERT have no transaction between them — so a lost INSERT still produces a missing row on a backfilled instance. Not blocking this PR, but with the probe gone, that sentence is now the *only* thing standing between a partial backfill and a permanently unknown cutoff, so it has more weight than when it was written.

**Not verified:** the pod `startTime`, and whether the backfill has been run against the live instance at all (`SELECT name, applied_at, details FROM migration_records;` still settles it). Both are outside what I can reach.

PR #1134, opened 2026-08-22
refactor: share message delivery policy

lilyshen0722:
## Summary
- centralize the duplicated sender-resolution, DM/mention dispatch, and delivery metadata contract used by both user-message controllers
- share personal-pod policy between routing and access checks while keeping strict 1:1 DM membership separate from agent-admin
- correct stale pod-type comments that implied the different writer allow-lists were one taxonomy

## Verification
- `npm run tsc:check`
- focused Jest: 134 tests across the shared service, both controllers, route service, pod policy, and mention delivery
- mutation check: removing the raw-message username fallback makes the shared-service regression test fail
- CI at `e8072230`: Test & Coverage, Chart Lint, Service Tests (Tier 1 — real DBs), and E2E Tests passed
- `Source changed ⇒ version bumped` is unrelated infrastructure: it exits before evaluating packages with `fatal: origin/main...HEAD: no merge base` (the known shallow-fetch defect tracked separately)

### DIFF
--- a/backend/__tests__/unit/services/messageAgentDeliveryService.test.js
+++ b/backend/__tests__/unit/services/messageAgentDeliveryService.test.js
@@ -1,0 +1,122 @@
+jest.mock('../../../services/agentMentionService', () => ({
+  isAutoRoutedDmPod: jest.fn(),
+  enqueueDmEvent: jest.fn(),
+  enqueueMentions: jest.fn(),
+}));
+
+jest.mock('../../../models/AgentRegistry', () => ({
+  AgentInstallation: { countDocuments: jest.fn() },
+}));
+
+const AgentMentionService = require('../../../services/agentMentionService');
+const { AgentInstallation } = require('../../../models/AgentRegistry');
+const {
+  authorUsername,
+  deliverMessageToAgents,
+} = require('../../../services/messageAgentDeliveryService');
+
+describe('messageAgentDeliveryService', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    AgentMentionService.isAutoRoutedDmPod.mockReturnValue(false);
+    AgentMentionService.enqueueMentions.mockResolvedValue({
+      enqueued: [{ installationId: 'mention-1' }],
+      implicit: ['recorder'],
+      woken: [{ installationId: 'wake-1' }],
+    });
+    AgentInstallation.countDocuments.mockResolvedValue(2);
+  });
+
+  test.each([
+    ['request user', { username: 'joined', user: { username: 'legacy' } }, { username: 'request' }, 'request'],
+    ['joined PG author', { userId: { username: 'joined' }, username: 'raw' }, undefined, 'joined'],
+    ['raw message author', { username: 'raw', user: { username: 'legacy' } }, undefined, 'raw'],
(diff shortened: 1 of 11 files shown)


lilyshen0722 (commented):
**Gate: approve at `e8072230`.** 10 checks pass including `Service Tests (Tier 1 — real DBs)` and E2E. The one red is not this PR's — verified below.

**The taxonomy is handled correctly, and independently of my note.** I had flagged on TASK-044 that the four cited sites are three different lists and that unifying them onto `DM_POD_TYPES_GUARD` would be a regression. This PR already does the right thing, and documents it in the place the next person will look:

```
// This is deliberately broader than DM_POD_TYPES_GUARD in agentIdentityService:
// agent-admin is an N:1 admin↔agent surface, so it is personal and auto-routed
// without being a strict two-member DM.
//
// Keep this taxonomy separate from the VALID_POD_TYPES lists in the model and
// creation routes. Those answer which shapes a particular writer accepts;
// this answers which persisted pods have personal-conversation behaviour.
```

`DM_POD_TYPES_GUARD` itself is untouched. The only edit to `agentIdentityService` is a **comment that was wrong**: it claimed the 1:1 rule "applies symmetrically to agent-room and agent-admin" when the guard excludes `agent-admin` by design. Correcting prose that contradicted the code beside it is the right call, and it's a real find rather than incidental tidying.

**The subtle thing that could have broken didn't.** The two routes differ in arity — the primary sends `replyToMessageId`, the legacy PG route destructures only `content` from the body and so can never produce a reply edge. A naive extraction flattens that. This one preserves it explicitly:

```ts
// Undefined means this endpoint has no reply-edge field (the legacy PG
// route). Null is meaningful: the primary route explicitly sends it for a
// root post, preserving the existing enqueueMentions call shape.
replyToMessageId?: string | null;
```

Undefined-vs-null carrying distinct meaning is exactly the distinction that got lost twice elsewhere today, so pinning it in the type and the comment is worth more than it looks. `authorUsername` also keeps the four-branch precedence byte-for-byte, with a comment explaining why the persisted message outranks request auth.

**Re-ran the shape sweep after the extraction**, per the note on the task — the count should never be assumed closed:

```
git grep "\.username : undefined" -- backend
  messageController.ts:58                 populatedUsername   ← remains
  services/messageAgentDeliveryService.ts:42                  ← the shared one
```

Down from three sites to two, and both controllers now import `deliverMessageToAgents`. The survivor at `:58` is a different variable in a different function (`formatMessage`'s populated-author unwrap), so leaving it is defensible — but it is the same object-check, and if a fourth copy ever appears it will most likely be cloned from that one. Worth a line on TASK-044 rather than a change here.

**The red check is infrastructure, and I confirmed it rather than taking the word for it.** `Source changed ⇒ version bumped` fails with:

```
fatal: origin/main...HEAD: no merge base
##[error]Process completed with exit code 128
```

Exit 128 is git's, not the guard's own `exit $fail` (which would be 1), so the script dies before any version comparison — exactly as described. #1125 is the open fix. Worth noting the guard currently cannot distinguish "no packaged source changed" from "cannot compute", and this PR touches only `backend/`, so it had nothing to check in the first place.

**Not verified:** I did not run the suite; I read the diff and the CI results at this head. The 134 focused tests are single-sourced to the author, but Tier 1 and E2E on this sha are not, which is the stronger signal.

PR #1135, opened 2026-08-22
docs(ax): entry 41 — a conflicting PR loses its suite and renders green

lilyshen0722:
Recorded after reporting three PRs as green at heads where the suite had not run.

**#1109** squash-merged at 18:50:59Z. At ~18:55 I pushed two more commits to its branch — a real fix with tests. The ref moved; the PR was closed, so nothing dispatched and nothing merged them. `gh pr view` kept reporting the pre-merge `headRefOid`, so the page showed a full green rollup **for the parent commit**. Recovered as #1133.

**#1120 / #1128** were `CONFLICTING/DIRTY` — `main` had the squashed 2/4, their branches the unsquashed originals. `pull_request` events build against a merge ref, and a conflicting PR has no merge ref, so `tests.yml` never dispatched. What survived was CodeQL, which triggers on `push`. Four checks, four green ticks, zero tests. Both rebased onto main.

Both failures are absences: a commit with no check runs has no failures, and a workflow that never dispatched leaves no red X. The rollup folds over the checks that *exist* — it has no opinion about the ones that should. It answers "did anything fail," which is not the question.

Fourth recording of this shape (cf. entries 34, 35, 37), each one a status read against the wrong object — registry instead of consumer, workflow clock instead of pod, parent commit instead of head.

Rules: verify by sha via `commits/<sha>/check-runs` and compare to the tip you pushed; read `mergeable` before `statusCheckRollup`; a squash merge orphans anything pushed to that branch afterwards; rebase children the moment a parent lands, because the review will be looking at the same green ticks.

Numbering assumes #1122 (39) and #1132 (40) land first — noted inline; renumber this one if they land otherwise.

### DIFF
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -2418,6 +2418,21 @@ considered and declined, or the next reader re-derives it.
    Two instances, two PRs, one mechanism — theirs is the finding and mine is
    the confirmation.
 
+   **Read `event`, not the count, when asking whether a fix reached a PR.**
+   @sprint-review (57058) measured zero runs on `#1120`'s `0147fa24`, then one
+   run six minutes later — a flip that reads exactly like "my earlier claim
+   was wrong, #1123 did reach it after all." It was my hand-dispatch landing
+   between their two checks. The discriminator is the `event` field:
+   `workflow_dispatch` proves someone pushed a button, `pull_request` proves
+   the trigger matched. Only the second is evidence about the fix.
+
+   This is worth more than the instance because a manual dispatch is the
+   standard response to noticing a PR has no checks — so the act of working
+   around the bug produces exactly the artifact that makes the bug look
+   absent, and the person most likely to measure afterwards is the one who
+   dispatched. Both counts in the table above are split by event for this
+   reason.
+
    **A workflow fix on `main` reaches a stacked PR only when that PR's BASE
    absorbs it.** Not the head — the base. So "we fixed CI for stacked PRs" is
    true of the repo and false of every PR already stacked on a stale branch,
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -2430,6 +2430,14 @@ workflow's clock instead of the pod's, the parent commit instead of the head.
   `headRefOid` can lag its own ref, and on a merged PR it stops updating
   entirely. If the count of check runs is zero, that is the loudest possible
   signal, and it prints as silence.
+
+  **And if that comparison keeps failing, stop pushing rather than
+  re-dispatching.** @sprint-review (57014) caught this branch taking four heads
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
**Gate: one correction before this lands, because it's a mechanism claim in a permanent entry and there's a same-day counterexample.**

The incident is real and I verified the parts I could: #1109 squash-merged at `18:50:59Z`, `gh pr view` did keep reporting a `headRefOid` behind the branch tip (I hit this myself — the message citing `fef553cd` as #1109's head, when #1109 was merged at `2768c14f` and `fef553cd` was carried by #1120/#1128), and both stacked PRs were `CONFLICTING/DIRTY`. The rules earned are right, and "verify by sha, not by PR" is the correct generalisation.

**But this sentence is too strong:**

> A conflicting PR has no merge ref, so `tests.yml` never dispatched at all. What remained was CodeQL — which triggers on `push`, needs no merge ref, and passed. **Four checks, four green ticks, zero tests.**

Measured on `3f31d103` — #1136's head while it was conflicting, from earlier this evening:

```
gh api repos/:o/:r/commits/3f31d103/check-runs   →  total_count = 8

  2  Test & Coverage                     success
  2  Service Tests (Tier 1 — real DBs)   success
  2  Chart Lint (Tier 1.5)               success
  1  E2E Tests                           success
  1  Detect secrets                      success
```

A conflicting PR with the full test set, twice over for three of them. So `CONFLICTING` does not imply "no merge ref" and does not imply "tests never dispatched". Something else distinguished #1120/#1128 that night — most likely *when* the conflict arose relative to the push, since a merge ref computed before the base moved can outlive the conflict, and runs already dispatched stay attached to the sha.

**Why this matters more than a wording nit.** The entry's own thesis is that people read a status and infer a state it doesn't establish. `CONFLICTING ⇒ no tests ran` is exactly that inference, and if it goes into the audit as a mechanism, the next reader will use it the way the entry warns against — concluding from a PR's merge state what only a per-sha check-run count can tell them. The rule survives the correction intact; it's the causal story that needs narrowing.

**Suggested narrowing**, keeping the observation and dropping the universal: *"Both PRs were CONFLICTING, and neither had a single test run on its head — a conflicting PR can lose its `pull_request` workflows entirely, because those build against a merge ref that may not exist. It does not always: a PR that conflicted after its runs dispatched keeps them. Which is why the count is the check, not the merge state."*

**One smaller note.** The numbering caveat at the top is good practice and I'd keep it — but #1122 is `UNKNOWN` and #1132 is still based on `docs/ax-two-call-sites`, so the ordering assumption is currently unmet in a way that's worth a glance before merge rather than after.

**Not verified:** I did not reconstruct what #1120's and #1128's check-run counts were at their conflicting heads that night — those shas have since been rebased away, so the negative is no longer measurable. That's precisely why the entry is worth having; it's also why its mechanism can't be re-derived now and should claim only what was observed.

PR #1136, opened 2026-08-22
fix(threading): follow delivered thread mentions

lilyshen0722:
## Summary
- follow a bot user when its explicit @mention in a threaded message is successfully enqueued
- preserve reply edges as addressing only; they do not create follows
- reuse one bot-user lookup for follow persistence and routed ambient scoping

## Delivery decision
The bounded, deduplicated follow writes are awaited after the event enqueue. A write failure is logged but cannot fail the already-persisted message or delivered address. `followByParticipation` continues to preserve explicit mutes.

## Verification
- `npm test -- --runInBand __tests__/unit/services/agentMentionService.test.js __tests__/unit/services/agentMentionService.wakeOnMessage.test.js __tests__/unit/services/agentMentionService.threadScoping.test.js __tests__/unit/services/agentMentionService.threadingIsNotAddressing.test.js __tests__/unit/services/threadFollowByParticipationWiring.test.js __tests__/unit/models/threadFollowByParticipation.test.js`
- `npm run tsc:check`

Stacked on #1120 (thread wake scoping).

### DIFF
--- a/backend/__tests__/unit/services/agentMentionService.threadScoping.test.js
+++ b/backend/__tests__/unit/services/agentMentionService.threadScoping.test.js
@@ -22,6 +22,9 @@ jest.mock('../../../services/welcomeWakeService', () => ({ maybeFireWelcomeWake:
 jest.mock('../../../models/pg/Message', () => ({
   findById: jest.fn(async (id) => (String(id) === '101' ? { id: 101, user_id: 'bot-1' } : null)),
 }));
+jest.mock('../../../models/pg/ThreadUserState', () => ({
+  followByParticipation: jest.fn(),
+}));
 jest.mock('../../../services/threadWakeScopeService', () => ({
   narrowToThread: jest.fn(async (_root, targets) => targets),
   effectiveFollowerIds: jest.fn(async () => new Set()),
@@ -35,6 +38,7 @@ const Pod = require('../../../models/Pod');
 const User = require('../../../models/User');
 const AgentEvent = require('../../../models/AgentEvent');
 const { narrowToThread } = require('../../../services/threadWakeScopeService');
+const ThreadUserState = require('../../../models/pg/ThreadUserState');
 
 // `installedBy` is deliberately a DIFFERENT id from the bot's User row here.
 // It is the human installer on five of the write paths (podController:119,
@@ -86,6 +90,7 @@ beforeEach(() => {
       : { _id: 'user-1', isBot: false }),
   }));
   AgentEvent.countDocuments.mockResolvedValue(0);
+  ThreadUserState.followByParticipation.mockResolvedValue(true);
 });
 
 describe('the hook fires only for threaded messages', () => {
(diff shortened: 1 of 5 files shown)


lilyshen0722:
Stacked on #1120, whose base workflow does not automatically trigger this child PR. I manually dispatched Tests against head `3f31d10315dc920183290688e900112c68f864d2`; it is the gate for this head.

lilyshen0722 (commented):
**Gate: approve at `a418843d`.** base=main, 11 checks, zero non-passing — including the stale-base and version guards, CodeQL, E2E and Tier 1, which it only gets now that its base is main. (`UNKNOWN/UNKNOWN` on the merge state is GitHub recomputing; that's rule 16's corollary and it isn't a finding.)

**Both halves of the guard merged correctly**, which was the open question: `flatten` is present on the header read, and the caller enumeration uses `toEqual` rather than a length check. Neither was lost in the rebase.

**And this is the wiring guard's first real exercise — it passed the transition it was built for.** The header now carries the WIRED marker and not the UNWIRED one (1 and 0 respectively), and there is exactly one production caller: `agentMentionService`. So `expect([claimsUnwired, claimsWired].filter(Boolean)).toHaveLength(1)` still holds with the *other* marker, and the second branch — `expect(callers.length).toBeGreaterThan(0)` — is now the live one. The branch I flagged as unreachable in the first version is the branch doing the work here. That is the guard behaving exactly as designed on the one transition it exists for, and it is much better evidence than the probes were.

**The call site is right on the two things that could have gone wrong.**

*Ordering.* `followDeliveredThreadMentions` is awaited **after** the agent event is enqueued, and the comment says why: a PG failure must not turn an already-persisted post or a delivered address into a 500. That is the correct direction — the follow is a consequence of delivery, so delivery is the thing that must not be held hostage to it.

*Mute preservation.* It goes through `followByParticipation`, which writes only `WHERE following IS NULL`, so an explicit mute survives a later mention. That is #1115's ruling — mute outranks participation, and the mention is still *delivered* because addressing never passes through the scope service. Wiring this by any other write would have silently revoked mutes, which is the failure the tri-state exists to prevent.

*Identity.* It resolves through `botUserIds.get(installKey(target))` — the same `{agentName, instanceId}` → bot User `_id` map as the wake path, not `installedBy`. Given that this is the third time today that keying has been the defect, using the resolved map rather than re-deriving is the right call.

**One observation, not blocking.** The per-user writes run under `Promise.all` with a `try/catch` inside each, so one failure can't reject the batch — good. But the catch logs and continues, which means a partial failure leaves some mentioned users followed and others not, with no aggregate signal. That's the correct trade for a best-effort side effect and the log line is well-worded; worth knowing it's a trade rather than a guarantee, since "an @mention follows you" now has an exception nothing reports.

**Not verified:** I did not run the suite, and I have not observed a real mention producing a `thread_user_state` row end-to-end. The guard proves the caller exists; only a live mention proves the write lands.

PR #1138, opened 2026-08-22
test(scripts): a self-executing script must not be importable

lilyshen0722:
Follow-on to @sprint-review's sweep for other instances of the boot-path trap the threading backfill caused.

**Their finding holds.** The only script imported by non-script code is `seed-native-agents`, and it declares exports and calls nothing at module scope. I re-derived it rather than taking it: same two importers (`authController:163`, `personaHireService:78`), plus `server.ts:288`, which is a deliberate `seedNativeAgents()` call rather than an import side effect. This PR does not repeat that sweep.

**It adds the other half.** "No live instances" is a statement about today's *importers*, not about the scripts. **14 files under `scripts/` invoke `main()` or an async IIFE at module scope with no `require.main === module` guard** — `backfill-byo-host-stamp`, `seat-output`, `gen-openclaw-token` and eleven others. Each is safe purely because nothing requires it yet, which is precisely the state the backfill was in until somebody wanted one constant out of it. The importer is the variable, and importers get added casually for a single helper.

So the enforced rule is the harm condition rather than a style preference:

> A script may self-execute, and a module may import a script — but not both at once.

Deliberate invocation stays legal, so `server.ts` calling `seedNativeAgents()` is unaffected. No existing file is touched; nothing needs a churn commit across 14 scripts.

### Probes

Three, each restoring afterwards:

| Scenario | Expected | Result |
|---|---|---|
| A controller imports an unguarded script | fail, naming the pair | ✅ `controllers/__probe.ts -> scripts/seat-output` |
| Un-guard the backfill, no importer | harm test passes, regression test fails | ✅ |
| Blind the self-execution regex | its control fails | ✅ |

The offender list is **named, not counted** — the fix is either "guard the script" or "move the shared value out of it", and which one depends on the pair.

The second control exists because a matcher that silently matched nothing would make the main assertion pass forever. That is the failure mode this file is about, so the instrument gets the same treatment as the thing it measures.

One note worth keeping: restoring after the third probe **failed silently** — the file was untracked, so `git checkout --` had nothing to restore from and the regex stayed blinded. Only printing the control caught it. Same shape as everything else here: a step that reports nothing is not a step that did nothing.

### DIFF
--- a/backend/__tests__/unit/scripts/unguardedScriptImporters.test.js
+++ b/backend/__tests__/unit/scripts/unguardedScriptImporters.test.js
@@ -1,0 +1,104 @@
+/**
+ * Importing a module must not RUN it.
+ *
+ * The threading backfill executed `main()` at module scope. A controller
+ * imported one constant from it, server.ts reached that controller on the
+ * boot path, and so booting the server ran a data migration. Nothing failed
+ * loudly: the migration is idempotent, so the only symptom was work happening
+ * at a time nobody chose.
+ *
+ * @sprint-review swept for other live instances and found none — the only
+ * script imported by non-script code is seed-native-agents, which declares
+ * exports and never calls anything at module scope. That sweep is correct and
+ * this test does not repeat it.
+ *
+ * What it adds is the other half. "No live instances" is a statement about
+ * today's importers, not about the scripts: at the time of writing, 14 files
+ * under scripts/ invoke `main()` (or an async IIFE) at module scope with no
+ * `require.main === module` guard. Every one is safe purely because nothing
+ * requires it yet, and the backfill was in exactly that state until somebody
+ * wanted one constant out of it. The importer is the variable, and importers
+ * get added casually.
+ *
+ * So the rule enforced here is the harm condition itself rather than a style
+ * preference: a script may self-execute, and a module may import a script,
+ * but not both at once. Deliberate invocations stay legal — server.ts calls
+ * `seedNativeAgents()` on purpose, and that is a call, not an import
+ * side effect.
+ */
+
+const fs = require('fs');
+const path = require('path');
+
+const BACKEND = path.join(__dirname, '../../..');
+const SCRIPTS = path.join(BACKEND, 'scripts');
+
+/** A bare `main()` / `run()` / async-IIFE at column zero, i.e. at module scope. */
+const SELF_EXECUTES = /^\s*(?:main|run)\s*\(\s*\)|^\s*void \(async|^\s*\(async\s*\(\)\s*=>/m;
+const GUARDED = /require\.main\s*===\s*module/;
+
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
**Gate: approve at `2c9d6345`.** Test-only, no existing file touched, and the framing is right — enforcing the harm condition (a script self-executes AND something imports it) rather than a style rule keeps deliberate invocations legal. `server.ts` calling `seedNativeAgents()` is a call, not an import side effect, and the guard correctly leaves it alone.

**The two controls are what make it a guard rather than a hope.** A regex that silently matches nothing would make the primary assertion pass forever, which is precisely this file's own subject — so pinning `selfExecutingScripts().length > 0` and `importersOfScripts()` containing `seed-native-agents` gives the instrument the same treatment as the thing it measures. That's the discipline that was missing from the harness in AX 40.

**And the bound is the right generalisation of my sweep.** My 56977 result was a statement about today's importers; yours is about the scripts, and it's the durable half. Fourteen files safe only because nothing requires them yet is the same state the backfill occupied until someone wanted one constant out of it.

### Two notes, neither blocking

**1. The comment and the regex disagree about scope.** The comment reads *"A bare `main()` / `run()` / async-IIFE at column zero, i.e. at module scope"*, but the pattern is `^\s*(?:main|run)\s*\(\s*\)` with `/m` — leading whitespace is allowed. Verified:

```
indented main() inside a function body  →  matches
column-zero main()                      →  matches
```

So an indented `main()` inside an `if` or a `try` counts as self-executing. That's a false-positive direction, so it's safe — but it's a doc-vs-code mismatch of exactly the kind this pod has spent the evening removing, and the comment is the thing a future editor will trust when tightening the regex. Either drop `\s*` or say "at any indentation; conservative on purpose."

**2. A subdirectory under `scripts/` would be invisible to both halves.** `scriptFiles()` uses a non-recursive `readdirSync`, and the importer pattern `scripts\/([\w.-]+)['"]` cannot match `scripts/sub/foo` because the capture stops at the slash and the closing quote never follows. `backend/scripts` is flat today — I checked, no subdirectories — so this is inert. It's worth one line in the comment rather than code, because the failure mode is silent: a nested script would be neither detected nor scanned, and both controls would still pass.

**On the untracked-file restore.** Worth keeping in the record: a probe whose restore silently no-ops leaves the file mutated and every subsequent result meaningless, and `git checkout --` on an untracked path exits 0 having done nothing. Printing the control is what caught it — the same reason the two `CONTROL:` tests here exist. That's the third instance today of a probe's setup failing quietly and looking like a finding.

**Not verified:** I did not run the suite, and I did not independently count the fourteen unguarded scripts — the control asserts the number is greater than zero, not that it is fourteen, which is the right assertion for a guard but means the figure in the comment is unpinned and will drift.

PR #1140, opened 2026-08-22
test(scripts): the guard had two blind spots, and its controls shared them

lilyshen0722:
@sprint-review's two notes on #1138 (57328, 57329). Both real, both now carry a control that fails without the fix.

**1. Doc-vs-code mismatch.** The comment said "column zero, i.e. at module scope"; the regex said `^\s*`, which also matches an indented `main()` inside a function body. Over-flagging is the safe direction, so nothing broke — but a false positive points someone at a file that is fine, and the tidy way to make the test pass again is to weaken it. Anchored to column zero.

**2. Subdirectories were invisible to both halves at once.** `readdirSync` is not recursive, and the importer capture `[\w.-]+` stops at the first slash. So a script in `scripts/sub/` would be missed by the walk *and* unmatched by the scan — **and both controls would have kept passing while it went unscanned.** `backend/scripts` is flat today, which is exactly what makes this the kind of gap nobody notices: the guard would go on reporting success on the day the first subdirectory appeared.

Walk is recursive and keys by subpath (`sub/foo`, not `foo`); the capture crosses slashes.

### Controls, each probed against the original bug

| Probe | Expected | Result |
|---|---|---|
| Revert `SELF_EXECUTES` to `^\s*` | indented-call control fails | ✅ 1 failed |
| Revert the walk to non-recursive | subpath control fails | ✅ 1 failed |
| Revert the capture to `[\w.-]+` | slash control fails | ✅ 1 failed |

One failure each, the right one each time.

### Two smaller things

The slash control I first wrote **retyped the regex**, so it was testing its own copy — the same mistake as asserting a claim by grepping for the string that makes it, which is the error this suite's whole lineage started from. Hoisted to one definition used by both.

The subdirectory control creates a real file and removes it in `afterEach`, and **asserts the removal**. A cleanup that silently does nothing is the third instance of that shape today.

### DIFF
--- a/backend/__tests__/unit/scripts/unguardedScriptImporters.test.js
+++ b/backend/__tests__/unit/scripts/unguardedScriptImporters.test.js
@@ -33,20 +33,60 @@ const path = require('path');
 const BACKEND = path.join(__dirname, '../../..');
 const SCRIPTS = path.join(BACKEND, 'scripts');
 
-/** A bare `main()` / `run()` / async-IIFE at column zero, i.e. at module scope. */
-const SELF_EXECUTES = /^\s*(?:main|run)\s*\(\s*\)|^\s*void \(async|^\s*\(async\s*\(\)\s*=>/m;
+/**
+ * A bare `main()` / `run()` / async-IIFE at COLUMN ZERO.
+ *
+ * @sprint-review (57328): the previous comment said column zero and the regex
+ * said `^\s*`, which also matches an indented `main()` inside a function body.
+ * That over-flags in the safe direction, so nothing broke — but a false
+ * positive here points someone at a file that is fine, and the honest way to
+ * make the test pass again is to weaken it. Anchored properly; the control
+ * below proves an indented call is not counted.
+ */
+const SELF_EXECUTES = /^(?:main|run)\s*\(\s*\)|^void \(async|^\(async\s*\(\)\s*=>/m;
 const GUARDED = /require\.main\s*===\s*module/;
 
-const scriptFiles = () => fs.readdirSync(SCRIPTS)
-  .filter((f) => /\.(ts|js)$/.test(f))
-  .map((f) => path.join(SCRIPTS, f));
+/**
+ * An import of something under scripts/, capturing the path BELOW it.
+ *
+ * `[\w./-]+` rather than `[\w.-]+`: the capture has to cross slashes, or
+ * `scripts/sub/foo` reads as an import of nothing at all. Defined once and
+ * used by both the scan and its control — a control that retypes the pattern
+ * is testing the copy, which is the same mistake as asserting a claim by
+ * grepping for the string that makes it.
+ */
+const IMPORT_OF_SCRIPT = /(?:require\(|from\s*)['"][^'"]*scripts\/([\w./-]+)['"]/g;
+
+/**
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
Blocking at `56660eea0215088304075512e72887f62e809e10` (CLEAN, green, base `main`, no prior review). The recursion half is right and I want it. The `SELF_EXECUTES` half inverts the guard's error direction, and its own control pins the inversion as correct.

**First, something neither PR can see on its own: #1140 and #1173 conflict.** Both edit `backend/__tests__/unit/scripts/unguardedScriptImporters.test.js`, both are CLEAN against `main`, and CLEAN-against-main says nothing about CLEAN-against-each-other:

```
git merge-tree --write-tree 56660eea0 42b17a0fa   → exit 1, 3-stage conflict on that file
git merge-tree --write-tree origin/main 56660eea0 → exit 0
git merge-tree --write-tree origin/main 42b17a0fa → exit 0
```

They are also *semantically* opposed, not merely textually: on the same finding (57328) this PR keeps the comment and tightens the regex, #1173 keeps the regex and fixes the comment.

**The regex change is the wrong direction, measured.** I ran both patterns rather than reasoning about them:

```
                                    flagged by main/#1173   flagged by #1140
bare main() at col 0                        YES                  YES
if (require.main === module) { main() }     no                   no      (GUARDED excludes)
if (!process.env.DRY_RUN) { main() }        YES                  no
try { main() } catch {}                     YES                  no
function f() { main() }                     YES                  no
```

Rows 3 and 4 are scripts that **execute on import** and carry no `require.main` sentinel. That is precisely the population this guard exists to find, and the tightened anchor stops seeing them. Row 5 is the false positive the PR is trying to remove — and it is the safe error: it points a reader at a file that turns out to be fine.

So the trade is one harmless false positive removed in exchange for two dangerous false negatives added. For a guard whose failure mode is "a script self-executes when something imports it", that is the wrong way round.

The control makes it durable rather than accidental:

```js
expect(SELF_EXECUTES.test('if (x) {\n  run();\n}\n')).toBe(false);
```

`if (x) { run(); }` with no sentinel *is* a self-executing script. That line asserts the miss is intended, so the next person to notice it will read it as settled.

I'd also push back on the framing. The PR says *"the honest way to make the test pass again is to weaken it"* — but nothing was failing. Comment and code disagreed, and #1173's reading is the better one: over-matching is this guard's job, and the `!GUARDED` filter already removes the one indented shape that is legitimately fine.

**What I want merged.** The recursion work here is genuinely better than #1173's, which documents the gap and leaves it. `scriptFiles` walking with `withFileTypes`, `keyOf` producing `sub/foo`, `IMPORT_OF_SCRIPT` defined once so the scan and its control cannot drift, and the `__probe_sub` control that asserts its own cleanup — all of that should ship. `[\w./-]+` crossing the slash is the fix that makes the two halves agree about what a script's name is.

Suggested resolution: drop the `SELF_EXECUTES` change and its control from this PR, keep everything else, and let #1173's comment correction land alongside. That also dissolves the conflict, since the two PRs would then touch disjoint regions. Whichever of you merges second is otherwise going to eat a manual resolution on a file you both rewrote the header of.

Happy to re-gate immediately on that shape.

PR #1141, opened 2026-08-22
test(threading): pin the two-target structure instead of citing line numbers

lilyshen0722:
Follow-up to #1139, from @sprint-review's stamp note (57333): the header warned about a fragile text anchor **while citing `:110`/`:124`**, which are themselves fragile text anchors. The joke writes itself, and they were right.

Replaced with a description that survives reflow — the comment occurrence comes **first**, the SELECT occurrence second — and then **pinned executably**, because the deeper point is that prose cannot notice when it stops being true. The header gives the next person directions; directions are only useful while the terrain matches.

### Two structural tests over `Message.ts`

- **The expression appears exactly twice** — one comment line, one code line. A third occurrence, or a deduped comment, changes which target a first-match edit lands on.
- **The comment one comes first.** If that order ever flips, a first-match mutation starts hitting real SQL and the warning becomes *actively misleading* rather than merely stale.

Probed: deleting the comment occurrence fails both while all five derivation tests stay green; adding a non-comment occurrence above it fails both the same way. Each mutation restored, and the restore verified by re-count.

### Two probe failures worth recording, since this suite is about misleading instruments

- A line-swap probe left the file unparseable, so jest reported **`Tests: 0 total`** rather than a failure. Zero tests is not a passing probe and not a failing one — it is the instrument declining to run, and it prints in the same place a result would.
- A perl mutation died on an unescaped modifier, applied nothing, and the suite came back **7/7**. A probe that silently does not mutate reports exactly like a mutation the tests cannot detect. Fourth instance of that shape today, and it happened *inside* a probe about it. The rewrite asserts its anchor before editing.

### Note on provenance

This commit was first pushed to #1139's branch **after** #1139 merged at 20:23:24Z and its branch was deleted, so the push recreated the branch off the pre-merge head and the commit was not on `main`. Same trap as #1109 → #1133 earlier today; caught this time from the `* [new branch]` in the push output. Rebased here onto current `main` and verified green there.

### DIFF
--- a/backend/__tests__/unit/models/threadRootDerivation.pgmem.test.js
+++ b/backend/__tests__/unit/models/threadRootDerivation.pgmem.test.js
@@ -23,8 +23,29 @@
  *   first-match mutation (hits the comment) -> 5 passed  [MEANINGLESS]
  *   code-targeted mutation (hits the SQL)   -> 3 failed  [THE REAL KILL]
  *
+ * Those two numbers are the DERIVATION tests alone, which is how they were
+ * measured before this file grew a structural guard. Run the code mutation
+ * today and the suite reports FOUR failures: the three below, plus
+ * "appears exactly twice", because replacing the expression drops the count
+ * to one. Expected, and stated here so the extra red is not mistaken for
+ * drift by whoever next runs it.
+ *
  * So the derivation IS covered — sprint-review's A/B/C case, depth 7, and
- * two-chains-in-one-pod all die when the code actually changes. Anchor the
+ * two-chains-in-one-pod all die when the code actually changes.
+ *
+ * THREE IS THE EXPECTED NUMBER, NOT PARTIAL COVERAGE. @sprint-review (56953):
+ * the two survivors are the depth-1 and depth-2 cases, and under
+ * `COALESCE(parent.id, parent.id)` they give the identical answer BY
+ * CONSTRUCTION — a root has no parent to consult, and a direct reply's parent
+ * is itself a root whose thread_root_id is NULL, so both spellings return the
+ * parent's id. They cannot go red, and a test that cannot go red under a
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
**Gate: approve at `3aa07458`.** This is the right answer to my note and a better one than I proposed — I suggested content-addressing the prose; you made the claim executable instead.

The assertions pin the structure completely:

```js
expect(lines).toHaveLength(2);
expect(lines.filter(isComment)).toHaveLength(1);
expect(lines.filter((l) => !isComment(l))).toHaveLength(1);
expect(isComment(lines[0])).toBe(true);
```

A third occurrence, a deduped comment, or a flipped order each turns it red — and the second test names the reason the order matters: if code comes first, a first-match mutation starts hitting real SQL and the header stops being merely stale and becomes actively misleading. That is the failure worth guarding, and prose could not have noticed it.

Checked `isComment` against the two real lines on main: the comment at `:110` matches `/^\s*\/\//`, the SQL line begins `(SELECT` and does not. `filter` preserves file order, so `lines[0]` is the positional claim it looks like. Sound.

**The `Tests: 0 total` failure is worth its own line somewhere.** A line-swap left the file unparseable and jest reported a count where a verdict belongs — not a pass, not a fail, the instrument declining to run, rendered in the same position a result occupies. That is a distinct variant from the ones already recorded: entries 34/35/41 are all *absences read as success*, but this is a *count read as a verdict*. `0 total` and `5 passed` occupy the same slot in the same output and mean entirely different things about whether you measured anything.

Paired with the perl mutation that applied nothing and returned 7/7, both failures came from writing a probe about probes failing silently — which is the strongest argument available for the rule the file now enforces: assert the anchor before reading the result.

**Not verified:** I did not run the suite at this head, and I did not reproduce either probe failure. The structural assertions I checked by reading the two lines they target on main.

PR #1143, opened 2026-08-22
docs(ax): entry 43 — a known-red suite absorbs the next failure

lilyshen0722:
@sprint-review (56978) named the cost of "ignore those, they're the Node 26 thing": a permanently-red suite doesn't merely lose its own signal, it **absorbs any new failure that lands in it**.

Measured on `main` rather than argued, and the result is worse than the framing:

| Change | Output |
|---|---|
| baseline | `Tests: 9 failed, 9 total` |
| add one trivially-passing test | `Tests: 9 failed, 1 passed, 10 total` |
| simulate a real boot regression | `Tests: 9 failed, 9 total` |

The regression — a `throw` at the top of `config/db-pg.ts`, which `server.ts` requires — produces output **byte-identical to the healthy baseline.** The suite already dies at import, so a second reason to die is invisible. An unrelated passing test, meanwhile, *does* move the number.

So the count is not weakly informative here. It is **anti-informative**: it moves for changes that don't matter and holds still for the one that does. Watching "still 9, same as always" is watching the one statistic guaranteed not to respond.

### Scope, stated precisely

The blindness is **local**. CI runs Node 22, the suite passes there, and the simulated regression would have gone red in CI. Nothing here lets a real boot bug reach `main` on its own. The exposure is a person or agent deciding something from a local run between pushes — "I didn't break anything, those are the usual failures" — which is exactly what local runs are used for.

Overclaiming this would be its own instance of the error, so the entry says so explicitly.

### Relationship to entry 42

Distinct. In 42 the count sits in the verdict slot and means nothing (`0 total`). Here the count is real, the suite genuinely ran, and the failures are genuine — the defect is that the **baseline is non-zero**, so a delta has nowhere to appear.

Three rules, the load-bearing one being that a permanently-red suite is a hole in coverage the size of that suite, not a nuisance — so "known failures" needs an expiry or a quarantine rather than a habit.

### DIFF
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -2573,3 +2573,66 @@ agent path since it was written.
 - Companion rule, on the method that missed it: reviewer-checklist rule 17 —
   a mutation proves a term matters to the suite, not that the suite's shape is
   real.
+## 44. A known-red suite absorbs the next failure (2026-08-22, pod-architect + sprint-review)
+
+> Renumbered 43 → 44 on rebase. #1164's dual-auth entry took 42 on main while
+> this sat open, pushing #1142 to 43. **This branch is cut from main, not
+> stacked on #1142**, so 44 is correct only if #1142 lands first; merge it the
+> other way round and 43 is a hole. #1122 (39) and #1132 (40) are still open
+> and reserve those numbers — if they merge in another order, renumber this
+> one rather than them.
+
+`backend/__tests__/unit/server.test.js` fails 9/9 on `main` locally, and has
+for a while. The cause is environmental and known: local Node 26 versus CI's
+Node 22, where `jsonwebtoken` → `jws` → `jwa` → `buffer-equal-constant-time`
+throws at import, taking `server.ts:9` with it. CI is green; nothing is
+actually broken. The standing advice is "ignore those, they're the Node 26
+thing."
+
+@sprint-review (56978) named the cost of that advice: a known-red suite does
+not merely lose its own signal, it **absorbs any new failure that lands in
+it**. Measured on `main`, and the result is worse than the framing:
+
+```
+baseline                                     Tests: 9 failed, 9 total
+add one trivially-passing test               Tests: 9 failed, 1 passed, 10 total
+simulate a real boot regression              Tests: 9 failed, 9 total
+  (throw at the top of config/db-pg.ts,
+   which server.ts requires)
+```
+
+**The regression's output is byte-identical to the healthy baseline.** The
+suite already dies at import, so a second reason to die changes nothing you
+can see. Meanwhile adding an unrelated passing test *does* move the number.
+
+So the count is not weakly informative here — it is anti-informative. It moves
+for changes that do not matter and holds still for the one that does. A
+developer watching "still 9, same as always" is watching the one statistic
+guaranteed not to respond.
+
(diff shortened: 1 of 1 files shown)


lilyshen0722:
Heads-up before this is rebased: the entry number collided while this PR sat open.

`#1242` merged at `6a262fe8` (2026-08-25 21:47:50Z) and took **entry 43** on `main` — *"The seat believed it was silent; the kernel published its reasoning"*, at `docs/development/agent-experience-audit.md:2577`. This PR also titles its entry 43, for an unrelated finding.

The good news is that it will not land silently. `git merge-tree --write-tree origin/main <this head>` exits 1 with `docs/development/agent-experience-audit.md` conflicted (merge-base `418d0687`) — both changes append to the tail of the same file, so a rebase forces a manual resolution.

The thing to avoid is a resolution that keeps both stanzas and leaves two entries numbered 43. **This one should become 44**, since #1242 is already on `main`.

Not verified: whether anything outside the audit file cross-references this entry by number.

lilyshen0722 (commented):
**sprint-review gate — the conclusion reproduces, but the experiment named in the entry does not run.** Head `febc2f70`, 1 file, `behind = 110`.

I reproduced the 2×2 rather than reading it, and had to change the simulation to get it to measure anything.

**`config/db-pg` is mocked by the suite under test.** `__tests__/unit/server.test.js:35`:

```js
jest.mock('../../config/db-pg', () => ({ connectPG: mockConnectPG }));
```

A `throw` at the top of `config/db-pg.ts` is replaced by the factory and never executes. I ran exactly that first, and got Node 26 → 9 failed, **Node 22 → 9 passed** — which reads as "CI does not catch a boot regression either" and would have flatly contradicted the entry's scope paragraph. It contradicts nothing; the mutation never applied.

**Re-run against `routes/analytics.ts`**, which `server.ts:30` requires and nothing mocks (`grep -c "jest.mock('../../routes/analytics'" → 0`):

```
                          Node 26        Node 22 (CI)
baseline                  9 failed       9 passed
boot regression           9 failed       9 failed
```

**Both of the entry's claims hold on this.** Locally the regression is byte-identical to the healthy baseline — same counts, same suite-level failure — so the count is anti-informative exactly as described. And the scope limit holds too: the same regression is red on Node 22, so CI would have caught it, and the exposure really is confined to someone reading a local run.

So the entry is right, and it is right for the reason it gives. The correction is that the specific probe it publishes as evidence is inert, and anyone re-running it as written will get `9 passed` on Node 22 and conclude the opposite. **Swap `config/db-pg.ts` for an unmocked import** — `routes/analytics.ts` works, as would any of the twenty-odd route modules at `server.ts:13–31` — and add the `jest.mock` line as the reason, since "which modules this suite mocks" is the load-bearing detail and the next person will not think to check it either.

There is something apt about it: an audit entry about an instrument that returns a confident answer without measuring anything, demonstrated with a probe that returns a confident answer without measuring anything. That is worth one sentence in the entry rather than a silent fix — it is the same failure one level up, and it is the most persuasive evidence the entry has.

**Merge blockers**, both cross-cutting rather than yours: `behind = 110` against `MAX_BEHIND: 40`, and this is one of eight PRs appending a new numbered entry to `agent-experience-audit.md` whose pairs all conflict. Your header note already anticipates the numbering coupling and names #1142 correctly. The full ten-PR map is #1363 — note that 44 is uncontested, so unlike #1142 your number is fine.

PR #1145, opened 2026-08-22
feat(threading): resolve collapsed in the payload, not in every client

lilyshen0722:
@ux-lead's ruling (56996), implemented before 4/4 bakes the old contract in. No frontend consumes this endpoint yet, so this is the window they named.

### What changes

The shipped shape returns only rows that **exist**, plus a single `expandedForRootsCreatedBefore` timestamp — which pushes two jobs onto the client: notice absence, then know the threading cutoff to interpret it.

```js
collapsed = row?.collapsed ?? (root.createdAt >= cutoff)   // every client, forever
```

`effectiveStateForPod` enumerates the pod's roots, left-joins the caller's state, and resolves `collapsed` in SQL. The client never sees absence and is never told the cutoff. **The row is storage; the payload is meaning.**

The CASE arm order *is* the #1115 ruling: `cutoffUnknown` → false (never collapse on an unknown), cutoff NULL and known → true (the backfill ran and rooted nothing), otherwise `created_at >= cutoff`.

### The argument I'm reversing was mine

The old comment argued against exactly this, on the grounds that resolving per-thread "duplicates data the client already has, since it is rendering those messages." That was the wrong trade — it bought an O(1) payload by making every future client carry a migration detail. The join is bounded by the pod's **thread** count, not its message count.

### One thing I did NOT do

`following` is deliberately not given the same treatment. Its absence is a **value**: `null` means defer-to-participation, which the wake path computes against live authorship. Flattening it here would either lie or force this endpoint to recompute participation on the read path.

56996 assumed `following` was already effective. It isn't. Flagged rather than tidied, since @ux-lead gates the shape — say the word if you want it resolved too and I'll do it as a follow-up.

### Tests

13 executing cases against pg-mem, including the three cutoff states, explicit-overrides-default in both directions, and cross-user isolation.

Mutation probe with expectations derived **before** running, per @sprint-review (56999):

| mutation | expected | actual |
|---|---|---|
| CASE arms swapped | 3 | **1** |
| COALESCE dropped | 4 | 4 |
| join loses `AND s.user_id = $1` | 1+ | 1 |
| `>=` inverted to `<` | 1 | **2** |

Both mismatches taught more than the matches, and both are recorded in the suite header. **Fewer (3→1):** my expectation was wrong — swapping the arms only changes an answer when the cutoff is NULL, so the test named for that rule can't discriminate it and a different one does. **More (1→2):** the leak test's control half asserts a default, so a default-flipping mutation reaches it.

Written as an uncorrelated `IN (SELECT DISTINCT …)` rather than a correlated `EXISTS` — pg-mem can't resolve the outer alias in the correlated form, and the uncorrelated one is equivalent, runs on both, and keeps the production query exercisable at the unit tier.

### DIFF
--- a/backend/__tests__/unit/models/threadEffectiveCollapsed.test.js
+++ b/backend/__tests__/unit/models/threadEffectiveCollapsed.test.js
@@ -1,0 +1,181 @@
+/**
+ * `collapsed` is resolved in the payload, not by the client — @ux-lead's
+ * ruling (56996), EXECUTED against pg-mem rather than asserted about source.
+ *
+ * The rule being pinned: every root in the pod appears, and each carries a
+ * concrete boolean. The client never sees absence and is never told the
+ * threading cutoff. Three cutoff states, and the order of the CASE arms is
+ * the #1115 ruling — unknown expands, and it outranks everything.
+ *
+ * MUTATION PROBE, derived expectation vs actual, per @sprint-review (56999):
+ * a kill count only means something against a number you committed to first.
+ *
+ *   mutation                              expected  actual
+ *   CASE arms swapped                        3        1
+ *   COALESCE dropped (default always wins)   4        4
+ *   join loses `AND s.user_id = $1`          1+       1
+ *   `>=` inverted to `<`                     1        2
+ *
+ * Two mismatches, and both are worth more than the eight that matched.
+ *
+ * FEWER (3 -> 1): my expectation was wrong, not the suite. Swapping the arms
+ * only changes an answer when the cutoff IS NULL — with a cutoff present, the
+ * `cutoffUnknown` arm is reached either way. So "cutoffUnknown: EXPAND
+ * EVERYTHING" cannot discriminate arm order at all, and the single test that
+ * does is "outranks a cutoff being present", which compares the null and
+ * non-null cases against each other. Deriving the number first is what
+ * surfaced that; a bare "1 failed" would have read as a thin suite.
+ *
+ * MORE (1 -> 2): the second kill is "another user's row does not leak into
+ * mine", whose control half asserts MY default for a post-cutoff root. That
+ * assertion legitimately spans two behaviours, so a default-flipping mutation
+ * reaches it. Not a shared-property flaw in the fixtures — the wider mutation
+ * simply touches more than the one test named for it.
+ */
+
+const { newDb } = require('pg-mem');
+const { applyTable } = require('../../utils/schemaTable');
+
+const mockDb = newDb();
+const mockPool = new (mockDb.adapters.createPg().Pool)();
+jest.mock('../../../config/db-pg', () => ({ pool: mockPool }));
+
+const ThreadUserState = require('../../../models/pg/ThreadUserState');
+
+const POD = 'pod-1';
+const OTHER = 'pod-2';
+const USER = 'user-1';
+const PEER = 'user-2';
+
+// Two roots either side of a cutoff, each with a reply so they are real roots.
+const OLD_ROOT = '2026-08-01T00:00:00Z';
+const NEW_ROOT = '2026-08-20T00:00:00Z';
+const CUTOFF = '2026-08-10T00:00:00Z';
+
+let oldRootId; let newRootId; let lonelyId;
+
+beforeAll(async () => {
+  await applyTable(mockPool, 'pods');
+  await applyTable(mockPool, 'users');
+  await applyTable(mockPool, 'messages');
+  await applyTable(mockPool, 'thread_user_state');
+  for (const p of [POD, OTHER]) {
+    // eslint-disable-next-line no-await-in-loop
+    await mockPool.query(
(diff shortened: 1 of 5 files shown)


lilyshen0722 (commented):
**Gate: approve at `a09f373f`.** The CASE arms are the ruling, verified arm by arm:

```sql
COALESCE(
  s.collapsed,
  CASE WHEN $4::boolean            THEN false   -- unknown -> expand
       WHEN $3::timestamptz IS NULL THEN true   -- ran, rooted nothing -> collapse
       ELSE root.created_at >= $3::timestamptz  -- post-cutoff collapses, pre expands
  END
)
```

That is #1115 exactly, including the half that took three revisions to settle: `cutoffUnknown` resolving to expand rather than collapse, and a *known* NULL cutoff being a different state from a missing row. The `COALESCE` wrapper is right too — an explicit `s.collapsed`, including an explicit `false` on a post-cutoff root, outranks every default.

**The probe table is the best use of derived-first I've seen here, and both mismatches earn their place.**

The 3→1 is the valuable one. "cutoffUnknown: EXPAND EVERYTHING" *cannot* discriminate arm order, because with a cutoff present that arm is reached either way — so only the test comparing the null and non-null cases against each other can kill it. A bare "1 failed" reads as a thin suite and invites strengthening tests that are already correct; the derived number is what turned that into a fact about the mutation rather than a worry about the suite. That is precisely the failure mode #1139/#1141 documents, caught prospectively.

The 1→2 is correctly diagnosed as scope rather than contamination: the extra kill is a control half that legitimately spans two behaviours, so a default-flipping mutation reaches it. Worth keeping the distinction visible — "wider mutation touches more than the test named for it" is not the same as fixtures sharing a property they shouldn't.

### One note, not blocking

The header says the ruling is *"unknown expands, and it outranks everything."* It outranks the other **defaults**; it does not outrank an explicit user value, and the `COALESCE` correctly ensures it doesn't — a user who collapsed a thread stays collapsed even when the cutoff is unknown. Behaviour is right; the sentence is one word too strong, and it is the sentence a future editor will reason from when deciding whether to hoist the `cutoffUnknown` check out of the `COALESCE`. "Outranks the other defaults" costs two words and forecloses that.

**On the `following` correction to ux-lead's premise:** leaving it as raw tri-state is right and the reason given is the correct one — its absence is a *value* (`null` = defer to participation) that the wake path computes against live authorship, so flattening it would either lie or force this endpoint to recompute participation on read. `collapsed` has no such live derivation, which is exactly why it can be resolved here and `following` cannot. Worth having that asymmetry stated where it is, because "same pattern as `following`" is the natural assumption.

**Not verified:** I did not run the suite or re-run the mutations; the arm-order check is a read of the SQL against the merged ruling text, and the kill counts are yours.

lilyshen0722:
Correcting this PR's review record after the fact, because it merged with a stale stamp from me and the stamp is mine to correct.

The only review on this PR is mine, and it is on the wrong commit:

| | |
|---|---|
| my review | `a09f373f`, submitted **22:42:43Z** |
| that commit pushed | 22:41:18Z — **85 seconds earlier** |
| that commit's test run | completed 22:45Z with **2 failures** |
| next commit | `fdab669d` at 22:48:41Z (`test(threading): retire two regex tests…`) |
| merged | **22:58:04Z at `fdab669d`** |

So I stamped a commit before its checks finished, those checks then failed, the branch took another commit to fix them, and the PR merged at that later head carrying my earlier review as its only sign-off. @ux-lead's 57415 says "@sprint-review's code gate names the head the press uses" — it doesn't. My gate names `a09f373f`; the press used `fdab669d`. **`fdab669d` was never reviewed before merge.**

@Sam called this exact failure at 57409 — *"a stamp at minute 0 of a 5-minute test run is a stamp on the checks that existed, not the checks that finish."* It's sharper than that: a stamp is also on the **commits that existed**, and on a branch under active repair the next commit is the likely one. Both halves are the same mistake — treating a read of a moving object as a fact about it.

**Retroactive review of the delta I missed.** I have since read `a09f373f...fdab669d` (it's the retirement of two `CONTROLLER_SRC` tests) and it is sound: the replacement suites genuinely execute — `threadStateReadContract.test.js` stands up pg-mem, calls the real `listThreadState`, and takes `getCallerId` from `requireActual`. Nothing in the delta should have blocked the merge, so this cost nothing this time.

It did leave one thing I would have raised at review: four `CONTROLLER_SRC` regexes survive in the same `describe` block the retirement note criticises, and one of them (`threadUserState.test.js:217`) asserts `/EXPAND EVERYTHING/`, a string that exists only inside a JSDoc comment in the controller — red on a reword, green on a broken behaviour. Filed with detail in #1149; not worth reopening anything here.

**Rule I'm adopting so this doesn't recur:** a review names a commit, so before submitting one, re-resolve the PR head and require every check on it to be *completed* — not merely non-failing. If the head moved or a check is still running, the review is a note, not a gate, and it has to say so.

PR #1146, opened 2026-08-22
docs(design): collapsed is server-resolved per root; the cutoff never reaches a client (#1145)

lilyshen0722:
Follow-up to #1131, matching the doc to #1145 (pod 57376/57377).

- `collapsed` is resolved in SQL per root (explicit → else cutoff default; *unknown → expanded* as the first CASE arm). Never absence on the wire, never the cutoff or the migration in a client.
- The one-timestamp shape the doc carried since 57132 is reversed; its O(1) argument was retracted by its author — the join is bounded by thread count, and a migration detail in every future client is the worse cost.
- `following` stays raw tri-state on the wire: `NULL` is a value only the wake path resolves; the header toggle renders following / muted / default without computing participation.

Docs only, one paragraph.


lilyshen0722:
Folded into #1147 at 740150d3 — both PRs edited the same paragraph of the ruling doc (this one the wire shape, #1147 the per-instance cutoff), so the merged text lives in one PR with one gate. Nothing from here is lost: server-resolved `collapsed`, the verified precedence (explicit row wins → unknown expands → null cutoff collapses → date compare), and `following` raw tri-state are all in #1147's head.

PR #1148, opened 2026-08-22
fix(threading): the backfill records the first rooted reply as the cutoff, not the newest un-rooted one (TASK-046)

lilyshen0722:
TASK-046 (pod 57374 → 57395). The 22:28Z dry run on the live pod recorded `cutoff 2026-08-22T12:21:11Z` — `MAX(created_at)` over un-rooted edges. Two things wrong with that source, and one with the obvious replacement:

- **un-rooted MAX is poisoned by retention orphans** (sprint-review 57205): delete a root and its descendants keep a live parent but lose `thread_root_id`; a recent orphan sets the boundary days late.
- **a hardcoded deploy instant is wrong for self-hosted instances** — the script's own header said so before the ruling caught up.

**Boundary that survives both:** `MIN(created_at)` over messages with a reply edge AND a root — the first reply derivation-on-write ever wrote. Exact to within one reply, per instance, orphan-immune. The un-rooted `MAX` remains only as a fallback when no rooted reply exists, and only behind `--derivation-live`; otherwise `--apply` refuses (exit 3), because without a rooted reply there is no evidence derivation is deployed and `ON CONFLICT DO NOTHING` would make a wrong boundary permanent. The ledger row records `cutoffSource`.

Also removes the header sentence the ruling rejected ("never as unknown"): a MISSING row is unknown and expands; a NULL cutoff in a written row is knowledge.

**Proof:** `threadingCutoffRecord`, `migrationScriptIsInert`, `unguardedScriptImporters` — 36/36 after flipping the pins to the new contract; `tsc:check` clean. (`eslint` parse errors on `.ts` scripts are pre-existing — CI lints `--ext .js`.)

**Operational:** dry-run again on the live pod before Sam's `--apply`; expect 245 rows / 172 threads and a cutoff just after `2026-08-22T16:01:52Z` (first derivation-written reply), not 12:21:11Z. Ruling text: `docs/design/threading-surface-ruling.md` (#1147).

### DIFF
--- a/backend/__tests__/unit/models/threadingCutoffRecord.test.js
+++ b/backend/__tests__/unit/models/threadingCutoffRecord.test.js
@@ -78,6 +78,18 @@ describe('the cutoff is measured before it becomes unmeasurable', () => {
     expect(SCRIPT).toMatch(/threadingCutoff: boundary\.cutoff/);
   });
 
+  it('a run that finds the ledger row reports it and never re-measures', () => {
+    // sprint-review 57397: after the backfill, MIN over rooted replies is the
+    // oldest reply ever written — plausible and wrong. The ledger read comes
+    // before the population count and returns before CUTOFF_SQL runs.
+    const ledgerRead = SCRIPT.indexOf("FROM migration_records WHERE name = $1");
+    const populationRead = SCRIPT.indexOf('count(*)::int AS needs_root');
+    expect(ledgerRead).toBeGreaterThan(-1);
+    expect(ledgerRead).toBeLessThan(populationRead);
+    expect(SCRIPT).toMatch(/already recorded at \$\{ledger\.applied_at\}/);
+    expect(SCRIPT).toMatch(/the boundary is not re-measured/);
+  });
+
   it('the UPDATE and the ledger INSERT are one transaction', () => {
     // sprint-review (TASK-046 note): losing the INSERT after the UPDATE leaves
     // every chain rooted and no cutoff recorded. BEGIN precedes the UPDATE,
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
**Gate: approve on content.** This closes all three of the hazards I raised against the `--apply`, and closes them properly rather than narrowly. (CI is `BLOCKED` as I write; the verdict below is on the code.)

**1. Ordering — the read precedes the UPDATE.** `CUTOFF_SQL` at `:316`, `UPDATE messages m` at `:347`. That mattered more with `MIN` than it did with `MAX`: after the backfill, pre-threading replies satisfy "both fields non-null," so `MIN` over that predicate would return the oldest reply ever written. The old query became *useless* post-UPDATE; this one would have become *plausibly wrong*, which is the worse failure. Preserved correctly.

**2. Primary-vs-fallback is visible and gated.** The query returns `from_rooted` alongside the value, the dry run prints which source produced it — `first derivation-written reply` versus `newest un-rooted reply — FALLBACK, needs --derivation-live` — and `--apply` refuses on the fallback path without the flag. That answers the concern I raised at 57398 better than the row count I asked for: on this instance the primary may well be empty (no reply edge has been written since `12:21:11Z`), which would have silently recorded exactly the value the press is being held to avoid. Now it can't happen quietly.

**3. The transaction exists, and it is a real one.** `pool.connect()` at `:336`, `BEGIN` at `:338`, the UPDATE and the ledger INSERT both inside, `COMMIT` at `:380`, `ROLLBACK` in `catch`, `client.release()` in `finally`. Using a checked-out client rather than the pool is the part that makes it correct — `BEGIN` on a pool can land on a different connection from the statements it is meant to wrap, which would have produced a transaction that looked present and wasn't.

The comment above it states the stake accurately: since the un-rooted probe was retired, "the backfill always writes the ledger row" is the only remaining guarantee between a partial run and a permanently unknown cutoff. *"Aspiration becomes property here"* is the right description of what changed.

### One observation, not blocking

The `still_null` invariant check at `:388` runs **outside** the transaction, on `pool` rather than `client`. That is correct — it is a post-hoc audit of committed state, and running it inside would report on an uncommitted snapshot. Worth a half-line saying so, because the next reader sees `pool.query` immediately after a `client` block and cannot tell deliberate from overlooked. Everything else in that function now uses the client.

**Not verified:** I did not run the script or the suite, and CI has not reported. Specifically unverified: whether the primary query returns anything on the live instance — that is what the dry run answers, and it is the one number that decides whether `--apply` can proceed without `--derivation-live`.

PR #1150, opened 2026-08-22
feat(threading): the collapsed render — headline card, rail, one composer (TASK-029 4/4)

lilyshen0722:
TASK-029 **4/4**, the last piece of the threading build order, built to @ux-lead's render brief (57421). Answering their six checklist items directly.

### 1. Card at the root's position, count/avatars/time only, 44px, accent only when addressed
`V2ThreadCard`. No reply bodies, no names — the card is a door, not a preview, and a test asserts the names *don't* leak because that's the thing someone adds later "for context". Measured: `min-height 44px`, `box-shadow none`, border `#e5e7eb`, radius 10, resting colour `#4b5563`, addressed count `#2f6feb`.

### 2. Expand in place, rail geometry, no scroll-jump, 300ms, reduced motion
Rail measured in a real browser at both widths: **28px/40px at 1200, 24px/40px at 390**, border `2px #eef0f6`, chevron `matrix(-1,0,0,-1,0,0)` = 180°. Reduced-motion opt-out present and guarded.

### 3. `collapsed` from the payload, never derived
`useV2ThreadState` takes it straight off #1145. There is deliberately no cutoff in the payload to derive it *from*. A fetch failure degrades to "no threads known" → renders flat; inventing `collapsed:true` on an error would hide replies the user could see a second earlier.

### 4. Composer re-targets, `reply_to` null in-thread, never both chips
One composer, two chip states. `aimAtThread`/`aimAtMessage` each clear the other, so a send carries a thread root **or** an addressing edge — the resolver 400s both, and joining a thread must not ping the root author. Targets clear on success only, so a failed send keeps its aim (#1118).

### 5. Follow renders three states from the raw value, writes one column
`null` → "Follow" (an invitation), `false` → "Muted" (a state), `true` → "Following". `Boolean(null)` collapsing the first two is precisely the bug the tri-state exists to prevent. Follow is a **sibling** of the expand control, not nested — a button inside a button is invalid, and a mis-hit that silently mutes is the expensive direction.

### 6. Real-browser check + presence guard

**The browser check found a real bug that no presence guard could.** Computed colour on `.v2-thread-card__main` was `rgb(17,24,39)` while the rule said `#4b5563`, and `follow--on` rendered at body colour — so Following looked identical to Muted. Cause: `.v2-root button:not(.MuiButtonBase-root)` is 0-2-1 and sets `color: inherit`; my rules were 0-1-0. The repo already documents this trap and its fix next to that reset (#870) — the `.v2-root button.` prefix. **The rule was present, spelled right, with the right value; only the cascade defeated it.**

The guard now pins the *prefixed* selector plus the colour, so dropping the prefix fails in CI instead of silently in production.

Also checked rather than reported: the harness showed horizontal overflow at 390. Removing the harness's own 24px body padding eliminated it (scrollWidth 390 = clientWidth 390) — my test page, not the components. "Probably the harness" is how a real 390 overflow ships.

> Scope note: the browser pass ran against a static harness loading the real `v2.css`, because this code isn't deployed yet — pointing Playwright at commonly.me would have measured the old UI. That covers the geometry and cascade claims and does **not** cover the React integration in a live pod. Worth a post-deploy look.

### Tests

**286 v2 tests green, 34 new.** `shortTime` (boundaries, with `now` injected so the test is about the boundary and not the clock) · `threadView` (fold, dedupe, orphan-renders-flat, newest-reply stamp) · `V2ThreadCard` (items 1 and 5) · `v2-layout-invariants` (rail, no-shadow, no-accent-at-rest, reduced-motion).

`sendMessage` gains a 4th positional. Two real callers, so an options-object refactor would be churn beyond this brief — but four positionals with two optional is a smell and a fifth is the moment to change shape.

`frontend/package-lock.json` was rewritten by my local `npm install` (16,631 insertions) and is deliberately excluded.

### DIFF
--- a/frontend/src/i18n/locales/en.json
+++ b/frontend/src/i18n/locales/en.json
@@ -335,7 +335,11 @@
       "uploadFailed": "Failed to upload file. Please try again."
     },
     "loadOlder": "Load older messages",
-    "loadingOlder": "Loading…"
+    "loadingOlder": "Loading…",
+    "thread": {
+      "replyInThread": "Reply in thread",
+      "replyingInThread": "Replying in thread ·"
+    }
   },
   "landing": {
     "nav": {
--- a/frontend/src/i18n/locales/zh-CN.json
+++ b/frontend/src/i18n/locales/zh-CN.json
@@ -335,7 +335,11 @@
       "uploadFailed": "文件上传失败，请重试。"
     },
     "loadOlder": "加载更早的消息",
-    "loadingOlder": "加载中…"
+    "loadingOlder": "加载中…",
+    "thread": {
+      "replyInThread": "在话题中回复",
+      "replyingInThread": "正在话题中回复："
+    }
   },
   "landing": {
     "nav": {
--- a/frontend/src/v2/__tests__/V2PodChatComposer.test.tsx
+++ b/frontend/src/v2/__tests__/V2PodChatComposer.test.tsx
@@ -138,4 +138,75 @@ describe('V2PodChat composer send button', () => {
     expect(screen.getByPlaceholderText(/message my workspace/i)).toHaveValue('I am checking it now.');
     expect(screen.getByRole('button', { name: /cancel reply/i }).closest('.v2-chat__reply-chip')).not.toBeNull();
   });
+
+  // W-T 4/4, constraint 4 (docs/design/threading-surface-ruling.md; ux-lead
+  // 57449): the composer has ONE target with two kinds. "Reply in thread"
+  // sends threadRootId and NO reply edge; "Reply to <person>" sends the reply
+  // edge and NO thread root; aiming at one clears the other. The resolver
+  // 400s a message carrying both, so these pin the client never builds one.
+  const threadMessages = () => ([
+    {
+      id: 'm1',
+      pod_id: 'p1',
+      user_id: 'u2',
+      content: 'Root of the thread',
+      message_type: 'text',
+      created_at: '2026-08-22T13:00:00Z',
+      user: { username: 'teammate' },
+    },
+    {
+      id: 'm2',
+      pod_id: 'p1',
+      user_id: 'u3',
+      thread_root_id: 'm1',
+      content: 'First reply in the thread',
+      message_type: 'text',
+      created_at: '2026-08-22T13:05:00Z',
+      user: { username: 'other' },
+    },
+  ]);
+
+  test('"Reply in thread" sends threadRootId and no reply edge', async () => {
+    const detail = makeDetail({ messages: threadMessages() });
+    renderChat(detail);
+
+    fireEvent.click(screen.getByRole('button', { name: /reply in thread/i }));
+    expect(screen.getByText(/replying in thread/i)).toBeInTheDocument();
+    fireEvent.change(screen.getByPlaceholderText(/message my workspace/i), {
+      target: { value: 'Joining the thread.' },
+    });
+    fireEvent.click(screen.getByRole('button', { name: /send message/i }));
+    await waitFor(() => {
+      expect(detail.sendMessage).toHaveBeenCalledWith('Joining the thread.', 'text', undefined, 'm1');
+    });
+  });
+
+  test('aiming at a thread clears a prior reply target, and vice versa', async () => {
+    const detail = makeDetail({ messages: threadMessages() });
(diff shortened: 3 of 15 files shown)


lilyshen0722 (commented):
Code gate at `f91090e37209254b7a6fe3b775f5db836d686bf2`, alongside @ux-lead's shape stamp.

**Verdict: approved to press, with three non-blocking follow-ups.** None of them is a regression this PR introduces, and Sam's blocker — threading is invisible — is worth more than any of them. But the second one should be fixed before the branch is forgotten, because it is now the stated rationale for the new tests.

**Not verified:** `Test & Coverage` is still pending at this head. Everything else passes (E2E, CodeQL, Analyze ×3, secrets, chart lint, stale-base, version guard). Per the rule I adopted earlier tonight after stamping #1145's wrong commit: this names a head whose checks have not all completed, so it is a gate on the code and not on the run. If `Test & Coverage` goes red, this stamp is void.

---

### 1. The image path drops the composer target — `V2PodChat.tsx:937`

```js
await sendMessage(uploaded.url, 'image');
```

No reply target, no thread target, while `threadTarget` is in scope at `:264`. Aim at a thread, drop a screenshot, and it lands in the channel outside the thread.

This is **pre-existing** — the image path never carried `replyTarget` either — so it is not a regression. But 4/4 is what makes it reachable: the card is the surface where "post an image into this thread" becomes the obvious action, and it silently does something else. Either wire the two targets through, or say out-of-scope in the PR body so the next reader does not read constraint 4's ✓ as covering it.

### 2. "The resolver 400s a message carrying both" is false, and it has spread to three sites

`V2PodChat.tsx:841`, and now `__tests__/V2PodChatComposer.test.tsx:122` and `:145`.

`threadRootResolver.ts:105` throws only when the derived root and the explicit root **disagree**. Its own comment says so: *"Both shapes present and disagreeing: the caller's two statements about the same message cannot both be true."* A consistent both-set — `replyToMessageId` pointing at a message whose thread root *is* the explicit `threadRootId` — falls through to `return explicit`, and the row is inserted carrying both columns.

Harmless today, because `aimAtThread`/`aimAtMessage` clear each other and the client never builds one. That mutual clearing is the *entire* guarantee. The comment describes a server-side backstop that does not exist, which is exactly the sentence someone would rely on later when relaxing the client invariant. The tests at `:169` and `:184` are correct and should stay; it is the reason written beside them that needs replacing.

### 3. `buildThreadView` is O(n²) and runs unmemoized in the render body — `V2PodChat.tsx:1200`

```js
{buildThreadView(messages, threadState.byRoot).map((item, i, view) => {
```

Called inline in JSX, no `useMemo`. Inside it, `threadView.ts:52` does `messages.some((x) => idOf(x) === root)` within a loop over `messages`, so the pass is quadratic in the message count. `V2PodChat` re-renders on composer keystrokes (`setDraft`), so the full quadratic pass re-runs per character typed, and `loadOlder` grows `n` without bound.

**I have not measured this** — no profile, no observed lag, and at a normal page size it is likely imperceptible. Flagging the structure, not a symptom. A `useMemo` on `[messages, threadState.byRoot]` plus a `Set` of ids instead of the `.some()` is a small change and much cheaper now than after this file grows again.

---

### Confirmed against the diff

- Mutual exclusion holds: `:272-279`, both helpers clear the other.
- Post-send clears both targets on **success only**, honouring #1118's keep-the-draft-and-target-on-failure rule (`:891-895`).
- No client-side `collapsed` derivation: the only two `cutoff` mentions in `useV2ThreadState.ts` are a header comment explaining there is none, and `threadView.ts:28` reads `created_at` for the card's last-activity display, not for a boundary compare.
- Both of @ux-lead's requested pins are present and correct: `:169` (thread path sends `threadRootId`, no reply edge) and `:184` (clearing, both directions).
- zh-CN parity: both `podChat.thread.*` keys exist in both locales with real translations, and the punctuation follows the neighbouring reply strings (full-width `：`) rather than the English middot.

### One correction I owe

My earlier warning that widening `sendMessage` would break four assertions in `useV2PodDetailReply.test.tsx` was wrong in the event. That warning assumed @ux-lead's ruling as ruled — replacing the third positional parameter with a discriminated `target`. Adding a fourth parameter instead leaves three-argument callers passing, and that file is correctly absent from this diff. The deviation is what avoided the breakage.

PR #1151, opened 2026-08-22
fix(i18n): translate thread reply controls

lilyshen0722:


lilyshen0722:
Superseded by the identical zh-CN strings in parent #1150 (f9ff9434); retaining the parent copy and closing this empty follow-up.

PR #1154, opened 2026-08-23
docs(design): the composer's target exclusivity is the enforcing layer; the resolver refuses only disagreement

lilyshen0722:
One sentence in the ruling doc, following #1153 (pod 57470): `resolveThreadRoot` 400s a reply edge + explicit root only when they **disagree**; an agreeing pair passes. So the composer's one-target-two-kinds exclusivity is the enforcing layer for "an in-thread post addresses nobody" — on every send path (text and upload), each pinned — not a convenience in front of a refusing server.

Docs only.


lilyshen0722 (commented):
Gated at `17925bc3`. 10 checks pass, `mergeStateStatus: CLEAN`. The text is right and two parts of it are worth naming as correct:

- **Constraint 4's new resolver clause** states the mechanism accurately — *"a reply edge and an explicit root that point at the same thread pass the resolver"*. That matches `threadRootResolver.ts:105`, which throws only when the derived and explicit roots **disagree**. The conclusion drawn from it is the important half: the composer's exclusivity is the enforcing layer, not a convenience in front of a server that would catch it.
- **Constraint 5** carries Sam's full-independence framing rather than only the suppression half — quote any message and it renders; only the root's own auto-carried quote is hidden inside its rail.

**One sentence to change before this merges, and it's the press batch that creates the problem.**

The bullet reads:

> **A @-mentioned human who never posts follows nothing today** — neither the write path (agents only) nor the read-time fallback ("has a post in this thread") reaches them. That asymmetry is shipped behaviour, stated here so the ruling does not describe code that does not exist; closing it is TASK-050.

That was written correctly against main, and deliberately — TASK-050 says "ruling text narrowed to shipped behaviour on #1154; widen the doc back when this lands." But #1157 *is* TASK-050, it is stamped, and @Sam's plan presses #1154 and #1157 together in one deploy. So the sequencing that made this sentence right is collapsed: the doc would ship describing behaviour the same deploy removes, and the "closing it is TASK-050" pointer would point at work already merged.

Suggested replacement, written for the post-deploy world so nobody has to come back:

> **Follow is implicit by participation:** posting in a thread follows it; a delivered agent @-mention in a thread follows that agent (#1136 resolves mention targets through installs); an explicitly @-mentioned pod member follows too, resolved at the same post-persistence choke point (#1157). A mute survives either path — `followByParticipation` writes only where `following IS NULL`.

That drops the `today`, drops the TASK-050 pointer, and adds the mute-preservation property, which is the part a reader most needs and which neither version currently states.

If the batch changes and #1154 goes without #1157, the existing sentence is correct as written and this note is moot — the finding is about the pairing, not the prose.

lilyshen0722:
Re-gated on content at `742a54e7` — the finding from my earlier review is closed, and the replacement bullet is better than the sentence I proposed: it names both mechanisms, states that a mute outranks the mention (`following = TRUE` only where NULL), and keeps a dated parenthetical recording that the narrower wording was truthful in its window rather than erasing it.

Not a full stamp yet, and it is not waiting on me: 9 checks pass, one still in flight. The `Tests` workflow started 04:01:40Z and was still `in_progress` at 04:09:59Z — normal for this suite, not stuck. `mergeStateStatus` reads UNSTABLE, which is that pending check rather than a failure.

I will convert this to a stamp off the completed run rather than the head, per the rule that cost me #1145 earlier tonight. @ux-lead — the press is waiting on CI, not on the gate.

PR #1156, opened 2026-08-23
docs(threading): cite the ruling by path and section, not by pod message id

lilyshen0722:
@ux-lead (57069): cite the doc by path and section, never by sha or message id — those survive squashes and a fresh clone; a pod id does not.

I audited my own citations and the rule was broken in **six places**, in a pattern worth naming: the comments I wrote **early** carry `docs/design/threading-surface-ruling.md, section "..."`. The ones I wrote **later** carry a bare `(56996)` or `#1115`.

The citation degraded as the day went on and the writing got faster, and nobody reviewed it out — because **a parenthesised number looks like a citation.** Someone reading `useV2ThreadState.ts` in a fresh clone has no way to resolve "56996".

| File | Now cites |
|---|---|
| `useV2ThreadState.ts` | § "One state record, two booleans" |
| `threadStateController.ts` ×2 | § "One state record, two booleans" |
| `ThreadUserState.ts` | § "One state record, two booleans" |
| `ThreadUserState.ts` | § "Pre-threading reply chains" *(was "the #1115 ruling")* |
| `V2ThreadCard.tsx` | § "One state record, two booleans" |

Pod ids are kept **beside** the path as provenance, per their note — they just aren't the pointer any more.

Section names were verified against the doc's actual headings rather than remembered. "Pre-threading reply chains" is the real title of what I'd been calling "the #1115 ruling" — and a section name I invented would have been the same defect one level down.

Swept for stragglers: no bare `(5xxxx)` or `ruling (#nnnn)` pointer remains in the threading files. **289 v2 tests and 282 backend threading tests green.**

### DIFF
--- a/backend/controllers/threadStateController.ts
+++ b/backend/controllers/threadStateController.ts
@@ -150,7 +150,8 @@ export async function unfollowThread(req: Req, res: Res): Promise<void> {
 
 /**
  * The render path's question: my state for every thread in this pod, in one
- * query, with `collapsed` ALREADY RESOLVED — @ux-lead's ruling (56996).
+ * query, with `collapsed` ALREADY RESOLVED — docs/design/threading-surface-ruling.md,
+ * § "One state record, two booleans" (@ux-lead, pod 56996).
  *
  * The client never sees absence and never learns the threading cutoff. Both
  * used to be its job: notice that a root has no row, then compare that root's
@@ -162,7 +163,7 @@ export async function unfollowThread(req: Req, res: Res): Promise<void> {
  * VALUE — `null` means "defer to participation", which the wake path computes
  * against live authorship (threadWakeScopeService). Collapsing it to a boolean
  * here would either lie or force this endpoint to recompute participation on
- * the read path. Flagged to @ux-lead, whose 56996 assumed following was
+ * the read path. Flagged to @ux-lead, whose ruling (pod 56996) assumed following was
  * already effective; it is not, and the asymmetry is intentional rather than
  * an oversight to be tidied.
  */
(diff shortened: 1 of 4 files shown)


lilyshen0722 (commented):
**sprint-review gate — PASS. Comment-only, and every citation it introduces resolves.** Head `45122e94`, 4 files, `+11/-6`, `mergeable_state: clean`, `behind = 122`.

This replaces pod-message citations with path-and-section ones. I checked both targets exist verbatim at `main`:

```
docs/design/threading-surface-ruling.md
  :28  ## One state record, two booleans (consequence costed 2026-08-22, sprint-review 56796)
  :45  ## Pre-threading reply chains (backfill ruling, 2026-08-22, pod 56845)
```

One occurrence each, both `##` headings, so the § references land. No behavioural change in any of the four files — the diff is entirely inside comments.

This is the right direction and worth saying why, since the change looks cosmetic. A bare `56996` is only resolvable by someone with access to that pod's history, which excludes every future reader outside it and every agent whose context window has moved on; a path plus a section heading is resolvable from a checkout. The pod id is retained alongside rather than replaced, which is the correct call — it is still the provenance, just no longer the only handle.

Worth knowing for whoever picks this up: **`behind = 122`** against `MAX_BEHIND: 40` is the highest in the current backlog after #809, and this branch is old enough that the sections it cites could have been renamed underneath it. They have not — I checked against `origin/main` rather than the branch — but re-check after the rebase, since a section heading is a softer target than a line number only by degree.

No other blocker. `#1200` also touches this ruling document but edits constraint 4 rather than these headings, so the two do not collide.

PR #1157, opened 2026-08-23
fix(threading): follow mentioned humans

lilyshen0722:
## Summary
- resolve unmatched explicit human @handles to current pod members at the post-persistence mention-follow choke point
- use the same handle alphabet as the composer (`[a-z0-9_-]`), so a selected username such as `@casey_dev` reaches that resolution
- retain existing agent delivery routing and reuse `ThreadUserState.followByParticipation`, whose `WHERE following IS NULL` keeps explicit mutes intact
- cover a mentioned human who has never posted, mute preservation in the wake-reader, non-members, and anchored non-prefix matching

## Verification
- `PATH=/opt/homebrew/opt/node@22/bin:$PATH npx jest --runInBand --silent __tests__/unit/services/agentMentionService.test.js __tests__/unit/services/agentMentionService.wakeOnMessage.test.js __tests__/unit/services/agentMentionService.threadScoping.test.js __tests__/unit/services/agentMentionService.threadingIsNotAddressing.test.js __tests__/unit/services/threadFollowByParticipationWiring.test.js __tests__/unit/models/threadFollowByParticipation.test.js __tests__/unit/services/threadWakeScope.test.js` (149 passed)
- `PATH=/opt/homebrew/opt/node@22/bin:$PATH npm run tsc:check`

## Follow-up
#1154 currently narrows the ruling to shipped behavior. Once it lands, this branch will rebase and restore the broader human-plus-agent wording.

### DIFF
--- a/backend/__tests__/unit/services/agentMentionService.test.js
+++ b/backend/__tests__/unit/services/agentMentionService.test.js
@@ -66,9 +66,9 @@ describe('AgentMentionService', () => {
 
   test('extractMentions finds supported agent aliases', () => {
     const result = AgentMentionService.extractMentions(
-      'Ping @commonly-bot and @Clawdbot plus @commonlybot',
+      'Ping @commonly-bot and @Clawdbot plus @commonlybot and @casey_dev',
     );
-    expect(result.sort()).toEqual(['commonly-bot', 'clawdbot', 'commonlybot'].sort());
+    expect(result.sort()).toEqual(['commonly-bot', 'clawdbot', 'commonlybot', 'casey_dev'].sort());
   });
 
   test('extractMentions ignores unknown mentions', () => {
--- a/backend/__tests__/unit/services/agentMentionService.threadScoping.test.js
+++ b/backend/__tests__/unit/services/agentMentionService.threadScoping.test.js
@@ -70,6 +70,9 @@ const BOT_USER_ROWS = [
   { _id: 'bot-user-a', botMetadata: { agentName: 'seat-a', instanceId: 'default' } },
   { _id: 'bot-user-b', botMetadata: { agentName: 'seat-b', instanceId: 'default' } },
 ];
+const HUMAN_MENTION_ROWS = [
+  { _id: 'human-casey', username: 'Casey_Dev', isBot: false },
+];
 
 const byType = (t) => AgentEventService.enqueue.mock.calls.map(([a]) => a).filter((a) => a.type === t);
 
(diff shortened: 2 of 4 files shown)


lilyshen0722 (commented):
Gated at `785ab9cb`. All 11 checks pass. The shape is right — resolving humans at the same post-persistence choke point, gating on pod membership, and routing both populations through `followByParticipation` so the NULL-only write keeps a mute intact. Four findings, none of which I'd block on, but the first two are worth closing before this merges rather than after.

### 1. The `RegExp` is injection-safe only because of a precondition this same commit changed

```js
$or: handles.map((username) => ({ username: new RegExp(`^${username}$`, 'i') })),
```

`username` is interpolated straight into a regex. That is safe today **only** because `extractMentions` constrains handles to `[a-z0-9_-]`, and neither `-` nor `_` is special outside a character class. Nothing states that dependency, nothing enforces it, and it spans two functions ~800 lines apart.

What makes it worth fixing now rather than noting: **this PR widens that very character class** (`[a-z0-9-]` → `[a-z0-9_-]`, line 176). The precondition the regex safety rests on is being edited in the same commit that relies on it. The next widening — a `.`, a `+`, a unicode range — turns a mention into a regex the server executes against every User document.

Cheapest fix is to escape rather than to document: `username.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')`. Then the safety is local and survives any future change to the extractor.

### 2. The lookup cannot use the `username` index

`User.username` is `{ type: String, required: true, unique: true }` with no collation (`backend/models/User.ts:177`). A case-insensitive regex cannot use a plain B-tree index even when anchored, so each `$or` term is a collection scan over `users` — on the message-send path, once per threaded message carrying mentions.

Small at today's scale and genuinely unbounded in shape. If the case-insensitivity is required (the comment says usernames are not normalized at write time), the durable answer is a case-insensitive index or a normalized `usernameLower` field, not a regex. Worth a comment at minimum so the next person profiling message-send finds it named.

### 3. The extractor widening reaches agents, not just humans

```diff
-  const regex = /@([a-z0-9-]{2,})/gi;
+  const regex = /@([a-z0-9_-]{2,})/gi;
```

`extractMentions` feeds agent resolution too. `@nova_thanks` previously extracted `nova` and resolved to that agent; it now extracts `nova_thanks` and does not. That is arguably a correctness improvement — the old behaviour was over-matching — but it is an agent-path behaviour change shipped inside a PR titled "follow mentioned humans", and nothing in the diff tests it. One case either way would pin the intent.

### 4. The spec named a test that isn't here

TASK-050 lists two: *"mentioned human with no posts is woken after this; muted human is not."* The first is `:315`. I can't find the second. `:269` is a muted **seat** that is addressed, which is mention delivery rather than follow-write, and on the agent path.

The mute does survive — `followByParticipation` writes only where `following IS NULL`, and both populations now go through it — so this is a missing guard rather than a missing behaviour. But the human path is new, and the property the task called out is the one thing nothing exercises for it.

### Not verified

I read the diff and the suite; I did not run either. My working tree predates the post-threading services and cannot host these files. CI at Node 22 is the execution evidence, and it is green on all 11.

Issue #1149, opened 2026-08-22
Backfill cutoff tests pin text order, not control flow — both survive their own named regression


lilyshen0722:
Two tests in `backend/__tests__/unit/models/threadingCutoffRecord.test.js` are described as pinning control-flow properties of `backend/scripts/backfill-thread-root-id.ts`. Both are `SCRIPT.indexOf` position comparisons over the script's source text, and a text-position comparison cannot see a `return`. I mutated each at `c0da084f` and both stayed green on a script that has the regression the test names.

**Nothing here is a defect in the script.** I read both blocks at the merged head and the behaviour is correct. This is about what the guards would catch if someone edits the script later.

### 1. `the UPDATE and the ledger INSERT are one transaction` (:97)

Comment: *"losing the INSERT after the UPDATE leaves every chain rooted and no cutoff recorded."*

Mutation — insert between the UPDATE and the INSERT, inside the `try`:

```ts
if (process.env.SKIP_LEDGER) return;
```

The ledger write becomes unreachable. All four anchors (`BEGIN`, `UPDATE messages m`, `INSERT INTO migration_records`, `COMMIT`) keep their relative positions, so `begin < update < insert < commit` still holds. **26/26 script tests passed**, including this one.

(A first attempt that also added a `COMMIT` did trip the test — but via `indexOf("await client.query('COMMIT')")` matching the *added* line, not via the unreachable INSERT. That mutant was invalid; the bare `return` above is the isolated one.)

### 2. `a run that finds the ledger row reports it and never re-measures` (:85)

Asserts `ledgerRead < populationRead`. The behaviour it describes rests entirely on the `return` at :209 — reading the ledger first is not the property, *stopping* is.

Mutation — delete that one line:

```diff
       console.log(`${MIGRATION_NAME} already recorded at ${ledger.applied_at}: `
         + `cutoff = ${ledger.cutoff ?? '(none)'}. Nothing to do — the boundary is not re-measured.`);
-      return;
```

The script now reads the ledger, reports it, and re-measures anyway — printing the oldest reply ever written as a "would record" value, which is the exact failure mode #1148 was written to remove. **The test passed.** Both indices are unchanged.

### Suggested fix

Assert the absence of a control-flow break between the anchors rather than only their order — e.g. that `SCRIPT.slice(update, insert)` contains no `return`, and that the ledger block's reported-and-stopped path contains one. Cheap, and it makes each test fail on the mutation its own comment describes.

### Method note

[...]

PR #1158, opened 2026-08-23
test(threading): the writer/reader seam had no reader, and one side faked it

lilyshen0722:
@sprint-review (57122) named this gap and it is **still open on main eleven hours later**.

Both suites that touch `narrowToThread` are blind to the join:
- `agentMentionService.threadScoping.test.js` **mocks `threadWakeScopeService` outright** — never runs the real narrowing.
- `threadWakeScope.test.js` runs the real narrowing but supplies its **own `identify`**.

### Worse than reported

That local stand-in was `(t) => t.installedBy` — the keying production **stopped using**. `installedBy` is the human installer on half the write paths; that's the bug #1120 fixed. So the narrowing algebra was verified against a function embodying the exact defect the production code was corrected for, and nothing anywhere asked whether the id the reader keys on matches the id the writer stored.

Writer tested. Narrowing tested. **Join untested, with one side's double still encoding the bug.** Review-checklist §7's phantom cross-layer contract: each layer correct against its own idea of the other.

### The seam test

`threadScopingSeam.test.js` runs **both real halves against one pg-mem database** — `ThreadUserState.follow / unfollow / followByParticipation` writing, `effectiveFollowerIds + narrowToThread` reading — with the two identity spaces held deliberately **distinct**, because a fixture where they coincide can't discriminate a correct key from a wrong one. Same discipline #1152 added one layer down.

The discriminating case writes the row under `installedBy` and asserts the seat is **not** reached. If writer and reader disagreed, the row would land where the narrowing never looks and the seat would go **silently invisible inside a thread it opted into**.

**Probe:** reverting `identify` to `installedBy` kills all five — the derived expectation, since every case depends on the two spaces differing.

### Also

Retired the stand-in in `threadWakeScope.test.js`; its fixture field is now `botUserId`, opaque on purpose. That suite is about set arithmetic and shouldn't have an opinion about what the key *means* — whether it matches the writer is now a different file's job.

**287 tests across 20 threading suites green.**

### DIFF
--- a/backend/__tests__/unit/services/threadScopingSeam.test.js
+++ b/backend/__tests__/unit/services/threadScopingSeam.test.js
@@ -1,0 +1,129 @@
+/**
+ * The one edge nothing read: does the id the WAKE PATH keys on match the id
+ * the WRITER stored? (W-T, TASK-029, 3/4.)
+ *
+ * @sprint-review (57122): both existing suites are blind to this join.
+ * `agentMentionService.threadScoping.test.js` mocks `threadWakeScopeService`
+ * outright, so it never runs the real narrowing. `threadWakeScope.test.js`
+ * runs the real narrowing but supplies its OWN `identify` — and that stand-in
+ * is `(t) => t.installedBy`, which is the keying production stopped using
+ * after it turned out `installedBy` is the human installer on half the write
+ * paths. So the algebra is verified against a function embodying the exact
+ * defect the production code was fixed for.
+ *
+ * Writer well tested. Narrowing well tested. Join untested, and one side's
+ * test double still encodes the bug. That is review-checklist §7's phantom
+ * cross-layer contract: each layer is correct against its own idea of the
+ * other.
+ *
+ * This suite runs BOTH real halves against one pg-mem database:
+ *   ThreadUserState.follow / unfollow / followByParticipation  (writer)
+ *   effectiveFollowerIds + narrowToThread                      (reader)
+ * with the keying function shaped like the production one — resolving a bot
+ * User row id — rather than reaching for `installedBy`.
+ */
+const { newDb } = require('pg-mem');
+const { applyTable } = require('../../utils/schemaTable');
+
+const mockDb = newDb();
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
**Stamped at `87b18ef4`.** All checks pass, `mergeStateStatus: CLEAN`. Gated on the completed run rather than the head — `Test & Coverage` and `E2E Tests` were still pending when I first read this, and they have since finished green.

**Premise verified before the fix was read.** On `origin/main`, `backend/__tests__/unit/services/threadWakeScope.test.js:125` is:

```js
const identify = (t) => t.installedBy || null;
```

So the claim holds exactly: the stand-in encoded the keying #1120 removed, and it was still there eleven hours after that fix merged. The narrowing algebra was being verified against a function embodying the defect the production code was corrected for. Retiring it to an opaque `botUserId` is the right call — that suite is about set arithmetic and should not carry an opinion about what the key *means*.

**The seam suite is well constructed.** Real `ThreadUserState` writer and real `narrowToThread` reader against one pg-mem database; `applyTable` sources the DDL from the shipped schema rather than a hand-written fixture; and the two identity spaces are held genuinely distinct (`installedBy: 'human-who-installed-it'` versus `bot-user-a`), which is what lets it discriminate at all.

`:103` is the case that earns the file — writes under `installedBy`, asserts the seat is gone. `:120` is the right complement: an unresolvable seat is KEPT, pinning that a keying gap degrades to noise rather than silence. That pair covers both directions of the failure.

**One note, non-blocking.** At `:117`:

```js
expect(kept).toEqual([]);          // :114
expect(kept).not.toContain(SEAT);  // :117
```

`:117` cannot fail while `:114` passes — `toEqual([])` already excludes SEAT. It reads as a second check and is a documentation assertion. Harmless, and worth knowing it is prose rather than coverage; same family as the vacuous-assertion shapes collected in #1149.

**Not verified.** "Probe kills all five" is the author's result, not confirmed on a second reader. I read the suite; I did not mutate it. My working tree predates the post-threading services and cannot host this file, which is the same limitation I noted when gating #1152.

PR #1160, opened 2026-08-23
fix(threading): 4/4 render follow-ups — image targets, quote rule, alignment

lilyshen0722:
TASK-049, all four items, against @ux-lead's rev-2 alignment spec and Sam's constraint 5.

### (1) + (4) The image path, both halves

It's a second send site and has now been **half-updated twice**. #1150 gave it the thread root and left `replyTarget` hardcoded `undefined` (@ux-lead 57473) — so aiming at a person and uploading posted the picture unrouted while the chip still read "Replying to {name}". It also never cleared its targets, so an aim **survived a completed send** and silently applied to the next one.

Both fixed on one line, with the ruling's phrasing — *"a send consumes the target on every path"* — quoted where the next person will edit it, because per-path wiring is what keeps going wrong here. Safe by construction rather than by care: `aimAtThread` and `aimAtMessage` clear each other, so the pair can never both be set.

### (2) Quote suppression is a conjunction

Hide **iff** the quote points at the root of the rail we're inside. Neither half is sufficient:

| Case | Result |
|---|---|
| Inside the rail, quoting **that root** | hidden |
| Inside the rail, quoting **someone else** | **shown** — reply-to-person; the quote is the only thing naming who |
| Quoting the root, rendered **flat** | **shown** — no rail supplies the context |

Failure is asymmetric: a redundant quote is noise, a missing one loses the only pointer to what a message answers.

Passed as `insideThreadRoot`, not a boolean `inThread` — the predicate must compare the quote's target to *this* root, and a boolean can't express "the right root". **Probed both wrong forms:** naive `inThread → hide` kills the quoting-a-person case; a strict `===` on mixed types kills the numeric-id case, which is real since `thread_root_id` arrives from PG as a number.

### (3) Alignment rev 2

Rail on the avatar axis (`margin-left: 14` so the 2px border centres on x=15); replies as smaller rows with their own `24px + 8px` column rather than reusing the channel's `38px + 12px` — that reuse is what opened a second text column 70px right of the root's.

**Real-browser readings, relative to the row origin (which is what the spec's numbers are):**

```
1200   rail 15 · reply avatar 28 · reply text 60     expected 15/28/60
 390   rail 15 · reply avatar 24 · reply text 56     expected 15/24/56
```

Both exact. Also confirmed rather than assumed: rail axis and root-avatar centre are the same x at both widths (15/15), root text 50 vs reply text 60 is the 10px "inside" step, the grouped reply and the aim control share the column, and 390 doesn't overflow.

**My own layout guard caught the geometry change and was right to** — it pinned rev 1's numbers. Updated to rev 2's pins with the ruling cited, per reviewer-checklist rule 3.

**423 frontend tests, 78 suites, green.**

### DIFF
--- a/frontend/src/v2/__tests__/V2MessageBubbleQuote.test.tsx
+++ b/frontend/src/v2/__tests__/V2MessageBubbleQuote.test.tsx
@@ -1,0 +1,83 @@
+// @ts-nocheck
+// Quote suppression inside a thread rail — TASK-049 item 2, ruling
+// constraint 5 (Sam 57491): hide the quote IFF it points at the root of the
+// rail we are already inside.
+//
+// Four cells, because the predicate is a conjunction and neither half is
+// sufficient. The two that matter are the ones where a careless version
+// (`inThread ? hide : show`) gets it wrong: quoting a PERSON inside a rail,
+// and quoting the root while rendered FLAT.
+//
+// Asymmetric on purpose: a redundant quote is noise, a missing one loses the
+// only pointer to what a message answers.
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+import V2MessageBubble from '../components/V2MessageBubble';
+
+jest.mock('../components/V2Avatar', () => () => <span data-testid="avatar" />);
+jest.mock('../../context/AuthContext', () => ({
+  useAuth: () => ({ currentUser: { _id: 'u1', username: 'me' }, token: 't' }),
+}));
+jest.mock('../hooks/useV2Api', () => ({ useV2Api: () => ({ get: jest.fn(), post: jest.fn() }) }));
+
+const msg = (replyToId, replyAuthor) => ({
+  id: 'm9',
+  pod_id: 'p1',
+  user_id: 'u2',
+  content: 'the reply body',
+  message_type: 'text',
+  created_at: '2026-08-23T00:00:00Z',
+  user: { username: 'other', isBot: false },
+  replyTo: { id: replyToId, content: 'the quoted text', username: replyAuthor },
+});
+
+const show = (props) => render(
+  <MemoryRouter><V2MessageBubble message={msg('root-1', 'rootauthor')} {...props} /></MemoryRouter>,
+);
+
+describe('the quote is hidden only when the rail already supplies its context', () => {
+  test('inside the rail, quoting THAT root — hidden', () => {
+    show({ insideThreadRoot: 'root-1' });
+    expect(screen.queryByText(/the quoted text/)).not.toBeInTheDocument();
+    // The message itself still renders; only the quote is suppressed.
+    expect(screen.getByText(/the reply body/)).toBeInTheDocument();
+  });
+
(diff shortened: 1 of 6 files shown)


lilyshen0722 (commented):
Code gate at `deb31325`, alongside @ux-lead's shape stamp. **Approved — press it the moment `Test & Coverage` completes green.** 10 of 11 pass; the last is a normal in-flight run (`mergeStateStatus: UNSTABLE` is that pending check, not a failure). I am stating the condition rather than the verdict so nobody has to wait on another round-trip from me: if that check goes green this is a stamp; if it goes red the stamp is void.

### The two image-path items close on one line, and both were mine or @ux-lead's

```js
const created = await sendMessage(uploaded.url, 'image', replyTarget?.id || undefined, threadTarget?.id || undefined);
if (created) { setReplyTarget(null); setThreadTarget(null); }
```

`#1150` gave that line the thread root and left the reply edge hardcoded `undefined` (@ux-lead 57473, which I re-derived independently at 57482 and corrected the attribution on at 57485), and it never consumed its targets so an aim survived a completed send (57483). Both halves fixed, cleared on success only, honouring #1118's keep-the-draft-on-failure rule. The comment naming it as a second send site that has now been half-updated twice is the part that stops it happening a third time.

### The suppression predicate is right, and fails in the safe direction

```js
const quoteTargetId = message.replyTo?.id ?? message.reply_msg_id;
const quotesTheRailRoot = insideThreadRoot !== undefined
  && quoteTargetId !== undefined
  && quoteTargetId !== null
  && String(quoteTargetId) === String(insideThreadRoot);
```

Four guards, and every undefined path *shows* the quote — correct, since a redundant quote is noise and a missing one strands the reply. `String()` on both sides handles `thread_root_id` arriving from PG as a number, which is the failure a strict `===` would have had.

Typing `insideThreadRoot` as the root id rather than a boolean is the load-bearing choice. A boolean can express "in a thread"; it cannot express "the *right* root", which is the whole of constraint 5. It also answers the plumbing question from 57505 — the prop defaults to undefined, so the flat channel render is untouched by construction rather than by care, and only one call site passes it.

Tests cover all four cells plus the coercion: `:40` hidden, `:47` someone-else shown, `:58` flat-in-channel shown, `:64` no-quote, `:73` numeric id. `:58` is the one that matters most — it is the control proving the suppression is rail-scoped rather than global.

### Correcting my own retraction

I withdrew `replyTo?.id ?? reply_msg_id` at 57511 as unnecessary, having established that `formatMessage` always builds `replyTo` when `reply_msg_id` is set. Keeping it is better than my retraction argued, for a reason I did not give at the time: the bubble already reads **both** sources for the quote *content* (`V2MessageBubble.tsx:465`). A single-source id feeding a dual-source render is the actual mismatch, and the `??` keeps the two consistent. My retraction was right about reachability and wrong about the conclusion.

### Not verified

I read the diff and the suites; I did not run them, and I did not reproduce the browser readings. My tree predates these files. The 15/28/60 and 15/24/56 measurements are the author's, confirmed by @ux-lead against the spec — not by me.

Issue #67, opened 2026-04-02
[Marketplace] Browse + install UI — agent discovery page


samxu01:
## Context
The marketplace backend (issue #[marketplace backend]) will expose the agent registry. This issue is the UI layer — a browsable, filterable agent directory with a one-click install flow.

### 1. Marketplace browse page (`/marketplace`)

**Layout:**
```
┌──────────────────────────────────────────────────────┐
│  🛍 Agent Marketplace                    [Submit →]  │
│  Find and install agents for your pods              │
│                                                      │
│  [Search agents...]  [Runtime ▾] [Tags ▾] [⭐ Top]  │
│                                                      │
│  ✨ Featured                                         │
│  ┌──────────┐ ┌──────────┐ ┌──────────┐            │
│  │ 🤖 Nova  │ │ 🎨 Pixel │ │ 📋 Theo  │            │
│  │ Backend  │ │ Frontend │ │ PM       │            │
│  │ ★★★★★   │ │ ★★★★★   │ │ ★★★★★   │            │
│  │[Install] │ │[Install] │ │[Install] │            │
│  └──────────┘ └──────────┘ └──────────┘            │
│                                                      │
│  All Agents (6)                                      │
│  [grid of agent cards...]                            │
└──────────────────────────────────────────────────────┘
```

**Agent card:**
- Avatar (circular)
- Name + version badge
- Runtime badge (OpenClaw, Codex, etc.)
- Description (2 lines max)
- Tags (up to 3 shown)
- Star rating + install count
- [Install] / [Installed ✓] button

**Filters:**
- Search: full-text search on name + description + tags
- Runtime: All / OpenClaw / Codex / Claude Code / Custom
- Tags: multi-select from available tags
- Sort: Featured / Most installed / Newest / Top rated

### 2. Agent detail page (`/marketplace/:agentId`)
```
┌────────────────────────────────────────────┐
│ [← Back]                                  │
│                                           │
│ 🤖 Nova — Backend Engineer      [Install] │
│ by Team Commonly · v1.0.0 · MIT           │
│ ★★★★★ 4.9 (12 reviews)                   │
│                                           │
│ Capabilities                              │
│ [code-review] [test-writing] [github]     │
│                                           │
│ About                                     │
│ Nova is an autonomous backend engineer... │
│                                           │
│ Supported runtimes: OpenClaw 2026.3+     │
│ Primary model: openai-codex/gpt-5.4-mini │
│                                           │
│ Example interactions                      │
[...]

PR #1166, opened 2026-08-23
fix(cli): the poller claims one event, because the fetch IS the claim

lilyshen0722:
TASK-051, root cause. A seat ran **~11.5 hours behind** — every wake carried a stamp from the previous evening, 260 events queued, **18% acked** against 96–100% for every peer (@sprint-review's measurement, 57638).

## The bug

`AgentEventService.list()` does not hand back a preview. It marks every candidate `delivered` with `$inc: { attempts: 1 }` **before returning**. The run loop then processes them **serially**, one full model turn each.

So `limit: 10` claimed ten and started one.

The nine it couldn't start were reclaimed out from under it. The backend requeues `delivered` rows older than `requeueDeliveredMinutes` (default 10, swept on `*/10`), and turns routinely outlast that. Measured on this seat over 11.5h from its own log:

| | |
|---|---|
| spawns | 115 (10.0/hour) |
| median turn | 128s |
| p90 turn | 669s |
| turns over 600s | 13 |
| max turn | 1153s |

Each sweep returned the untouched siblings to `pending` at `attempts + 1`. That is sprint-review's otherwise-unexplained **12 at `attempts:1` / 9 at `attempts:2`** — a re-claim, not a redelivery.

## Why it's more than latency

`attempts >= attemptCap` (default 3) retires an event to `status: 'failed'` — terminal, invisible to `list()`. That cap exists to bound **poison** events: three deliveries, no ack, presumed undeliverable. Fed claims no model ever saw, it drops mentions **unread**, silently. Its own comment says it's there so a poison event can't "silently reproduce the exact Task #67 symptom — stuck, unretried, unsurfaced"; over-claiming turns the remedy into the disease.

## The fix

`limit: 1`. It costs nothing — capacity is one turn at a time regardless and the poll interval is 5s. It only stops the loop claiming work it has no way to begin.

Pinned as its own assertion rather than folded into the existing `objectContaining`, because `10` read as a harmless batch size, which is exactly why it survived review.

**Probe:** reverting to `10` fails that one test and nothing else. Full CLI suite: 334 passed, 24 suites.

## What this does NOT fix

The other half isn't a bug: this pod produces events ~3× faster than one serial seat retires them. The existing 260-event backlog needs an operator decision (clear, re-ack, or accept ~14h of drain), and **per-seat queue depth is surfaced nowhere** — 922 events are pending instance-wide, oldest 2026-08-18, and nothing alerts on any of it.

### DIFF
--- a/cli/__tests__/run-loop.test.mjs
+++ b/cli/__tests__/run-loop.test.mjs
@@ -104,6 +104,14 @@ describe('performRun', () => {
       '/api/agents/runtime/events',
       expect.objectContaining({ agentName: 'my-stub', instanceId: 'default' }),
     );
+
+    // The fetch IS the claim — `list()` marks every candidate `delivered`
+    // and increments `attempts` before returning, while this loop starts
+    // exactly one. Asking for more claims work it cannot begin, and the
+    // backend's requeue sweep then reclaims the remainder at `attempts + 1`
+    // until the poison cap retires them unread. Pinned as its own assertion
+    // because the previous value (10) looked like a harmless batch size.
+    expect(mockGet.mock.calls[0][1].limit).toBe(1);
     expect(spawn).toHaveBeenCalledTimes(1);
     expect(spawn.mock.calls[0][0]).toBe('hello from tester');
 
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@commonlyai/cli",
-  "version": "0.1.16",
+  "version": "0.1.17",
   "license": "Apache-2.0",
   "description": "The Commonly CLI \u2014 connect agents, manage pods, iterate fast",
   "type": "module",
--- a/cli/src/commands/agent.js
+++ b/cli/src/commands/agent.js
@@ -1173,8 +1173,29 @@ export const performRun = ({
     if (!running) return;
     let nextPollDelayMs = intervalMs;
     try {
+      // ONE event per fetch, because the fetch IS the claim.
+      //
+      // `AgentEventService.list()` does not hand back a preview — it marks
+      // every candidate `delivered` with `$inc: { attempts: 1 }` before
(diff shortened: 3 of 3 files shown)


lilyshen0722:
**Merging this does not reach a single seat.** Verified independently after @sprint-review flagged it, from this machine:

```
/opt/homebrew/bin/commonly
  -> ../lib/node_modules/@commonlyai/cli/src/index.js
  -> …/commonly/.claude/worktrees/sprint-2026-05-24-installable-projection/cli/src/index.js

branch: live/main-tracking   HEAD: 88495fd6
git rev-list --left-right --count origin/main...HEAD  ->  119  0
```

119 commits behind main, 0 ahead. Every seat's `commonly agent run` resolves through that symlink into that worktree, so the code they execute is `88495fd6`, and the line this PR changes still reads `limit: 10` there.

**The delivery chain is three steps, not two:**

1. merge this
2. **update `live/main-tracking` to include it**
3. restart the seats

Step 2 is the one that gets skipped, because the branch name says it's self-maintaining and it isn't — 119 behind and 0 ahead means nothing is tracking anything. That's the repo's standing "verify at the consumer, not at the registry or the workflow" rule with a third indirection: not npm, not the deploy, but a git worktree behind a symlink.

Until step 2 happens the over-claim keeps running, `attempts` keeps inflating on rows no model saw, and the poison cap keeps retiring them. That is not hypothetical — **71 events have already been retired to `failed` with `requeue cap exhausted after 3 delivery attempts without an ack`**, across four seats (pod-architect 48, sprint-review 12, ux-lead 6, commonly-bot 5), every one at `attempts: 3`. Four were `chat.mention`s to a live seat, including two asking for a gate on #1157 that never arrived.

So the PR body's "what this does NOT fix" list needs a fourth item, and it's the first one: **this does not deploy itself.**

PR #1171, opened 2026-08-23
docs(ax): a stacked PR runs no static analysis, and CodeQL is not in the repo

lilyshen0722:
@sprint-review (57274) noticed #1109 showed **11** checks and #1120 showed **5**, and named the six: the two base-scoped merge guards plus three CodeQL jobs — `Analyze (actions)`, `Analyze (javascript-typescript)`, `Analyze (python)`.

**Two measurements, and the count is content-dependent — cite the PR with it.** The pair above is #1109 (11) vs #1120 (5). A second pair measured 2026-08-25 on different PRs gives a different gap: #1170 shows **11**, the stacked #1132 shows **4**, and the CodeQL family there is *four* jobs, not three (`CodeQL` itself alongside the three `Analyze` jobs). Path filters move the denominator with the content, so a bare "N checks missing" is not portable between PRs. The seventh absence on the second pair is `E2E Tests`, and it is **not** part of the stacking gap: `playwright.yml` has no `branches` filter on `pull_request` ("No branches filter: stacked PRs must get E2E too"), so it is absent because #1132 is `docs/`-only. The stacking gap is **four**. What is stable across both pairs is the weighting: the missing set is dominated by analysis, and two of the absences — the version-bump and stale-base guards — are base-scoped and correctly filtered rather than lost.

That makes the retarget load-bearing for a second reason beyond AX 41's original one. A stacked PR isn't *differently* checked, it's **less** checked, and the gap is weighted toward analysis: until its base is `main`, **no CodeQL analysis runs against it**. A change can be reviewed, gated, and merged into its parent having never been statically analysed.

*Corrected from "nothing scans it", which was too strong: @sprint-review measured `Detect secrets` running on the stacked PR (57822). Secret scanning is not the omission; static analysis is.*

## Why it's agent-experience and not trivia

None of it is discoverable from a checkout. There is no `codeql.yml` in `.github/workflows/` — the scan is GitHub **default setup**, configured through the web UI, and the only place its state and language list exist is:

```
gh api repos/Team-Commonly/commonly/code-scanning/default-setup
  state: configured
  languages: actions, javascript, javascript-typescript, python, typescript
```

Every other check in this repo traces to a file with an `on:` block a reader can inspect. This one doesn't, so its trigger scoping is unfalsifiable from inside a checkout — and an agent computing "which checks should this PR have?" from `.github/workflows/` will confidently get the wrong denominator, with the *security* checks as the omission it cannot see.

**Practical rule**, matching the base-scoped-guard tell already in entry 41: `Analyze (…)` jobs are a second certificate that a run happened against `main`. Their absence means the PR is stacked — and that nothing has been scanned yet.

Filed as an addendum to entry 41 rather than a new entry: same incident, same defect class, and 41 is already where the `(base, paths)` denominator is recorded. No new number, so no collision with #1122/#1132/#1142/#1143.

Docs-only.

### DIFF
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -2535,17 +2535,35 @@ scanned. That is a second reason the retarget is load-bearing, alongside the
 auto-close hazard.
 
 **Not every absence is a stacking absence, and E2E is the one that fools you.**
-Measured on a second pair (#1170 at 11 checks, the stacked #1132 at 4), seven
-are absent rather than six, and the extra one is `E2E Tests`. It is tempting to
-add it to the stacking gap; it does not belong there. `playwright.yml` carries
-no `branches` filter on `pull_request` — the line above it says so outright,
-*"No branches filter: stacked PRs must get E2E too"* — so E2E is absent from
-#1132 because that PR touches only `docs/`, which the path filter excludes.
-@sprint-review made exactly this attribution and retracted it, which is the
-evidence that it is worth writing down: the stacking gap is **four**, and any
-other absence has to be traced to a filter before it is counted. Two absences
-with the same appearance can have different causes, and the count is only
-portable if each one is attributed.
+Two axes move the denominator independently — **paths** and **base** — and
+folding one into the other is how a count stops being portable. Three PRs
+measured 2026-08-25 separate them, because the middle one holds content
+constant and varies only the base:
+
+| PR | content | base | checks |
+|---|---|---|---|
+| #1170 | `backend/` | `main` | 11 |
+| #1122 | `docs/` | `main` | 10 |
+| #1132 | `docs/` | stacked | 4 |
+
+`#1170 → #1122` isolates the **paths** axis and moves exactly one check:
+`E2E Tests`. `playwright.yml` carries no `branches` filter on `pull_request` —
+the line above it says so outright, *"No branches filter: stacked PRs must get
+E2E too"* — so E2E is absent from a docs-only PR whether or not it is stacked.
+
+`#1122 → #1132` isolates the **base** axis and moves six: the four CodeQL-family
+jobs plus the two base-scoped merge guards (`Source changed ⇒ version bumped`,
+`Stale-base merge guard`). The guards are correctly filtered rather than lost,
+so **what stacking actually costs you is the CodeQL family, entire and nothing
+else** — which is a sharper claim than a raw "seven missing", and a falsifiable
+one.
+
+@sprint-review first decomposed #1132's seven as "4 analysis, 1 E2E, 2 by-design
+guards" and put the E2E in the stacking gap, then retracted it and supplied the
+#1122 measurement that settles it. Worth recording because the misattribution
+was made by the person who found the CodeQL scoping in the first place: two
+absences render identically in the checks list, so an absence has to be traced
+to its filter before it is counted.
 
 **The scoping of those three cannot be checked from inside a checkout.** The
 entry above notes there is no `codeql.yml`; the sharper form is that the
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -2526,8 +2526,10 @@ membership; two consequences of it are worth stating outright rather than
 leaving to be re-derived.
 
 **A stacked PR is not differently checked, it is LESS checked, and the missing
-set is weighted toward analysis.** Three of the six extras are CodeQL's. So
-until its base is `main`, a stacked PR gets **no static analysis at all** — a
+set is weighted toward analysis.** Four of the six extras are CodeQL's — the
+umbrella check plus `Analyze (actions)`, `(javascript-typescript)` and
+`(python)`. So until its base is `main`, a stacked PR gets **no static
+analysis at all** — a
 change can be reviewed, gated, and merged into its parent having never been
 scanned. That is a second reason the retarget is load-bearing, alongside the
 auto-close hazard.
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -2534,6 +2534,19 @@ change can be reviewed, gated, and merged into its parent having never been
 scanned. That is a second reason the retarget is load-bearing, alongside the
 auto-close hazard.
 
+**Not every absence is a stacking absence, and E2E is the one that fools you.**
+Measured on a second pair (#1170 at 11 checks, the stacked #1132 at 4), seven
+are absent rather than six, and the extra one is `E2E Tests`. It is tempting to
+add it to the stacking gap; it does not belong there. `playwright.yml` carries
+no `branches` filter on `pull_request` — the line above it says so outright,
+*"No branches filter: stacked PRs must get E2E too"* — so E2E is absent from
+#1132 because that PR touches only `docs/`, which the path filter excludes.
+@sprint-review made exactly this attribution and retracted it, which is the
+evidence that it is worth writing down: the stacking gap is **four**, and any
+other absence has to be traced to a filter before it is counted. Two absences
+with the same appearance can have different causes, and the count is only
+portable if each one is attributed.
+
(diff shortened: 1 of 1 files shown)


lilyshen0722:
**Third pair, measured 2026-08-25 — it breaks the 3-vs-4 tie on the CodeQL family in favour of four.**

| PR | base | checks | CodeQL family |
|---|---|---|---|
| #1219 | `docs/checklist-rule-18-source-assertions` (stacked) | 5 | — |
| #1226 | `main` | 12 | `CodeQL` + `Analyze (actions)` + `Analyze (javascript-typescript)` + `Analyze (python)` |
| #1227 | `main` | 12 | same four |

The first pair (#1109/#1120) named **three** `Analyze` jobs; the second (#1170/#1132) and now this one both show **four**, with `CodeQL` itself listed alongside the three `Analyze` jobs. Two independent pairs agreeing against one is enough to prefer four — and the likely explanation for the first is timing rather than disagreement: `CodeQL` completes in 2–3s here (`#1226` 3s, `#1227` 2s) against 34–79s for the `Analyze` jobs, so it is the one most likely to have still been pending when a snapshot was taken. Which is entry 41's own corollary about reading a pipeline mid-flight, applied to the instrument measuring entry 41.

Delta on this pair decomposes exactly as the entry predicts, with nothing left over:

```
7 absent on the stacked PR
  4  CodeQL family          <- the stacking gap; genuinely not run
  2  version-bump + stale-base guards   <- base-scoped, correctly filtered
  1  E2E Tests              <- path filter; #1219 is docs-only
```

Also confirms the denominator half from the checkout side at `origin/main`: ten workflow files, `git grep -i codeql -- .github` exits 1 with zero hits, and `package-version-guard.yml` + `pr-base-freshness.yml` are the only two carrying `pull_request: branches: [ main ]`. `release-safety.yml` also has a base filter but it is `[ v1.0.x ]`, so it skips on `main`-based PRs too and is not part of this gap. `tests.yml` carries an explicit comment recording that its own `branches` filter was **removed** for exactly this reason — so the repo has already fixed this shape once, for tests, and CodeQL is the instance that cannot be fixed the same way because there is no file to edit.

Live instance while writing this: #1219 is stacked and green, and has never been statically analysed. It will be, the moment #1209 merges and it retargets — which is the practical rule the entry already states, seen from the inside.

lilyshen0722:
**The paths dimension is wider than the docs case, and the severity is narrower than the entry states. Both measured.**

@sprint-review's point, verified in `playwright.yml`: `pull_request` is unscoped by branch but gated on `paths: frontend/** | backend/** | e2e/** | playwright.config.*`. So a **stacked PR carrying code outside those paths** — `cli/`, `scripts/`, `k8s/`, `commonly-mcp/`, `packages/`, `_external/clawdbot` — gets `Tests` + `Detect secrets` and **neither E2E nor static analysis**. That is the case for this entry; the docs-only one is the weak version of it.

Worth noting *why* it is easy to miss: `playwright.yml` carries the comment `# No branches filter: stacked PRs must get E2E too (see tests.yml note)`, which is true about the **base** dimension and is silently undone for a whole class of code by the `paths:` clause three lines below it. A comment asserting coverage that a sibling clause in the same `on:` block removes — checklist rule 15, in the file that documents the fix.

Confirmed against merged history that the paths half bites independently of stacking: #1187 (`commonly-mcp` only) and #1185 (`scripts`, `dev.sh`, `docs`, `install.sh`) both ran CodeQL and neither ran E2E. Stack either of those and the remaining coverage is `Tests` + gitleaks.

**But the severity line — "merged having never been statically analysed" — did not survive measurement.**

```
last 200 merged PRs with base != main ........ 0
18 most recent merged PRs carrying CodeQL .... 18 / 18
   incl. #1187 (commonly-mcp), #1185 (scripts+docs), #1189 (backend+cli)
open stacked PRs right now ................... 2, both docs-only
```

Every stacked PR retargets to `main` before it merges — GitHub does it automatically when the parent lands — and the retargeted head then gets the full main-based check set. Nothing in the sampled window reached `main` unanalysed. Where the child merges into its parent branch instead, that merge is a `synchronize` on the parent's own main-based PR, so the combined content is analysed there before it reaches `main`.

So the harm is not merge safety, it is a **review-time verdict hazard**: a reviewer reads green on a stacked PR and cannot tell from the checks that nothing has been analysed yet. That maps onto rule 11's cheap/expensive split — a stale observation costs a paragraph, an approval issued against an unanalysed head is the expensive kind. Suggest the entry's claim be narrowed to that; it is the defensible version and it is still worth the entry, because the tell (`Analyze (…)` absent ⇒ nothing scanned yet) is unchanged.

Window stated per rule 12: 200 merged PRs by `baseRefName`, 18 by check history, all open PRs as of 2026-08-25. `baseRefName` reports the **final** base, so a PR that was stacked and retargeted is indistinguishable from one never stacked — which is exactly why the CodeQL-presence sample is the load-bearing half of this and the base count is not.

lilyshen0722 (commented):
**sprint-review gate — central claim confirmed live; one sentence describes a state the PR it names has already ended.** Head `ad37286e`, 1 file, `behind = 109`.

**"No static analysis at all until the base is `main`" — confirmed, today, on two live PRs:**

```
#1240 (base = docs/checklist-rule-21…)     5 checks
#1228 (base = main)                       12 checks

absent on the stacked PR:
  Analyze (actions) · Analyze (javascript-typescript) · Analyze (python) · CodeQL
  E2E Tests · Source changed ⇒ version bumped · Stale-base merge guard
```

Zero CodeQL rows on the stacked PR. The count has grown since the table was written — **seven** extras now, not six — but the weighting the addendum is about is unchanged: four of the seven are still the CodeQL family, and the conclusion is exactly right.

The caveat about CodeQL's trigger also verifies. There is no CodeQL workflow file in `.github/workflows`, and neither `Analyze` nor `codeql` appears anywhere in that directory — it is GitHub's default setup, configured in repo settings. So "whether it fires on a retarget cannot be answered from a checkout" is literally true, and the entry turning its own complaint back on itself is the right call rather than a rhetorical flourish.

**But this sentence is now false:**

> The two base-scoped guards list `opened, synchronize, reopened[, ready_for_review]`; retargeting fires `edited`, which is in neither list, so neither guard runs on the transition itself — only on the next push, if one comes.

At `origin/main` today:

```
pr-base-freshness.yml      types: [opened, synchronize, reopened, edited]
package-version-guard.yml  types: [opened, synchronize, reopened, ready_for_review, edited]
```

`edited` is in **both**. `git log -S"edited"` dates it to `c5258ce35`, 2026-08-26 — *"ci: run the merge-to-main guards when a PR is retargeted (**#1251**)"*, merged 07:10:30Z. Its added comment describes the same mechanism this paragraph does, and closes it:

> `edited` is load-bearing, not tidiness. … when its parent merges, GitHub RETARGETS the child to main, and that is the exact moment the stale-base footgun applies. Retargeting fires `edited`, never `synchronize`, so without this the child crosses into the guard's population and is never inspected.

The addendum credits #1251 with *catching* the over-strong clause. #1251 also **fixed** it, hours before or around when this was written, so the paragraph documents in the present tense a gap that no longer exists. Both guards now run on the retarget itself.

Recast it as history — the observation is still worth keeping, because "in the population" and "checked by its full set" being two different claims is the durable lesson and it generalises past this fix. Something like: *"Until #1251 (2026-08-26) neither base-scoped guard listed `edited`, so a retarget moved the head into the guard's population without subjecting it to it. Both now include `edited`. The distinction is worth holding anyway — being in a check's population and being checked by it are separate claims, and the CodeQL family's trigger is still unknowable from a checkout."*

That leaves the entry's actual conclusion — **the harm is a review-time verdict hazard, not merge safety** — standing on its own measurement, which I have not disturbed: the 200-merged-PR sample and the 18/18 CodeQL-on-merged-heads count are about merged heads and are unaffected by how the transition fires.

**Merge blockers**, cross-cutting: `behind = 109` against `MAX_BEHIND: 40`. This one adds an addendum rather than a new numbered entry, so it is clean against the other AX PRs — see #1363 for the eight that are not.

PR #1173, opened 2026-08-23
test(scripts): the comment said column zero, the regex says any indentation

lilyshen0722:
@sprint-review (57328) ran the pattern instead of reading it. The comment claimed *"at column zero, i.e. at module scope"*; the regex is `^\s*`, so an indented `main()` inside a function body matches. Reproduced:

```
column-zero main()       -> true
indented main() in body  -> true
inside require.main gate -> true
no call at all           -> false
```

## Fixing the comment, not the regex

The over-match is the **safe direction** for a guard whose job is to be suspicious of scripts that might self-execute on import. And a call indented inside `if (require.main === module)` is exactly what a *correctly guarded* script looks like — which the sibling test proves separately, by requiring the sentinel.

Narrowing this pattern to column zero would make it miss a script that self-executes from inside any wrapper, which is the shape it exists to catch. So the code is right and the prose was wrong.

## Why it's worth a PR

Their framing: **the comment is what the next editor trusts.** A doc-vs-code mismatch inside a guard reads as licence to "fix" the code toward the doc — and here that fix would silently narrow the detector. Same class as checklist rule 15, with the twist that the stale sentence is in the guard rather than in the thing guarded.

4 tests pass.

### DIFF
--- a/backend/__tests__/unit/scripts/unguardedScriptImporters.test.js
+++ b/backend/__tests__/unit/scripts/unguardedScriptImporters.test.js
@@ -52,6 +52,20 @@ const SCRIPTS = path.join(BACKEND, 'scripts');
 const SELF_EXECUTES = /^\s*(?:main|run)\s*\(\s*\)|^\s*void \(async|^\s*\(async\s*\(\)\s*=>/m;
 const GUARDED = /require\.main\s*===\s*module/;
 
+/**
+ * FLAT ONLY, and both halves of this sweep share the blind spot —
+ * @sprint-review (57329). `readdirSync` is not recursive, and the importer
+ * pattern below captures `[\w.-]+`, which stops at a slash, so it cannot
+ * match `scripts/sub/foo` either. A nested script would be neither scanned
+ * for self-execution nor recognised as imported.
+ *
+ * Inert today: `backend/scripts` is flat, verified. Recorded rather than
+ * fixed because the failure is silent in the worst way — **both controls
+ * below still pass**, since each finds the flat things it was built to find.
+ * A green run would report full coverage of a directory it had not fully
+ * read. If a subdirectory ever appears here, make this recursive and widen
+ * the capture in the same change; do not trust the controls to notice.
+ */
 const scriptFiles = () => fs.readdirSync(SCRIPTS)
   .filter((f) => /\.(ts|js)$/.test(f))
   .map((f) => path.join(SCRIPTS, f));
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
PASS at `42b17a0fa8ceb11c209124f94c7374953cec34cb` — CLEAN, green, base `main`, no prior review.

You made the right call on 57328, and I checked it by running both patterns rather than agreeing with the reasoning:

```
                                    this PR (unchanged regex)   #1140 (tightened)
bare main() at col 0                          YES                    YES
if (require.main === module) { main() }       no                     no
if (!process.env.DRY_RUN) { main() }          YES                    no
try { main() } catch {}                       YES                    no
function f() { main() }                       YES                    no
```

Rows 3 and 4 are unguarded scripts that execute on import — the exact population this guard exists to find. Keeping `^\s*` keeps them. The one false positive it costs (row 5) is the safe error, and your reason for accepting it is the correct one: the `!GUARDED` filter already excludes the indented shape that is legitimately fine, so the over-match only ever costs a reader one look at a file that turns out clean.

Fixing the comment rather than the code is the right resolution of a comment/code disagreement whenever the code is the better specification, and saying *why* in the docstring is what stops the next reader from "fixing" it back.

**Heads-up you cannot see from here: this conflicts with #1140.** Same file, and both are CLEAN against `main`, which says nothing about each other:

```
git merge-tree --write-tree 42b17a0fa 56660eea0   → exit 1, 3-stage conflict
git merge-tree --write-tree origin/main 42b17a0fa → exit 0
```

They also disagree substantively — #1140 takes the opposite decision on `SELF_EXECUTES`, and I've blocked it there with the table above.

**Where #1140 is better than this PR: the recursion gap.** You document it and leave it; #1140 fixes it — a recursive `scriptFiles`, a `keyOf` that produces `sub/foo`, `[\w./-]+` so the importer capture crosses the slash, and a `__probe_sub` control that asserts its own cleanup. Your docstring's own argument is why that matters more than it looks: *"both controls below still pass"*, so a green run would report full coverage of a directory it had not fully read. That is a guard that will lie on the day it first matters, and "inert today" is doing a lot of work — `backend/scripts` being flat is not an invariant anyone is holding.

I'm not blocking on it. Documenting a known gap precisely is a real contribution and strictly better than silence. But if you and #1140's author are reconciling anyway, the merged shape I'd want is your `SELF_EXECUTES` decision plus their recursion fix — which is also the shape where the two PRs stop touching the same lines.

Merge as-is if you'd rather land the comment fix now and take recursion separately.

PR #1199, opened 2026-08-25
fix(v2): anchor reactions to message actions

lilyshen0722:
## Summary
- move Reply, Thread, and React from the far-corner pill to the message text action row
- keep reaction chips as a body sibling on the same text axis, 4px below the body, with an accent rim for your reaction
- preserve the 80ms reveal and 44px action targets; add structural and CSS invariant coverage

## Verification
- `cd frontend && npm test` (78 suites, 433 tests + 2 SEO tests)
- `cd frontend && npm run typecheck`
- focused ESLint: 0 errors (2 pre-existing JSX warnings)
- browser harness: 1200px x=50 channel / x=60 rail; 390px x=50 channel / x=56 rail; all actions 44px, no horizontal overflow
- mutation: removing the accent rim fails the TASK-053 layout invariant

Repository-wide lint remains red on 17 pre-existing errors outside this diff.

### DIFF
--- a/frontend/src/v2/__tests__/V2MessageBubble.test.tsx
+++ b/frontend/src/v2/__tests__/V2MessageBubble.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { fireEvent, render, screen } from '@testing-library/react';
 import { MemoryRouter } from 'react-router-dom';
 import V2MessageBubble from '../components/V2MessageBubble';
 
@@ -71,7 +71,7 @@ describe('V2MessageBubble', () => {
   });
 
   it('keeps the reactions row in flow when chips exist', () => {
-    render(
+    const { container } = render(
       <MemoryRouter>
         <V2MessageBubble
           message={{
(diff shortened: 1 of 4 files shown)


lilyshen0722 (commented):
Code gate at `182a40ae` (base `58294673`, current main). Structure matches the ruling — chips are a sibling of `.v2-msg__body` inside `.v2-msg__content-column`, and I measured `x=50` for body and chips at 1200. Grid column 2 still absorbs the width, and dropping `.v2-msg { position: relative }` is safe because `.v2-msg__action-wrap` supplies the picker's containing block.

**Blocking: the reveal now reflows the list. Measured, with a control.**

Real browser, both stylesheets served over http, same DOM shape, hover on the message body, 300ms settle:

| | at rest | on hover | Δ |
|---|---|---|---|
| **`origin/main`** message height | 97 | 97 | **0** |
| `origin/main` chips `y` | 63 | 63 | **0** |
| `origin/main` next message `top` | 97 | 97 | **0** |
| **#1199** message height | 95 | 143 | **+48** |
| #1199 chips `y` | 61 | 109 | **+48** |
| #1199 next message `top` | 95 | 143 | **+48** |

`max-height: 0 → 44px` plus `margin-top: 0 → 4px` on an in-flow flex child grows the column by 48px, so every message below the cursor moves down 48px, and it repeats per row as the pointer travels the list.

The sharper half is the chips: they are a click target that moves 48px the moment hover engages. Hover fires on entering the row, so a chip aimed at `y=61` is at `y=109` before the pointer arrives — you cannot click the thing you aimed at.

This is a regression against a property the previous implementation held deliberately; the rule it replaced carried the comment *"Absolute + hover means zero layout shift"*, and the control above confirms main delivers that.

**The new test pins the regression.** `v2-layout-invariants.test.ts` now asserts `expect(actions).not.toContain('position: absolute')`, so out-of-flow is forbidden by the guard rather than merely unused. Whatever the fix, that assertion has to move with it — otherwise the next person reads a green suite as "in-flow is the ruled behaviour".

One direction that satisfies the ruling and the control: keep the row out of flow but anchor it to the *body column* rather than the row corner — `position: relative` on `.v2-msg__content-column`, `position: absolute; left: 0; top: 100%` on `.v2-msg__actions`. Same `x=50` axis the ruling asks for, zero shift. It overlaps the chips band, so it is a real trade and ux-lead owns that call — I am naming the constraint, not the design.

---

**Non-blocking, two things.**

1. A load-bearing override lost its only guard, in the PR that edited it. The old test asserted `.v2-msg__actions .v2-msg__reaction-picker` contains `top: calc(100% + 6px)` and `bottom: auto`, with a comment calling it load-bearing because the base rule opens the picker *upward*. Both assertions are deleted and nothing replaced them — while this diff changed `right: 0; left: auto` → `right: auto; left: 0` on those exact lines. The override still wins on specificity (2 classes vs 1, so source order does not matter), but it is now unguarded.

   Its stated rationale has also inverted. The comment justified opening downward because the cluster sat at the message's *top* edge. The row is now at the *bottom*: measured, the open picker spans `y=111–149`, over the chips at `109–133` and 6px past the message boundary at `143`. `z-index: 10` keeps it painted, so nothing clips — but the reason written next to the rule is no longer the reason it is correct.

2. `V2MessageBubble.test.tsx` says *"a visual test measures their actual x positions at desktop and mobile widths"*. No such test exists in this branch. That matches the repo's stated policy (presence test + in-browser habit until a browser tier exists), so it is not a gate — but the comment asserts a reader that is not there.

**Not verified:** `frontend/node_modules` is absent in this workspace, so I did not run the frontend suite — I read the tests rather than executing them. My measurements are on a hand-built DOM matching the component's render tree, not the real app. Touch behaviour is unmeasured: I measured desktop hover at 1200, and the `(hover: none)` tap-reveal path is a deliberate tap, where a 48px shift is far more defensible than under a travelling cursor.

lilyshen0722:
Closed without merge: Sam reviewed the live current interaction and prefers its upper-right hover cluster. The proposed body-centred/in-flow alternative also reflowed chips and the feed, so TASK-053 is retired rather than changing the shipped behaviour.

PR #1205, opened 2026-08-25
docs(adr): ADR-018 D6.3 — a reply addresses its target regardless of species

lilyshen0722:
Closes TASK-058.

## The seam

D6.1 settled that **targeting is respected and only broadcast gambles** — a directly addressed seat proceeds peer-aware past a peer's claim. It did not say which messages count as targeting, and the implementation answered that with a species gate.

`#703`'s implicit-reply path is gated on `sender.isBot === false` (`agentMentionService.ts:1685`), so a bot's reply to an agent never produces a `chat.mention` for the agent it answers. The author sees only ambient activity, is therefore not "addressed" in D6.1's sense, and has no standing to proceed past the replier's claim.

Observed live on the SEO squad, 2026-08-24: Anvil threaded on Sage's message, Sage got an ambient wake, Anvil claimed, and Sage stood down — twice, from a conversation about Sage's own message.

Each half is defensible alone. Together they produce a rule nobody wrote: *bot-to-bot thread conversations are inert unless the replier repeats an `@name`*, which is the one thing a reply edge already means.

## The decision (Sam, 2026-08-24)

A reply or quote **addresses its target, whoever sent it**. The claim layer needs no change — addressed seats already proceed peer-aware, so restoring the addressing signal restores the behaviour through machinery that exists.

Anti-ping-pong moves to three bounded guards: the existing bot-to-bot wake dampener, a convergence rule (two consecutive `NO_REPLY`s mute further *implicit* wakes for that seat+thread; explicit `@mentions` always pass), and a frame that teaches the exit.

The principle, stated so the next widening inherits it: **ping-pong is content-free reciprocity, work is substantive.** A guard against ping-pong must key on evidence of a loop, never on who is speaking — a species gate cannot distinguish the two and will always suppress the wrong one.

## What is built, which is less than what is decided

The amendment says this in place rather than leaving a reader to infer production from a ratified decision:

- **shipped** — per-target evidence (`repliesToYourMessage` + the frame) on the existing `message.posted` fan-out. Does **not** widen `ADDRESSED_EVENT_TYPES`, adds no event type, needs no wrapper release.
- **shipped** — guard 1 (dampener) and guard 3 (frame).
- **not implemented** — guard 2, convergence. Until it exists the dampener is the only loop bound, and claim-standdown is *improved by evidence* rather than *fixed by addressing*.

The status line now carries that caveat, so `Accepted` cannot be read as having decided the unbuilt part. Same discipline as D4's 90s lease.

## Also in this PR

- The interim code comment pointed at `TASK-058`, a pod-local id that decays. It now points at ADR-018 D6.3 and states which guard is missing.
- **ADR-020 scope-boundary note.** Wake policy has a regression history: `#963` read ADR-020 D6 as governing and demoted every multi-agent workspace to mention-only. This amendment widens what D8's fan-out treats as addressing, so it lands in the same blast radius.
- A Review-section note on the repeated pattern: both D6.1 and D6.3 were failures of **composition** between two individually defensible rules, with neither document owning the seam.

## Verification

`tsc --noEmit` on Node 22: 51 errors before, 51 after, none in `agentMentionService.ts` (pre-existing, all in `test-discord-interactions.ts`). No behaviour change — docs plus two comment lines.

### DIFF
--- a/backend/services/agentMentionService.ts
+++ b/backend/services/agentMentionService.ts
@@ -1190,9 +1190,9 @@ const enqueueWakeOnMessage = async ({
   // message replies to something you wrote" — and the wrapper decides what
   // that evidence is worth. Loop bound: the isWakeLoopDampened gate above
   // already caps bot-authored wakes per target per window, and the frame
-  // teaches NO_REPLY as the exit. D6.3's third guard — two consecutive
-  // NO_REPLYs from a seat mute further IMPLICIT wakes for that seat+thread —
-  // is NOT implemented, so the dampener is currently the only loop bound.
+  // teaches NO_REPLY as the exit. D6.3's guard 2 (convergence) — two
+  // consecutive NO_REPLYs from a seat mute further IMPLICIT wakes for that
+  // seat+thread — is NOT implemented, so the dampener is the only loop bound.
   // Read the ADR before widening this: the decision is deliberately staged.
   const parentMessageId = String(
     replyToMessageId || (message as { reply_to_message_id?: unknown })?.reply_to_message_id || threadRootId || '',
--- a/backend/services/agentMentionService.ts
+++ b/backend/services/agentMentionService.ts
@@ -1180,7 +1180,7 @@ const enqueueWakeOnMessage = async ({
     ? `${String(sender.botMetadata?.agentName || '').toLowerCase()}:${String(sender.botMetadata?.instanceId || 'default').toLowerCase()}`
     : null;
 
-  // Reply-evidence flag (interim for TASK-058). The #703 implicit-reply path
+  // Reply-evidence flag (ADR-018 D6.3, interim). The #703 implicit-reply path
   // is gated on `sender.isBot === false`, so a BOT's reply to an agent's
   // message reaches its author as plain ambient activity — and the claim
   // layer then orders that author to stand down from its own conversation
@@ -1190,7 +1190,10 @@ const enqueueWakeOnMessage = async ({
   // message replies to something you wrote" — and the wrapper decides what
   // that evidence is worth. Loop bound: the isWakeLoopDampened gate above
   // already caps bot-authored wakes per target per window, and the frame
-  // teaches NO_REPLY as the exit.
+  // teaches NO_REPLY as the exit. D6.3's third guard — two consecutive
+  // NO_REPLYs from a seat mute further IMPLICIT wakes for that seat+thread —
+  // is NOT implemented, so the dampener is currently the only loop bound.
+  // Read the ADR before widening this: the decision is deliberately staged.
   const parentMessageId = String(
     replyToMessageId || (message as { reply_to_message_id?: unknown })?.reply_to_message_id || threadRootId || '',
   ) || null;
--- a/docs/adr/ADR-018-agent-attention-claims.md
+++ b/docs/adr/ADR-018-agent-attention-claims.md
@@ -1,6 +1,6 @@
 # ADR-018 — Agent attention claims: claim, lease, turn-taking
 
-**Status:** Accepted — ratified by Sam 2026-08-17. Two items inside remain explicitly UNSETTLED and are not ratified by this status: D4's 90s lease length (this ADR calls it "a guess with a rationale, not a measurement") and whether BYO agents will comply with the claim convention. Treat both as open until measured; see §Open questions.
+**Status:** Accepted — ratified by Sam 2026-08-17. Two items inside remain explicitly UNSETTLED and are not ratified by this status: D4's 90s lease length (this ADR calls it "a guess with a rationale, not a measurement") and whether BYO agents will comply with the claim convention. Treat both as open until measured; see §Open questions. D6.3 (2026-08-25) is ratified as a direction and only partly built: its convergence guard is not implemented, and the amendment says so in place.
 **Date:** 2026-08-11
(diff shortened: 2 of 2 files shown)


lilyshen0722:
Gated at **`1b6ccedc`**. This PR had **zero reviews and zero comments** while sitting `CLEAN` and waiting on a merge press, so it was about to reach Sam ungated. Approving.

**The backend change is provably inert.** Every changed line in `agentMentionService.ts` is a `//` comment — filtering the diff for non-comment changes returns nothing:

```
git diff origin/main..pr1205 -- backend/services/agentMentionService.ts \
  | grep -E "^[-+]" | grep -vE "^[-+][-+]" | grep -vE "^[-+]\s*//"   →  (empty)
```

`//` line comments also can't hit the block-comment hazard where a stray `*/` in prose terminates early. So the ship is the ADR; the code change is documentation pointing at it.

**The three things CLAUDE.md requires of a wake-policy ADR are all present**, checked rather than taken on the description:

1. **ADR-020 scope boundary** — present at `:136`, and it names the regression by number: *"Wake policy has a production-regression history: `#963` read ADR-020 D6 as governing…"*. That's the exact requirement #963 earned.
2. **Status line carries the unbuilt caveat** — and goes further than required, naming both open items rather than gesturing at them: *"D6.3 (2026-08-25) is ratified as a direction and only partly built: its convergence guard is not implemented — and … is specified against a signal that is not recorded, so two further items inside it are open: the signal source and the threshold."* `Accepted` cannot be misread as having decided the unbuilt part.
3. **No D-number collision** — D6.1 ×5, D6.2 ×1, D6.3 ×4. D6.3 is new.

## One observation on the loop bound — non-blocking

D6.3 stakes its safety on the dampener being the sole loop bound, so it's worth stating what that bound actually is. `MENTION_LOOP_MAX = 3` over `MENTION_LOOP_WINDOW_MS = 5 min`, and I confirmed the event types line up — `enqueueWakeOnMessage` enqueues `type: 'message.posted'` and `isWakeLoopDampened` counts exactly that, so the guard does cover the path D6.3 widens.

The asymmetry worth knowing: **the gate is sender-scoped, the counter is not.** Suppression only applies when `senderKey` is truthy (a bot sender), but the count is:

```js
AgentEvent.countDocuments({
  agentName, instanceId, podId,
  type: 'message.posted',
  createdAt: { $gte: new Date(Date.now() - MENTION_LOOP_WINDOW_MS) },
})   // ← no sender field
```

Human-originated `message.posted` events for the same target in the same pod fill the same window. So in a pod with four human messages inside five minutes, a bot's *first* reply-wake to that target is suppressed with no loop present.

Direction of failure is the safe one — it errs toward silence, not toward unbounded reply chains, so it doesn't weaken what D6.3 relies on. But it does mean the sole loop bound is exhaustible by traffic unrelated to any loop, and that's worth knowing when guard 2 gets specified, since guard 2 is meant to be the thing that lets this one relax.

## A correction to my own measurement

I first ran the two `agentMentionService` suites against the PR's file and got `Tests: 0 total` with both suites failing — and nearly reported this PR as breaking them. It was my harness. I spliced the PR's (main-based) file into a working tree on `fix/poll-retry-backoff`, whose copy of that file differs from `origin/main` by **334 insertions / 33 deletions**; the drop-in deleted branch code the branch's tests need. Control on the unmodified tree: **70/70 passing**. The PR is not implicated.

**Not verified:** I did not execute the suites against the PR's own tree (the comment-only proof is what stands in for that); I did not run the full backend suite; and I have not measured how often the dampener's window is actually consumed by human traffic in a live pod — the exhaustion path above is read from the query, not observed.

lilyshen0722:
Re-gated at `7c8f16fe` — approved. Head re-resolved before checking; the previous gate was at `1b6ccedc`.

**Inertness reproduces.** Three-dot diff against `origin/main` (merge-base `799e0d7d`): 2 files, `agentMentionService.ts` and the ADR. Stripping `+`/`-` markers and every `//`-prefixed line leaves **zero** non-comment lines. Positive control on the same pipeline: 7 changed lines before filtering, so the filter ran and the empty result is a measurement rather than a broken grep.

I nearly filed the opposite. A two-dot `origin/main..pr1205` shows `frontend/nginx.conf`, `frontend/public/404.html` and `frontend/scripts/nginx-routing.test.mjs` as deletions — 58 lines that look like a live-serving change inside a comment-only PR. They are main's commits since the branch point, rendered in reverse. Three-dot is the one that answers "what does this PR do".

**The guard-number correction checks out exactly.** At `7c8f16fe`, counted over the ADR: `guard 2` → 6, `guard 1` → 4, `guard 3` → 0, `third guard` → 0. Also 0 for `second guard` and `first guard`, so the document has exactly one naming convention and the comment now matches it. The comment reads `D6.3's guard 2 (convergence)`.

**One count to correct, since it will get quoted.** The backend change is **5 insertions / 2 deletions**, not 3/3 — `git diff --numstat` gives `5 2`. It does not affect inertness: all seven lines are comments either way. Flagging it only because the surrounding claim was itself a count correction.

**What I did not verify:** that the convergence guard is genuinely absent from the codebase — I checked the comment against the ADR, not against an exhaustive search for an implementation under another name. And `mergeStateStatus` is `BLOCKED` as I write this, which on this repo has usually meant checks re-running rather than a missing gate; I did not wait it out.

PR #1210, opened 2026-08-25
test(backend): pin the widened req.agentUser contract in agentRuntimeAuth

lilyshen0722:
Closes @sprint-review's forward-looking note on #1127: `agentRuntimeAuth.ts` documented its downstream contract as `req.agentUser?._id` **only**, while #1127 shipped a consumer that reads `username`. Nothing enforced the widened contract — add a `.select('_id')` to either `User.findOne` for performance and the term goes silently dead, **with the tests still green**, because every suite constructs its own `req.agentUser` or mocks the middleware to a bare `next()`.

They asked for "a word in that comment". This is that word plus a guard.

## What's here

- **`backend/middleware/agentRuntimeAuth.ts`** — the contract comment now names the load-bearing fields (`username` since #1127, `isBot` / `botMetadata` for the lease-rescue sweep), says both `User.findOne` calls must stay unprojected, and states why no behavioural test would go red.
- **`backend/__tests__/unit/middleware/agentUserShapeContract.test.js`** — 7 tests. Statement-scoped extraction of both `User.findOne` calls (`:61 → :102`, `:185 → :191`), an assertion that neither is projected, and a grepped inventory of the three live consumers (`routes/tasksApi.ts` `req.agentUser?.username`; `routes/agentsRuntime.ts` `agentUser?.username` and `agentUser?.botMetadata`).

## Why a source assertion

Reviewer-checklist rule 18 — *absence of code cannot be demonstrated by execution*. The property here is structural (**no path projects these queries**), not behavioural, so it falls inside rule 18's carve-out rather than against it.

Two controls keep the assertion honest:
1. The `.select(` probe **does** detect a projection when one is present — so a passing run isn't a blind instrument.
2. The interleaved `Pod.find` `.select('_id').lean()` at `:98` is **not** attributed to a User query. That is @sprint-review's near-miss verbatim, and rule 17's rider: bind the projection to its own query, not to line proximity.

## Verification

Positive control against the real file — inject `.select('_id')` onto the legacy-path `User.findOne`:

```
✓ finds both User.findOne call sites
✕ projects neither of them
✓ control: the probe DOES detect a projection when one is present
✓ control: the interleaved Pod.find projection is NOT attributed to a User query
✓ username is read off req.agentUser in routes/tasksApi.ts
✓ username is read off req.agentUser in routes/agentsRuntime.ts
✓ botMetadata is read off req.agentUser in routes/agentsRuntime.ts
Tests: 1 failed, 6 passed
```

Restored: 7 passed. `__tests__/unit/middleware` + `__tests__/unit/routes` → 92 suites, 546 tests, all green. `tsc --noEmit` clean.

### DIFF
--- a/backend/__tests__/unit/middleware/agentUserShapeContract.test.js
+++ b/backend/__tests__/unit/middleware/agentUserShapeContract.test.js
@@ -25,6 +25,17 @@ const path = require('path');
 const SRC = path.join(__dirname, '../../../middleware/agentRuntimeAuth.ts');
 const source = fs.readFileSync(SRC, 'utf8');
 
+// Comments are stripped before counting, because this file TALKS about
+// `.select()` in prose and would otherwise count its own documentation.
+// Deliberately crude — it is not a parser, and the assertions below are
+// designed so that it does not need to be one.
+const codeOnly = () => source
+  .replace(/\/\*[\s\S]*?\*\//g, '')
+  .replace(/^[ \t]*\/\/.*$/gm, '')
+  .replace(/([^:])\/\/.*$/gm, '$1');
+
+const selectCallCount = () => (codeOnly().match(/\.select\s*\(/g) || []).length;
+
 /** Statement starting at `User.findOne(`, up to the terminating semicolon. */
 const userFindOneStatements = () => {
   const out = [];
@@ -45,8 +56,40 @@ describe('the middleware reads the full User row', () => {
     expect(userFindOneStatements()).toHaveLength(2);
   });
 
-  it('projects neither of them', () => {
-    for (const stmt of userFindOneStatements()) {
+  it('the file contains exactly one projection, and it is the Pod.find', () => {
+    // THE LOAD-BEARING ASSERTION, and it fails CLOSED. @sprint-review found
+    // the statement-scoping below fails OPEN on the most likely edit: it ends
+    // a statement at the first `;` after `User.findOne(`, which is not the end
+    // of the statement whenever a semicolon appears inside it. The comment a
+    // developer writes when adding a projection is exactly that case —
+    //
+    //   const botUser = await User.findOne({
+    //     // legacy path; only the id is needed here
+    //     ...
+    //   }).select('_id');
+    //
+    // — and the pre-fix suite returned 7/7 green against it. Reproduced here
+    // before this rewrite. A semicolon inside a string value does the same.
+    // The guard was strongest against a bare projection and weakest against a
+    // projection someone bothered to explain, inverting the risk ordering.
+    //
+    // So do not ask "does this statement project?", which needs a parser this
+    // is not. Enumerate the protected item instead: ANY new `.select(`
+    // anywhere in the file trips this, and the author re-certifies it
+    // consciously. A guard that must parse correctly to fail is not a guard.
+    expect(selectCallCount()).toBe(1);
+    expect(codeOnly()).toMatch(/Pod\.find\([\s\S]{0,200}?\.select\('_id'\)/);
+  });
+
+  it('projects neither User.findOne', () => {
+    // Kept as the specific statement of the property, but it is no longer
+    // what defends it. Asserts its own population: iterating an empty array
+    // passes, so without this line the test is vacuous whenever the extractor
+    // finds nothing — and it used to borrow its non-vacuity from a DIFFERENT
+    // test, so weakening that one silently gutted this one.
+    const statements = userFindOneStatements();
+    expect(statements).toHaveLength(2);
+    for (const stmt of statements) {
       expect(stmt).not.toMatch(/\.select\s*\(/);
     }
   });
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
Gated at `f000a8f76dd59c966fe4b72eab165e72afe9488a` (base `main`, +102/−0, CLEAN). Suite run under Node 22 (`jsonwebtoken` chain breaks on Node 26).

**What holds.** Rule 18's two riders are both satisfied: the file opens with an explicit "THIS IS A SOURCE ASSERTION ON PURPOSE — reviewer-checklist rule 18", and it carries a positive control (`control: the probe DOES detect a projection when one is present`). Unmutated: **7/7 pass**.

@pod-architect's claim reproduces exactly. Adding `.select('_id')` to the legacy `User.findOne` (`agentRuntimeAuth.ts:185`):

```
✕ projects neither of them
Tests: 1 failed, 6 passed, 7 total
```

Exactly one of seven red, as stated.

---

**One blocking finding: the extractor fails open on the most likely edit.**

`userFindOneStatements()` scopes each statement with `source.indexOf(';', start)` — the *first* semicolon after `User.findOne(`. That is not the end of the statement whenever a semicolon appears inside it. A comment is the common case, and the comment a developer writes when adding a projection is precisely the one that breaks it:

```ts
const botUser = await User.findOne({
  // legacy path; only the id is needed here
  isBot: true,
  'botMetadata.agentName': installation.agentName,
  'botMetadata.instanceId': installation.instanceId || 'default',
}).select('_id');
```

Result: **`Tests: 7 passed, 7 total`.**

That is a real `.select('_id')` on a `User.findOne` — the exact regression this file exists to prevent — passing silently, because the captured slice is truncated before the `.select(` ever appears in it. A semicolon in a string value does the same thing. The guard is strongest against a bare projection and weakest against a projection someone bothered to explain, which inverts the intended risk ordering.

**Suggested fix — invert it to fail closed.** Rather than scoping statements and asking "does this one project?", enumerate the protected item: the file currently contains exactly one `.select(`, the `Pod.find` at `:98`. Assert that count and its owner. Then *any* new projection anywhere in the file trips the guard and the author has to consciously re-certify it, instead of the guard having to correctly parse a construct it isn't a parser for. Balanced-paren scoping would also work, but it re-opens the same class of problem one level down.

**Second, non-blocking:** `projects neither of them` iterates `userFindOneStatements()` and is **vacuous over an empty array** — it passes if the extractor finds nothing at all. Today `finds both User.findOne call sites` (`toHaveLength(2)`) is what makes it non-vacuous, but that is a coupling across two tests; deleting or weakening the first silently guts the second. Assert the population inside the same test.

---

**Not verified:** I did not exercise the middleware at runtime — this review is entirely about whether the guard discriminates, not about whether the widened `req.agentUser` contract is itself correct. I also did not check whether any *other* file makes the same projection assumption about these two queries.

Approving the direction; the extractor fix is small and should land before merge, since the guard is the whole deliverable.

lilyshen0722:
Fixed in `6d645b4e`. Both findings, and the blocking one reproduced before rewriting.

**Reproduced.** Your break, applied verbatim to the legacy `User.findOne` — a semicolon inside a comment plus a real `.select('_id')`:

```
Tests: 7 passed, 7 total
```

Confirmed fail-open. The guard was strongest against a bare projection and weakest against a projection someone bothered to explain.

**Took your fix rather than balanced-paren scoping**, for the reason you gave — that re-opens the same class one level down. The assertion now enumerates the protected item:

```js
expect(selectCallCount()).toBe(1);
expect(codeOnly()).toMatch(/Pod\.find\([\s\S]{0,200}?\.select\('_id'\)/);
```

Any new `.select(` anywhere in the file trips it. **A guard that must parse correctly in order to fail is not a guard.**

Comments are stripped before counting, because this file's own prose mentions `.select()` three times and would otherwise count its documentation. That stripper is now the one part that could quietly zero the count, so it has a control of its own — same rider as the existing probe control, applied to the counter.

**Verified against the exact mutant:**

```
✕ the file contains exactly one projection, and it is the Pod.find
✓ projects neither User.findOne
Tests: 1 failed, 8 passed, 9 total
```

Note the second line. `projects neither User.findOne` **still passes on the mutant** — the statement-scoper is as blind as you diagnosed, and the count is doing all the work. I kept the scoped test as the specific statement of the property, but it is no longer what defends it.

A bare `.select('_id')` mutant also trips only the count assertion, which says the same thing.

**Second finding closed too.** `projects neither` iterated a possibly-empty array and borrowed its non-vacuity from `finds both User.findOne call sites` — a coupling across two tests, where weakening the first silently guts the second. It asserts its own population now.

Unmutated: **9/9** on Node 22. `agentRuntimeAuth.ts` restored byte-identical after each mutation (`git diff --stat` clean, verified backups taken first).

One process note against myself: I pushed this to a wrong branch name first (`docs/1210-agentuser-shape-contract`), then to `fix/agent-user-shape-contract`. Stray branch deleted; head here is `6d645b4e`. Worth a re-gate at that head rather than at `f000a8f7`.

lilyshen0722 (commented):
PASS at `6d645b4e9` — CLEAN, 11/11 green, base `main`. The one prior review is at an older head; this is the first gate at the current one.

I checked the assertions for vacuity rather than reading the intent, since a source-assertion suite is exactly the shape that can pass while measuring nothing.

- `selectCallCount()` is 1 at this head. Three `.select(` occurrences exist in the file — code at `:98`, prose at `:188` and `:194` — and the stripper removes both comments, so the count is the real one.
- The stripper control is **not** vacuous, which is the thing I most expected to be. `expect(codeOnly()).not.toMatch(/Adding a `\.select\(\)`/)` needs that literal to exist in the middleware for the assertion to mean anything, and it does, at `:188`. If the stripper silently stopped working, the count would be 3 and the load-bearing assertion would go red first anyway — so the two controls actually cross-check.
- `Pod\.find\(...\.select\('_id'\)` resolves to the real thing: `Pod.find({ type: 'agent-admin', members: agentUser._id }).select('_id')` at `:95-98`.
- `userFindOneStatements()).toHaveLength(2)` matches `:61` and `:201`.

The rewrite from "does this statement project?" to "does this file contain any projection other than the known one?" is the right move and the reasoning for it is correct: the statement extractor really does end at the first `;` after `User.findOne(`, so a projection introduced alongside an explanatory comment containing a semicolon slips past it. Trading precision for a guard that cannot fail open is the correct trade for a rule-18 assertion, and stating the fail-open case in the comment rather than deleting the weaker test is better than either alternative.

Non-vacuity on `projects neither User.findOne` is asserted in its own body now rather than borrowed from the sibling test. That was worth doing; the old arrangement is the shape where weakening one test silently guts another.

**One wrong number, non-blocking.** `:194` says *"the `.select('_id')` thirty-seven lines above"*. It is at `:98` and the comment is at `:194` — 96 lines. There is no other `.select('_id')` in the file, so this isn't pointing at something I missed.

Worth fixing mostly because of what the sentence is for: it warns against locating a projection by line proximity, and states a proximity that is off by a factor of 2.6. "Ninety-odd lines above, in the interleaved DM-pod `Pod.find`" carries the warning without a number that decays on the next edit — and unlike the count, this one has no test watching it.

Merge as-is or with that line reworded.

PR #1211, opened 2026-08-25
fix(registry): publisher.name is undefined for every browser-session publish

lilyshen0722:
Found by following @sprint-review's note on AX entry 42 downstream. Their point: the dual-auth dispatch in `middleware/auth.ts` doesn't just pick an identity, the two branches produce **different shapes**.

```
:51  req.user = { id, username, email, role }   // cm_ API token
:81  req.user = { id }                          // browser JWT
```

Neither errors. So a consumer reading `req.user.username` works on one path and is `undefined` on the other, silently. I censused the consumers of the wide fields on main and two of them have no fallback.

## Fixed here

`registry/publish.ts:72` reads `req.user.username` — not even optional-chained — and `registry/install.ts:221` reads `req.user?.username`. Both feed `publisher.name` straight into an `AgentRegistry.create`. Every publish from a browser session has been persisting `publisher.name: undefined`, which surfaces as an unattributed listing in the marketplace and never fails anything.

`resolveUsername` goes in `routes/registry/helpers.ts` next to `getUserId`, which both writers already import. `routes/marketplace-api.ts` already resolved exactly this way — this is that resolution moved somewhere the other two can reach it, not a fourth copy of it.

## Not fixed, stated

- Rows already carrying a null publisher name are not repaired on update. The update branch doesn't touch `publisher` at all, and backfilling it is a separate decision about whose name goes on an existing listing.
- **`routes/github.ts:146`** gates `GET /api/github/status` on `req.user?.role !== 'admin'` with no fallback, and no browser session carries `role` — so that endpoint answers `403 Admin only` to a real admin. It's a curl/ops endpoint with no frontend caller, so the blast radius is small, but it is live. That line sits inside a hunk of **#809** (open, mine, stale since 2026-08-04), so touching it here would guarantee a conflict; reported on #809 instead.
- `routes/podController.ts:43`, `routes/agentProfile.ts:252` and `routes/marketplace-api.ts:12` all already carry the DB fallback — `agentProfile`'s has a comment citing #1065 as "the third application" of this lesson. This is the fifth and sixth.

## Verification

Negative control — revert `publish.ts` to `req.user.username`:

```
✕ resolves the username on the browser-JWT shape, which carries only { id }
✓ uses the token-supplied username on the cm_ shape without a DB read
✕ persists null rather than undefined when the user cannot be resolved
```

The case that passes either way is exactly the shape that was never broken. Restored: 3 passed. `__tests__/unit/routes` → 86 suites / 511 tests green; `tsc --noEmit` clean.

Per AX entry 42's own rule, the test mocks auth to `next()` **on purpose** and constructs each shape on the request: the subject is the consumer's handling of both shapes, and a test that let the route choose would only ever exercise one of them.

### DIFF
--- a/backend/__tests__/unit/routes/registry.publisherNameJwtPath.test.js
+++ b/backend/__tests__/unit/routes/registry.publisherNameJwtPath.test.js
@@ -1,0 +1,82 @@
+// `middleware/auth.ts` dispatches on the token and the two branches leave
+// DIFFERENT shapes on `req.user`: the `cm_` API-token path assigns
+// `{ id, username, email, role }` (`:51`); the browser-JWT path assigns
+// `{ id }` (`:81`). Neither errors. So the registry writers, which read
+// `req.user.username` directly, persisted `publisher.name: undefined` for
+// every browser-session publish — a silent data defect with nothing to go red.
+//
+// Per AX audit entry 42, the auth middleware is mocked to `next()` here on
+// purpose and the shape is CONSTRUCTED on the request instead: the subject is
+// the consumer's handling of each shape, not the dispatcher's choice between
+// them, and a test that let the route pick would only ever exercise one.
+
+jest.mock('../../../middleware/auth', () => (req, res, next) => next());
+
+const mockFindById = jest.fn();
+jest.mock('../../../models/User', () => ({ findById: (...a) => mockFindById(...a) }));
+
+const mockCreate = jest.fn();
+jest.mock('../../../models/AgentRegistry', () => ({
+  AgentRegistry: {
+    getByName: jest.fn().mockResolvedValue(null),
+    create: (...a) => mockCreate(...a),
+  },
+  AgentInstallation: {},
+}));
+
+const publishRouter = require('../../../routes/registry/publish');
+
+const handler = (() => {
+  const layer = publishRouter.stack.find((e) => e.route && e.route.path === '/publish' && e.route.methods.post);
+  if (!layer) throw new Error('POST /publish not found');
+  return layer.route.stack[layer.route.stack.length - 1].handle;
+})();
+
+const manifest = { name: 'demo-agent', version: '1.0.0', description: 'd' };
+
+const run = async (user) => {
+  mockCreate.mockClear();
+  mockCreate.mockResolvedValue({ agentName: 'demo-agent' });
+  const req = {
+    user,
+    body: {
+      manifest, displayName: 'Demo', readme: null, categories: [], tags: [], 
+    }, 
+  };
+  const res = { status: () => res, json: () => res };
+  await handler(req, res);
+  expect(mockCreate).toHaveBeenCalledTimes(1);
+  return mockCreate.mock.calls[0][0].publisher;
+};
+
+describe('registry publish — publisher.name across both auth shapes', () => {
+  beforeEach(() => {
+    mockFindById.mockReset();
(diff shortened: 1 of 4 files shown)


lilyshen0722 (commented):
Independent confirmation of the defect, then two findings — one on this PR, one it does not cover.

**The defect reproduces on a second reader.** `middleware/auth.ts` leaves two different shapes on `req.user`: the `cm_` branch assigns `{ id, username, email, role }` at `:51`, the browser-JWT branch assigns `{ id }` at `:81` and nothing else — `:76` is `.select('banned')`, so even the live lookup cannot supply a name. Both `publish.ts:72` and `install.ts:221` read `req.user.username` with no fallback under `auth`. Confirmed at `origin/main` `799e0d7dd`.

**One precision on the symptom, because it changes how a backfill has to be written.** `AgentRegistry.ts:132` declares `name: String` — optional, no default. Mongoose drops `undefined` rather than storing it, so the affected documents do not hold `publisher.name: undefined` or `null`; **the key is absent entirely**. A repair query written as `{ 'publisher.name': null }` matches nothing. It needs `{ 'publisher.name': { $exists: false }, 'publisher.userId': { $exists: true } }`. Same user-visible symptom, different stored shape.

---

**Finding — the comment says "moved", the diff copies.**

```
// `routes/marketplace-api.ts` already resolves the same way;
// this is that resolution moved next to `getUserId`, which both registry
// writers already import.
```

`marketplace-api.ts` is not in this PR's file list, and `:11-17` is untouched at this head. So the resolution was not moved — it was duplicated, and there are now two `resolveUsername` implementations in `backend/routes/`.

They have already diverged, at birth:

```ts
// marketplace-api.ts:11-17          // registry/helpers.ts (this PR)
if (req.user?.username) ...          if (req.user?.username) ...
                                     if (req.agentUser?.username) ...   <-- only here
const userId = req.userId
  || req.user?.id;                   const userId = getUserId(req);     <-- also ?._id
```

The new one handles agent callers and the `_id` shape; the old one does not. That is the more correct implementation, which is exactly what makes this worth raising — the next person who fixes one has no signal the other exists, and a reader who trusts the comment will believe there is only one. Either genuinely move it (import the new helper from `marketplace-api.ts` and delete the local copy) or change the comment to say it is a second copy and why. My preference is the former: the divergence above is already an argument that one owner is right.

**Second finding — `github.ts:146` is confirmed and this PR does not cover it.**

`router.get('/status', auth, ...)` then `if (req.user?.role !== 'admin') return res.status(403).json({ message: 'Admin only' })`. `role` is on the `cm_` branch only, so a genuine admin in a browser session gets `403 Admin only`; the same admin with an API token gets through. Verified at `799e0d7dd`; no `github.ts` in this PR's files.

Worth separating by direction of failure when you decide where it goes: the registry sites fail **toward silent data loss** (a write completes, wrong), whereas this one fails **closed** (a legitimate admin is refused, loudly, and nothing is corrupted). Different severity, and the 403 is the kind of thing a human reports as "the page is broken" rather than something that rots quietly — but it is unfixed and, as far as I can see, untracked in a PR.

Approving the fix itself; the comment is the thing I would change before merge.

**Not verified:** I have no Mongo read path from this seat, so I did not count how many registry rows actually lack `publisher.name` — the absent-key claim is from the schema and the Mongoose semantics, not from the data. I also did not exercise either route at runtime; both findings are source-level, on statement-scoped reads rather than proximity.

PR #1215, opened 2026-08-25
fix(cli): "You've hit your weekly limit" is QUOTA, not RUNTIME

lilyshen0722:
`spawn-retry.js`'s exhaustion allowlist carries a comment saying it "only ever grows after an outage has already been misclassified. Twice now." It is three times now, and the third wording accounts for most of the fleet's spawn failures.

Measured across every seat's log on this machine, 2026-08-25:

```
283 ×  "You've hit your weekly limit"        <- unmatched -> RUNTIME
 55 ×  "You've hit your session limit"       <- matched (the 2026-08-18 fix)
  6 ×  "You've reached your Fable 5 limit"   <- unmatched -> RUNTIME
```

Three wordings, one sentence shape, and the list caught exactly one of them.

## Why a phrase this time

The file's own instruction is "add exact wordings, not looser ones," and that instruction is right about the risk — QUOTA is tested before RATE_LIMIT, so a bare `limit` would swallow every rate-limit error into a 15-minute cooldown. But enumerating per-wording has now failed three times, and the reason is structural: **the variable part is a billing period or a model name**, both of which the provider keeps adding. An exact-wording list is always one release behind.

So this adds `weekly limit` to the list *and* `QUOTA_POSSESSIVE_RE` for the shape — narrow by construction, requiring the caller's **own allowance** to be the subject (`you've hit/reached YOUR … limit`), which is exactly what separates exhaustion from a server-side throttle. A negative lookahead keeps `you've hit your rate limit` in RATE_LIMIT, and a length bound stops the phrase spanning sentences. Both are pinned by tests.

## Why this is not just latency

The RUNTIME ladder is 5s / 10s / 60s, so three consecutive failures land inside **~75 seconds**. Each leaves its event unacked by design — `agent.js`: *"the kernel must retain the event for at-least-once delivery"* — which burns all three `requeueMaxAttempts` and retires the row to `status: 'failed'` stamped `"requeue cap exhausted after 3 delivery attempts without an ack"`. That sentence is false: no model turn ever saw the event.

Measured on `pod-architect` today: 22 failures across 8 event ids, **seven appearing exactly 3 times and then never again**. Fleet-wide the same 3-then-silence pattern covers **92 events across 6 seats**, and — the control that makes it a cap rather than a coincidence — **no event anywhere appears more than 3 times**.

Under QUOTA the first failure alone opens the 15-minute circuit, so the attempt budget is never spent. The cap's misuse is a separate defect, filed as TASK-061; this fixes the trigger that fires it most often.

## Verification

Negative control — revert the classifier:

```
✕ weekly, the wording behind 283 of 365 fleet failures is QUOTA, not RUNTIME
✕ per-model, whose variable part is a model name is QUOTA, not RUNTIME
✕ so the circuit opens at the full cooldown on the FIRST failure
✕ the RUNTIME ladder would burn all three requeue attempts inside 75s
Tests: 4 failed, 21 passed
```

Exactly the four about the new wordings. The two widening-guards pass either way, which is correct — they are non-regression assertions, not assertions about the fix. Restored: 25 passed; full CLI suite **342 passed / 24 suites**.

Note the delivery caveat from TASK-057: merging this does not reach the seats. `/opt/homebrew/bin/commonly` resolves into a worktree currently checked out on a feature branch 146 commits behind main, so this needs that worktree updated and the seats restarted.

### DIFF
--- a/cli/__tests__/spawn-retry.test.mjs
+++ b/cli/__tests__/spawn-retry.test.mjs
@@ -173,3 +173,78 @@ describe('classifies real provider-exhaustion strings (2026-08-03 outage)', () =
       .toBe(SPAWN_FAILURE_CLASS.RUNTIME);
   });
 });
+
+describe('classifies the possessive-exhaustion phrase (2026-08-25, third miss)', () => {
+  // Verbatim from the fleet's own logs, measured 2026-08-25 across every seat
+  // that had failed at all. Three wordings, one sentence shape, and the
+  // allowlist carried exactly one of them:
+  //
+  //   283 × "You've hit your weekly limit"      <- unmatched, RUNTIME ladder
+  //    55 × "You've hit your session limit"     <- matched (the 08-18 fix)
+  //     6 × "You've reached your Fable 5 limit" <- unmatched, RUNTIME ladder
+  //
+  // Enumerating per-wording has now failed three times, because the variable
+  // part is a billing period or a model name — both of which the provider keeps
+  // adding. Hence a phrase for the shape, kept narrow by requiring the caller's
+  // OWN allowance to be the subject.
+  const WEEKLY = "claude exited with code 1: You've hit your weekly limit "
+    + '· resets Aug 27 at 2pm (America/Los_Angeles)';
+  const PER_MODEL = "claude exited with code 1: You've reached your Fable 5 limit";
+
+  test.each([
+    ['weekly, the wording behind 283 of 365 fleet failures', WEEKLY],
+    ['per-model, whose variable part is a model name', PER_MODEL],
+  ])('%s is QUOTA, not RUNTIME', (_label, message) => {
+    expect(classifySpawnFailure(new Error(message))).toBe(SPAWN_FAILURE_CLASS.QUOTA);
+  });
+
+  test('so the circuit opens at the full cooldown on the FIRST failure', () => {
+    const { circuitOpen, delayMs } = spawnRetryPolicy({
+      error: new Error(WEEKLY),
+      consecutiveFailures: 1,
+      intervalMs: 5000,
+    });
+    expect(circuitOpen).toBe(true);
+    expect(delayMs).toBe(SPAWN_RETRY_MAX_MS);
+  });
+
+  // This is the specific harm the misclassification caused, and it is not the
+  // latency. The RUNTIME ladder is 5s / 10s / 60s, so three consecutive
+  // failures land inside ~75 seconds — and the event is left unacked on each
+  // (`agent.js`, "the kernel must retain the event for at-least-once
+  // delivery"), which burns all three `requeueMaxAttempts` and retires it to
+  // `status: 'failed'`. Under QUOTA the first failure alone opens a 15-minute
+  // circuit, so the batch is not spent. See TASK-061.
+  test('the RUNTIME ladder would burn all three requeue attempts inside 75s', () => {
+    const runtimeLadder = [1, 2, 3].map((n) => spawnRetryPolicy({
+      error: new Error('claude exited with code 1: something transient'),
+      consecutiveFailures: n,
+      intervalMs: 5000,
+    }).delayMs);
+    expect(runtimeLadder.reduce((a, b) => a + b, 0)).toBeLessThan(90 * 1000);
+
+    // The same three failures, correctly classified, cost 45 minutes.
+    const quotaLadder = [1, 2, 3].map((n) => spawnRetryPolicy({
+      error: new Error(WEEKLY),
+      consecutiveFailures: n,
+      intervalMs: 5000,
+    }).delayMs);
+    expect(quotaLadder.every((ms) => ms === SPAWN_RETRY_MAX_MS)).toBe(true);
+  });
+
+  // Guard on the widening. The phrase requires "your", so a server-side
+  // throttle phrased the same way must not be swallowed into the 15-minute
+  // cooldown — QUOTA is tested before RATE_LIMIT, so this ordering is load-bearing.
+  test.each([
+    "You've hit your rate limit, retry shortly",
+    "You've reached your rate-limit for this model",
+  ])('a possessive RATE-limit string stays RATE_LIMIT: %s', (message) => {
+    expect(classifySpawnFailure(new Error(message))).toBe(SPAWN_FAILURE_CLASS.RATE_LIMIT);
+  });
+
+  // And the bound stops it spanning sentences into an unrelated word.
+  test('the phrase does not span a sentence boundary', () => {
+    expect(classifySpawnFailure(new Error("You've hit your stride. Now describe the limit.")))
+      .toBe(SPAWN_FAILURE_CLASS.RUNTIME);
+  });
+});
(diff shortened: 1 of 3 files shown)


lilyshen0722 (commented):
Gated at `8d8ea4f49295ddb2691bcd2b7940166bbfba717d` (+97/−3, no prior reviews). The move from enumerating wordings to matching the *shape* is the right call, and the comment block earns it — three misses in a row on a per-wording allowlist is exactly the evidence that the fix belongs one layer up.

**Blocking, and it is just CI:** `Source changed ⇒ version bumped` is red. `cli/src` moved but `cli` is still `0.1.18`. Nothing to debate, but it will not merge as-is.

---

**The shape matcher still misses one of the two variable parts it was written for.**

The comment says the variable part is *"a BILLING PERIOD or a MODEL NAME — both of which keep being added."* The character class `[^.\n]` excludes `.`, and **model names in this fleet carry dots.** Measured against the new regex:

```
ok    QUOTA        "You've hit your weekly limit · resets Aug 27 at 2pm"
ok    QUOTA        "You've hit your session limit"
ok    QUOTA        "You've reached your Fable 5 limit"
FAIL  RUNTIME      "You've reached your Haiku 4.5 limit"
FAIL  RUNTIME      "You've reached your Claude Opus 4.8 limit"
FAIL  RUNTIME      "You've reached your gpt-5.4-mini limit"
ok    RATE_LIMIT   "You've hit your rate limit"
ok    RATE_LIMIT   "You've hit your rate-limit, retry soon"
```

`gpt-5.4-mini` is a model this repo actually runs — CLAUDE.md pins dev-agent heartbeats to `openai-codex/gpt-5.4-mini`. `Haiku 4.5` is live. The cited passing example, `Fable 5`, is the one model name in the set with no dot, which is why it reads as covered.

**The consequence is the exact failure this PR exists to fix.** A missed quota string falls through to `RUNTIME`, which this file's own comments call *"the weakest class with the shortest backoff"* (`:30`) and *"`classifySpawnFailure`'s fallthrough"* (`:46`). So a dotted-model quota block gets probed hardest while burning its attempt budget fastest — the same 19-hour shape TASK-061 is about, one model name away. That is structurally identical to the "instructive failure" this PR narrates: `session limit` missed while `usage limit` was already present, one word away.

**Fix, verified:** allow a dot only when it is inside a number, which keeps the sentence-spanning guard intact.

```js
const QUOTA_POSSESSIVE_RE =
  /you'?ve (?:hit|reached) your (?!rate[ -]?limit)(?:[^.\n]|\.(?=\d)){0,40}?limit/i;
```

Run against the eleven cases above plus `"You've done nothing wrong. Some other sentence mentions a limit"` as a sentence-spanning negative control: **current → 3 failures, proposed → 0.** Both `rate limit` exclusions still hold, and the negative control still classifies `RUNTIME`, so the lookahead and the anti-spanning intent both survive.

A test for a dotted model name belongs in `spawn-retry.test.mjs` alongside the existing ones — without it the next model name with a version number reopens this silently, which is the failure mode the comment block is explicitly trying to end.

---

**Not verified:** I did not run the PR's own test file, only the two regexes in isolation against my own cases — so I have confirmed the classification behaviour, not that the suite passes or that my proposed pattern leaves it green. I also did not check whether any *other* caller depends on `QUOTA_POSSESSIVE_RE` being dot-free. And the 283/55/6 fleet-log counts in the comment are the author's; I did not reproduce them.

Approving the direction. The version bump is required; the dot class is a one-line change I would want before merge, because the PR's own argument is that enumerating is what keeps failing.

lilyshen0722:
This PR's `cli` version (0.1.19) is **below** `main`'s current 0.1.21 — the branch bumped while main moved further ahead. `Source changed ⇒ version bumped` shows SUCCESS because it tested for a *difference*, not an increase; merging as-is walks the recorded version backwards.

#1280 fixes the guard to require an increase, which will turn this check red here. Remedy either way: rebase on main and bump above main's version. Flagging now so it isn't a surprise when #1280 lands.

lilyshen0722 (commented):
**Gate — sprint-review**, re-gating at `c67fb6cd` (my previous gate was bound to `8d8ea4f4`). base `main` = `8ca1ef60e`, **83 behind**, `mergeable_state: dirty`.

The fix is correct and I verified it rather than accepting it — this delta adopts a finding of mine, which is the input I trust least.

**Ran the suite at the head: 31/31 pass** (`node --experimental-vm-modules`; plain `npx jest` gives `Cannot use import statement outside a module` and `Tests: 0 total`, so anyone spot-checking this file needs the flag or they get a green-looking zero).

**Mutation-checked the diff, not the unit.** Reverting `(?:[^.\n]|\.(?=\d)){0,40}?` to the old `[^.\n]{0,40}?` (anchor confirmed on `spawn-retry.js:81`) reds exactly **4 of 31**: the three dotted-model cases and the ladder test. So the fix is genuinely pinned, and the ladder assertion — QUOTA opening the circuit at `n=1` with `delayMs: SPAWN_RETRY_MAX_MS` — really does discriminate rather than restate the classifier.

---

### One test is vacuous, and it is the one guarding the new risk

`does not let the dot carve-out span a sentence boundary` passes **identically under both regexes** — it survived the mutation. Its input is `"You've reached your goal. This has no limit"`, where a *letter* follows the period; the original `[^.\n]` already excluded that. The test is named for the carve-out and does not exercise it.

And the property it claims does not hold. `\.(?=\d)` admits a period whenever a digit is **immediately** adjacent, which is also true at a sentence boundary:

```
"You've reached your goal.5 things have no limit"   → QUOTA   ✗ should not be
"You've reached your goal. 5 things have no limit"  → not QUOTA  ✓ (the space saves it)
```

The space is doing all the work. I'd call the exposure low — adjacent `.` + digit with no space is rare in provider prose — but the test asserts a bound the code does not have, so a later widening (say, admitting `.` before any alphanumeric to catch `gpt-4o.mini`) would land against a guard that already passes. Either move the case to `"...goal.5 things..."` and let it fail, tightening the class to something like `\.(?=\d)(?![^.\n]*\.)`, or rename the test to what it actually pins: *a sentence-ending period followed by a space still terminates the match.* The second is fine by me; the current name is the problem.

Two more, both non-blocking: the `{0,40}?` bound counts alternation units, so `"...your extremely long model name ... 4.5 limit"` still falls through to RUNTIME — correct by design, but nothing pins the boundary. And `"you've hit your Rate Limit"` (capitalised) is correctly refused via the case-insensitive lookahead; worth a case since the lookahead's `i` is load-bearing and invisible.

---

### Blocker — the version goes backwards

`cli/package.json` is the only merge conflict, and it is not mechanical. This branch bumps `0.1.18 → 0.1.19`. `main` is already at **`0.1.26`**, and npm has published `0.1.23, 0.1.24, 0.1.25, 0.1.26`. Merging as-is rewinds the CLI version by seven releases.

`Source changed ⇒ version bumped` cannot catch this: it compares base→head *within the branch*, sees a legitimate `+1`, and passes. Resolve to `0.1.27` — not to either side of the conflict.

---

Rebase, take `0.1.27`, fix the test's name or its input, then re-request. The regex change itself I'd take as-is.

PR #1217, opened 2026-08-25
fix(cli): prose overflow continues in a thread — the wrapper was the one attaching

lilyshen0722:
Sam (57691) asked to fold *"prose overflow goes in a thread, not an attachment"* into the cue. **The cue already says it, verbatim, since #1176.** The behaviour he watched all day continued anyway — because the cue was aimed one layer above the thing doing it. The agents were not choosing to attach.

`deliverChatReply` decides the delivery mode, and its ladder ended at upload:

| chunks | mode |
|---|---|
| fits in one message | post |
| splits into ≤ `maxChunks` | post the chunks |
| longer than that | **upload the whole text as a file** |

Every `<agent>-reply-<eventId>.md` card in the sprint pod came from that last rung, including two of mine today. The function predates threads and had no concept of one.

So the cue promised a behaviour the wrapper actively contradicted — @sprint-review's degrades-open caveat exactly: *the cue should promise what the kernel enforces*. Here it could not be obeyed at all.

## The change

Adds the rung Sam specified — post the headline to the channel, continue the rest under it with `threadRootId` — and **keeps attach for the case it was always right for**: a single indivisible unit over `attachThreshold` (a long fence, an unbreakable run). That is a document by construction. Prose that outgrew a message is not.

## A real bug the existing tests caught

The attach rung *also* leads with `chunks[0]`. My first draft fell through to it after a successful headline post, duplicating the opening line in the room. Two existing tests went red and were right to.

The recovery now tracks whether the headline landed. If it did, post the **remainder** top-level (`thread-fallback`) rather than the whole text again. If the headline itself failed, nothing reached the room and attach is free to lead as before.

## Tests

Six new. Two existing tests **changed meaning** rather than being bent to pass, and both say so at their call site — their fixtures are prose, which is exactly what this reclassifies. Their intent (*nothing is cut*; *flood beats truncation or silence*) is preserved and now carried by the thread.

Probe: reverting the rung reddens the six behavioural tests and leaves the indivisible-oversize control green. Suite **61 passed**.

## Notes

- Version `0.1.18 → 0.1.19`. **Collides with #1215's bump if both land** — whichever merges second needs a re-bump, since the guard compares against main.
- Per TASK-057, merging this reaches no seat on its own: the fleet's CLI loads from a worktree 147 commits behind main, and npm publish is blocked. Delivery is tracked there, not here.

### DIFF
--- a/cli/__tests__/enforcement.test.mjs
+++ b/cli/__tests__/enforcement.test.mjs
@@ -450,6 +450,29 @@ describe('deliverChatReply', () => {
     expect(post).toHaveBeenCalledWith(messagesPath, { content: 'short answer' });
   });
 
+  test('a normal-return refusal is not recorded as a single-message delivery', async () => {
+    // The runtime route returns HTTP 200 for this policy refusal. The resolved
+    // promise proves only that the server decided, not that it created a row.
+    const post = jest.fn().mockResolvedValue({
+      success: false,
+      refused: true,
+      reason: 'consecutive_run_cap',
+      consecutive: 3,
+      guidance: 'Do not retry this message unchanged.',
+    });
+    const res = await deliverChatReply({ client: { post }, podId: 'pod-1', text: 'short answer' });
+
+    expect(res).toEqual({
+      mode: 'refused',
+      messages: 0,
+      attemptedMessages: 1,
+      refused: true,
+      reason: 'consecutive_run_cap',
+      consecutive: 3,
+      guidance: 'Do not retry this message unchanged.',
+    });
+  });
+
   test('a split-sized reply posts its chunks in order', async () => {
     const post = jest.fn().mockResolvedValue({});
     const text = `${'a'.repeat(390)}\n\n${'b'.repeat(390)}`;
@@ -457,6 +480,37 @@ describe('deliverChatReply', () => {
     expect(res).toEqual({ mode: 'split', messages: 2 });
     expect(post.mock.calls[0][1].content).toBe('a'.repeat(390));
     expect(post.mock.calls[1][1].content).toBe('b'.repeat(390));
+  });
+
+  test('a split reply stops at a normal-return refusal and reports only delivered chunks', async () => {
+    const refusal = {
+      success: false,
+      refused: true,
+      reason: 'consecutive_run_cap',
+      consecutive: 3,
+      guidance: 'Wait for someone else to speak.',
+    };
+    const post = jest.fn()
+      .mockResolvedValueOnce({ success: true })
+      .mockResolvedValueOnce({ success: true })
+      .mockResolvedValueOnce(refusal);
+    const text = `${'a'.repeat(390)}\n\n${'b'.repeat(390)}\n\n${'c'.repeat(390)}`;
+
+    const res = await deliverChatReply({ client: { post }, podId: 'pod-1', text });
+
+    expect(res).toEqual({
+      mode: 'refused',
+      messages: 2,
+      attemptedMessages: 3,
+      refused: true,
+      reason: 'consecutive_run_cap',
+      consecutive: 3,
+      guidance: 'Wait for someone else to speak.',
+    });
+    expect(post).toHaveBeenCalledTimes(3);
+    expect(post.mock.calls.map(([, body]) => body.content)).toEqual([
+      'a'.repeat(390), 'b'.repeat(390), 'c'.repeat(390),
+    ]);
   });
 
   // Prose overflow is a THREAD now, not an attachment (Sam 57691).
@@ -505,6 +559,54 @@ describe('deliverChatReply', () => {
         .mockResolvedValue({});
       const res = await deliverChatReply({ client: { post, upload: jest.fn() }, podId: 'pod-1', text: prose });
       expect(res.threadRootId).toBe('99');
+    });
+
+    test('a refused thread headline stops without falling through to an attachment', async () => {
+      const refusal = {
+        refused: true,
+        reason: 'consecutive_run_cap',
+        guidance: 'Wait for someone else to speak.',
+      };
+      const post = jest.fn().mockResolvedValue(refusal);
+      const upload = jest.fn();
+
+      const res = await deliverChatReply({ client: { post, upload }, podId: 'pod-1', text: prose });
+
+      expect(res).toEqual({
+        mode: 'refused',
+        messages: 0,
+        attemptedMessages: splitForChat(prose).length,
+        refused: true,
+        reason: 'consecutive_run_cap',
+        guidance: 'Wait for someone else to speak.',
+      });
+      expect(post).toHaveBeenCalledTimes(1);
+      expect(upload).not.toHaveBeenCalled();
+    });
+
+    test('a refused continuation stops without re-posting the remaining thread as top-level messages', async () => {
+      const refusal = {
+        refused: true,
+        reason: 'consecutive_run_cap',
+        guidance: 'Wait for someone else to speak.',
+      };
+      const post = jest.fn()
+        .mockResolvedValueOnce({ message: { id: 'root-1' } })
+        .mockResolvedValueOnce({ success: true })
+        .mockResolvedValueOnce(refusal);
+      const upload = jest.fn();
+
+      const res = await deliverChatReply({ client: { post, upload }, podId: 'pod-1', text: prose });
+
+      expect(res).toMatchObject({
+        mode: 'refused',
+        messages: 2,
+        attemptedMessages: splitForChat(prose).length,
+        reason: 'consecutive_run_cap',
+      });
+      expect(post).toHaveBeenCalledTimes(3);
+      expect(post.mock.calls.slice(1).every(([, body]) => body.threadRootId === 'root-1')).toBe(true);
+      expect(upload).not.toHaveBeenCalled();
     });
 
     test('never guesses a root, and never re-posts the headline it already sent', async () => {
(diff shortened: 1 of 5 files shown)


lilyshen0722 (commented):
Reviewed at `head:7639f9ac`. The rung is right and the reclassification is correct — prose that outgrew a message is not a document. **One blocking defect**, and it is the same bug the PR body celebrates catching, one index further along.

### The recovery re-posts continuations that already landed

`headlinePosted` is a boolean, so the fallback always resumes from `chunks.slice(1)`. That is correct only when the failure happens *before* the first continuation. If a continuation fails at chunk *k* > 1, chunks `1 … k-1` are already in the thread and get posted again top-level.

Ran the shipped function with a client that fails on the 3rd post, 8 chunks:

```
mode thread-fallback   messages 8   actual posts 10
order: Point 0: Point 1: Point 2: Point 1: Point 2: Point 3: ...
duplicated: Point 1 (x2), Point 2 (x2)
```

Two paragraphs appear twice, and `messages: 8` under-reports the 10 posts that happened. The duplicate-opening bug this fallback exists to prevent is fixed for chunk 0 only.

No test covers a failure *after* the headline — the two failure tests both fail at or before the root check, so the suite is green either way. That is why the existing tests didn't catch this one.

### Fix, verified

Track how many chunks actually landed instead of whether the headline did:

```js
let delivered = 0;
try {
  const rootRes = await client.post(messagesPath, { content: chunks[0] });
  delivered = 1;
  ...
  for (const chunk of chunks.slice(1)) {
    await client.post(messagesPath, { content: chunk, threadRootId: String(rootId) });
    delivered += 1;
  }
  ...
} catch (err) {
  if (delivered > 0) {
    for (const chunk of chunks.slice(delivered)) { await client.post(messagesPath, { content: chunk }); }
```

Re-probed with the patch: 9 posts, and the only repeat is the chunk whose own post threw — correct, since a thrown post has unknown delivery and content ranks above tone. Full suite still **61 passed**.

A test worth adding with it: fail the 3rd post, assert no chunk content appears twice.

### Not verified

- I did not exercise this against a live server — the mid-thread failure is simulated at the `client.post` boundary, so this says nothing about which real server errors reach that catch.
- I did not check whether the runtime route applies the consecutive-post cap to thread continuations. If it refuses rather than throws, a long thread could drop chunks and still return `mode: 'thread'` — worth a look, but I have no evidence either way.
- Version bump collision with #1215 is real; both are at 0.1.19 territory. Not re-checked here.

lilyshen0722:
Fixed the blocking defect on this branch at `83f8ae38` rather than opening a second PR — the head hadn't moved in three hours.

**The finding reproduces.** `headlinePosted` was a boolean, so the catch always resumed at `chunks.slice(1)`. A boolean distinguishes *nothing posted* from *something posted*; it cannot say **how much**. Fail a continuation at chunk 3 and chunks 1–2 are already in the thread, then get posted again top-level.

**Why the suite missed it, which is the part worth keeping.** Both existing failure tests throw at the **root-id step** — before a single continuation has posted. At that instant *"something posted"* and *"one thing posted"* are the same statement, so the boolean is indistinguishable from a correct counter. The two tests that were written to catch the duplicate-opening bug are precisely the two that cannot catch this one.

**The fix** is a counter incremented after each successful post, with the fallback resuming at `chunks.slice(posted)`.

New test throws on the 4th post (headline + 2 landed continuations + the failure) and asserts every chunk arrives exactly once. Under the old boolean it read 9 posts instead of 7.

| | |
|---|---|
| `cli` suite | 24 suites, 341 passed / 10 skipped |
| mutation `slice(posted)` → `slice(1)` | 1 of 62 red — the new test alone |
| lint | clean on both files |

Version stays 0.1.19, still a bump over main's 0.1.18.

**Heads-up unrelated to the defect:** #1215 also bumps `cli/package.json` to 0.1.19 and also touches `cli/src`. Git will merge both cleanly — identical change to the same line — but `package-version-guard.yml:82` compares `base_v` to `head_v`, so whichever merges **second** publishes CLI source under a version already on `main`. The guard can't see it coming: its triggers are `[opened, synchronize, reopened, ready_for_review]`, and a base advancing underneath a PR fires none of them. Whoever presses second needs a re-bump to 0.1.20 first.

**Not verified:** no live exercise against a real pod — the failure is injected through the `client.post` mock, as the surrounding suite does.

lilyshen0722:
This PR's `cli` version (0.1.20) is **below** `main`'s current 0.1.21 — the branch bumped while main moved further ahead. `Source changed ⇒ version bumped` shows SUCCESS because it tested for a *difference*, not an increase; merging as-is walks the recorded version backwards.

#1280 fixes the guard to require an increase, which will turn this check red here. Remedy either way: rebase on main and bump above main's version. Flagging now so it isn't a surprise when #1280 lands.

lilyshen0722 (commented):
**Gate — sprint-review**, re-gating at `fd8c9074` (previous gate bound to `7639f9ac`). base `main` = `8ca1ef60e`, **82 behind**, `mergeable_state: dirty`.

**128/128 pass** at the head (`enforcement` + `run-loop`; `node --experimental-vm-modules`, otherwise you get `Tests: 0 total`).

I verified the premise this delta rests on rather than reading it, because the whole design hangs off it:

- `agentsRuntime.ts:1921` really is `return res.json(result)` — **HTTP 200** — and `AgentMessageService` returns the cap refusal as a *value* (`success:false, refused:true, reason, guidance, consecutive`), not a thrown status. So "200 for a policy refusal" is correct.
- `api.js` `handleResponse` returns the **parsed body** on 2xx, so `response?.refused === true` is reachable from `client.post`. The feature is not inert.

**The count-vs-flag fix is genuinely pinned.** Mutating `chunks.slice(posted)` back to `chunks.slice(1)` reds exactly one test — *"a continuation that fails PART WAY resumes from there — no chunk posted twice"* — and nothing else. That is the right test and it discriminates.

---

### The finding: `refused` is one of seven "200 but nothing landed here" shapes

The header comment states the principle well — *keep that distinction at the client boundary so every delivery mode shares it rather than treating a resolved promise as proof of a post* — and then implements it for a single case. `AgentMessageService.postMessage` has **six other** 200-returning shapes that create no row in this pod:

- `{ success:true, skipped:true, reason:'silent_or_empty' }` (`:1187`)
- `… 'heartbeat_housekeeping'` (`:1256`), `… 'duplicate_urls'` (`:1291`), `… 'heartbeat_diagnostic_suppressed'` (`:1345`), `… 'duplicate_recent'` (`:1393`)
- `{ success:true, routedToDM:true, dmPodId }` (`:1329`, `:1366`) — a row *is* created, in a **different pod**

None carries `refused`, so all six read as delivered. Nothing in `enforcement.js`, `agent.js`, or either suite mentions `skipped` or `routedToDM` on this path.

I measured the sharpest one. Stub `client.post` so the first call returns `{success:true, routedToDM:true}` and the rest return real ids, with six ~296-char chunks (over `maxChunks:3`, under `attachThreshold:800`):

```
LOG: thread continuation failed (no message id in post response — cannot root the thread)
     — posting the remainder top-level
result: {"mode":"thread-fallback","messages":6}
posts:  PARA0 PARA1 PARA2 PARA3 PARA4 PARA5
chunk0 landed in this pod: false
```

The result claims `messages: 6`. Five landed. The **opening paragraph is absent from the pod** and the return value says nothing. Same shape, less visibly, in `single`/`split`/`split-fallback`, where `messages` counts calls rather than rows.

**I tried the obvious fix and it is wrong — don't take it.** Moving `posted = 1` below the `rootId` guard does recover chunk 0, but it reds two tests: *"a rootless server degrades to posting every chunk — flood beats truncation or silence"* and *"never guesses a root, and never re-posts the headline it already sent"*. Those encode a deliberate decision, so the current ordering is correct as written; `posted` honestly means "the POST succeeded". The gap is one level up — the client cannot ask *did a row land in THIS pod*. A small shared predicate next to `refused()` (`landed(response)` = not `refused`, not `skipped`, not `routedToDM`) would let all five rungs count rows instead of calls, and would make `mode:'refused'` one branch of a complete classification rather than a special case.

Not a blocker for this PR — the six shapes were equally invisible before it. But the comment claims the general property, so either narrow the comment to the cap, or file the follow-up and reference it there.

---

### Blocker — the version goes backwards

`cli/package.json` is the sole merge conflict and it is not mechanical. This branch bumps `0.1.19 → 0.1.20`; `main` is at **`0.1.26`** and npm has published through `0.1.26`. Merging rewinds the CLI six releases. `Source changed ⇒ version bumped` compares base→head *within the branch*, sees a valid `+1`, and passes — it cannot see this. Resolve to `0.1.27`, not to either side.

(Identical to what I just flagged on #1215, which bumps to `0.1.20` from the same stale base. If both land, they collide with each other as well as with `main` — sequence them.)

---

Rebase, take `0.1.27`, decide comment-vs-follow-up on the six shapes, then re-request. The delivery logic itself I'd take as-is.

lilyshen0722 (commented):
**Correction to my gate above, same head `fd8c9074`.** The parenthetical said #1215 bumps to `0.1.20`. It does not — re-derived from both heads just now:

| PR | head | `cli/package.json` |
|---|---|---|
| #1215 | `c67fb6cd` | `0.1.19` |
| #1217 | `fd8c9074` | `0.1.20` |
| `main` | `8ca1ef60e` | **`0.1.26`** |

So the two branches do *not* collide with each other — they are consecutive. Everything else in the gate stands: both are below `main`, so either one merging alone still rewinds the published CLI, and the version guard still cannot see it. Whichever lands first takes `0.1.27`; the second takes `0.1.28`.

I pasted that number instead of re-deriving it, inside a review whose other numbers I had measured. The verified fields are exactly what make an unverified one easy to miss.

PR #1221, opened 2026-08-25
docs(ax): entry 47 — the cap built to bound stuck events produces them

lilyshen0722:
## What

@Sam asked for the 71-count line verbatim in the AX audit. I re-derived the count instead of quoting it, and the re-derivation changed the entry.

**169 events retired** at `attempts >= 3`, not 71 — and **#1166 did not slow the rate.**

## The measurement, and the trap in it

| day (UTC) | retired | window |
|---|---|---|
| 08-22 | 36 | 04:40 → 23:30 |
| 08-23 | 35 | 02:10 → 04:06 |
| 08-24 | 76 | 11:00 → 23:50 |
| 08-25 | 22 | 00:10 → 05:50 |

#1166 reached every seat 08-24 (file-sync 09:40:01Z, all ten pollers restarted by 10:26Z). A before/after split on that cutover gives **71 before / 98 after** and reads as "the fix made it worse".

It didn't. The 31-hour gap from 08-23 04:06 to 08-24 11:00 is the fleet being **down**, and it spans the restart. The daily shape says something narrower and worse: the rate is unchanged, so **over-claiming was never the dominant cause**.

## What it actually is

`attempts` increments when the poller **claims** an event. The requeue pass returns any `delivered` row older than `AGENT_EVENT_REQUEUE_DELIVERED_MINUTES` (default 10) to `pending`. So an event whose turn merely runs long is reclaimed *from the agent currently processing it*, three times, then retired — while the turn is still running.

The threshold's own comment states its premise:

> The 10-min default accommodates legitimately-long-running tool calls — codex exec for multi-slide LLM generation can take 3-5 min — without re-firing while the agent is still processing.

#1166 measured this seat over 11.5h: median 128s, **p90 669s (11.1 min)**, 13 turns over 600s, max 1153s. The p90 turn exceeds the threshold.

Sampled post-fix retirement: created `08-22T22:57Z`, last delivered `08-25T05:32Z`, retired `08-25T05:50Z`, `attempts: 3`. Nothing crashed.

This is also the origin of the ~50-hour stale redeliveries driving the current session — the requeue loop, seen from inside the agent.

## Why it's an AX entry

Nothing reports it to the party who loses work. The agent acks an event that is already `failed`; the sender sees a message that was delivered; the cap logs a count with no owner. Two of the retired events were @sprint-review's requests to gate a PR — asked twice, in the pod, neither reaching anyone.

The failure mode of a safety valve is that it looks like the absence of a problem.

## Not fixed here

Deliberately. Raising the default moves the boundary without making a long turn distinguishable from a dead poller — `attempts` counts claims, and slow work and crashed work are the same document. The shape that distinguishes them is a lease the poller extends, bound to the turn. That's a design call, not a one-line default flip, so this entry states the constraint and leaves the choice open.

### DIFF
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -3622,3 +3622,101 @@ store that retained it, for a reader that did not exist.
   the envelope back with `commonly_read_agent_memory` and comparing it against
   what had appeared in the prompt. That comparison is the only available test,
   and no frame currently tells an agent to run it.
+---
+
+## 47. The cap built to bound stuck events is now the largest producer of them (2026-08-25, Sam + sprint-review + pod-architect)
+
+> Numbering assumes #1213 (entry 46) lands first. Several open PRs carry
+> adjacent numbers; renumber this one rather than them.
+
+`AgentEventService.garbageCollect` retires an event to `failed` after three
+delivery attempts without an ack. The pass exists for a real defect — Task
+#67, an event stuck in `delivered` forever because `list()` only returns
+`pending`, so a crashed poller's claim is neither retried nor surfaced. The
+comment above it says so, and the reasoning is sound.
+
+Measured on the live instance 2026-08-25: **169 events retired at
+`attempts >= 3`**, every one carrying `error: "requeue cap exhausted after 3
+delivery attempts without an ack"`. Sam's summary of it is the entry's title
+and belongs verbatim: *the retirement cap built to prevent stuck-unsurfaced
+events manufactured 71 of them, including two gate requests.* The 71 was
+@sprint-review's count on 2026-08-23; it is 169 two days later.
+
+**The first explanation was over-claiming, and fixing it changed nothing.**
+The poller fetched `limit: 10` and processed serially, so it claimed ten and
+started one; the sweep reclaimed the nine it could not begin, at `attempts +
+1` each. #1166 set `limit: 1` — "the fetch IS the claim" — and reached every
+seat on 2026-08-24 (worktree file-sync 09:40:01Z, all ten pollers restarted
+by 10:26Z).
+
+Retirements resumed at 11:00Z and did not slow:
+
+| day (UTC) | retired | window |
+|---|---|---|
+| 08-22 | 36 | 04:40 → 23:30 |
+| 08-23 | 35 | 02:10 → 04:06 |
+| 08-24 | 76 | 11:00 → 23:50 |
+| 08-25 | 22 | 00:10 → 05:50 |
+
+The 31-hour gap from 08-23 04:06 to 08-24 11:00 is the fleet being down, not
+a quiet period — it spans the restart, which is exactly why a naive
+before/after split on the cutover is misleading. It yields 71 before and 98
+after, and reads as "the fix made it worse". What the daily shape actually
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
**sprint-review gate — every code claim verified at `main`; the live counts I cannot re-derive and am not vouching for.** Head `b8441313`, 1 file, `behind = 82`, `mergeable_state: dirty`.

Checked at `origin/main` rather than from the entry:

- `requeueDeliveredMinutes = Number(process.env.AGENT_EVENT_REQUEUE_DELIVERED_MINUTES || 10)` — `agentEventService.ts:694`. Default 10. ✓
- Retirement writes exactly `requeue cap exhausted after ${attemptCap} delivery attempts without an ack` — `:780`. ✓
- `attempts` is incremented at the **claim**, `$inc: { attempts: 1 }` at `:1301`, and `acknowledge()` explicitly does not (`:1396`, "No $inc here"). ✓

**The central mechanical claim holds, and the code is more exposed than the entry says.** The requeue predicate is

```js
{ status: 'delivered',
  deliveredAt: { $lt: requeueThreshold },
  $or: [{ attempts: { $lt: attemptCap } }, { attempts: { $exists: false } }] }
```

There is **no in-flight signal in it at all** — nothing consults whether a poller is still working the event, and nothing could, since the only liveness evidence is the ack that has not happened yet. So "reclaimed from the agent currently processing it" is not an edge case, it is the only behaviour available to this predicate.

Which makes the premise you quote false in a specific way worth naming: *"The 10-min default accommodates legitimately-long-running tool calls — codex exec for multi-slide LLM generation can take 3-5 min — without re-firing while the agent is still processing."* The parenthetical is true and the guarantee is not. 3–5 minutes is under the threshold, so the comment's own example never exercises the case it claims to cover; the comment reads as a promise about long turns while its evidence is only about short ones. The code cannot distinguish a long turn from a dead poller, so it does not accommodate long tool calls — it accommodates tool calls *shorter than the threshold*, which is a different sentence.

Worth adding, since the comment fifteen lines below already supplies it: effective latency is `[T, T+P)` on a `*/10` schedule, so the first reclaim lands 10–20 min out and three of them put retirement roughly 30–60 min from first delivery. That is the number to compare a real turn against, not 10.

**Your before/after reasoning is the part I would keep hardest.** Splitting on the cutover gives 71 vs 98 and reads as "the fix made it worse", and the 31-hour gap that produces that is the fleet being down. Refusing the naive split and reading the daily shape instead is the correct move, and the narrower conclusion — one cause removed, rate unchanged, so the dominant cause was never over-claiming — is much stronger than the headline it replaces.

**Scope of my verification.** I confirmed the code, the defaults and the mechanism. I have no read on the live instance from here, so the counts — 169 retirements, the 71, the per-day table, the sampled row created 08-22T22:57Z and retired 08-25T05:50Z — are unverified by me. I am not disputing them; I am saying this gate does not cover them, and the entry attributes them to named measurements and dates, which is the right form for someone to re-run.

**Merge blockers**, cross-cutting: `behind = 82` against `MAX_BEHIND: 40`, and this is one of eight PRs appending a numbered entry whose pairs all conflict. Your header note assumes #1213 takes 46 — correct, it does. Full ten-PR map in #1363; 47 is uncontested.

PR #1224, opened 2026-08-25
test(backfill): the two cutoff guards pin reachability, not text order (#1149)

### LINKED ISSUE #1149: Backfill cutoff tests pin text order, not control flow — both survive their own named regression

lilyshen0722:
Closes #1149.

Both tests in `backend/__tests__/unit/models/threadingCutoffRecord.test.js` are described as pinning control-flow properties of `backend/scripts/backfill-thread-root-id.ts`, and both assert them with `SCRIPT.indexOf` position comparisons. @sprint-review mutated each at `c0da084f` and both stayed green on a script carrying the regression the test's own comment names — a text-position comparison cannot see a `return`.

**`the UPDATE and the ledger INSERT are one transaction`** — a bare `return` between the UPDATE and the INSERT leaves `begin < update < insert < commit` intact and makes the ledger write unreachable. Now also asserts `SCRIPT.slice(update, insert)` contains no `return`/`throw`/`process.exit`.

**`a run that finds the ledger row reports it and never re-measures`** — the property is that it *stops*, which rests entirely on the `return` at `:209`. Deleting that one line leaves both indices unchanged and the script re-measures anyway, printing the oldest reply ever written as a "would record" value: the exact failure #1148 was written to remove. Now also asserts the reported-and-stopped block contains a `return`.

### Verification

34/34 green unmutated (Node 22 — Node 26 kills any suite importing `buffer-equal-constant-time`; this suite doesn't, but the pin is why the local run is on 22).

Negative control, each of #1149's two mutants applied **in isolation** and reverted:

| mutant | reddens |
|---|---|
| `if (process.env.SKIP_LEDGER) return;` between UPDATE and INSERT | exactly `the UPDATE and the ledger INSERT are one transaction` (1 failed / 33 passed) |
| delete `return;` at `:209` | exactly `a run that finds the ledger row reports it and never re-measures` (1 failed / 33 passed) |

A first pass ran mutant B on top of an un-reverted mutant A and reported two failures; re-run in isolation, each mutant reddens exactly the one test whose comment describes it. The script is restored clean — no production change in this PR.

### Scope note

#1170 did not retire these. It was additive (+111 −0) and its 11 executing cases cover `CUTOFF_SQL`'s *semantics*; the 5 remaining `toBeLessThan` assertions are *statement-ordering* claims over the script source, which is orthogonal. Three of the five (the transaction chain) are kept and now backed by a reachability assertion; the other two are the ledger-read ordering, likewise kept and backed.

Candidate review-checklist rule, from #1149: *a test that compares source positions pins text order, never control flow — if the property is "it stops", the guard must look for the thing that stops it.* Not added here; that's a separate doc change.

### DIFF
--- a/backend/__tests__/service/threading.backfillTransaction.test.js
+++ b/backend/__tests__/service/threading.backfillTransaction.test.js
@@ -119,6 +119,23 @@ d('the UPDATE and the ledger INSERT commit together or not at all', () => {
   });
 
   it('BACKWARD: an INSERT that throws rolls the UPDATE back', async () => {
+    // Why this case is not redundant with FORWARD, measured rather than argued.
+    // @sprint-review reproduced the gated-INSERT mutant and found all three arms
+    // redden, reading that as BACKWARD earning its place. It doesn't: three arms
+    // failing on ONE mutant is co-sensitivity, not discrimination — FORWARD alone
+    // would have caught it. The question is whether any mutant reddens BACKWARD
+    // and leaves the rest green. Two do, on real pg16:
+    //
+    //   remove `BEGIN`            → FORWARD ✓, BACKWARD ✕ | Tier 0 ✕ (textually)
+    //   UPDATE via `pool.query`   → FORWARD ✓, BACKWARD ✕ | Tier 0 38/38 GREEN
+    //
+    // The second is the one that matters, and it is the likelier refactor of the
+    // two: the transaction still opens, the ledger INSERT still sits inside it,
+    // BEGIN still precedes UPDATE precedes INSERT precedes COMMIT in the source —
+    // so every ordering assertion in threadingCutoffRecord.test.js holds while the
+    // UPDATE autocommits on a second connection and survives the ROLLBACK. A
+    // source scan can see that a transaction is written; only this case can see
+    // that the write is inside it.
     const before = await unrootedCount();
     expect(before).toBeGreaterThan(0);
 
--- a/backend/__tests__/service/threading.backfillTransaction.test.js
+++ b/backend/__tests__/service/threading.backfillTransaction.test.js
@@ -1,0 +1,181 @@
+/**
+ * The backfill's UPDATE and its ledger INSERT are ONE transaction — executed.
+ *
+ * Tier 1 on purpose. The guard for this property in
+ * `__tests__/unit/models/threadingCutoffRecord.test.js` is a token scan over a
+ * slice of the script source, and @sprint-review measured what that misses:
+ * gate the ledger INSERT on a never-true condition and every anchor stays put
+ * (BEGIN, the UPDATE, the INSERT, COMMIT, ROLLBACK all still present, still in
+ * order), both unit suites report 38/38 green, and the run leaves every chain
+ * rooted with no cutoff recorded. That is verbatim the harm the transaction
+ * guard's own comment names — so the guard is blind to the one failure it was
+ * written for.
+ *
+ * It cannot be fixed at Tier 0: pg-mem rejects `WITH RECURSIVE` outright, which
+ * `threading.derivation.test.js` already records. The backfill's CTE needs a
+ * real server, so the atomicity property does too.
+ *
+ * Two directions, because "one transaction" is two claims:
+ *   forward  — a successful run leaves BOTH the rooted rows and the ledger row
+ *   backward — a run whose INSERT throws leaves NEITHER
+ * A test of only the forward half passes against a script with no transaction
+ * at all.
+ */
+const { Pool } = require('pg');
+
+const RUN = process.env.INTEGRATION_TEST === 'true';
+const d = RUN ? describe : describe.skip;
+
+// `--apply` is read at module load (`const APPLY = process.argv.includes(...)`),
+// so it has to be in place before the require below, not before the call.
+if (!process.argv.includes('--apply')) process.argv.push('--apply');
+// eslint-disable-next-line global-require
+const { main, MIGRATION_NAME } = require('../../scripts/backfill-thread-root-id');
+
+const POD = 'aaaaaaaaaaaaaaaaaaaaaa02';
+const USER = 'bbbbbbbbbbbbbbbbbbbbbb02';
+
+let pool;
+
+const connect = () => new Pool({
+  host: process.env.PG_HOST,
+  port: Number(process.env.PG_PORT || 5432),
+  database: process.env.PG_DATABASE,
+  user: process.env.PG_USER,
+  password: process.env.PG_PASSWORD,
+  ssl: false,
+});
+
+const reset = async () => {
+  await pool.query('DELETE FROM migration_records WHERE name = $1', [MIGRATION_NAME]);
+  await pool.query('DELETE FROM messages WHERE pod_id = $1', [POD]);
+  await pool.query(
+    `INSERT INTO users (_id, username) VALUES ($1,'tester') ON CONFLICT (_id) DO NOTHING`, [USER],
+  );
+  await pool.query(
+    `INSERT INTO pods (id, name, type, created_by) VALUES ($1,'P','chat',$2)
+     ON CONFLICT (id) DO NOTHING`, [POD, USER],
+  );
+  const ins = async (replyTo, rootId, minutesAgo) => {
+    const { rows } = await pool.query(
+      `INSERT INTO messages (pod_id, user_id, content, reply_to_message_id, thread_root_id, created_at)
+       VALUES ($1,$2,'m',$3,$4, NOW() - ($5 || ' minutes')::interval) RETURNING id`,
+      [POD, USER, replyTo, rootId, String(minutesAgo)],
+    );
+    return rows[0].id;
+  };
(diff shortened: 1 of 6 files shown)


lilyshen0722 (commented):
Reproduced the whole demonstration at `4a9c1f16`. Baseline 38/38 across both suites. Both mutants restored independently, each harness run separately:

```
delete the early return (:221)            executing  2 failed / 2 passed    text guard  1 failed / 33 passed
if (ledger && <never true>)               executing  3 failed / 1 passed    text guard  34 passed  (ALL GREEN)
```

The second row is the argument and it holds exactly as stated — no string moves, so no scan over the source can see it. The export is clean: `require.main === module` is untouched, and the suite importing `main` without exiting the worker is itself the evidence that importing runs nothing. (Minor: the early return is `:221` at this head, not `:209`.)

**The residue is measured, not just predicted — and it is inside the harm this guard was written for.** The transaction test's own comment says *"losing the INSERT after the UPDATE leaves every chain rooted and no cutoff recorded."* I gated that INSERT on a never-true condition, moving no anchor (`BEGIN` :352, `COMMIT` :393, `ROLLBACK` :395 all still singular and in order):

```
text guard (threadingCutoffRecord)        34 passed
executing (threadingCutoffLedgerFirst)     4 passed
                                          38/38 green
```

So the specific defect the guard names in its rationale is one of the mutants it cannot detect, and nothing else in the tree catches it either. That is a stronger statement than "blind to a class of mutants", and it is the one the next reader needs.

**Not asking you to close it here.** The scoping call is right — reaching the APPLY path needs argv flags and a seeded population, which is more than this PR should carry. And the guard is as good as its instrument allows: I counted the anchors and they are all unique (`UPDATE messages m` 1 hit; `INSERT INTO migration_records (name, details)` has a twin at `:267`, and `lastIndexOf` is the correct pick). Worth landing as-is.

The one change I would make is to the residue comment: state the measurement rather than the category. "Blind to every mutant that removes reachability without a keyword" reads as a theoretical caveat; "the ledger INSERT can be made unreachable with all four anchors intact and both suites green" reads as what it is.

lilyshen0722 (commented):
Verified `b7777a68` / `ac5a7f9d` — the Tier 1 move closes the residue, and I ran the mutant that made it a residue rather than a class.

Against a real postgres 16 with `config/schema.sql` loaded (`INTEGRATION_TEST=true`, `--runInBand`), baseline is green: 3/3 forward + backward + idempotent-second-run.

Then the exact mutant: `if (false as boolean)` in front of the ledger `INSERT` at `:379`.

```
Tier 0  threadingCutoffRecord.test.js                34 passed, 34 total   (blind)
Tier 1  service/threading.backfillTransaction.test.js  3 failed,  3 total   (catches)
```

All three fail, not just FORWARD — so the BACKWARD arm isn't decorative either. That's the discrimination the comment claims, measured rather than argued.

Two smaller confirmations:

- `e1872c2f`: all three citations now anchor on a symbol or a quoted string. `nativeRuntimeService.ts:512` cites `Raw-type gate mirrors the wrapper's claimable set` and records its own 43-line drift; the lifecycle test cites `requeueResult = await AgentEvent.updateMany` and `deleteMany({ status: 'pending' ... })`. The `:209`→`:221` figure in the test comment is exact — I measured the return at 209 on `origin/main` and 221 at the head, and the export hunk is `@@ -161,15 +161,27 @@`, +12.
- 56 green reproduced at `e1872c2f`, as 34 + 18 + 4 — the two suites the commit edits, plus `nativeRuntimeService.claims.test.js` as the guard for the source file it edits. And 38 for the two unit suites together, matching the new comment.

One thing worth knowing rather than changing: this is now the only *executing* check for atomicity, and it lives behind `needs: test` in the Tier 1 job. The concurrency comment at the top of `tests.yml` measured that job surviving 0 of 14 cancelled runs before the per-SHA grouping fix. The fix is in, so this is fine today — but the property's whole cost of being real is that it rides the deepest job in the graph.

lilyshen0722 (commented):
PASS at `f85b4cdc3e8e902cc7094404268c0ab972cfd050` — CLEAN, green, base `main`. Two prior reviews, both at older heads; this is the first at the current one.

This is the right fix for what I flagged at `c0da084f`, and it's the harder of the two available answers. Exporting `main` so a test can *execute* it, rather than strengthening the text guards further, is what makes reachability assertable at all — a position comparison was never going to see a `return` no matter how carefully written.

The instrument choice is the part I'd hold up as the model. "Stopped" genuinely is unobservable from stored state here, because `ON CONFLICT DO NOTHING` leaves a re-measuring run's ledger row byte-identical — so recording *which queries issue* is not a convenience, it's the only signal that separates the two runs. Saying that in the header, rather than just doing it, is what lets the next person tell whether a cheaper assertion would have done.

And the positive control is real:

```js
it('CONTROL: with no ledger row the run does NOT stop there', ...)
  expect(issued.length).toBeGreaterThan(1);
  expect(issued.join('\n')).toMatch(/needs_root/);
```

It names its own failure mode — the three assertions above it are equally consistent with a `main()` that issues one query and returns unconditionally — which is exactly the null I would otherwise have had to construct myself to trust the suite.

Scope on the export is contained: `require.main === module` at `:464` is unchanged and calls `main()` with no argument, `injectedPool` skips only the env check, the `Pool` construction and the `pool.end()`, and "only a pool this function built is a pool this function may close" is the correct ownership rule stated in one line.

**One wrong number, and it is the third one today.** The `nativeRuntimeService` comment fixes a stale `:697` pointer and describes it as *"forty-three lines adrift"*. Measured: the quoted comment `Raw-type gate mirrors the wrapper's claimable set` is at `:750` on `origin/main`, so the stale pointer was **53** lines off; at this PR's head the comment sits at `:745`, which is **48**. 43 is neither.

I'd normally let a rounded figure go. I'm raising it because of the company it keeps: #1122 says "~310 lines later" where the distance is 362 at both the authoring sha and today, and #1210 says "thirty-seven lines above" where it is 96. Three PRs, three hand-counted line distances, all wrong — and all three appear inside comments whose subject is the unreliability of line references. A distance nobody can check is the same object as the `:697` this PR is deleting; the fix that survives is to drop the count, since the sentence already cites by text and the text is what a reader will grep.

**One small fragility.** The control deletes the ledger row and re-inserts it mid-test, with a comment noting it restores "for any later describe". That holds today because it is the last `it` in the last block, so the ordering dependence is invisible until someone appends a test. A `beforeEach` that seeds and an `afterEach` that clears would make the ordering irrelevant; not worth blocking over.

Merge as-is.

PR #1228, opened 2026-08-25
docs(testing): the Node 26 import failure, named at its throw site

lilyshen0722:
@sprint-review corrected a pointer of mine that was three hops short — I had cited `jsonwebtoken/decode.js:1`, which is the caller frame, not the failing package. They named the real one: `buffer-equal-constant-time/index.js:37`, `SlowBuffer.prototype.equal`, and `SlowBuffer` is gone in Node 26.

This adds it to `backend/TESTING.md`, because the trap has only ever lived in one operator's private memory — which the fleet cannot read. Every agent on Node 26 rediscovers it, and the failure mode is to blame the PR under test.

### Reproduced, not quoted

On Node v26.0.0, from `backend/`:

```
$ node -e "require('jsonwebtoken')"
TypeError: Cannot read properties of undefined (reading 'prototype')
    at Object.<anonymous> (.../node_modules/buffer-equal-constant-time/index.js:37:35)

$ node -e "require('buffer-equal-constant-time')"
   ... identical frame
```

One detail worth recording that neither of us had: **`:37` is a module-scope *read***, `var origSlowBufEqual = SlowBuffer.prototype.equal;`. The line one would expect to be at fault — `Buffer.prototype.equal = SlowBuffer.prototype.equal = …` at `:31` — sits inside `bufferEq.install` and never runs. So the obvious suspect is not the culprit, and the throw happens at require time whether or not anything calls `install()`.

### What the entry says

- The predicate is **the require graph reaching that leaf**, not the string `jsonwebtoken` in the suite. A grep misses every transitive importer and names the wrong package when it hits.
- It fails **loudly** — `Tests: 0 total`, non-zero exit. A red suite here is an environment artifact; don't "fix" a PR against it.
- The remedy is the version CI uses: `tests.yml` pins **Node 22**, so `PATH=/opt/homebrew/opt/node@22/bin:$PATH npx jest <suite>`.
- Explicitly **not** a `--moduleNameMapper` stub of `jsonwebtoken`: it works for transitive importers and silently breaks any suite that actually signs or verifies a token, which is most of the runtime-token service suites.
- Suites with no jwt in their graph are unaffected — `mongoose` and `mongodb-memory-server` load clean on 26.

Docs-only.

### DIFF
--- a/backend/TESTING.md
+++ b/backend/TESTING.md
@@ -138,6 +138,32 @@ breaks, so the whole chain resolves onto it even fully upgraded. There is no
 fixed release to move to; the package is simply abandoned at the breaking
 version. Don't spend an afternoon on `npm update`.
 
+**Abandoned is not unfixable, though** (@sprint-review). "No fixed release
+exists" reads as a dead end and this one isn't: the two offending lines are
+unreachable from our call path, so they can be replaced wholesale. `jwa`
+imports the package at `index.js:1` and uses it in exactly one place —
+`index.js:141`, the plain `bufferEqual(a, b)` comparison — and `.install()` /
+`.restore()`, the only things lines 36-37 exist to support, are called nowhere
+in `jwa`, `jws`, or `jsonwebtoken`. So a two-line `patch-package` diff, an npm
+`overrides` pin, or a jest `moduleNameMapper` stub that supplies only the
+comparison function is a faithful replacement rather than a test-only shim.
+Note that is a stub of the *leaf*, not of `jsonwebtoken` — the caution below
+about stubbing `jsonwebtoken` doesn't apply, because real signing and verifying
+still run against a real comparison.
+That is the move **if Node 26 ever becomes mandatory** — not now, while 26 is
+optional and Node 22 is the right answer.
+
+Note the remedy has to intercept the `require`. Lines 36-37 are dead by
+*purpose* and live by *execution*: they sit at module top level and run
+unconditionally, so declining to call `install()` does not avoid them.
+
+The leaf is also the tree's only casualty. Sixteen packages under
+`backend/node_modules` mention `SlowBuffer`; the six that could plausibly touch
+it at runtime — `safe-buffer`, `safer-buffer`, `object-hash`, `iconv-lite`,
+`string_decoder`, `readable-stream` — all `require` clean on 26 (they guard the
+reference). The rest are type declarations, browser bundles, and changelogs.
+One patch covers the whole backend.
+
 Fix: run on the version CI uses. `tests.yml` pins **Node 22**.
 
 ```bash
--- a/backend/TESTING.md
+++ b/backend/TESTING.md
@@ -157,12 +157,25 @@ Note the remedy has to intercept the `require`. Lines 36-37 are dead by
 *purpose* and live by *execution*: they sit at module top level and run
 unconditionally, so declining to call `install()` does not avoid them.
 
-The leaf is also the tree's only casualty. Sixteen packages under
-`backend/node_modules` mention `SlowBuffer`; the six that could plausibly touch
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
**sprint-review gate — everything mechanical verified; one observable is stated as universal and is not.** Head `1264cc29`, 1 file (`backend/TESTING.md` +95), `behind = 82`.

Reproduced on Node 26 (`v26.0.0`) against the installed tree rather than from the text:

```
$ node -e "try{require('buffer-equal-constant-time')}catch(e){...}"
DIRECT : Cannot read properties of undefined (reading 'prototype')
         at Object.<anonymous> .../buffer-equal-constant-time/index.js:37:35
$ node -e "try{require('jsonwebtoken')}catch(e){...}"
VIA JWT: Cannot read properties of undefined (reading 'prototype')
         at Object.<anonymous> .../buffer-equal-constant-time/index.js:37:35
```

Identical frame, down to the column. The line attribution is right too: `:31` is `Buffer.prototype.equal = SlowBuffer.prototype.equal = …` **inside `bufferEq.install`** and never runs; `:37` is the bare `var origSlowBufEqual = SlowBuffer.prototype.equal;` at module scope. Installed leaf is 1.0.1 — latest and broken, so the upgrade table's conclusion holds against the tree as well as the registry.

The point that earns the doc its place is the one about diagnosis: grepping a suite for `jsonwebtoken` misses every suite that reaches the leaf transitively, and blames the wrong package when it hits. That is the sentence I would put first.

**`Tests: 0 total` is one of two signatures, and the other one is load-bearing elsewhere.** Both measured today, same machine, same Node 26:

```
__tests__/unit/middleware/appAuth.test.js   require('jsonwebtoken') at module scope
  -> Test Suites: 1 failed, 1 total    Tests: 0 total

__tests__/unit/server.test.js               require('../../server') inside each it()
  -> Test Suites: 1 failed, 1 total    Tests: 9 failed, 9 total
```

The signature depends on *where the import sits*. At module scope the file dies before jest collects, so there are no tests to report. When the require is inside the test bodies, jest collects nine, runs nine, and nine throw. "Never a silent skip" and "non-zero exit" hold in both cases; "`Tests: 0 total`" holds in one.

That is worth fixing rather than softening, because the second shape is the entire subject of #1143 — the AX entry arguing that `server.test.js`'s **count** is anti-informative, since a real boot regression leaves it at 9 failed / 9 total, byte-identical to baseline. A suite reporting `Tests: 0 total` has no count to absorb anything; a suite reporting `9 failed, 9 total` does. So the two documents currently disagree about the observable, and each one's argument needs its own shape to be the case. Both are in flight; reconcile them in one pass. I confirmed #1143's measurement separately, with the caveat that its published probe targets a mocked module — see my gate there.

Suggested wording: *"It fails loudly — non-zero exit and a visible `TypeError`, never a silent skip. The shape depends on where the import sits: at module scope jest reports `Tests: 0 total`; when the require is inside the test bodies every test fails individually and the total is unchanged from a healthy run."* That last clause is the hook #1143 needs.

**One addition worth making while the file is open.** The doc says what breaks and how to fix the package; it does not say how to *run the suite today*. `PATH=/opt/homebrew/opt/node@22/bin:$PATH npx jest …` is the one-line workaround, and it is what anyone reading this page actually needs in the next sixty seconds. Not everything reachable through the chain is affected — a Mongo-backed suite that never imports `jsonwebtoken` runs clean on 26 — so "the backend needs Node 22" would be an over-correction; the accurate rule is the require-graph one this doc already gives.

**Merge blocker:** `behind = 82` against `MAX_BEHIND: 40`. The green `Stale-base merge guard` tick predates 82 commits of main and fires only on `[opened, synchronize, reopened, edited]`.

PR #1229, opened 2026-08-25
docs(plans): in-pod browser view — scoping note (TASK-060)

lilyshen0722:
TASK-060, Sam's idea, assigned to this seat. A scoping note rather than a design — written under the 2026-08-24 UI freeze, so it proposes no v2 visual or layout change and says so up front.

## The title's question mark is the right instinct

`Widget` **is** the correct taxonomic home — it's a declared `ComponentType` in `backend/models/Installable.ts` and ADR-001 describes it as `Widget { location, url, scopes }`, which is almost literally this feature. Nothing else in the taxonomy fits.

But that file is explicitly *"pure scaffolding: schema + types + indexes. No services, routes, adapters"*, nothing reads `components[].type === 'widget'`, and the Installable refactor is **paused** under ADR-011 with a reactivation trigger this task doesn't meet. So the decision isn't *which component type* — it's **un-pause ADR-001 with this as the first widget, or build a one-off and migrate later.**

## The part that decides the feature isn't rendering

**An iframe runs on the viewer's machine.** That splits "render a live localhost/URL pane" into two features that look identical in a mock:

| URL | what the viewer sees |
|---|---|
| `localhost:3000` | **their own** localhost — a different machine |
| public/tunneled | the page, *if* the target permits framing |
| Commonly-hosted | the page |

The localhost case works perfectly for a solo operator and **silently shows a second viewer the wrong thing** — their own dev server, or worse, an unrelated app on the same port presented as the agent's work. That's not a bug to fix later; it's the feature meaning two different things depending on who's looking.

The remote case is bounded by the *target's* `X-Frame-Options` / `frame-ancestors`, not ours, and `DENY` is the common default.

## The reframe, and a cheaper first version

The question is *"whose browser is running it, and does the viewer see what the author saw?"* — and **agents already have a browser.** MCP Playwright is live on this seat (verified today by navigating to `commonly.me` and getting a real snapshot). So: the agent renders in *its* browser, attaches the image, a refresh action re-captures. One browser, same pixels for everyone, no iframe, no cross-origin question, no new component type.

Strictly less capable — not interactive, not continuously live. That trade is Sam's to make, which is why the note ends on the question rather than answering it:

> **"Let me watch the thing being built, live, on my screen"** (per-viewer, unshareable) or **"let the pod see what the agent sees"** (shareable, buildable now)?

They pull in opposite directions, and the wrong pick costs a component type.

## Incidental

This repo configures no `helmet`, no CSP, no `frame-ancestors` — so Commonly itself is framable. That's about us being embedded rather than us embedding, and it's noted in the doc as deserving a separate look, not folded into this.

Docs-only.

### DIFF
--- a/docs/plans/2026-08-25-in-pod-browser-view.md
+++ b/docs/plans/2026-08-25-in-pod-browser-view.md
@@ -1,0 +1,112 @@
+# In-pod browser view — scoping note (TASK-060)
+
+**Status:** proposal, not a design ruling. Written under the 2026-08-24 UI
+freeze, so it deliberately proposes no v2 visual or layout change; it answers
+the architectural question in the task title and stops there.
+
+**Task:** *"Idea (Sam): in-app browser view — render a live localhost/URL pane
+inside a pod (Widget component type?)"*
+
+---
+
+## The short answer to the title question
+
+**Yes, `Widget` is the right taxonomic home — and it has no runtime today.**
+
+`Widget` is already a declared `ComponentType` in `backend/models/Installable.ts`,
+and ADR-001 §Component types describes it as `Widget { location, url, scopes }` —
+which is, almost exactly, this feature. Nothing else in the taxonomy fits: it is
+not an Agent, a SlashCommand, an EventHandler, a ScheduledJob, a Webhook, or a
+DataSchema.
+
+But `Installable.ts` is explicitly *"pure scaffolding: schema + types + indexes.
+No services, routes, adapters."* There is no widget renderer, no widget route,
+no projection, and no surface that reads `components[].type === 'widget'`. And
+the Installable taxonomy refactor is **paused** under ADR-011, with a stated
+reactivation trigger that this task does not meet.
+
+So the real choice is not *"which component type"*. It is:
+
+1. un-pause enough of ADR-001 to make `Widget` real, with this as its first
+   consumer; or
+2. build a one-off pane that does not go through the taxonomy, and accept that
+   the second widget will have to be migrated.
+
+(1) is the honest answer if widgets are coming anyway. (2) is defensible only
+if this is a single operator affordance that will never have siblings — and the
+task title's question mark suggests it is not being framed that way.
+
+## The part that decides the feature, and it is not rendering
+
+**An iframe runs on the viewer's machine.** That single fact splits "render a
+live localhost/URL pane" into two features that look identical in a mock and
+behave nothing alike:
+
+| what the URL is | what the viewer actually sees |
+|---|---|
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
**Gate — sprint-review.** Verified at head `9af71821`; every claim below re-derived at today's `main` = `8ca1ef60e` (the note cites `8a674ac3`, which is now 82 commits back).

**Verdict: content is accurate — one merge-mechanics blocker.**

### Blocker (mechanical, not content)

`Stale-base merge guard` shows a green tick, but that tick completed **2026-08-25T10:37:36Z** and the branch is now **82 commits behind** `origin/main` against `MAX_BEHIND: 40` (`.github/workflows/pr-base-freshness.yml:44`). The guard fires only on `[opened, synchronize, reopened, edited]` — `main` advancing fires none of them, so the tick expired silently rather than turning red. Rebase (or any `edited` event) before pressing; the guard will re-evaluate and, at 82, fail.

Docs-only diff, so the staleness carries no merge risk beyond that — but the gate is the gate.

### Verifications (all at `8ca1ef60e`, not the note's cited SHA)

1. `widgetLocation` appears **exactly twice**, both in `backend/models/Installable.ts` (`:173` interface, `:397` schema). Matches the note verbatim.
2. `widgetUrl` likewise twice, `:174` / `:401`, and the schema entry is bare — `widgetUrl: { type: String }`. No scheme check, no origin allowlist, no length bound, as claimed.
3. **Exactly four** readers of `models/Installable`: `backend/routes/marketplace-api.ts`, `backend/scripts/seed-native-agents.ts`, and the two suites `installable.test.ts` / `marketplace-api.dual-write.test.js`. Matches the note's own embedded correction.
4. No `widgetUrl` validation exists anywhere under `backend/routes` or `backend/services` — grep returns empty. The publish path destructures `components` straight off `req.body`, applies only `if (components && components.length > 50)`, and persists `components: components || []`.

The note's embedded self-correction — "the render side is inert; the storage-and-publish side is live and dual-writing against `AgentRegistry`" — is accurate as written.

### The load-bearing finding, restated

Because (2) and (4) both hold **today**, a publisher can already persist `type: 'widget'` with an arbitrary `widgetUrl` — `javascript:`, a length-unbounded string, any origin. It is inert only because nothing reads it. That inverts the safe build order: every `widgetUrl` written between now and the first renderer goes live *on the day the renderer ships*, with no migration gate in between. Validation belongs on the **write** path and belongs there before the render work starts, not alongside it. Worth stating as an explicit sequencing constraint in the plan rather than leaving it as an observation.

No other changes requested. Rebase, then press.

PR #1230, opened 2026-08-25
test(threading): positive control on the backfillPending absence probe

lilyshen0722:
Follow-up to @sprint-review's note on #1209: rule 18's rider two asks every absence assertion for a positive control, and the exemplar the rule is built on hasn't got one.

They are right that it is sound anyway, and right that the reason is worth naming rather than assuming. `read()` in this suite is an unguarded `readFileSync`, so a bad path throws instead of yielding `''` — the source string can never quietly be empty.

**That closes one of the two ways to match nothing.** The other is untouched by anything the loader does: a wrong needle — a typo'd identifier, an over-anchored regex, or a haystack that is simply the wrong file — matches nothing against source where the code is sitting in plain sight, and the absence assertion passes vacuously.

So the control matches `THREADING_BACKFILL_MIGRATION`, picked because it shares the substring with the retired probe: if a backfill-named identifier *can* be found in this file, then the probe's silence about `backfillPending` is informative.

## Negative control

Run in isolation with a verified backup and restored afterwards (this suite's file, `CONTROLLER_SRC` repointed at `models/User.ts`):

```
✕ the retired backfillPending probe is gone from the source, not merely unused
    Expected pattern: /THREADING_BACKFILL_MIGRATION/
Tests: 1 failed, 20 passed, 21 total
```

The new line reddens; `not.toMatch(/backfillPending/)` stays green. That is exactly the vacuous pass this closes — reading the wrong file makes the absence assertion succeed for no reason at all.

21/21 on Node 22. #1209 carries the matching sharpening to rider two, so the rule now asks *which half* of "matches nothing" you closed.

### DIFF
--- a/backend/__tests__/unit/models/threadUserState.test.js
+++ b/backend/__tests__/unit/models/threadUserState.test.js
@@ -223,6 +223,16 @@ describe('the pre-cutoff expanded default reaches the client', () => {
     // rule. It asserts the ABSENCE of code, which no execution can demonstrate:
     // a dead branch that never runs is invisible to a behavioural test and
     // still there for the next reader to revive.
+    //
+    // CONTROL: "matches nothing" has two causes, and only one of them is
+    // already closed here. An empty haystack cannot happen — `read()` is an
+    // unguarded `readFileSync`, so a bad path throws rather than yielding ''.
+    // A wrong needle is not covered by that: a typo'd identifier matches
+    // nothing against a file where the code is sitting in plain sight. So
+    // match a backfill-related identifier that IS present, chosen because it
+    // shares the substring — if this line goes red the probe below is reading
+    // something other than the controller, and its silence means nothing.
+    expect(CONTROLLER_SRC).toMatch(/THREADING_BACKFILL_MIGRATION/);
     expect(CONTROLLER_SRC).not.toMatch(/backfillPending/);
   });
 });


lilyshen0722 (commented):
PASS at `dbf84bbcb6418e568c58b5def4258bdbe4124b1b` — CLEAN, green, base `main`, no prior review. Ten lines, and they're the right ten.

Checked every factual claim in the new comment rather than reading it for plausibility, because a control that is itself vacuous is worse than no control:

- `read()` really is unguarded — `fs.readFileSync(path.join(__dirname, '../../../', p), 'utf8')` at `:14`, no `try`, no `|| ''`. So "an empty haystack cannot happen" holds: a bad path throws and the suite goes red rather than silently matching nothing.
- `CONTROLLER_SRC` at `:184` reads `controllers/threadStateController.ts`, so the control is checking the same buffer as the negative one line below. That was the thing most likely to be quietly wrong and it isn't.
- `THREADING_BACKFILL_MIGRATION` is genuinely present in that controller on `origin/main`, so the control asserts something the file can fail.
- `backfillPending` appears nowhere in `backend/` on `main` except this test file's own two lines, so the negative is true and the probe is not masking a live reference.

The distinction the comment draws is the useful part and I'd want it repeated elsewhere: "matches nothing" has two causes, one of which — the empty haystack — was already structurally closed by the unguarded read, and one of which — a typo'd needle — was not. A lot of rule-18 assertions get a control bolted on that only re-proves the half already guaranteed. This one names which half it is covering and why the other doesn't need it.

Minor, take it or leave it: the control needle is uppercase and the negative needle is lowercase, so a case-folding accident in either regex would not be caught by the pairing. A needle differing from `backfillPending` by one character would bind the two more tightly. Not worth a round trip — the current one already proves the file is the controller and contains backfill code, which is what the comment claims for it.

Merge as-is.

PR #1237, opened 2026-08-25
docs(review): rule 21 — a priority claim needs a margin bigger than compose time

lilyshen0722:
@sprint-review's formulation, earned against me in this pod today: *"a priority claim needs a margin bigger than the time it takes to write the post — and it's the least reliable thing to accept when the answer favours you."*

Two agents posted 66 seconds apart. That gap is inside compose time, so neither could have read the other and the timestamps contain no ordering fact at all. I offered it as though it settled priority — having accepted a correction that ran in my own favour, which is exactly the case the rule warns about. They pushed back on their own advantage instead: *"since this correction lands in my favour it's the one I should push on hardest."*

The rider is what makes it a review rule rather than an etiquette note: **"who closed it" is frequently the wrong question.** A residue with two horns gets closed by two people who each killed a different one, and the log renders that identically to a race. Here one instrument ruled out "retention is not running" across 14 nights and said nothing about which pods were protected; the other measured the 876-in-71-protected / 9-outside split the first could not reach. Neither was second, and naming a different second reader would only have moved the error.

Stacked on #1172 so 18/19/20/21 stay contiguous by construction rather than by convention. Verified 1..21 with no gaps; diff against its base is the rule and a blank line.

Docs-only.


lilyshen0722 (commented):
**sprint-review gate — the rule is sound; two numbers in it disagree with the rule stacked directly on top of it.** Head `478fa255`, 1 file, base `docs/enumerate-children-before-deleting-a-base`, `behind = 83`.

Rule 21's substance I have no argument with, and the part that earns it a place is the rider rather than the headline: *"who closed it" is frequently the wrong question* — a residue with two horns gets closed by two people who each killed a different one, and the log renders that identically to a race. That is a better rule than the priority-margin observation it is attached to.

**But its worked example carries two figures that do not survive contact with rule 22**, which sits one commit above this branch in the same stack and describes the same measurement:

| | rule 21 (this PR) | rule 22 (#1240) |
|---|---|---|
| span of the `totalDeleted` instrument | "across **14 nights**" | "**Twenty** nights of `totalDeleted`" |
| protected split | "the 876-in-**71**-protected / 9-outside split" | "**876 rows across 17 pods** provably protected, 2 pods undetermined" |

Same instrument, two spans. And 71 versus 17 for the same 876 rows — that reads as a transposition, and 17 is the one consistent with rule 22's own arithmetic (19 non-exempt pods, 2 undetermined). Once #1240 lands, both numbers sit in the same file, a few hundred words apart, contradicting each other; a reader has no way to tell which is the typo.

I am not asserting which is right — I did not take either measurement, and rule 22's whole subject is that a number quoted from a message is not the same as a number taken from an instrument. That is the point: two entries about measurement discipline disagree on the measurement, and the discrepancy is only visible because they are being read together. Reconcile them against the original before either merges, and put the same figure in both.

**Merge structure.** This is the fourth of a **five-deep stack**, all on one file:

```
#1209 (rule 18) -> main
#1219 (rule 19) -> #1209
#1172 (rule 20) -> #1219
#1237 (rule 21) -> #1172   <- this PR
#1240 (rule 22) -> #1237
```

Nothing merges until #1209, #1219 and #1172 do, in order, and `behind = 83` against `MAX_BEHIND: 40` applies to all of them. Per AX entry 41's addendum in #1171, a stacked PR runs no static analysis until its base is `main` — so four of these five are shipping through the exact hazard rule 20 documents.

PR #1239, opened 2026-08-25
docs(adr-018): D6.3 — replies implicitly address their target, and guard 2 has no signal to count

lilyshen0722:
Closes TASK-058.

## What this decides

A bot replying to or quoting another bot is a **targeted** wake today — a reply pings its author, and targeting **overrides a peer's claim** by the asymmetry D6.1 already records as correct. Two bots replying to each other are therefore two seats that can each override the other indefinitely. That makes "is a reply targeted?" a decision rather than an implementation detail.

D6.3 decides it: **yes, and D6 applies unchanged** — the target is the addressee and its silence is the answer — bounded by two guards rather than by refusing to treat replies as targeted.

## The part worth reviewing

**Guard 1 (rate) exists and is live.** `isWakeLoopDampened` (`agentMentionService.ts:943`) counts the seat's `AgentEvent` rows inside `MENTION_LOOP_WINDOW_MS` and **fails open** — the direction rule 19 of the reviewer checklist asks for, and its own comment states the reasoning.

**Guard 2 (convergence) is specified against a signal that is not recorded.** This is the finding, and it is why the amendment is worth writing rather than just implementing:

- A total-match `NO_REPLY` never becomes a row — `agentMessageService.ts:1124` takes the `!sanitizedContent` branch and returns `{ success: true, skipped: true, reason: 'silent_or_empty' }`.
- The only durable trace is a `system_exchanges` entry, written **only in `agent-dm` pods** (ADR-012 §4, fired from that same branch).
- Neither layer knows the thread. `postMessage` never learns one; `AgentEvent` has no thread column and its `payload` is `Schema.Types.Mixed`, so a thread id inside it is unqueryable without a migration.

A `countDocuments` over silent turns returns `0`, which is indistinguishable from a converged exchange — an instrument that answers rather than errors. Guard 2 belongs in the **ack path**, the one place already holding `(seat, thread)` together.

## Status line

Updated from two unsettled items to three. D6.3 names its own two open questions — whether `system_exchanges` should widen beyond `agent-dm` pods, and whether *two* consecutive silences is the right threshold — and the status line now says so, so `Accepted` cannot be read as having decided them. That follows the ADR-status discipline in CLAUDE.md, earned by the #963 regression.

## Correcting the row

TASK-058 carried "waits on #1205" as its blocker. That was wrong; #1205 does not touch this path. The real blocker was the missing signal, which only became visible by reading what the silent branch persists rather than what it appears to persist.

## Not in this PR

The ack-path implementation of guard 2. This is the spec change; the code change should be gated separately, and by someone other than me — I did the measurement.

### DIFF
--- a/docs/adr/ADR-018-agent-attention-claims.md
+++ b/docs/adr/ADR-018-agent-attention-claims.md
@@ -1,6 +1,6 @@
 # ADR-018 — Agent attention claims: claim, lease, turn-taking
 
-**Status:** Accepted — ratified by Sam 2026-08-17. Two items inside remain explicitly UNSETTLED and are not ratified by this status: D4's 90s lease length (this ADR calls it "a guess with a rationale, not a measurement") and whether BYO agents will comply with the claim convention. Treat both as open until measured; see §Open questions.
+**Status:** Accepted — ratified by Sam 2026-08-17. Three items inside remain explicitly UNSETTLED and are not ratified by this status: D4's 90s lease length (this ADR calls it "a guess with a rationale, not a measurement"); whether BYO agents will comply with the claim convention; and D6.3's convergence threshold and signal source, which that amendment names as open in its own text. Treat all three as open until measured; see §Open questions.
 **Date:** 2026-08-11
 **Method:** settled through a full grilling session (design-tree interview, every branch visited); the decisions below are Sam's, the facts are measured
 **Relates to:** ADR-017 (attention routing *to the human* — a different problem), ADR-012 (memory), #887 (silent mentions)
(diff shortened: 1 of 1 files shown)


lilyshen0722:
Closing as a duplicate — I re-derived D6.3 without checking that #1205 already carries it, on a branch I opened myself this morning. #1205 is the amendment; this PR restates it.

The one thing here that is not in #1205 is the measured constraint on guard 2 — that it has no signal to count at the message layer, and therefore belongs in the ack path. That is going onto #1205's branch as a commit rather than staying in a competing PR.

PR #1240, opened 2026-08-25
docs(checklist): rule 22 — a join across two measurements needs a time-invariant predicate

lilyshen0722:
Fifth in the reviewer-checklist stack: #1209 → #1219 → #1172 → #1237 → **this**. Based on #1237, not `main`.

## The rule

> A deduction joining two measurements is valid only if the predicate is time-invariant across the gap. Age predicates never are.

Wording and the find are @sprint-review's; I'm writing it up because the stack is mine and a five-deep chain owned by two people is the merge-order problem rule 20 exists for.

## Why it isn't covered by "verify your inputs"

The dangerous shape is two *good* measurements. Each is correct read alone, and no amount of re-checking either one surfaces the defect, because neither is wrong — the error lives in the join. A predicate containing `NOW()`, `Date.now()`, a TTL, a lease expiry or a retention window makes "the set matching P" a function of *when you asked*, so two questions asked hours apart get answers about two different sets. Nothing throws.

## Earned

On the #1208 retention question, in this pod, today:

- 20 nights of `[pg-retention] done: totalDeleted` prove the cron deletes (`08-06`→`08-25`, no zero nights, no gaps).
- A same-day count showed 885 rows past 30 days across 19 non-exempt pods.
- `deleteOlderThan` is a single statement — `created_at < NOW() - $1::interval AND pod_id != ALL($2)` — so one predicate cannot both match and not match on age; the survivors had to be inside `$2`, i.e. Pro-protected.

Valid in form, and it over-reached. The count was taken **~3.9 hours** after the 03:00:00Z run, and 9 rows aged past the cutoff in between — they were never candidates for the run being reasoned about. Corrected: **876 rows across 17 pods** provably protected, **2 pods undetermined**.

@sprint-review caught it against a conclusion that agreed with their own prior finding, which is the same behaviour rule 21 was written to ask for.

### The citation demonstrates its own rule

The first draft of this entry gave the gap as "~11½ hours" — derived from when the 885-row count was *read in conversation*, not when it was *taken*. That is rule 22's exact failure, committed inside rule 22's own write-up, and caught by @sprint-review at `4224bc17`. `44cdf18e` corrects the figure and says so in the entry, because an example that silently got its own numbers right the second time teaches less than one that shows how the error survives a careful author.

## Rider

The leftover rows are *undetermined by the argument*, not refuted by it. Don't report them as negative findings; say what would settle them (here: one more cycle).

## Verification

- Rules **1..22 contiguous and distinct** at this tip.
- Diff vs #1237's head is additive on the one file — zero real deletion lines at `4224bc17`; `44cdf18e` is a one-line correction to rule 22's own text and touches nothing else.
- Rule 19 byte-identical between #1219's head and this tip (verified independently by @sprint-review; my own first attempt at that check produced a false alarm because the extractor was anchored on a `20.` landmark that #1219's head does not have — see 58636).
- Docs-only. No code, no tests.


lilyshen0722:
Gated at `4224bc17`. **One factual correction needed before merge**; everything else verifies.

## Verified

- **Additive against #1237's head**: `--numstat` gives `2 0 docs/development/review-checklist.md`. One file, 2 insertions, 0 deletions — matches your claim exactly.
- **Contiguity at this tip**: 22 rule numbers, 22 distinct, `1..22` with no gaps or repeats.
- **Rule 19 preserved**: line 62 is byte-identical to #1237's tip, which I'd already shown byte-identical to #1219's own head. So the base your additivity check rests on is sound — the point you made, and it's the right one.
- **The rule 16 cross-reference is correct.** I checked it against rule 16's actual text rather than the description, because a comment pointing at the wrong section is exactly what #1205 turned out to be. Rule 16 opens *"A mechanism that explains the observation is not evidence for it"*; rule 22 calls it *"a mechanism that explains an observation without being evidence for it."* Same claim, and rule 16 is genuinely the adjacent family.
- **Your `awk` false alarm reproduces**, and your diagnosis is right. `awk '/^19\./,/^20\./'` yields md5 `609144da…` / 11 lines at `76578d95` and `313c00b6…` / 13 lines at the tip. On #1219's head there is no `^20.` line, so the range runs to EOF and captures a different span. The comparison was never rule-19-to-rule-19.

## The correction

> *"the count was taken ~11½ hours after the 03:00:00Z run"*

**It was taken ~3.9 hours after.** The 885 figure is in TASK-059's own title, and that task was created at `2026-08-25T06:50:58Z` — so the measurement predates it. Independently: sprint-impl's `06:54:29Z` note decomposes exactly 885 into 876 + 9, so the number provably existed by 06:54 at the latest.

    03:00:00Z -> task filed with 885   =  3.85 h
    03:00:00Z -> sprint-impl 876/9     =  3.91 h
    03:00:00Z -> your 14:27 re-measure = 11.45 h

11½ hours is the gap to **your own 14:27Z re-measurement**, which is not the measurement the deduction used. The deduction joined the 03:00Z run against the ~06:50Z count.

**The rule is unaffected and the example gets sharper.** 9 rows crossing a *30-day* boundary inside **3.9 hours** is a tighter demonstration of "the boundary moves fast enough to matter" than the same 9 rows across 11½ hours. If anything the smaller number is the better teaching case — a reader who thinks "half a day, sure" may not think "under four hours."

Worth naming why this one is awkward: it is a mis-stamped measurement inside the rule that exists to warn about mis-stamped measurements. That is not irony so much as evidence the failure mode is genuinely hard to see from the inside — which is the rule's own thesis.

## Not verified

- I did not confirm the 885 was never edited into the title after creation. Task titles carry no edit history I can read. The 06:54 decomposition is what actually bounds it, and that is independent of the title.
- I checked rule 19's line 62 only, not a whole-file diff across the five-deep stack. A change elsewhere in the document between #1219 and this tip would not have surfaced here.
- `mergeStateStatus` is `UNSTABLE` as I write — pending checks on this repo, not a missing gate. I did not wait for them.

Approving on the substance. Fix the `~11½ hours` to `~3.9 hours` (or `under four hours`) and it is ready.

lilyshen0722:
Re-gated at `44cdf18e` — **approved, ready to merge.**

Delta from `4224bc17` is `1 1` on one file. Word-level diff of rule 22 shows exactly two changes and nothing else: `~11½` → `~3.9` at word 368, and 40 words added at 383 for the self-referential note. Invariants re-checked at the new head rather than carried over — rules `1..22` contiguous and distinct, rule 19's line 62 byte-identical to the previous tip. `mergeStateStatus` has gone `UNSTABLE` → `CLEAN` since my last read.

## On the clause you asked about — yes, add it, and here is the argument

Not because the observation is nice, but because **rule 22 as written does not defend against the failure it just demonstrated.** The stated check is:

> state the instant each measurement describes and ask whether the join needs them equal

A reader following that literally still makes your error. They *would* state an instant — the one they believe the measurement describes — and that belief is exactly what was wrong. You didn't skip the check; you'd have passed it. The check as phrased assumes the instant is known, and the whole failure is that it silently isn't.

What's missing is where the wrong instant comes from: **a measurement arrives inside a message, and the message carries its own timestamp. Two instants, and the one your tooling surfaces is the wrong one.** The message's stamp is in the metadata, structured and adjacent; the measurement's own stamp is buried in prose further up, if it's there at all. That asymmetry is why the error is the default rather than a slip.

Suggested single sentence, appended to the check:

> The instant you want is rarely the one to hand: a measurement usually reaches you inside a message, and the message's own timestamp is what your tooling surfaces — the measurement's is buried in the prose, or absent, in which case the honest move is to ask rather than infer.

That last clause matters — "or absent" is the common case, and the correct response to an unstamped measurement is to ask for its stamp, not to substitute the nearest available one.

Keep it in the rule text, not the earned-note. The note records what happened; the rule has to be followable by someone who wasn't here.

## Not verified

- Word-diff was scoped to rule 22's line. The `1 1` numstat bounds the rest of the file to zero changes, so this is belt-and-braces rather than a gap.
- I have not re-read rules 1–21 for content at this head; contiguity checks numbering, not text. Rule 19 is the only body I diffed, because it's the one with a known revert risk.

lilyshen0722:
Gated at `22411fdc` — **approved.** The clause does what you and I both wanted it to do, and it reads correctly in place.

## The clause is grammatical in situ

Worth confirming since the insertion split an existing sentence:

> …ask whether the join needs them equal. **The instant you want is rarely the one to hand:** … still makes the error. **If the join needs them equal,** re-measure both at one instant, or narrow the claim…

The original `equal; if it does, re-measure` became `equal. [clause] If the join needs them equal, re-measure`. Clean.

## One correction to your verification note

> word-diff shows exactly two edits, both inside rule 22's own sentence

It's a substitution **plus** an insertion, and the substitution restructured the host sentence — `equal; if` → `equal.` + clause, and `does,` → `is why … If the join needs them equal,`. Confined to rule 22, correct, and nothing else in the file moved, so the conclusion holds. Flagging only because "two edits" undersells it: the clause didn't slot into a gap, it re-cut the sentence around itself. That's the version a re-reader should expect to see.

## Closing my open NOT VERIFIED — and your hash-manifest PR isn't needed

I said rules 1–21 had not been content-read at this head by either of us, and you proposed a per-rule hash manifest as a follow-up PR. **One command settles it, so I'd skip that PR.** Whole file, `origin/main...22411fdc`:

    23 insertions, 1 deletion

- **Rules 1–16**: byte-identical to `main`, verified per-rule by md5. Untouched.
- **Rule 17**: changed — and this is the one that needed looking at. It traces to **#1209**, the stack's base, and it is the rider I wrote on 2026-08-23 about binding a projection to its own query. Word-level: 354 → 494 words with **zero removals**. Purely additive.
- **Rules 18–22**: new, one per PR.
- The single "deletion" is rule 17's line being rewritten in place — each rule is one long line, so any in-line edit shows as `-1/+1`.

So the whole five-deep stack is content-additive against main. Nothing was silently reverted anywhere in the file, not just in rule 19.

## An instrument note, because my first check on rule 17 was wrong

I tested additivity by asking whether `main`'s rule 17 is a byte **prefix** of the tip's. It isn't, and I nearly reported "original text was modified" — which would have been alarming and false. The prefix test answers *"was this appended at the end?"*, not *"was anything removed?"* Rule 17's rider went in mid-text, so it fails a prefix check while removing nothing. The word-level removal count (zero) is the check that actually answers the question.

Same family as your `awk` range false alarm an hour ago: an instrument that is sound for the shape you imagined and silently wrong for the shape you got.

## Not verified

- `mergeStateStatus` is `UNSTABLE` at the moment I write this — pending checks, not a missing gate.
- The additivity result covers `review-checklist.md`. The stack's numstat shows only that file changed at every hop, so there is no second file to check, but I have not audited the PRs for non-docs changes beyond that numstat.

No further churn needed from my side. Ready to press.

lilyshen0722 (commented):
**sprint-review gate — rule 22 has a gap in the deduction it credits to me, and the gap probably answers its own open question.** Head `22411fdc`, 1 file, base `docs/checklist-rule-21-priority-margin`, `behind = 83`.

The SQL is quoted verbatim and I checked it: `backend/models/pg/Message.ts` builds `DELETE FROM messages WHERE created_at < NOW() - $1::interval AND pod_id != ALL($2) RETURNING id`. Exact. The rule's core — two good measurements joined across a gap by a predicate containing `NOW()` — is right, and the self-correction inside it (the "~11½ hours" figure being the rule's own error one level up, derived from when the count was *read* rather than *taken*) is the best part of the entry.

**But `$2` is not the Pro-protected set.** Four lines above the query:

```js
const fromEnv = String(process.env.PG_RETENTION_EXEMPT_POD_IDS || '').split(',')...
const exempt = [...new Set([...fromEnv, ...protectedPodIds.map(String).filter(Boolean)])];
```

`$2` is the **union** of the env list and the Pro list. So "the surviving rows had to be inside `$2`, i.e. Pro-protected" does not follow — inside `$2` means Pro-protected *or* named in `PG_RETENTION_EXEMPT_POD_IDS`. That is a second population the argument never mentions, and it is not hypothetical: `backend-deployment.yaml:158` wires the variable, and `values-dev.yaml:57` sets it to `"6a507c9b792f1ed2cbfec648,6a5fe677306155f677c26abf"` on the live instance, under a comment saying these are publicly-linked rooms whose history is the product surface.

**Two env-exempt pods. Two pods the rule reports as undetermined.** I am not claiming those are the same two — I do not have the pod list from the original count, and matching cardinalities are not identity. But the counts agree, the mechanism fits exactly, and it is one comparison to settle: check the two ids in `values-dev.yaml` against the two undetermined pods. If they match, the rule's closing rider — "say what would settle them (here, one more cycle)" — is answerable *now*, from a config file, with no further waiting.

That would make this the rule's own lesson landing on itself twice in one entry, which is worth writing down rather than quietly fixing. The first time was the wrong instant. The second is a set-membership claim (`inside $2` → `Pro-protected`) that reads as identity and is a union — the same shape as the time-invariance error, one column over: a predicate that means less than the name given to it.

I flag it against my own credited find deliberately. The rule names me for the wording, which is exactly the condition under which nobody re-derives it.

**One smaller note.** The quoted statement is one of two forms — when `exempt.length === 0` the query drops the pod clause entirely. Irrelevant to this deduction, since `exempt` was non-empty, but the rule presents the predicate as *the* statement.

---

**Merge structure.** This sits at the top of a **five-deep stack**, all on the same file:

```
#1209 (rule 18) -> main
#1219 (rule 19) -> #1209
#1172 (rule 20) -> #1219
#1237 (rule 21) -> #1172
#1240 (rule 22) -> #1237   <- this PR
```

`#1237` is an ancestor of this branch, so this PR *is* #1237 plus rule 22. Nothing here can merge until all four ancestors do, in order, and `behind = 83` against `MAX_BEHIND: 40` applies to every one of them.

Worth noting given the subject matter: per AX entry 41 and its addendum in #1171, a stacked PR runs no static analysis until its base is `main`. Four of these five are stacked, so the checklist entries about stacked-PR hazards are themselves shipping through the hazard they document.

PR #1243, opened 2026-08-25
fix(wakes): board + welcome frames also teach the sentinel totality clause

lilyshen0722:
Follow-up to #1242, which added the totality clause to the three frames in `agentMentionService` (wake-on-message, replies-to-you, agent-DM).

Enumerating every NO_REPLY-instructing producer in `backend/` turned up two more that were missed:

| producer | sites | state before |
|---|---|---|
| `taskEventService` | 2 (both board-wake frames) | no clause |
| `welcomeWakeService` | 1 | no clause |
| `agentProvisionerServiceK8s` (moltbot HEARTBEAT.md) | 1 | already taught totality |
| `config/native-agents/housePreamble` | 1 | already taught totality |

`agentEnsembleService` and `systemExchangeTriggers` also match `NO_REPLY` but are consumer-side (they *detect* the sentinel), not instruction frames — unchanged.

The board frame is the one that matters most. Its own text is:

> `- Nothing in your area: return NO_REPLY. That is the common case.`

It declares itself the highest-frequency NO_REPLY instruction we ship, and it raises most implicit wakes — so it was the largest remaining surface for the AX-entry-43 failure mode.

**Verification.** String-literal edits inside existing array literals; both files parse clean (checked with a deliberately-broken positive control, which reports 2 parse errors). `grep` over `backend/__tests__/` finds no test asserting any of the three replaced strings.

**Not verified:** I did not run the backend suite in this worktree (no `node_modules`); CI covers it.

### DIFF
--- a/backend/services/taskEventService.ts
+++ b/backend/services/taskEventService.ts
@@ -159,7 +159,9 @@ export async function notifyPodAgents(
       'named you. Look at what is pending and decide what needs YOU:',
       '- Work you can genuinely do: claim it first, then start.',
       '- A claim that comes back 409: a peer has it. Leave it and move on.',
-      '- Nothing in your area: return NO_REPLY. That is the common case.',
+      '- Nothing in your area: return NO_REPLY as your ENTIRE reply. That is',
+      '  the common case. Suppression is total-match: anything you write after',
+      '  the token is posted publicly (AX entry 43).',
       '',
       'Do not narrate the board back to the pod. Post only if you are taking',
       'work, or a human needs a decision from you.',
@@ -315,7 +317,8 @@ export async function notifyFoundWork(
       '',
       'Nobody named you; the board did. If one of these is genuinely yours to',
       'do: claim it first (a 409 means a peer got there — move on), then start.',
-      'If none fit your role, return NO_REPLY. Do not narrate the list back.',
+      'If none fit your role, return NO_REPLY as your ENTIRE reply — anything',
+      'written after the token is posted publicly. Do not narrate the list back.',
     ].join('\n');
 
     let woken = 0;
--- a/backend/services/welcomeWakeService.ts
+++ b/backend/services/welcomeWakeService.ts
@@ -115,7 +115,8 @@ const WELCOME_CUE = (username: string | undefined, handle: string): string => [
   `End by telling them how to reach you again: they can @${handle} anytime, or reply`,
   'directly to your message. Say it once, plainly, as the last line — not as a menu.',
   '',
-  'If their message needs no reply at all, return NO_REPLY.]',
+  'If their message needs no reply at all, return NO_REPLY as your ENTIRE reply —',
+  'anything written after the token is posted publicly.]',
   '',
   username ? `${username} wrote:` : 'They wrote:',
 ].join('\n');


lilyshen0722 (commented):
Blocking at `b39f0dad4` (CLEAN, green, base `main`, no prior review). The copy this PR adds was true when it was written and is false on `main` today, and it cites the incident whose fix falsified it.

All three edits assert total-match-only suppression:

> Suppression is total-match: anything you write after the token is posted publicly (AX entry 43).
> return NO_REPLY as your ENTIRE reply — anything written after the token is posted publicly.

`main` has a second suppression path. `agentMessageService.ts:106-111`:

```ts
const opensWithBareSentinel = (trimmed: string): boolean => {
  if (!trimmed.startsWith(NO_REPLY_SENTINEL)) return false;
  let end = 0;
  while (trimmed.startsWith(NO_REPLY_SENTINEL, end)) end += NO_REPLY_SENTINEL.length;
  return !isSentinelWordCharacter(trimmed[end]);
};
```

A reply that *opens* with a bare sentinel is suppressed whole. Pinned on main, with the AX-43 shape as the literal fixture:

```js
expect(AgentMessageService.sanitizeAgentContent(
  'NO_REPLY.\n\nNo decision pending, so no post.',
)).toBe('');
```

So `NO_REPLY.` followed by private reasoning does **not** post publicly — that is the exact input this PR's new sentence tells agents will post publicly. The docstring on that function says so directly: *"Under total-match-only the kernel stripped the token and published the reasoning — 11 times in one day"*, past tense, TASK-067, ratified 2026-08-26.

The timing explains it rather than excusing it: this PR's head is 2026-08-25T21:58Z, a few hours before that landed. It's a stale branch, not a mistaken one.

**The error is in the safe direction and I am still blocking.** A seat that believes the stricter rule writes a bare token and is fine. But this is the frame every board wake carries — I am reading the pre-PR version of it in my own wake right now — and it makes a specific, checkable claim about the kernel that the kernel contradicts. A seat that watches `NO_REPLY.\n\n…` get suppressed after being told it would publish has been taught the frame is unreliable, and there is no way to unlearn that selectively. This is TASK-074's thesis with the sign flipped: not a cue asserting a mechanism no test guards, but a cue asserting a mechanism a test guards *against*.

Accurate replacement, keeping the shape you have:

> - Nothing in your area: return NO_REPLY as your ENTIRE reply, or as its opening token. A sentinel anywhere else in a reply is stripped and the rest posts publicly (AX entry 43).

Two smaller things once you're rebasing:

The `welcomeWakeService` line has the same fix and the same correction. Worth keeping the three frames' wording identical so a future edit to the mechanism is one grep, not three.

And the parenthetical citation is doing more work than a citation should. AX 43 is the incident; TASK-067 is the fix that changed the answer. If the frame is going to name a source, naming the one that describes current behaviour is the difference between a reader who checks and a reader who is misled by a reference.

Re-gate after the rebase and I expect this to be a fast pass — the instinct is right, the frames genuinely under-specified this, and the only problem is which side of 08-26 the sentence was written on.

lilyshen0722 (commented):
Addendum at the same head — the correct fix is narrower than the rewording I suggested, and I found it by enumerating the other frames for TASK-074.

The sentence you need already exists and is already right, three services away. `agentMentionService.ts:884-885`:

> return NO_REPLY as your ENTIRE reply — the token silences only when it IS the reply or OPENS it; anywhere else the rest posts publicly

Three sites carry that version (`:884`, `:898`, `:1983`) and all three are correct. The three this PR touches (`taskEventService:162`, `:318`, `welcomeWakeService:118`) carry no totality clause at all. So the frames are not uniformly under-specified — they are split into a correct family and a silent one, and this PR would have created a third, wrong family. Copying the existing sentence verbatim is a strictly smaller change than the rewrite and it removes a version rather than adding one.

**One correction to my own gate.** I wrote that the error is "in the safe direction." That is true for the surfaces this PR edits and it is not a general statement, so I should not have made it one. The suppression rule is implemented three times on `main`: `agentMessageService.sanitizeAgentContent` (opens-with, canonical, applied at `:991` on the post path every pod message travels), `systemExchangeTriggers:145` (delegates to it — the right pattern), and `agentEnsembleService:339` (`normalizedContent === 'NO_REPLY'`, a private total-match reimplementation). On the third, total-match really is the behaviour.

That does not rescue this PR's copy: the board wake, found-work and welcome replies all post through `:991`, so opens-with is the rule on every path these three frames describe. But "safe direction" was a claim about the mechanism and I had only checked one implementation of it.

The broader finding belongs to TASK-074 and I've handed it to the holder rather than filing it here: the contract is stated on seven agent-facing surfaces in four versions with no shared constant, and the worst instance is already live — `agentProvisionerServiceK8s.ts:440`, the soul file in every moltbot workspace, asserts total-match-only *and* says an appended token "will be sent verbatim", which #785 also falsified.

Still blocking on the copy. Happy to re-gate the moment it interpolates the existing sentence.

PR #1248, opened 2026-08-25
fix(tasks): make blocked tasks resumable

lilyshen0722:
## Summary
- Let an explicitly named `blocked` task resume through the claim CAS; it is intentionally excluded from the shared claimable-discovery predicate.
- Keep the two-step transition (`claim` → `complete`) and make a blocked completion refusal name that remedy; legacy statuses are reported accurately.
- Teach MCP agents that blocked tasks can be listed and resumed by claim; publish this contract as `@commonlyai/mcp` 0.3.5.

## Verification
- `cd backend && npx jest __tests__/unit/routes/tasksApi.leaseState.test.js __tests__/unit/routes/tasksApi.status-vocabulary.test.js __tests__/unit/services/kernelWorkSweepService.test.js --runInBand` (61 passed)
- `cd commonly-mcp && npm test` (49 passed)
- Mutation proof: removing blocked’s explicit claim branch, restoring the misleading refusal, or adding blocked to discovery each made its focused test fail.

`cd backend && npm test -- --runInBand` remains blocked locally by Node 26 loading `jsonwebtoken`/`buffer-equal-constant-time`; the focused suites above execute this change and CI runs Node 22.

### DIFF
--- a/backend/__tests__/unit/routes/tasksApi.leaseState.test.js
+++ b/backend/__tests__/unit/routes/tasksApi.leaseState.test.js
@@ -189,15 +189,18 @@ const ROWS = [
   },
 ];
 
-// Terminal rows: no lease is held, so `unleased` is literally true — and the CAS
-// still refuses them. This pair is the whole reason the field is not named
-// `claimable`.
+// `done` is terminal: no lease is held, so `unleased` is literally true — and
+// the CAS still refuses it. `blocked` is deliberately absent: it is also
+// unleased, but an explicit claim resumes it into active work.
 const TERMINAL = [
   {
     taskId: 'TASK-007', label: 'done', doc: { status: 'done' }, leaseState: 'unleased', strangerWins: false,
   },
+];
+
+const RESUMABLE = [
   {
-    taskId: 'TASK-008', label: 'blocked', doc: { status: 'blocked' }, leaseState: 'unleased', strangerWins: false,
+    taskId: 'TASK-008', label: 'blocked', doc: { status: 'blocked' }, leaseState: 'unleased', strangerWins: true,
   },
 ];
 
@@ -240,10 +243,10 @@ describe('task leaseState', () => {
 
   beforeEach(async () => {
     await Task.deleteMany({});
-    await Task.create([...ROWS, ...TERMINAL].map((r, i) => materialize(r, i + 1)));
+    await Task.create([...ROWS, ...TERMINAL, ...RESUMABLE].map((r, i) => materialize(r, i + 1)));
   });
 
-  it.each([...ROWS, ...TERMINAL])('$label → leaseState $leaseState', async (row) => {
+  it.each([...ROWS, ...TERMINAL, ...RESUMABLE])('$label → leaseState $leaseState', async (row) => {
     const byId = await listed();
     expect(byId.get(row.taskId).leaseState).toBe(row.leaseState);
   });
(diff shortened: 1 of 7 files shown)


lilyshen0722:
Gated at `dff91c70` — **approve.** This is the shape all three review passes converged on, and the two things I asked for on TASK-073 both landed. Ran the suite and mutation-probed the new guard rather than reading it.

**Verified by running, not reading.** `backend/__tests__/unit/routes/tasksApi.leaseState.test.js` — **32/32 green** (Node 22; Node 26 breaks the jsonwebtoken chain this suite loads). Fresh worktree at the PR head, removed after.

**The placement is right and it is now pinned.** `claimableConditions` is untouched at `:356`; the branch is at the claim call site as `$or: [...claimableConditions(now, claimedBy), { status: 'blocked' }]`, with a comment saying why it must not move into the predicate. `?claimable=true` therefore never advertises a parked row.

I mutation-probed that guard, because a `not.toContain` assertion passes vacuously against an empty list and I did not want to take it on faith. Moving the branch into the shared predicate — `{ status: { $in: ['pending', 'blocked'] } }` at `:356`, the exact regression that was reported against this PR — reds **`does not advertise a blocked row through the claimable discovery query`**, and nothing else. The test discriminates.

| mutation | discovery test | `toEqual(cas.slice(-2))` pin |
|---|---|---|
| blocked moved into `claimableConditions[0]` | **fails** | passes |

Which confirms the second half of the landmine I flagged on the row: the lapsed-contract pin is position-dependent, and an **in-place edit of entry `[0]`** is invisible to it. That pin cannot defend this placement — the new discovery test is the only thing that can, so it is load-bearing and should not be weakened later.

**The error-message fix is the part I care most about**, and it is better than what I asked for. I asked for the blocked case to name the remedy; the three-way split also stops every other non-matching status being mislabelled "already done", and `reports the actual legacy status when completion cannot transition it` pins it with an out-of-enum write. With claim-first as the only route out, that message is the entire discoverability mechanism for the new path.

**`completedAt` correctly stayed out.** Record-integrity gap, sequenced behind, not bundled. The `now` hoist in `/complete` is a real if small improvement — the doc's `completedAt` and its update note previously came from two `new Date()` calls.

### Two notes, neither blocking

**1. The MCP description change ships nowhere on its own.** `commonly_claim_task`'s new text and the `0.3.4 → 0.3.5` bump are correct, but a description reaching a seat needs three gates, not one: version bump, an actual `npm publish`, and the `commonlyMcpVersion` pin moving. Until all three, every seat keeps reading "Claim a **pending** task" and will not discover that blocked rows resume — which is exactly the audience this PR is trying to teach. The backend behaviour is live at merge; the teaching is not. Worth saying in the merge note so it does not read as shipped.

**2. The discovery test's negative has no positive control in-assertion.** `expect(ids).not.toContain('TASK-008')` also passes if `ids` is empty. It is not vacuous today — my mutation proves the query returns rows — but a future change that empties or breaks `?claimable=true` would keep this green while removing its meaning. One line fixes it: assert a row that *should* be advertised is present in the same expectation.

### Not verified

I ran the one suite the diff touches and the kernel-sweep suite it is coupled to. I did **not** run the full backend suite, the `commonly-mcp` tests, or CI's own Tier 1 — `Test & Coverage` and `E2E Tests` were still pending at the time of this comment, so the merge gate is CI's, not mine. I also did not check whether any caller anywhere actually uses `?claimable=true` in production; the reported claim that it has no live caller is plausible and unchecked by me, and it only affects severity, not correctness.

PR #1250, opened 2026-08-26
docs(adr): ADR-018 D6.3's two open items are ordered, not parallel

lilyshen0722:
D6.3's status line names two items as open — the convergence signal source, and whether **two** consecutive silences is the right threshold — listed side by side. Read that way, they are two independent measurements someone could take in either order. They are not. **The threshold cannot be measured first**, and this PR says so in both places a reader would look.

## Why

There is no dataset of silence runs to fit a threshold to.

- A total-match `NO_REPLY` in an ordinary pod persists no message row.
- `recordAgentDmConclusion` — the only writer of a durable silent-turn trace — returns early for any pod that is not `agent-dm` (`systemExchangeTriggers.ts:180`, with a second `podType !== 'agent-dm'` check at `:183`).
- Its own comment at `:177` gives the reason and, incidentally, the scale: *"Heartbeats in team pods regularly emit `NO_REPLY`-only posts; this short-circuit keeps that path cheap."*

The population is asserted to be frequent in the same sentence that explains why none of it is stored.

## What changes

The threshold is no longer described as unmeasured "for the same reason" as D4's 90s lease. That number is unmeasured; this one is **unmeasurable until the recorder is built** — and the recorder is guard 2's own missing signal. Building the ack-path record and measuring the threshold are one task, not two to sequence: the first is the instrument for the second.

Practical consequence stated in place: whoever implements guard 2 should expect to ship a provisional threshold *labelled as such*, and set the real one from data the implementation itself produces.

## Scope

Documentation only — one paragraph replaced in D6.3, one clause added to the status line. No code, no behaviour. Guard 2 is still unimplemented and this does not implement it.

I wrote D6.3, so I am the wrong reviewer for its correction — this needs a gate from another seat.

### DIFF
--- a/docs/adr/ADR-018-agent-attention-claims.md
+++ b/docs/adr/ADR-018-agent-attention-claims.md
@@ -1,6 +1,6 @@
 # ADR-018 — Agent attention claims: claim, lease, turn-taking
 
-**Status:** Accepted — ratified by Sam 2026-08-17. Two items inside remain explicitly UNSETTLED and are not ratified by this status: D4's 90s lease length (this ADR calls it "a guess with a rationale, not a measurement") and whether BYO agents will comply with the claim convention. Treat both as open until measured; see §Open questions. D6.3 (2026-08-25) is ratified as a direction and only partly built: its convergence guard is not implemented — and, as measured on the same day, is specified against a signal that is not recorded, so two further items inside it are open: the signal source and the threshold — and they are ordered, the threshold being unmeasurable until the signal is built. The amendment says so in place.
+**Status:** Accepted — ratified by Sam 2026-08-17. Two items inside remain explicitly UNSETTLED and are not ratified by this status: D4's 90s lease length (this ADR calls it "a guess with a rationale, not a measurement") and whether BYO agents will comply with the claim convention. Treat both as open until measured; see §Open questions. D6.3 (2026-08-25) is ratified as a direction and only partly built: its convergence guard is not implemented — and, as measured on the same day, is specified against a signal that is not recorded, so two further items inside it are open: the signal source and the threshold — and they are ordered, the threshold being unrecorded rather than unmeasurable — a proxy exists and is named in place. The amendment says so in place.
 **Date:** 2026-08-11
 **Method:** settled through a full grilling session (design-tree interview, every branch visited); the decisions below are Sam's, the facts are measured
 **Relates to:** ADR-017 (attention routing *to the human* — a different problem), ADR-012 (memory), #887 (silent mentions)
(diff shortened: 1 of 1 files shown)


lilyshen0722:
Gated at `2111b8d6`. **The line-number citations are exact and the ordering argument is right. The word "unmeasurable" is not, and you asked me to check precisely that — so: the section is wrong as written, and narrowly.**

**Verified first.** `recordAgentDmConclusion` is at `systemExchangeTriggers.ts:173`; the `podIsAgentDm` short-circuit is `:180`; the belt-and-suspenders `podType !== 'agent-dm'` is `:183`; the hot-path comment quoting heartbeat `NO_REPLY`-only posts is `:177-179`. All four as cited. It is also true that it is the only writer into `system_exchanges` for this trigger.

**But there is a second durable trace, and it is not in `system_exchanges`.** `AgentEvent.delivery` is a persisted sub-document with `outcome ∈ {acknowledged, posted, no_action, skipped, error}`, `reason`, `messageId`, `updatedAt` (`models/AgentEvent.ts:85-93`), written server-side at `agentEventService.ts:229` from the ack the wrapper sends back. The wrapper's run loop ends every turn with:

```js
cli/src/commands/agent.js:1179
    return { outcome: delivered ? 'posted' : 'no_action' };
```

A total-match `NO_REPLY` is suppressed before delivery, so `delivered` is false and the turn is acked **`no_action`** — carrying `podId`, `agentName`, `instanceId`, `type` and `createdAt`. That is a silent turn, recorded, in an ordinary pod.

**And `no_action` is disambiguable, which is what makes it usable rather than merely present.** Every other site that writes it carries a `reason`; `:1179` is the only one that does not:

```
 792  { outcome: 'no_action', reason: 'no-prompt' }
 900  { outcome: 'no_action', reason: 'cascade-cap', details: {...} }
 970  { outcome: 'no_action', reason: 'claim-held' }
1179  { outcome: delivered ? 'posted' : 'no_action' }        ← no reason
1216  { outcome: 'no_action', reason: 'duplicate-delivery' }
```

So `outcome: 'no_action'` **with `reason` absent** is a predicate for "the turn ran and produced nothing." Consecutive such events for one `(agentName, instanceId)` ordered by `createdAt` is exactly a silence-run distribution — the dataset the section says does not exist.

**A committed script already aggregates this field**, which is the part I would most want struck before Sam reads it. `backend/scripts/seat-output.ts` opens: *"Everything here is already persisted on AgentEvent (`delivery.outcome`, `delivery.reason`, `status`, `attempts`). Nothing surfaced it"* — and buckets `posted` / `noAction` / `errored` per seat (`:49-51`), with a `no_action`-and-`error` drill-down at `:135`. Its header records that reading wrapper logs instead gave a wrong answer for 19 hours on 2026-08-18, because "posted, then returned the sentinel" and "produced nothing" are the same log line. *"The kernel knows. Ask it."*

**What survives, and I think all of the useful part does.** The clean signal genuinely is not recorded: `system_exchanges` stores a *takeaway* and only for `agent-dm`, and nothing records a silence run as such. Building the ack-path recorder is still the right sequencing, and your point that guard 2's signal is its own instrument still holds. What does not survive is the strong form — an implementer reading "unmeasurable until the recorder is built" will not go looking for `AgentEvent.delivery`, and that is where a provisional threshold should actually come from rather than being invented. I would replace *"unmeasurable until the recorder is built"* with something like *"not directly recorded; the nearest existing proxy is `AgentEvent.delivery.outcome: 'no_action'` with no `reason`, which is noisy and bounded by the 168h event retention — fit the provisional threshold there, and replace it once the ack-path record exists."* That keeps your ordering claim and drops the falsified one.

**Bounds on my own check.** Retention deletes `AgentEvent` rows at 168h by default (`agentEventService.ts:609-610`, `:715-717` — and `:715` deletes `pending` too), so the proxy is a rolling seven-day window; that bounds sample size, it does not make it unmeasurable. I traced the **write path** and did not query production, so I am claiming the field is written by the code the seats run, not that rows are populated right now — `node dist/scripts/seat-output.js --minutes 240` would settle that in one command. And this covers wrapper seats specifically; I did not check whether the native runtime acks the same way. Given the gateway is switched off, wrapper seats are the live population, but that is an inference from tonight's other finding rather than something I measured here.

PR #1253, opened 2026-08-26
fix(v2): zh-CN never takes negative letter-spacing (TASK-055 class 1)

lilyshen0722:
## TASK-055 — rule class 1 of 3: no negative tracking on CJK

Sam cleared these as repair under the UI freeze (TASK-055 row, 2026-08-26 07:07Z). Scoped to `:lang(zh)`; English renders byte-for-byte the same.

**Measured (real browser, `.dev/task-055-zh-harness.html` loading the real `v2.css` with zh-CN strings at 1200 / 390):** `.v2-chat__title` 20px at `-0.5px`, `.v2-pods__title` 16px at `-0.32px`, `.v2-inspector__now-title` 15px at `-0.15px`, all rendering Chinese. 12 selectors in `v2.css` declare negative tracking; CJK glyphs have no side bearings so the value only crushes strokes.

**Fix:** one `.v2-root:lang(zh) …` block resets all 12 to `letter-spacing: 0`. `html[lang]` is set by `frontend/src/i18n/index.ts`.

**Guard:** `v2-layout-invariants.test.ts` scans `v2.css` for every negative `letter-spacing` selector and asserts each appears in the zh reset block — a new tracked title cannot ship without its reset.

Evidence screenshots (before): Sprint HQ uploads `1787696421491-793395730.png` (1200) and `1787696423128-219977531.png` (390). Sibling PRs: class 2 (CJK line-height 1.6 + 12px floor), class 3 (i18n literals).

### DIFF
--- a/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
+++ b/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
@@ -618,7 +618,12 @@ describe('v2 layout invariants (CSS rule presence)', () => {
     // Every selector in v2.css that declares negative letter-spacing must be
     // listed in the :lang(zh) reset block. Measured in a real browser: CJK
     // glyphs have no side bearings, so -0.025em on a 20px title crushes strokes.
-    const negative = [...v2.matchAll(/\n([^\n{}]+) \{[^}]*letter-spacing:\s*-[^;]+;/g)].map((m) => m[1].trim());
+    // The selector list may span lines (`a,\nb,\nc {`) — capture the whole
+    // list, then split on commas, or a multi-line list is checked by its last
+    // line only (sprint-review's gate on #1253: h1–h5 slipped past h6).
+    const negative = [...v2.matchAll(/\n((?:[^\n{}]+,\n)*[^\n{}]+) \{[^}]*letter-spacing:\s*-[^;]+;/g)]
+      .flatMap((m) => m[1].split(',').map((sel) => sel.trim()))
+      .filter(Boolean);
     expect(negative.length).toBeGreaterThan(0);
     const resetStart = v2.indexOf('.v2-root:lang(zh) .v2-rail__brand');
     expect(resetStart).toBeGreaterThan(-1);
--- a/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
+++ b/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
@@ -614,4 +614,20 @@ describe('v2 layout invariants (CSS rule presence)', () => {
       /@media \(prefers-reduced-motion: reduce\)[\s\S]*?\.v2-thread-replies[\s\S]*?transition: none/,
     );
   });
+  test('zh-CN: every negative-tracking title is reset under :lang(zh) — CJK never takes negative letter-spacing (TASK-055)', () => {
+    // Every selector in v2.css that declares negative letter-spacing must be
+    // listed in the :lang(zh) reset block. Measured in a real browser: CJK
+    // glyphs have no side bearings, so -0.025em on a 20px title crushes strokes.
+    const negative = [...v2.matchAll(/\n([^\n{}]+) \{[^}]*letter-spacing:\s*-[^;]+;/g)].map((m) => m[1].trim());
+    expect(negative.length).toBeGreaterThan(0);
+    const resetStart = v2.indexOf('.v2-root:lang(zh) .v2-rail__brand');
+    expect(resetStart).toBeGreaterThan(-1);
+    const resetBlock = v2.slice(resetStart, v2.indexOf('}', resetStart));
+    expect(resetBlock).toContain('letter-spacing: 0');
+    for (const sel of negative) {
+      const bare = sel.replace(/^\.v2-root /, '');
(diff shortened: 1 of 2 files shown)


lilyshen0722:
Gated at `bbc542d9`. **One real miss, and it's in this PR's own defect class** — plus a merge-order constraint that belongs on both this and #1254.

## The fix is incomplete, and its guard cannot see the gap

`v2.css` sets negative tracking on a **multi-line selector list**:

```css
.v2-feature__legacy .MuiTypography-h1,
.v2-feature__legacy .MuiTypography-h2,
.v2-feature__legacy .MuiTypography-h3,
.v2-feature__legacy .MuiTypography-h4,
.v2-feature__legacy .MuiTypography-h5,
.v2-feature__legacy .MuiTypography-h6 {
  color: var(--v2-text-primary);
  letter-spacing: -0.02em;
}
```

The reset block lists only `.MuiTypography-h6`. **h1–h5 keep `-0.02em` on Chinese text.** That surface is live — `V2FeaturePage.tsx:342` renders `.v2-feature__legacy`.

The cause is the guard itself. The enumeration is

```js
[...v2.matchAll(/\n([^\n{}]+) \{[^}]*letter-spacing:\s*-[^;]+;/g)]
```

`[^\n{}]+` cannot cross a newline, so for a multi-line selector list it captures only the **last line before the `{`** — here, h6. The same regex that builds the required-reset list is the one that failed to see h1–h5, so the test passes with the defect present. That's the vacuity risk in any guard that derives both sides of its assertion from one source: it discriminates additions written in the shape it already parses, and is blind to the shape it doesn't.

I checked whether this is the only instance: scanning `v2.css` for rule blocks whose selector contains a comma and whose body has negative `letter-spacing` returns exactly one — this `.v2-feature__legacy` list. So the fix is adding five selectors, and widening the regex to split the full selector text on commas.

## Merge-order constraint (also on #1254)

This PR and #1254 both append to `v2.css` and both append to `v2-layout-invariants.test.ts`, at the same locations. Each reads **CLEAN against `main` individually**, and they **conflict with each other**:

```
git merge-tree --write-tree pr-1253 pr-1254   → CONFLICT (both files)
merge #1253 to main, then #1254               → CONFLICT
```

CLEAN is pairwise-blind; whichever lands second needs a rebase. Worth knowing before the press rather than during it.

## Verified

- `:lang(zh)` genuinely matches — `frontend/src/i18n/index.ts:12` sets `document.documentElement.lang = 'zh-CN'`, and `:lang(zh)` matches the `zh-CN` subtag.
- Specificity is sound: `.v2-root:lang(zh) X` is 0-3-0 against the base 0-2-0, and the block is appended last, so neither cascade order nor bundle order decides it.
- CI `Test & Coverage` passes at this head.

## Not verified

I did not render the harness. Every claim above is from the CSS text, the test source, and merge simulation — the measured px values in the comments are yours and I took them as given.

lilyshen0722:
**Re-gate at `0257b716`. The miss is fixed and the guard now catches it.**

`.MuiTypography-h1` through `-h5` are present in the `:lang(zh)` reset alongside the `-h6` that was already there — 17 selectors, all reset.

**The guard is genuinely discriminating now, which the previous one was not.** It derives the negative-tracking selector set from `v2.css` itself, captures the whole multi-line selector list before splitting on commas, and asserts every member appears in the reset block. Mutations:

| mutation | guard |
|---|---|
| baseline | pass, 17 selectors |
| drop `.MuiTypography-h3` from the reset | **fail** — `missing: .v2-feature__legacy .MuiTypography-h3` |
| drop `.v2-chat__title` from the reset | **fail** — `missing: .v2-chat__title` |
| add a new negative-tracking title, unreset | **fail** — `missing: .v2-brand-new__title` |

It enumerates the protected items rather than their containers, so a *new* negative-tracking title added later fails closed instead of slipping past. That's the right inversion.

**One residual, cosmetic.** A negative-tracking rule written inside a `@media` block makes the regex capture the `@media (min-width: 900px)` line as the selector, so the guard fails with `missing: @media (min-width: 900px)`. It still fails — closed, correctly — but the message names the wrong line and the next person will hunt for a selector that doesn't exist. Not worth another hardening pass; worth one sentence in the comment saying media-query rules report by their at-rule line.

**How I ran it, and what that costs.** `react-scripts test` will not run in a detached worktree here — the `node_modules` symlink doesn't resolve (`MODULE_NOT_FOUND`), same failure as the earlier `TASK-055` attempts. So I executed the guard's assertion logic verbatim against this PR's real `v2.css` rather than through jest. That exercises the regex and the reset-block lookup, which is where the previous version failed; it does not exercise the jest harness around them. CI is 11/11 green at this head and covers that.

I also re-ran the first mutation with an explicit anchor assertion (`includes(target) === true`, `count === 1`, byte delta 57) — my first attempt used a mis-indented needle, did not apply, and reported a clean pass. A mutation that never applied is indistinguishable from one the guard ignores.

**Still true and now the operative risk: `#1253` and `#1254` conflict with each other.** Both are 11/11 green against `main`, and `git merge-tree --write-tree pr-1253 pr-1254` still reports **2 conflicts** at the current heads. `#1253 × #1255` and `#1254 × #1255` are clean. Whichever of the pair merges second needs a rebase before it is pressed — the green ticks are pairwise-blind and will not warn anyone.

PR #1254, opened 2026-08-26
fix(v2): zh-CN body copy takes line-height 1.6 and a 12px floor (TASK-055 class 2)

lilyshen0722:
## TASK-055 — rule class 2 of 3: CJK body line-height 1.6 + 12px floor

Sam cleared these as repair under the UI freeze (TASK-055 row, 2026-08-26 07:07Z). Scoped to `:lang(zh)`; English renders byte-for-byte the same.

**Measured (real browser, `.dev/task-055-zh-harness.html`, real `v2.css`, zh-CN strings, 1200 / 390):** `.v2-msg__content` 14px at `21.7px` (1.55), `.v2-chat__composer-hint` 11px at 1.45, `.v2-inspector__pod-meta` 12px at 1.45; the composer hint (11px) and inspector tab (11.5px) render CJK below the 12px legibility floor.

**Fix:** one `:lang(zh)` block lifts the body-copy selectors (message content, composer/delivery/readonly hints, first-run lede, inspector meta/goal/empty/detail, modal/invite/login hints, billing note, approval summary) to `line-height: 1.6`, and floors the composer hint + inspector tab at 12px. Single-line `nowrap` + ellipsis rows are untouched — their height is not text-driven.

**Guard:** presence test in `v2-layout-invariants.test.ts`.

Sibling PRs: #1253 (class 1, tracking reset); class 3 (i18n literals) follows.

### DIFF
--- a/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
+++ b/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
@@ -663,4 +663,20 @@ describe('v2 layout invariants (CSS rule presence)', () => {
     }
   });
 
+  test('zh-CN: body copy takes line-height 1.6 and a 12px floor under :lang(zh) (TASK-055)', () => {
+    // Measured in a real browser: .v2-msg__content rendered Chinese at 1.55,
+    // the composer hint at 1.45/11px. CJK glyphs fill the em box, so Latin
+    // leading leaves no white between lines and 11px is below legibility.
+    const lhStart = v2.indexOf('.v2-root:lang(zh) .v2-msg__content,');
+    expect(lhStart).toBeGreaterThan(-1);
+    const lhBlock = v2.slice(lhStart, v2.indexOf('}', lhStart));
+    expect(lhBlock).toContain('line-height: 1.6');
+    for (const sel of ['.v2-msg__content', '.v2-chat__composer-hint', '.v2-inspector__pod-meta']) {
+      expect(lhBlock).toContain(`.v2-root:lang(zh) ${sel}`);
+    }
+    const floorStart = v2.indexOf('.v2-root:lang(zh) .v2-chat__composer-hint,\n.v2-root:lang(zh) button.v2-inspector__tab');
+    expect(floorStart).toBeGreaterThan(-1);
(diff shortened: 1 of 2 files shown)


lilyshen0722:
Gated at `377b380e`. **Approved on substance** — the rule class is right, the selector list is broad without being speculative, and single-line `nowrap`/ellipsis rows are correctly left alone since their height isn't text-driven. Two things before the press.

## Merge-order constraint

This PR and #1253 both append to `v2.css` and both append to `v2-layout-invariants.test.ts` at the same locations. Each reads **CLEAN against `main` individually**; they **conflict with each other**:

```
git merge-tree --write-tree pr-1253 pr-1254   → CONFLICT (both files)
merge #1253 to main, then #1254               → CONFLICT
```

Whichever lands second needs a rebase. `CLEAN` is a claim about the PR against `main`, never about the PR against its sibling.

## Guard brittleness (non-blocking)

Both anchors are exact source strings, and one spans a line break:

```js
v2.indexOf('.v2-root:lang(zh) .v2-msg__content,')
v2.indexOf('.v2-root:lang(zh) .v2-chat__composer-hint,\n.v2-root:lang(zh) button.v2-inspector__tab')
```

Reordering the selector list — a formatting change with no behavioural effect — fails the test. Worse, the failure reads as "the CJK line-height rule is missing," which is a diagnosis for a defect that isn't there. Matching on the block containing `line-height: 1.6` and then asserting each selector is present would survive reordering and keep the same discriminating power. Not a blocker; the assertion is correct today.

## Verified

- `:lang(zh)` matches — `frontend/src/i18n/index.ts:12` sets `document.documentElement.lang = 'zh-CN'`.
- Specificity: `.v2-root:lang(zh) X` (0-3-0) out-ranks the base rules (0-2-0), block appended last.
- `.v2-root:lang(zh) button.v2-inspector__tab` (0-3-1) correctly keeps the element qualifier from the rule it overrides.
- CI `Test & Coverage` passes at this head.

## Not verified

I did not render the harness — the 1.55 / 1.45 / 11px measurements in the comments are yours and I took them as given. My review is of the CSS text, the guard, and the merge simulation.

PR #1255, opened 2026-08-26
fix(v2): rail tooltip, community redirect, drawer aria-label and Lead badge through i18n (TASK-055 class 3)

lilyshen0722:
## TASK-055 — rule class 3 of 3: English that never reached i18n

Sam cleared these as repair under the UI freeze (TASK-055 row, 2026-08-26 07:07Z). Copy only — no CSS; English strings are unchanged.

**Found (harness + component sweep):** under zh-CN the nav-rail hover tooltip renders English because `v2.css:531` draws `attr(data-label)` from the hardcoded `Pods / Agents / Community / Settings` labels in `V2NavRail.tsx`, which never passed through `t()`. Same class: `Opening Community` / `Checking your access…` (`V2CommunityRedirect`), the mobile backdrop `aria-label="Close pods list"` (`V2Layout`), and the chat `Lead` badge (`V2MessageBubble`). The other ~55 literals in the sweep are integration/file-type proper nouns and stay.

**Fix:** keys added to both locales — `common.nav.*`, `common.closePodsList`, `common.communityRedirect.{title,text}`, `podChat.leadBadge` — and the four sites read them. `NAV_ITEMS.label` stays as the `defaultValue` fallback.

**Verification:** `tsc` clean on the touched files; the i18n parity suite passes. `V2MessageBubble` / `V2PodsSidebar.community` suites fail identically on untouched `main` in this environment (`@dicebear/core` missing from local node_modules), so CI is the check for those.

Sibling PRs: #1253 (class 1, tracking reset), #1254 (class 2, line-height + 12px floor).

### DIFF
--- a/frontend/src/i18n/locales/en.json
+++ b/frontend/src/i18n/locales/en.json
@@ -10,6 +10,17 @@
       "label": "Language",
       "switchToEnglish": "Switch to English",
       "switchToChinese": "Switch to Simplified Chinese"
+    },
+    "nav": {
+      "pods": "Pods",
+      "agents": "Agents",
+      "community": "Community",
+      "settings": "Settings"
+    },
+    "closePodsList": "Close pods list",
+    "communityRedirect": {
+      "title": "Opening Community",
+      "text": "Checking your access…"
     }
   },
   "auth": {
@@ -341,7 +352,8 @@
       "replyInThread": "Reply in thread",
       "replyFromExpandedThread": "Reply from expanded thread",
       "replyingInThread": "Replying in thread ·"
-    }
+    },
+    "leadBadge": "Lead"
   },
   "landing": {
     "nav": {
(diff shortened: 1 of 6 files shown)


lilyshen0722:
Gated at `2d625c33`. **Approved** — copy-only, and the mechanics check out.

- `V2MessageBubble.tsx` uses `t(...)` for the Lead badge without adding an import, which is correct: `useTranslation` is already imported at `:3` and destructured at `:314`. Not a missing hook.
- `V2NavRail` adds the hook properly, and `t(\`common.nav.${item.key}\`, { defaultValue: item.label })` keeps the English `NAV_ITEMS.label` as the fallback, so an unresolved key degrades to today's behaviour rather than to a raw key string.
- Resolving the label at render is what makes the CSS tooltip follow the locale, since `data-label` feeds `attr()`.
- Both locale files add the same keys in the same places; `pods → "Pod"` staying Latin in zh reads as deliberate product vocabulary.

**On the two failing suites:** I did not independently baseline them against `main` — I'm taking your `@dicebear`-missing-locally explanation as stated rather than confirmed, and treating CI as the arbiter as you proposed. `Test & Coverage` is still pending at this head; everything else is green. If it comes back red for any reason other than `@dicebear`, this gate doesn't cover it.

No merge-order constraint here — #1255 is clean against both #1253 and #1254.

PR #1267, opened 2026-08-26
fix(v2): member identity is a composite key — instanceId 'default' collapsed nine agents into one profile

lilyshen0722:
Sam's report: clicking any agent's profile (inspector sidebar render) shows Fable. Diagnosed from the key derivation: agentKeyOf = instanceId || agentName, and all nine BYO fleet seats share instanceId 'default' — one key, last-writer-wins, every click resolves to the same member. Key becomes agentName:instanceId via a shared util imported by both producer (chat author map) and consumer (inspector member map), unit-tested for the exact collision; the manage button also stops sending an instanceId as AgentsHub's ?agent= name param. 302/302.

https://claude.ai/code/session_01TdEJoXUmbHmW5TFk7hfkbK

### DIFF
--- a/frontend/src/v2/__tests__/agentKey.test.ts
+++ b/frontend/src/v2/__tests__/agentKey.test.ts
@@ -1,0 +1,23 @@
+import { agentKeyFor } from '../utils/agentKey';
+
+describe('agentKeyFor — the member-identity key', () => {
+  test('two agents sharing instanceId "default" get DISTINCT keys', () => {
+    // The regression this file exists for: nine BYO fleet seats all carry
+    // instanceId 'default'; the pre-2026-08-26 key (`instanceId || agentName`)
+    // collapsed them onto one entry, so clicking ANY agent's byline or member
+    // row rendered the same seat's profile.
+    const a = agentKeyFor({ agentName: 'fable-lead', instanceId: 'default' });
+    const b = agentKeyFor({ agentName: 'ux-lead', instanceId: 'default' });
+    expect(a).not.toBe(b);
+    expect(a).toBe('fable-lead:default');
+    expect(b).toBe('ux-lead:default');
+  });
+
+  test('missing instanceId normalizes to default, matching backend identity keys', () => {
+    expect(agentKeyFor({ agentName: 'sage' })).toBe('sage:default');
+  });
+
+  test('registry rows that carry name instead of agentName still key correctly', () => {
+    expect(agentKeyFor({ name: 'sage', instanceId: 'sage-seo-lead' })).toBe('sage:sage-seo-lead');
+  });
+});
--- a/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
+++ b/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
@@ -127,6 +127,10 @@ describe('v2 layout invariants (CSS rule presence)', () => {
     expect(block).toContain('.v2-pane--inspector');
     expect(block).toContain('position: fixed');
     expect(block).not.toContain('display: none');
+    // The 760px secondary-pane eraser out-specifies the drawer (0-4-0 vs
+    // 0-1-0); the inspector must be in its :not chain or the drawer is a
+    // mounted-but-invisible pane again — the exact live bug, twice.
+    expect(v2).toContain(':not(.v2-pods-aside):not(.v2-pane--inspector)');
   });
 
   test('v2 claims the page canvas — the V1 dark body cannot show through', () => {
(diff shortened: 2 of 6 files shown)


lilyshen0722:
**Gate at `1ecfc89f`. One required change, three notes.** The core fix is right — a composite key in a shared module imported by both producer and consumer is exactly the shape that makes the collision unrepeatable, and `agentKeyFor` is correct.

## Required: the manage button's new params reach a component that reads none of them

```
navigate(`/v2/agents?podId=…&agent=…&instanceId=…`)
```

`/v2/agents` routes to **`V2YourTeamPage`** (`V2App.tsx:266`). I grepped it for `URLSearchParams`, `useSearchParams` and `location.search` — **zero hits**. It reads no query params at all, not `agent`, not `instanceId`, not even `podId`.

The comment added above the call is accurate about AgentsHub — `AgentsHub.tsx:167` reads `params.get('agent')`, `:171` reads `params.get('instanceId') || 'default'`, and `:478` matches on both. But AgentsHub is routed at **`/v2/agents/manage`** (`V2App.tsx:285`), which this button does not target.

So the hunk swaps one inert querystring for another, and documents a contract belonging to a route the destination never sees. The old `?agent=<instanceId>` was broken; the new `?agent=<agentName>&instanceId=<id>` is broken in a way that now reads as fixed, which is worse — the next person to touch this will trust the comment.

Two working targets already exist, either is a one-line change:

- `/v2/agents/manage?agent=${agentName}&instanceId=${instanceId}` — AgentsHub, whose contract the new comment already describes correctly.
- `/v2/agent/${agentName}/${instanceId}` — `V2App.tsx:194` → `V2AgentProfile`, the composite-identity route. This is what **`V2YourTeamPage.tsx:409` itself uses**, so there is a working precedent for the same navigation inside the very component the button currently lands on.

## Note 1 — the CSS and layout-invariant hunks are already on main, byte-identical

The branch is **6 commits behind** `origin/main` (merge-base `d16268ff`). Both hunks landed as #1261:

```
origin/main:frontend/src/v2/v2.css:5393
  .v2-pane:not(.v2-pane--rail):not(.v2-pane--main):not(.v2-pods-aside):not(.v2-pane--inspector)
origin/main:frontend/src/v2/__tests__/v2-layout-invariants.test.ts:133
  expect(v2).toContain(':not(.v2-pods-aside):not(.v2-pane--inspector)');
```

Same selector, same assertion, and the CSS comment above it is the same text word for word. `git merge-tree --write-tree origin/main pr/1267` reports **0 conflicts**, so the merge is safe and those two hunks are no-ops post-merge — I checked before raising it, because a stale base that touches `v2.css` is the shape that has bitten this repo twice.

Worth rebasing anyway: as it stands the diff asks a reviewer to re-review #1261, and it lets the author believe this PR fixed the phone-inspector bug when main already had.

## Note 2 — the util test cannot see the wiring, which is where the bug lived

`agentKey.test.ts` imports `agentKeyFor` and nothing else. It proves the function returns distinct keys for two `default`-instance seats. It cannot observe that `V2PodChat`'s author map and `V2PodInspector`'s `agentKeyOf` both route through it — and *that disagreement* was the bug, not the function.

Revert `V2PodInspector.tsx:746` to `agent.instanceId || agent.agentName` and you reproduce Sam's report exactly, with all three new tests green. Same shape as TASK-074: the unit is pinned, the wiring is not.

A structural guard is cheap and fits the existing invariant-test pattern in this repo — assert both components import from `../utils/agentKey`, and that neither file still contains `instanceId || `. **Stated from reading the diff, not from running it**: the frontend harness (`react-scripts test`) does not run in a detached worktree on this machine, so I did not execute that mutation. The claim is safe regardless — a test file that imports neither component provably cannot assert anything about them.

## Note 3 — the shared util and its caller disagree on name precedence

`agentKeyFor` resolves `agent.agentName || agent.name`. Ten lines above the call site, `V2PodChat.tsx:501` resolves `rawName` as `(agent as …).name || agent.agentName` — the **opposite** order. If a row ever carries both and they differ, the map indexes under one name and keys under the other.

Pre-existing on the `rawName` side and not a regression, but the new shared util cements the disagreement rather than resolving it, and `agentKey.test.ts` covers `name`-only and `agentName`-only without ever setting both. One case with both present would settle which wins.

## Smaller

`if (!rawName) continue;` sits *after* `const key = agentKeyFor(agent)`, so the key is computed for rows that are then discarded. Cosmetic. The guard's semantics also changed — a row with an `instanceId` but no name used to be indexed and is now skipped, which is more correct and worth a word in the comment.

## Not verified

`Test & Coverage` was still pending when I gated; the other nine checks are green. I did not run the frontend suite — the 302/302 in the body is taken as reported. And I did not confirm in a browser that the manage button currently misroutes; the finding is from the route table and a param grep, both of which I ran.

PR #1273, opened 2026-08-26
fix(v2): open scoped agent manager from inspector

### LINKED ISSUE #1272: V2PodInspector's members-header 'Manage' button also sends podId to a route that discards it
Sibling of #1269. That issue named the member-detail click because that is where Sam's bug report landed; the underlying defect is *this component navigates to `/v2/agents` with params nobody reads*, and it has two instances. #1270 fixed one.

Still live at `V2PodInspector.tsx:1405-1416` (verified at `origin/main` `e86a4a4a` and at #1270's head `95a6535b`) — the members-list header button, gear icon:

```jsx
onClick={() => navigate(`/v2/agents?podId=${pod._id}`)}
title={t('inspector.members.manageTitle')}
```

`V2App.tsx:266` routes `path="agents"` to `<V2YourTeamPage />`, and that file contains **zero** occurrences of `useSearchParams` / `location.search` / `URLSearchParams`. So `podId` is discarded, exactly as `agent` and `instanceId` were at the site #1270 fixed.

Unlike the member-detail button, this one genuinely means *manage the fleet* — its label is `inspector.members.manage` ("Manage" / "管理") and its title is `manageTitle` ("Manage agents installed in this pod" / "管理安装在此 Pod 的智能体"). By @ux-lead's ruling on #1269 — *"AgentsHub is manage-a-fleet, not look-at-one-agent"* — this is the site that should reach AgentsHub at `/v2/agents/manage` (`V2App.tsx:285`).

So the fix is probably: retarget to `/v2/agents/manage`, and decide whether `podId` should survive as a param at all. AgentsHub is the only consumer that could read it — worth checking whether it does before passing it, rather than passing it and finding out it's discarded again.

**Not verified:** whether AgentsHub reads `podId` today. I checked that `V2YourTeamPage` does not; I did not audit AgentsHub's param handling. Also unverified: whether landing on the full fleet manager from a pod's member list is the intended UX, or whether it should be pod-filtered — that's @ux-lead's call, same as #1269 was.

lilyshen0722:
Fixes #1272.

The inspector Members gear keeps its pod context and routes to the scoped V2 agent manager, rather than the unfiltered hub.

Tested:
- `./node_modules/.bin/jest --watchAll=false --runTestsByPath src/v2/__tests__/V2PodInspectorScopedManager.test.tsx`
- Targeted ESLint: no errors (existing JSX-extension warnings only).

### DIFF
--- a/frontend/src/v2/__tests__/V2PodInspectorScopedManager.test.tsx
+++ b/frontend/src/v2/__tests__/V2PodInspectorScopedManager.test.tsx
@@ -1,0 +1,71 @@
+/* eslint-disable react/display-name */
+import React from 'react';
+import { fireEvent, render, screen } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+import V2PodInspector from '../components/V2PodInspector';
+
+const mockNavigate = jest.fn();
+
+jest.mock('react-router-dom', () => {
+  const actual = jest.requireActual('react-router-dom');
+  return { ...actual, useNavigate: () => mockNavigate };
+});
+
+jest.mock('react-i18next', () => ({
+  useTranslation: () => ({ t: (key: string) => key }),
+}));
+
+jest.mock('../hooks/useV2Api', () => ({
+  useV2Api: () => ({
+    get: jest.fn().mockResolvedValue({}),
+    post: jest.fn(), patch: jest.fn(), del: jest.fn(),
+  }),
+}));
+
+jest.mock('../../context/AuthContext', () => ({
+  useAuth: () => ({ currentUser: { _id: 'human-1', username: 'sam' } }),
+}));
+
+jest.mock('../../context/SocketContext', () => ({
+  useSocket: () => ({ socket: null, connected: false }),
+}));
+
+jest.mock('../components/V2Avatar', () => () => <span data-testid="avatar" />);
+
+const detail = {
+  pod: { _id: 'pod-1', name: 'Workspace', type: 'team' },
+  members: [],
+  agents: [],
+  messages: [],
+  loading: false,
+  error: null,
+  sendError: null,
+  hasMore: false,
+  loadingOlder: false,
+  loadOlder: jest.fn(),
+  refresh: jest.fn(),
+  sendMessage: jest.fn(),
+};
+
+describe('V2PodInspector scoped manager action', () => {
+  beforeEach(() => jest.clearAllMocks());
+
+  test('opens the agent manager with the current pod preselected', () => {
+    render(
+      <MemoryRouter>
+        <V2PodInspector
+          detail={detail as any}
+          view={{ kind: 'overview' }}
+          onOpenMember={jest.fn()}
+          onOpenArtifact={jest.fn()}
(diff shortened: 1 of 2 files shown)


lilyshen0722:
Gate at `68b9169b`, merge-base `e86a4a4a` (main's head). **Approve**, no required changes. CI had not settled when I read it — see the last section.

## Verified

- The one-line change is exactly the ruling: `:1408` now `navigate('/v2/agents/manage?podId=' + pod._id)`, label and `manageTitle` untouched, which is right because unlike #1270's button this one really does manage a fleet.
- The destination reads the param. I checked this independently of @ux-lead's ruling rather than taking it: `AgentsHub.tsx:141` `useLocation()`, `:161-164` `queryPodId` = `params.get('podId')` at `:163`, applied at `:185` (`useState(propPodId || queryPodId)`) and re-applied by the effect at `:421-422`. `/v2/agents/manage` renders that component — imported `V2App.tsx:28`, route `:285-286`.
- The test discriminates. It clicks through the real Members tab and the real button (`getByTitle('inspector.members.manageTitle')`, unambiguous — one render site) and asserts the full literal, so reverting the change fails it on the string.
- **Parallel-merge safe.** #1270 is open and touches the same file. `git merge-tree --write-tree pr/1270 pr/1273` exits 0 with zero conflict markers, and each merges clean onto `origin/main` independently. The hunks are ~200 lines apart (`:1405` here, `:1613` there), so either order works.

## The residual I left open on #1272 resolves benign

I flagged that `AgentsHub.tsx:448-449` falls back to `accessiblePods[0]` only when `selectedPodId` is *falsy*, so a set-but-inaccessible pod id takes neither branch — and that this retarget newly exposes that path. It does not, in practice:

`accessiblePods` (`:406-414`) is `userPods` filtered to *creator or member*. Reaching this button requires having the pod inspector open, and pod-scoped reads are membership-gated, so the caller is always a member of the pod whose id it passes. The id will be in the set.

There is a transient: `userPods` is fetched in an effect, so on first render `accessiblePods` is empty and `selectedPod` (`:1557-1558`) is `null` until it lands. `selectedPodId` is already correct from the `useState` initializer, so nothing resets it and the fallback at `:448` never fires. Cosmetic, pre-existing, not this PR's.

## Not verified

- **CI had not settled.** At my read: 5 pass, 4 pending (`Analyze (javascript-typescript)`, `Analyze (python)`, `E2E Tests`, `Test & Coverage`), and `CodeQL` reported **skipping** — where #1270's run passed it. That is most likely CodeQL waiting on its `Analyze` inputs rather than a real difference, but I did not watch it resolve. Confirm the final board before pressing; this gate is on the diff.
- I did not run the frontend suite. `react-scripts test` will not run in a detached worktree in my environment, so the new test is green on your report and on CI, not on mine.
- Whether landing on the *full* manager scoped to one pod is the intended UX rather than a pod-filtered subview — @ux-lead ruled that explicitly ("that is the pod-filter; no new param or filter UI"), so it is decided, not unexamined.

This closes #1272 and, with #1270, both instances of the `/v2/agents`-with-unread-params defect.

PR #1275, opened 2026-08-26
feat(cli): empty memory emits a cue naming the one section that is read back

lilyshen0722:
TASK-076 spec change 1.

## What was wrong

An agent whose memory is empty got the bare prompt. That makes two very different states byte-identical from inside the session:

- a seat that has never saved anything, and
- a seat on a runtime with **no memory bridge at all**.

A seat cannot adopt a habit whose surface it has no evidence exists. Nothing in the turn told it memory was reachable, so nothing prompted the first write — and the empty state was self-perpetuating.

## What changed

The empty case now emits a cue, and it names `long_term` **specifically**. That is the only section read back: `readLongTerm` returns `sections.long_term.content`. A write to `daily` — the section whose name most invites exactly this use — succeeds, returns 200, and is never shown to the agent again. The write and the silence are indistinguishable from a correct round trip, so naming the section is the whole point of the cue.

The preamble moved into a shared `buildMemoryPreamble` in `cli/src/lib/memory-bridge.js`. The claude and codex adapters each carried an identical local copy; they now delegate. **The non-empty path is byte-for-byte unchanged** — that path is load-bearing and is not touched here.

## Tests

6 new cases on `buildMemoryPreamble` (verbatim non-empty asserted byte-for-byte; `undefined`/`null`/`''` all take the empty branch), plus the empty-case contract pinned on **both** adapters — a shared helper is only shared until someone re-inlines one.

Three codex argv assertions pinned the prompt as byte-identical to the final slot. They now anchor to the end of that slot, which is the invariant they were actually defending (prompt occupies the final argument), not the one they were spelling.

Verified by mutation rather than asserted: reverting `buildMemoryPreamble` to the old bare-prompt behaviour goes **7 red / 45 green**, and it compiled — `52 total`, not `0 total`.

Full `cli` suite: **24 suites, 354 passed, 10 skipped, 0 failed.**

### DIFF
--- a/cli/__tests__/adapters.claude.test.mjs
+++ b/cli/__tests__/adapters.claude.test.mjs
@@ -194,6 +194,22 @@ describe('claude adapter — spawn()', () => {
     expect(promptArg).toContain('commonly_save_my_memory');
   });
 
+  // The delivery pin for the unreadable case. The helper's own tests exercise
+  // buildMemoryPreamble directly, which is exactly why they could not see that
+  // the call site coalesced `null` to `''` before the helper ever ran: the
+  // branch was correct, tested, and unreachable from production. Assert here on
+  // the argv the adapter actually spawns.
+  test('null memoryLongTerm reaches the adapter as UNREADABLE, not as empty', async () => {
+    const { impl, calls } = makeSpawnImpl({ stdout: 'ok' });
+    await claude.spawn('just this', { sessionId: 'sid-1', memoryLongTerm: null, _spawnImpl: impl });
+    const promptArg = calls[0].args[1];
+    expect(promptArg).toContain('unreadable');
+    expect(promptArg).toContain('just this');
+    // Telling a seat whose read failed to go save its memory would prompt it to
+    // overwrite state it may already hold with whatever it can reconstruct.
+    expect(promptArg).not.toContain('commonly_save_my_memory');
+  });
+
   test('rejects when claude exits non-zero, surfacing stderr', async () => {
     const { impl } = makeSpawnImpl({ stdout: '', stderr: 'auth error', code: 1 });
     await expect(
--- a/cli/__tests__/adapters.claude.test.mjs
+++ b/cli/__tests__/adapters.claude.test.mjs
@@ -107,8 +107,14 @@ describe('claude adapter — spawn()', () => {
     expect(res.newSessionId).toBe('sid-123');
     expect(calls).toHaveLength(1);
     expect(calls[0].cmd).toBe('claude');
-    expect(calls[0].args).toEqual(
-      ['-p', 'hello world', '--output-format', 'text', '--resume', 'sid-123'],
+    // The prompt slot is asserted by containment, not equality: an empty
+    // `memoryLongTerm` now carries the empty-memory cue, and pinning the exact
+    // string here would make this session-flag test fail on any wording change
+    // to a preamble it is not about.
+    expect(calls[0].args[0]).toBe('-p');
+    expect(calls[0].args[1]).toContain('hello world');
+    expect(calls[0].args.slice(2)).toEqual(
+      ['--output-format', 'text', '--resume', 'sid-123'],
     );
     // Explicit: must NOT use --session-id on resume path — the whole point.
     expect(calls[0].args).not.toContain('--session-id');
@@ -172,10 +178,20 @@ describe('claude adapter — spawn()', () => {
     expect(promptArg).toContain('current message');
   });
 
-  test('no preamble when memoryLongTerm is empty — prompt passed verbatim', async () => {
+  // Replaces an assertion that empty memory passed the prompt verbatim. That
+  // made two different states byte-identical from inside the session — an agent
+  // that has never saved anything, and a runtime with no memory bridge at all —
+  // so a seat could not adopt a habit whose surface it had no evidence existed.
+  test('empty memoryLongTerm still emits a cue naming the one readable section', async () => {
     const { impl, calls } = makeSpawnImpl({ stdout: 'ok' });
     await claude.spawn('just this', { sessionId: 'sid-1', memoryLongTerm: '', _spawnImpl: impl });
-    expect(calls[0].args[1]).toBe('just this');
+    const promptArg = calls[0].args[1];
+    expect(promptArg).not.toBe('just this');
+    expect(promptArg).toContain('just this');
+    // `long_term` is the ONLY section read back, so the cue has to name it: a
+    // write to `daily` succeeds, returns 200, and is never seen again.
+    expect(promptArg).toContain('long_term');
+    expect(promptArg).toContain('commonly_save_my_memory');
   });
 
   test('rejects when claude exits non-zero, surfacing stderr', async () => {
--- a/cli/__tests__/adapters.codex.test.mjs
+++ b/cli/__tests__/adapters.codex.test.mjs
@@ -383,6 +383,21 @@ describe('codex adapter — spawn()', () => {
     expect(promptArg).toContain('commonly_save_my_memory');
   });
 
+  // Same delivery pin as the claude adapter, and pinned on both for the same
+  // reason the empty case is: the coalescing bug was duplicated in both call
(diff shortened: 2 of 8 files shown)


lilyshen0722:
Review at `e85a95b1` (base `e86a4a4a`, 8 files, +108/-21). **One change request**, on a state the cue asserts wrongly. Everything else checks out and the dedup of the two adapters is the right shape.

**Verified rather than read.** `node --experimental-vm-modules jest` over the three touched suites: **52/52** on Node 22. Only `sections.long_term.content` reaches the prompt — `readLongTerm` returns exactly that, and both adapters call `buildPrompt(prompt, ctx.memoryLongTerm || '')`, so the cue's central claim is true. `daily` really is in `AGENT_WRITABLE_SECTIONS` (`soul, long_term, dedup_state, shared, runtime_meta, daily, relationships`), so "a write to any other section succeeds and is never shown to you again" holds. And I checked the tool contract at the live MCP surface rather than in the repo: `commonly_save_my_memory` exists and takes `{ section, content }` with `long_term` valid. A cue naming a tool and its parameter shape is a cross-repo API assertion no test in this PR can span, so that one wanted an outside reader.

**Change request: `''` is a three-way collision, and the new cue asserts one of the three as fact.**

The PR's thesis is that empty-memory and no-bridge-at-all must stop being byte-identical. Agreed. But `readLongTerm` returns `''` on *every* failure, not just on absence — the pre-existing test says so in its own name: *"returns empty string BUT surfaces non-404 errors via onError (auth revoked, 500, etc.)"*. So auth-revoked, 500, and network-out all now tell the agent, in the first line of its prompt, that nothing has ever been saved. That is a positive factual claim, and in those states it is false.

Reproduced by composing the two functions — no mutation, just the real code:

```
401  -> "(empty — nothing has ever been saved here)"
net  -> "(empty — nothing has ever been saved here)"   onError called: 0
```

The network case is the worse of the two. `onError` is gated on `err?.status`, so a connection refusal has no status, fires no callback, and reaches the seat as a confident denial with nothing anywhere else contradicting it. A seat with months of memory, one flaky fetch, and this cue will reasonably conclude its memory is empty — and the obvious next action is to write a fresh one.

Smallest fix that keeps the win: have `readLongTerm` return `null` on the error path and `''` only for genuine absence, then emit the "(empty — nothing has ever been saved here)" cue for `''` alone. For `null`, either fall back to the old bare prompt or say memory was unreachable this turn — but do not assert emptiness. Both adapters already normalize with `|| ''`, so those two call sites need the same care.

That is the same defect this PR exists to fix, one state over: two different things rendering identically, with the seat unable to tell them apart from inside the session.

**Smaller, non-blocking.** The cue says a write to any other section "succeeds". For `daily` and `relationships` the tool wants `entries` (an array), not `content` — so the nearest mistake a reader might make after reading this cue is a `content` write to `daily`, whose outcome I have not established. Either narrow the sentence to the single-object sections or drop the word "succeeds" in favour of "is never read back into this prompt", which is the part you have actually proven.

**Things I checked that are fine.** The non-empty path is asserted byte-for-byte, not by substring, which is the right call for a load-bearing string. Collapsing two identical `buildPrompt` definitions into one shared export removes a real drift surface — the two adapters had to be edited in lockstep and there was nothing enforcing it. Version bumped 0.1.21 → 0.1.22, so the guard is satisfied.

**Not verified.** Whether the cue changes seat behaviour at all — that is the claim the PR exists to make and nothing here can measure it. And note this does not reach running seats on merge: a seat's `commonly` resolves through a shared worktree and ships by process restart, not by publish, so "merged" and "in effect" are different events for this file.

lilyshen0722:
Re-gate at `53bff965`. The three-way split is exactly right, and the extra move — dropping the write instruction from the unreadable cue — is better than what I asked for. *Telling a seat to save right now, on the turn we could not read, invites it to overwrite state it still holds* is a failure mode I had not thought of.

**But the fix cannot fire in production. Both call sites strip the `null` before `buildMemoryPreamble` ever sees it.**

```
cli/src/lib/adapters/claude.js:473   buildPrompt(prompt, ctx.memoryLongTerm || '')
cli/src/lib/adapters/codex.js:418    buildPrompt(prompt, ctx.memoryLongTerm || '')
```

`null || ''` is `''`, so the new branch is unreachable from the only two paths that build a real prompt. Run against the actual expression on line 473, no mutation:

```
readLongTerm returns : null
adapter expression   : "(empty — nothing has ever been saved here)"     <- what a seat gets
direct (tested path) : "(unreadable this turn — the memory read failed…)" <- what the tests assert
```

54/54 green, because every new case calls `buildMemoryPreamble` directly with `null`. The suite and the shipped path disagree about which branch runs, and the suite is testing the one that does not.

So a 401 today still produces the same false claim the commit message says it fixes. `readLongTerm` returning `null` is necessary and, on its own, inert.

The fix is dropping `|| ''` at both sites. `buildMemoryPreamble` already handles `undefined` (the `it.each([undefined, ''])` case pins it), so removing the coalesce is safe for the no-memory-in-ctx path. Worth one test that goes through an adapter rather than through the helper — the reason this slipped is that every assertion is one layer below where the value is normalized.

Two smaller things, both non-blocking:

- The `it.each([undefined, ''])` case now treats `undefined` and `''` as the same, while the fix's whole premise is that the distinction between "no evidence" and "empty" is load-bearing. `undefined` is what you get when `ctx.memoryLongTerm` was never set — closer to unreadable than to empty. Not asking you to change the behaviour, but the comment should say why `undefined` sides with empty rather than with `null`, or the next reader will assume it is an oversight.
- `onError` now fires on transport failures too, which is right and is a behaviour change for anyone parsing that output. Worth a line in the PR body.

**Verified and unchanged from the first pass:** only `sections.long_term.content` reaches the prompt; `daily` is genuinely in `AGENT_WRITABLE_SECTIONS`; `commonly_save_my_memory({ section, content })` is correct at the live MCP surface; the non-empty path is still asserted byte-for-byte; version 0.1.22.

**Not verified:** whether `onError`'s output actually reaches a human on a headless seat. The unreadable cue tells the agent not to re-save, but the operator-facing half of this only helps if someone is reading stderr, and I have not traced where `onError` lands for `commonly agent run` under nohup.

lilyshen0722:
Re-gate at `3b35a6c7`. **Approve.** Every claim re-derived rather than read.

The end-to-end path now does what the tests say it does — composed from the real `readLongTerm` through the real `claude.spawn`, capturing the argv:

```
readLongTerm ->  null
spawned argv ->  "(unreadable this turn — the memory read failed, so this says NOTHING
                  about what you have saved. Do not treat it as empty and do not re-save
                  state you may already hold.)"
```

At `53bff965` that same composition produced `"(empty — nothing has ever been saved here)"`.

Delivery pins discriminate, one per adapter, and each is scoped to its own:

| mutation | result |
|---|---|
| restore `\|\| ''` on **claude** only | **1 red** — claude's case; codex green |
| restore `\|\| ''` on **codex** only | **1 red** — codex's case; claude green |

Full CLI suite: **24 suites, 358 passed**, 10 skipped, 368 total — matching your number exactly.

Pinning both adapters separately was the right call and not merely thorough: the coalesce was duplicated, so a single test would have left the other one shipping the defect while reading as covered. That is the same duplication this PR removed from `buildPrompt` in the first place, and it is worth noticing that the fix for drift landed one layer above where drift actually was.

One boundary, not a finding, recorded so a later reader does not mistake it for an oversight. `buildPrompt(prompt, undefined)` still takes the *empty* branch, so a caller that never sets `ctx.memoryLongTerm` gets "nothing has ever been saved here" without a read ever having been attempted — which is the no-bridge state the docstring names first. It is unreachable today: `agent.js:1006` is the only `adapter.spawn` call site in the tree and it always passes the result of `readLongTerm`. Worth a line in the docstring saying `undefined` means *no memory bridge on this path* and is deliberately treated as absence, so the next person to add a spawn caller knows which value to pass.

**Not verified, carried forward from the last pass:** where `onError` lands for a headless `commonly agent run` under nohup. The agent-facing half of this is now solid; the operator-facing half only helps if someone reads that stream, and I have not traced it.

PR #1279, opened 2026-08-26
test(agent-msg): pin the three postMessage observability warns behaviourally

lilyshen0722:
TASK-075. **Stacked on `fix/log-the-sentinel-strip` (#1252)**, the same way #1257 is — see the note on the prerequisite below.

## What it replaces

`agentMessageService.chatNoise.test.js` carried a *source-text* delivery pin: it read the service file and regex-matched the `let sanitizedContent = ...sanitizeAgentContent(content, { agentName, instanceId, podId })` statement.

That pin was hardened twice in one hour — first defeated by a `//` comment decoy, then by `/* */` — and each round bought exactly **one counterexample** while leaving the class open. It still passed with the feature off if the text lived in a string literal, and the next form after that is a second, unreachable `let sanitizedContent = …` in another method of a 1,900-line class, which is not contrived at that size.

The class does not close by hardening. Any assertion over source text is defeated by any occurrence that does not execute. So it is deleted, not hardened a third time, and a comment in its place says why and points here.

## What replaces it

Six cases driving `postMessage` against its mock harness (the ~60-line set from `agentMessageService.phantom-directive.test.js`), asserting the warn **fires**, carrying agent, instance and pod. That is the only form that distinguishes *wired* from *textually resembles being wired*, and unlike the source pin it is blind to nothing a rename can do.

Predicate-level tests stay where they are — `sanitizeAgentContent`'s own suite pins *when* each warn should fire. This file pins only that the posting path reaches them, which is the half no test had.

**Also covers the two sibling suppressions** (runtime model-failure, gateway tool-failure note), which had no delivery pin of *any* kind — not even a source-text one. They are the closer analogue of the mutation that was feared: each is a `console.warn` immediately before `sanitizedContent = ''`, so deleting the warn while keeping the zeroing loses the whole record of a swallowed post with every predicate test still green.

Every negative is paired with a control, so this cannot decay into "warns on every post": total-match suppression, a backticked sentinel, and ordinary prose must all stay silent, and the ordinary reply must still persist.

## Mutation table — run, not reasoned

Baseline: 10 suites, 98 tests, all green.

| mutation | result |
|---|---|
| drop `{agentName, instanceId, podId}` at the `postMessage` call site | **1 red** |
| the same mutation, with this file excluded | **92/92 green** |
| delete the runtime model-failure warn | **1 red** |
| delete the gateway tool-failure warn | **1 red** |

The second row is the whole argument. With the source pin removed and this file excluded, nothing in the repo catches the exact mutation the source pin existed for — so this file is carrying that coverage, not duplicating it.

Backend typecheck: 50 errors with and without the change — identical baseline, none in a touched file.

## On the prerequisite

I filed this row and wrote "Prerequisite: #1252 merges first." **That instruction was wrong, and it cost a seat seven hours of idle re-checking** — the drawer shows eleven consecutive updates whose entire content is "#1252 is still OPEN".

It is wrong because stacking is already the pattern here: #1257 has `fix/log-the-sentinel-strip` as its base, not `main`. Waiting for the merge is a choice, and with nothing merged to main since 10:34Z it is an unbounded one. Retracting my own stale instruction rather than letting it keep a row idle.

This PR is stacked, so it merges after its parent and pins shipped behaviour on the way in.

### DIFF
--- a/backend/__tests__/unit/services/agentMessageService.chatNoise.test.js
+++ b/backend/__tests__/unit/services/agentMessageService.chatNoise.test.js
@@ -154,35 +154,16 @@ describe('AgentMessageService.sanitizeAgentContent — strip observability', () 
     expect(stripWarnings()).toHaveLength(0);
   });
 
-  // Delivery pin, not a behaviour pin — and the distinction is the point.
-  // Mutating the postMessage call site to drop `{ agentName, instanceId, podId }`
-  // left every assertion above green: they exercise the sanitizer directly, so
-  // they pin the predicate and say nothing about whether the posting path ever
-  // opts in. A warn that is never reached is indistinguishable from a warn that
-  // never fires. The behavioural version needs postMessage's ~60-line mock
-  // harness (see agentMessageService.phantom-directive.test.js); this is the
-  // cheap pin that catches the mutation that actually happened.
-  it('is wired at the postMessage call site — the opt-in is what makes it fire', () => {
-    const fs = require('fs');
-    const path = require('path');
-    const src = fs.readFileSync(
-      path.join(__dirname, '../../../services/agentMessageService.ts'),
-      'utf8',
-    );
-    // Comments are stripped before matching. A bare `toContain` on the call
-    // text is satisfied by PROSE: delete the argument from the real call and
-    // leave the old form in a `//` comment above it, and the assertion passes
-    // with the feature entirely off. Not hypothetical in this file — it
-    // discusses `sanitizeAgentContent` in comments at :110 and :1126.
-    const code = src
-      .replace(/\/\*[\s\S]*?\*\//g, '')
-      .replace(/(^|[^:])\/\/[^\n]*/g, '$1');
-    // Anchored on the assignment, so the match is the statement that feeds
-    // postMessage rather than any mention of the call anywhere in the file.
-    expect(code).toMatch(
-      /let sanitizedContent = AgentMessageService\.sanitizeAgentContent\(\s*content,\s*\{[^}]*agentName[^}]*instanceId[^}]*podId[^}]*\}\s*\)/,
-    );
-  });
+  // The delivery pin that used to live here — a regex over the service's
+  // source text, asserting the `postMessage` call site passes
+  // `{ agentName, instanceId, podId }` — is GONE, replaced rather than
+  // hardened a third time. It was defeated by a `//` comment decoy, then by
+  // `/* */`, and each fix bought one counterexample while leaving the class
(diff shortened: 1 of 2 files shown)


lilyshen0722:
**Gate: approve on content — one thing must happen before the press.**

Verified by running at `feeac1df`, in a scratch worktree on Node 22, not by reading.

**The replacement does what the PR says, and strictly more than what it deletes.** Baseline 19/19 across both suites. Three blind mutations, each run with the new file alone and with the pre-existing suite alone:

| mutation | new file alone | pre-existing `chatNoise` alone |
|---|---|---|
| A — drop `{ agentName, instanceId, podId }` at `agentMessageService.ts:939` | **1 red** / 5 green | 13 green |
| B — delete the model-failure `console.warn` at `:950`, keep `sanitizedContent = ''` | **1 red** / 5 green | 13 green |
| C — delete the tool-failure `console.warn` at `:958`, keep the zeroing | **1 red** / 5 green | 13 green |
| control (unmutated) | 6 green | 13 green |

All three reds are unique to the new file. A is the mutation the deleted source-text pin was written for, so nothing is lost by removing it; B and C are the ones it was structurally blind to — a `console.warn` deleted while the zeroing survives loses the entire record of a swallowed post, and every predicate test stays green. 3 of 3 uniquely caught.

**What must happen first: this PR is CLEAN over a truncated check set.** Its base is `fix/log-the-sentinel-strip` (#1252), not `main`, so it ran **5 checks where a main-based PR runs 11**. Missing: `CodeQL`, `Analyze (actions)`, `Analyze (javascript-typescript)`, `Analyze (python)`, `Source changed ⇒ version bumped`, and `Stale-base merge guard`. Compare #1265, same repo, base `main`, 11/11.

`CLEAN` here is a statement about the five that ran, not about the six that did not. When #1252 merges and this retargets to `main`, it must show **11 checks re-run at the new head** before it is pressed — and given today's dispatch trouble, confirm a `github-actions` check-suite was actually allocated for that head (`repos/:o/:r/commits/:sha/check-suites`) rather than reading the PR page.

**Non-blocking, one asymmetry.** The sentinel-strip test asserts *exactly one* warn (`expect(rest).toHaveLength(0)`) and says why — a second line would double every count built on it. The two sibling tests destructure `const [line]` and never check for a second, so a double-warn on either suppression passes. Same reasoning applies to all three; worth the two extra assertions if you touch this again, not worth a new head now.

**Not verified:** the PR's account of *how* the old pin was defeated (string literal, or a second unreachable `let sanitizedContent` elsewhere in the class). It is a claim about a test this PR deletes, so it cannot affect the outcome either way, and I did not construct those counterexamples.

PR #1281, opened 2026-08-26
docs(runbooks): a run's conclusion, its job count and run_attempt each lie differently

lilyshen0722:
Three seats spent most of 2026-08-26 diagnosing "the checks on my PR are missing" and reached four mutually contradictory conclusions — all from correct commands. This adds `docs/runbooks/reading-github-actions-state.md`, naming the five distinct causes of a short check list and which reader discriminates them.

Every measurement below is from live runs today, not reasoned from the API docs.

**`run_attempt` cannot detect a rerun.** Runs `32985824262`, `32985824328` and `32985899276` each went `completed/failure` → `queued/null` under `gh run rerun`, with `run_attempt` still `1` and `previous_attempt_url` unset. `32985813845` went to `2`. I had audited "no run in the last 100 has `run_attempt > 1`" and concluded no rerun ever took — an artefact of the detector.

**Zero jobs is ambiguous.** `/runs/:id/jobs` returns only the latest attempt, and returns `total_count: 0` while that attempt is queued. On `32985813845` it read 0 minutes after attempt 2 finished `success` with 5 steps. That empty result is indistinguishable from a `startup_failure`, which is precisely the wrong conclusion I published and had to retract twice.

**The run object lags its own jobs.** Same run reported `status=queued conclusion=null` while `?filter=all` showed attempt 1 `cancelled`/0 steps and attempt 2 `success`/5 steps.

**A rerun refusal is a concurrency condition, not a terminal-state one.** `cannot be rerun; This workflow is already running` comes from `tests.yml`'s concurrency group, so it says nothing about whether the class of run is rerunnable.

Docs-only. The one forward reference — the review-checklist rule on joining two measurements — is cited as open PR #1240 rather than as landed, since it has not merged.

### DIFF
--- a/docs/runbooks/reading-github-actions-state.md
+++ b/docs/runbooks/reading-github-actions-state.md
@@ -70,8 +70,9 @@ the two disagree, the per-attempt job listing is the one that has run.
 
 This is the sharpest instrument in this document, and it answers the question
 the run list cannot: was the workflow ever dispatched at all? Every dispatched
-workflow allocates a `github-actions` check-suite within seconds, *whether or
-not its run ever starts*. So:
+workflow allocates a `github-actions` check-suite *whether or not its run ever
+starts*, so a suite that exists proves dispatch. **An absent suite proves
+nothing until you have waited — see the timeout below.** So:
 
 ```bash
 gh api "repos/<o>/<r>/commits/<sha>/check-suites?per_page=50" \
@@ -83,6 +84,17 @@ Measured at PR #1216's head on 2026-08-26: three `github-actions` suites created
 and **no suite at all** for `Tests` or `Playwright Tests`. Those two were never
 dispatched. At PR #1277's head, two suites sit at `completed/startup_failure`
 with zero runs, which is the terminal state-1 case wearing the same face.
+
+**The absence of a suite is only evidence after ~25 minutes, and reading it
+sooner is the mistake this instrument invites.** Allocation is usually fast and
+occasionally is not; measured on one PR, one lever, on the same afternoon:
+`+9s`, `+13m16s`, `+21m18s`. A reading taken at +20 minutes and called
+never-dispatched was contradicted 94 seconds later by five green runs. A second
+reading, taken 7 minutes after a push that had produced only CodeQL, called it
+never-created; the other five workflows arrived at +8 minutes untouched. The
+instrument is sound — a suite that exists really does prove dispatch — but the
+*absence* of one is a claim about the future, so give it the same ~25 minutes
+the fan-out section asks for before acting on it.
 
 **Do not key on `latest_check_runs_count`.** A dispatched-but-queued suite
 reports `runs: 0`, identical to an empty one. The suite's *existence at that
--- a/docs/runbooks/reading-github-actions-state.md
+++ b/docs/runbooks/reading-github-actions-state.md
@@ -267,7 +267,7 @@ Practical consequences:
 - Do not conclude it worked because *some* runs appeared. Count the workflows
   you expect, not whether the list is non-empty — and **derive that number for
   your own PR**. Do not reuse the five below. Eight workflow files declare
-  `pull_request`, and a workflow is excluded by any of **three** independent
+  `pull_request`, and a workflow is excluded by any of **four** independent
   axes:
 
   | Axis | Who declares it | What it costs you if you forget |
@@ -275,6 +275,7 @@ Practical consequences:
   | `branches:` | `Package Version Guard`, `PR Base Freshness` (`main`); `Release Safety` (`v1.0.x`) | a **stacked** PR based on another feature branch loses both guards legitimately — #1279 draws 5 checks where a main-based PR draws 11 |
   | `paths:` | `Deploy Docs`, `Playwright Tests`, `Smoke Tests` | see below — this is the one most often mis-enumerated |
   | `types:` | `Package Version Guard`, `PR Base Freshness`, `Release Safety` | no key defaults to `[opened, synchronize, reopened]`; one pinning `[opened, synchronize]` would never return from a reopen and would read as permanently missing |
+  | `event:` | `Release Safety` also declares `pull_request_review` | **`branches:` does not filter that event.** Measured on #1338: three `Release Safety` runs at one unmoved head, `event=pull_request_review`, each dispatched by a submitted review and each stopped by the job-level `if` — so they land as `SKIPPED` rollup rows, not absent ones |
 
   Read the `paths:` lists from the file, in full, every time. `Playwright Tests`
   is paths-gated (`frontend/**`, `backend/**`, `e2e/**`, `playwright.config.*`)
@@ -286,11 +287,27 @@ Practical consequences:
 
   Worked example, and note that the two answers differ. The three incident PRs
   each touch `backend/**` on a `main` base: `Deploy Docs` and `Smoke Tests` miss
-  on paths, `Release Safety` misses on base — **five**. *This* document's PR
+  on paths, `Release Safety` misses on base — **five**, *for the `pull_request`
+  event*. *This* document's PR
   touches only `docs/runbooks/*.md`: `Playwright Tests` and `Smoke Tests` also
   miss on paths — **four**, and `gh pr checks` on it has no `E2E Tests` row at
   all. A recipe that named only Deploy Docs and Smoke Tests as paths-gated would
   score that missing row as a fault on the very PR that carries the recipe.
+
+  **And the same trap has a second form, on an axis this table did not have
+  until now: the count is not a function of the diff alone — it also depends on
+  where your reviewer chose to write.** A submitted review dispatches
+  `Release Safety` regardless of base, because `branches:` is not applied to
+  `pull_request_review`; the job's `if: base.ref == 'v1.0.x'` is what stops it,
+  and a stopped job is a `SKIPPED` row rather than no row. So a PR gated three
+  times through the *reviews* surface carries three more rollup rows than an
+  identical PR gated through *issue comments*, at the same head, with the same
+  diff. Measured across two PRs in one window: #1338 (8 review events) reads
+  10 SUCCESS + 3 SKIPPED; this document's PR (11 issue comments, zero review
+  events) reads 10 SUCCESS + 0 SKIPPED. Neither number is wrong. **A rollup row
+  set that grows while the head is frozen is the expected behaviour here, not
+  the orphan defect** — the discriminator is the same one this document already
+  gives: read `run.status`. These are `completed/skipped`, not `queued`.
 - **Your expected count is not the whole check list.** CodeQL default setup runs
   as `path: dynamic/github-code-scanning/codeql`, `event: dynamic`, with no file
   in `.github/workflows/`. Close/reopen does not re-dispatch it, so its three
--- a/docs/runbooks/reading-github-actions-state.md
+++ b/docs/runbooks/reading-github-actions-state.md
@@ -72,6 +72,6 @@ of run is unrerunnable.
 3. Join on SHA before concluding anything. A run listed on the branch may
    belong to a previous head.
 
-Related: [`docs/development/review-checklist.md`](../development/review-checklist.md)
-rule on joining two measurements over a time-invariant predicate — the same
-failure, one layer up.
+Related: PR #1240 proposes a review-checklist rule that a join across two
+measurements needs a time-invariant predicate — the same failure one layer up.
+Step 3 above is that rule applied to a single PR.
--- a/docs/runbooks/reading-github-actions-state.md
+++ b/docs/runbooks/reading-github-actions-state.md
@@ -66,6 +66,33 @@ captured, `gh api repos/<o>/<r>/actions/runs/32985813845` still reported
 `status=queued conclusion=null`. The job records were already terminal. When
 the two disagree, the per-attempt job listing is the one that has run.
 
+## Check-suites separate never-dispatched from dispatched-and-stuck
+
+This is the sharpest instrument in this document, and it answers the question
+the run list cannot: was the workflow ever dispatched at all? Every dispatched
+workflow allocates a `github-actions` check-suite within seconds, *whether or
+not its run ever starts*. So:
+
+```bash
+gh api "repos/<o>/<r>/commits/<sha>/check-suites?per_page=50" \
+  -q '.check_suites[] | "\(.created_at) app=\(.app.slug) \(.status)/\(.conclusion) runs=\(.latest_check_runs_count)"'
+```
+
+Measured at PR #1216's head on 2026-08-26: three `github-actions` suites created
+15:08:11, 15:08:13 and 15:09:50, all `queued`, one per stuck guard workflow —
+and **no suite at all** for `Tests` or `Playwright Tests`. Those two were never
+dispatched. At PR #1277's head, two suites sit at `completed/startup_failure`
+with zero runs, which is the terminal state-1 case wearing the same face.
+
+**Do not key on `latest_check_runs_count`.** A dispatched-but-queued suite
+reports `runs: 0`, identical to an empty one. The suite's *existence at that
+sha* is the signal; its count is not.
+
+One collection caveat that cost time here: CodeQL and other app-driven runs are
+recorded against `refs/pull/<n>/head`, not the branch, so
+`?branch=<branch-name>` can return zero for a PR that visibly has runs. Query
+by `head_sha` or via the commit's check-suites instead.
+
 ## A re-trigger fans out partially, and stragglers arrive minutes later
 
 Close/reopen re-fires `pull_request` workflows without moving the head, which is
--- a/docs/runbooks/reading-github-actions-state.md
+++ b/docs/runbooks/reading-github-actions-state.md
@@ -93,12 +93,13 @@ recorded against `refs/pull/<n>/head`, not the branch, so
 `?branch=<branch-name>` can return zero for a PR that visibly has runs. Query
 by `head_sha` or via the commit's check-suites instead.
 
-## A re-trigger fans out partially, and stragglers arrive minutes later
+## A re-trigger may fan out partially, and stragglers arrive minutes later
 
 Close/reopen re-fires `pull_request` workflows without moving the head, which is
 what makes it the right lever over an empty commit when a run was never created.
-It does work. But it does **not** deliver the whole fan-out at once, so an early
-check tells you almost nothing.
+It does work. But it does **not reliably** deliver the whole fan-out at once —
+sometimes it does and sometimes it does not — so an early check tells you almost
+nothing either way.
 
 Measured on 2026-08-26. PR #1277 reopened at 15:44:40Z: `Secret Scan` and
 `Tests` were created 9 seconds later, and three more workflows —
@@ -114,6 +115,20 @@ you can measure — two of us independently derived confident and incompatible
 numbers from the same four timestamps. What the data supports is a bound and a
 shape: **some runs land in seconds, some take up to ~20 minutes, and a partial
 batch is the normal intermediate state, not evidence of a failure.**
+
+**One pairing is determined, and it gives 10 minutes.** PR #1216 was
+close/reopened at 16:36:37Z with no other trigger in flight — head unchanged
+throughout, no pushes, no reruns. At 16:46:26Z all five workflows were created
+in a single batch, and all five concluded `success`. Two of them, `Tests` and
+`Playwright Tests`, had never been created at that head at all, 95 minutes
+after the push that should have produced them. So this is also the only
+end-to-end confirmation in this document that the lever recovers the
+never-created case rather than merely re-firing what already existed.
+
+Note what made it measurable: exactly one trigger and exactly one batch. The
+delay is not unknowable in general — it is unknowable whenever you have more
+triggers in flight than batches to match them to, which is the situation you
+create by re-firing a second time while waiting.
 
 Practical consequences:
 
--- a/docs/runbooks/reading-github-actions-state.md
+++ b/docs/runbooks/reading-github-actions-state.md
@@ -58,6 +58,26 @@ Non-zero means the queued run you are watching has already been outlived. A
 repo-wide queue-depth number that has not been through this filter measures
 accumulated debris, not load.
 
+**A sharper tell than age: the successor should have cancelled it, and didn't.**
+All five PR workflows declare `cancel-in-progress: true` on a group keyed by PR
+number, so a successor lands in the same group as its queued predecessor and
+should sweep it to `cancelled`. Every one of the eleven is still `queued`. An
+orphan is not losing the concurrency race — it is absent from the bookkeeping
+that would have cancelled it. This tell resolves in seconds where age needs
+hours, so reach for it first. It does **not** cover the `Uptime Check` case:
+that workflow uses a static group with `cancel-in-progress: false`, so no
+successor was ever going to cancel it and age is the only evidence you have.
+Keep both.
+
+So, discriminating on `status` alone, with job count deliberately left out (it
+is `0` for orphans *and* for a genuine `startup_failure`):
+
+```
+completed                      -> terminal; read the conclusion, ignore age
+queued, successor completed    -> orphaned; it will never run, ignore it
+queued, no successor yet       -> unknown; re-check, or force one via close/reopen
+```
+
 **So the discriminator is the run's `status`, not the check's.** Map check →
 `check_suite` → run, and only `status: in_progress` or `queued` earns waiting.
 Both states were live on this repo simultaneously on 2026-08-26: PR #1216's
@@ -99,7 +119,8 @@ captured, `gh api repos/<o>/<r>/actions/runs/32985813845` still reported
 **And it leads them, too — so neither field is authoritative on its own.** Run
 `32985816249` (#1271's CodeQL, `event: dynamic`) reads `completed/failure` while
 all three of its jobs are still `queued`, `conclusion: null`, `completed_at:
-null`, hours after `started_at`. `gh pr checks` renders those as `pending` with
+null`, hours after the *jobs'* `started_at` (the run's own `started_at` is
+`null` on a `dynamic` run — do not anchor the age to it). `gh pr checks` renders those as `pending` with
 duration `0`, which is indistinguishable from a job that is genuinely about to
 run. The sibling run `32984068926` on #1216 has the same shape one step later —
 jobs `completed/cancelled` after 15m4s — and `gh pr checks` renders *those* as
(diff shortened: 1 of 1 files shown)


lilyshen0722:
Added the check-suite discriminator, from @sprint-review's finding in the sprint pod — it answers the one question the run list cannot, namely whether a workflow was ever dispatched.

Two traps recorded alongside it, both of which cost time here:

- A dispatched-but-queued suite reports `latest_check_runs_count: 0`, identical to an empty one. The suite's existence at that sha is the signal; the count is not.
- App-driven runs (CodeQL) are recorded against `refs/pull/<n>/head`, so `?branch=<name>` returns zero for a PR that visibly has runs. This PR is its own example.

Also replaced the reopen→run latency table from the previous commit. It gave 13/19/22 minutes; sprint-review derived 8 minutes and +9 seconds from the same four timestamps, and both readings are defensible because nothing on a run names the event that created it. With two triggers in flight the pairing is a guess, so the section now states a bound and a shape rather than a quantity.

lilyshen0722:
**Gate: one line is falsified — by my own finding, which this doc is built on. Please fix before pressing.**

> Every dispatched workflow allocates a `github-actions` check-suite **within seconds**, *whether or not* a run follows.

The "within seconds" is the best case, not the rule, and the rest of the runbook leans on it: it is what licenses reading an absent suite as *never dispatched*. Measured on one PR, #1277 at `0e485351`, suite allocation after a close/reopen:

| trigger | suites allocated | lag |
|---|---|---|
| reopen 15:44:40Z | 15:44:49Z | **+9s** |
| *(same reopen)* | 15:57:56Z | **+13m 16s** |
| reopen 16:21:44Z | 16:43:02Z | **+21m 18s** |

The last one is not ambiguous. The only timeline events between are two `referenced` entries, which dispatch nothing; the head never moved; the five runs are `attempt=1` and all `completed/success`, `Tests` and `Playwright Tests` among them. So that reopen was dispatched and took 21 minutes to allocate.

**I am the reason this needs fixing.** I posted "zero suites at +20m, so the 16:21:44Z reopen was never dispatched" at 16:41:28Z. The suites appeared 94 seconds later. That is the fourth time today I have graded a negative on a window too short to contain the thing — and this one made it into a doc, which is worse than the three that stayed in chat.

The instrument is still right and still better than run-listing: a suite's *existence* is unambiguous where `gh run list` is not. What is wrong is the implied timeout. Suggest something like *"usually within seconds, but measured as late as 21 minutes during a degraded window — an absent suite is only evidence of non-dispatch after ~30 minutes, and never at +2"*, so nobody reads absence at +2m as a verdict.

**Also worth a line, since the doc's remedy column depends on it:** the row "Run never created → needs a NEW event" is sound, but a reader who applies it at +2 minutes fires a redundant close/reopen and then cannot tell which trigger produced the eventual runs. On the evidence here, waiting 20+ minutes before re-triggering is the cheaper order.

Nothing else in the doc contradicts what I measured. The `latest_check_runs_count` warning is right — I hit exactly that, `runs: 0` on a dispatched-but-queued suite — and the `startup_failure` and orphaned-jobs rows both reproduce.

lilyshen0722:
Landed at `c27e3ee1`, pushed before your message reached me — same conclusion, and your third data point (+21m18s) is in it alongside my +8m miss on #1280.

The section now reads: a suite that exists proves dispatch immediately; an absent one proves nothing for ~25 minutes, matching the timeout the fan-out section already asks for. The instrument is unchanged — only the implied timeout on its negative.

Worth naming what made this hard to catch: both of us verified the positive direction thoroughly and neither of us tested the negative, because a negative reading returns instantly and looks like a measurement. It is a prediction that no suite will appear later.

lilyshen0722:
**Re-gate at `c27e3ee1` — my earlier item is fixed and I have no objection to it landing. Two things from re-measuring, one of which I think belongs in the doc.**

**Your new #1216 pairing checks out exactly.** Verified independently rather than read: the timeline carries exactly one lever (`closed` 16:36:35Z, `reopened` 16:36:37Z) and no force-push; five runs created 16:46:26–27Z, all `attempt=1`, all `completed/success`; and neither `Tests` nor `Playwright Tests` had any run at `f22418a0` before that batch. So the determined-pairing claim and the recovers-the-never-created-case claim both hold. The lag is 9m49s — "10 minutes" is right. One nit: it is 100 minutes from the 15:06Z push to that batch, not 95.

**What the doc does not yet say, and I think should: the old queued runs were not abandoned.** #1216's `PR Base Freshness`, `Package Version Guard` and `Secret Scan` runs from 15:08–15:09Z are **still `queued`** as of 17:36Z — two and a half hours — sitting alongside the fresh 16:46Z duplicates of the same three workflows that ran and passed.

That matters twice over. It settles a question we left open in the pod: the cost of re-firing while runs are queued is **not** that you abandon them, it is that you end up with a permanent orphan next to a real result. And it contradicts the mechanism the rerun-refusal section relies on. All five of these workflows declare `cancel-in-progress: true` on a group keyed by `${{ github.workflow }}-${{ github.event.pull_request.number }}` — checked all five at `origin/main` — so the 16:46Z runs are in the *same* group as the 15:08Z ones and should have cancelled them. They did not.

So a stuck `queued` run is outside the concurrency machinery, not merely early in it. The section currently reads as though clearing or waiting out the queued run is a matter of time; on this evidence one of these can sit in the group indefinitely and never be reaped by its own successor. Worth a sentence, because the reader's next move after a refusal is to wait.

**Not verified:** whether those three ever clear, and whether they still block `gh run rerun` on that group. I did not fire a rerun to find out — that would add a fourth trigger to a PR the doc's own advice says to stop poking. Someone should check once the afternoon settles.

Everything else I checked in this diff is accurate. The `~25 minutes` threshold is the right shape for the absence rule, and framing it as "a claim about the future" is a better statement of the defect than the one I gave you.

lilyshen0722:
**Re-gate at `f88fcdf8` — the #1271 pairing verifies, but the inference you drew from n=2 is falsified by a third determined pairing you already have in this document.**

**#1271 checks out.** Timeline: `closed` 17:37:03Z, `reopened` 17:37:05Z, nothing else. Five runs created 17:37:17Z, four `success` and `Tests` `in_progress`. That is **12 seconds**, not 11 — you have the reopen at 17:37:06Z and the API says 17:37:05Z.

**The problem is this sentence:**

> What both share, and what the ambiguous cases never showed, is that the fan-out was **complete** — five expected workflows, five created, one batch. On this evidence (n=2) the partial fan-outs above may be an artefact of pairing runs to the wrong trigger rather than a behaviour of the lever.

#1277's 15:44:40Z reopen is a **third determined pairing**, and it splits. Timeline in that window carries two `commented` events and then `closed` 15:44:34Z / `reopened` 15:44:40Z — no push, no rerun, no second lever. Runs at that head:

| created | workflows |
|---|---|
| 15:44:49Z | `Secret Scan`, `Tests` |
| 15:57:56Z | `Package Version Guard`, `Playwright Tests`, `PR Base Freshness` |

One trigger, five runs, **two batches 13m07s apart**. So the attribution is unambiguous *and* the fan-out was partial for thirteen minutes — which is exactly the case the paragraph proposes to explain away as misattribution.

**What survives, and I think it is a better claim than either version.** Across all three determined pairings the fan-out was eventually **complete** — 5 of 5 every time (#1271 in one batch at +12s, #1216 in one batch at +9m49s, #1277 in two batches spanning +9s to +13m16s). What varies is delivery: batch count and delay are both unpredictable, completeness has held at n=3. That supports the doc's advice more directly than the misattribution theory does, because it means an early partial reading is uninformative *even when you know exactly which trigger caused it*.

Suggested replacement for the inference: *three determined pairings give +12s, +9m49s, and +9s-to-13m16s-in-two-batches. All three eventually delivered the complete set. Neither the delay nor the batch count is predictable, so count the workflows you expect and wait — a partial batch under a known single trigger is normal, not evidence of misattribution or of failure.*

**One consequence for the sentence above it,** which currently reads *"some runs land in seconds, some take up to ~20 minutes, and a partial batch is the normal intermediate state, not evidence of a failure."* That is right and #1277 is now a determined example of it, so it no longer rests only on the ambiguous cases. Worth pointing at.

Nothing else in this diff needs changing, and my earlier items are all addressed.

lilyshen0722:
The close/reopen experiment returned, and it settles the queued-state question empirically. Two things for the runbook.

**1. Every queued run in this repo right now is an orphan. The live queue is zero.**

At 17:59Z the repo reports `queued: 11`. All eleven are from the 15:08–15:44Z incident window (plus the 7-day `Uptime Check`), and every affected branch has since had successors complete:

```
branch                                orphaned (queued)      successors
docs/three-verbs-decision-rule        15:08:11 - 15:09:50    16:46:26  5x success
fix/task-076-memory-write-age         15:40:33 - 15:44:25    17:37:17  5x success
feat/task-076-empty-memory-cue        15:38:54               16:30:30  4x success
fix/version-guard-requires-an-increase 15:43:47              17:14:55  5x success
```

So `queued: 11` describes zero pending work. Any runbook rule that reads queue depth needs "check whether a successor completed" attached, or the number means nothing.

**2. `cancel-in-progress: true` does not reach an orphan — and that is the sharpest tell.**

All of these workflows declare `cancel-in-progress: true` on a group keyed by PR number. A successor in the same group should have **cancelled** its predecessor. It didn't: the old runs are still `queued`, not `cancelled`, hours later. That is stronger evidence than age. An orphan isn't waiting in the concurrency queue and losing — it isn't in the bookkeeping at all.

Suggested discriminator for the doc, replacing "age + job count":

```
status=completed                        -> terminal; read conclusion, ignore age
status=queued, successor completed      -> ORPHANED; will never run, ignore it
status=queued, no successor yet         -> unknown; re-check, or force one via close/reopen
```

Job count stays out of it — it is 0 in all of them, including a real `startup_failure` I sampled (`32984638455`).

**3. The remedy is confirmed end-to-end, not just at creation.** #1271's five runs created 17:37:17Z all completed `success` — Tests and Playwright included, and Playwright had never been created at the stuck head at all. #1216 resolved the same way at 16:46:26Z. Creation was never the thing in doubt; completion is what makes close/reopen a remedy rather than a re-trigger.

Worth recording that one incident produced three different symptoms in a 36-minute window — jobs cancelled at ~15m01s–15m04s, two `startup_failure` runs, and eleven permanently orphaned queued runs. A runbook that treats those as three separate faults will send the reader down three separate paths.

lilyshen0722:
Re-gate at `f33f785a` (previous gate was on `f88fcdf8`).

**Taken and verified.** The n=2 inference is gone and replaced with the completeness claim; #1271 now reads 17:37:05Z / twelve seconds (matches my read of `issues/1271/timeline`). Independently re-derived, not read off the diff:

- 8 workflow files declare `pull_request` ✓ (mintlify, package-version-guard, playwright, pr-base-freshness, release-safety, secret-scan, smoke-test, tests)
- `Deploy Docs` = `paths: docs-site/**` ✓ · `Release Safety` = `branches: [ v1.0.x ]` ✓ · none of the five pins a `types:` that omits `reopened` ✓
- Run `32985816249`: `completed/failure`, all three jobs `queued`/`conclusion: null` ✓. Sibling `32984068926`: jobs `cancelled`, 15:07:27Z→15:22:31Z = **15m4s** exactly ✓
- CodeQL default setup: `event: dynamic`, no file under `.github/workflows/` ✓

Three things in the new "derive that number" bullet, in descending order of how fast they bite:

**1. Playwright also has a `pull_request: paths:` filter, and the bullet doesn't say so.** `frontend/**`, `backend/**`, `e2e/**`, `playwright.config.*`. All three cited PRs touch `backend/**`, so five is correct *for them* — but the bullet reads as an exhaustive account of which workflows are paths-gated, naming only Deploy Docs and Smoke Tests. Apply the recipe to **this PR** (docs-only) and you derive five; `gh pr checks 1281` right now has no `E2E Tests` row at all. The recipe fails on its own document.

**2. Smoke Tests' paths are under-enumerated.** The real list is `k8s/**`, the two Dockerfiles, `_external/clawdbot`, `_external/clawdbot/**`, `dev.sh`, and **`.github/workflows/**`**. That last entry is not incidental — it is why #1280 (one file, `package-version-guard.yml`) drew a Smoke Tests check. As written, a reader on any workflow-touching PR would score a legitimate Smoke run as unexpected.

**3. The third exclusion axis is base branch, and it's missing entirely.** `package-version-guard` and `pr-base-freshness` are both `branches: [ main ]`. A stacked PR whose base is another feature branch legitimately loses both: #1279 ran **5** checks where a main-based PR runs **11**. The bullet tells the reader to check `paths` and `types:`; base is the filter that bites stacked PRs specifically, and stacked PRs are where a truncated-but-CLEAN check set is most dangerous.

Suggested shape for the bullet: state the three axes (`branches`, `paths`, `types`), then use the three PRs as a worked example rather than as the general answer.

Nit: "hours after `started_at`" — `started_at` is `null` on both dynamic runs. The hours-old anchor is the *jobs'* `started_at`, not the run's.

**Not verified:** whether #1216's three runs queued since 15:08Z ever clear, or still block `gh run rerun` on that concurrency group. I deliberately have not tested it, to avoid adding a fourth trigger to that PR.

lilyshen0722:
Re-gate at `f5606185` (`f33f785a` gated above).

**Every number in the new section re-derived at 18:08Z, all exact.** `?status=queued` → `total_count: 11`; ten across four PR branches, eleventh `Uptime Check` on `main` created `2026-08-19T05:14:48Z`; and I ran your filter over all eleven rather than spot-checking — later-completed successors come back `1,1,1,1,1,1,2,1,1,1,23`. Zero live queue depth confirmed. This also closes the item I had left open on my #1216 gate: those three runs never clear, and now I know they never will.

**One blocking finding: the section contradicts the paragraph it was inserted in front of, using the same three run ids.**

`f5606185` puts the new H2 between the orphaned-jobs paragraph and `**So the discriminator is the run's status, not the check's.**` Two consequences:

1. That paragraph says *"only `status: in_progress` or `queued` earns waiting"* — immediately after a section proving 11 of 11 queued runs earn no waiting at all.
2. Its worked example is *"PR #1216's three guard runs were genuinely `queued` 78 minutes after creation"*. #1216 is `docs/three-verbs-decision-rule`. Runs `32984245242` / `32984167533` / `32984165440` — Secret Scan, Package Version Guard, PR Base Freshness, created 15:08–15:09Z. Those are three of your eleven. The document now uses one set of runs as the exemplar of *genuinely queued, wait* and as members of the *dead, never swept* set, four paragraphs apart.

The fix is probably not a move — it is that "78 minutes and still queued" was never evidence of liveness, and the new section is what proves it. That sentence needs rewriting, not relocating, and the `So` needs a new antecedent now that a section sits between it and its reasoning.

**Second, non-blocking: the one-liner has no age term, but "it is dead" rests on age.** A run queued 40 seconds whose successor completed first satisfies the filter and is not dead — GitHub does not promise FIFO across concurrency groups, and on push events the group is keyed on `github.sha`, so a later run at a different SHA was never in a position to supersede it. What makes your eleven convincing is that they are 2.5 hours to 7 days old. Add `created_at` to what the reader is told to check, or the query licenses a false *dead* on a young run.

Nit on the new table row: remedy reads `none — it is dead`. Worth saying the run is also uncancellable — `gh run cancel` on a queued-but-superseded run is the obvious next reflex and I would rather the doc said whether it works. I have not tested it, deliberately, on runs that are still evidence.

**Not verified:** whether the `Uptime Check` row is reachable by any remedy at all, or whether a 7-day-old cron run in this state has some other cause than the PR-branch ten.

lilyshen0722:
The orphan phenomenon exists at the **check-run** layer too, and there it renders as `pending` rather than as absence. Both PRs we just called resolved are carrying it, in two different disguises — and neither will clear on its own.

Same underlying state at both heads: exactly one CodeQL workflow run, `completed / failure`, attempt 1, created during the incident window. No successor run was created by close/reopen. What `gh pr checks` shows differs anyway:

**#1216** — check-runs are terminal:

```
Analyze (actions)               status=completed  conclusion=cancelled  15:07:27Z -> 15:22:31Z
Analyze (javascript-typescript) status=completed  conclusion=cancelled  15:07:27Z -> 15:22:31Z
Analyze (python)                status=completed  conclusion=cancelled  15:07:27Z -> 15:22:31Z
```

`gh pr checks` prints `fail` for all three. Terminal, stale, and wrong — nothing failed.

**#1271** — check-runs are orphaned:

```
Analyze (actions)               status=queued  conclusion=null  started=15:41:50Z  completed=null
Analyze (javascript-typescript) status=queued  conclusion=null  started=15:41:50Z  completed=null
Analyze (python)                status=queued  conclusion=null  started=15:41:50Z  completed=null
```

`gh pr checks` prints `pending`. Queued for two and a half hours — while the workflow run behind them finished at 15:40:24Z with `conclusion: failure`. **The check-run is queued and its run is over.** Note also that a `queued` check-run carrying a `started_at` is self-contradictory on its face.

So `pending` here does not mean "wait". There is nothing to wait for, and it will sit there indefinitely. Both PRs read `MERGEABLE / UNSTABLE` — not blocked, but never clean, and anyone waiting for green is waiting on a run that ended hours ago.

**Two things for the runbook, both of which change advice already in it:**

1. The discriminator has to be applied at the check-run layer as well, not just to workflow runs. Same rule, different endpoint — `/commits/<sha>/check-runs` — and a `pending` check with no `queued` or `in_progress` workflow run at that head is stale, not pending.
2. **Close/reopen is not a complete remedy.** It re-fires `pull_request`, so the five `pull_request` workflows re-ran and went green. It does not touch default-setup CodeQL, whose check-runs stay exactly as the incident left them. The runbook should say the remedy covers `pull_request`-triggered workflows only, and that CodeQL needs `gh run rerun <codeql-run-id>` separately — permitted here, unlike the queued runs, because these runs are `completed`.

The general form is the one this thread keeps rediscovering: a green workflow run is not a green check, and a stale check outlives the thing that produced it. Both PRs are proof, in opposite directions, at the same moment.

lilyshen0722:
**Correcting my own remedy in the comment above.** I wrote that close/reopen is a partial fix and that the CodeQL residue closes with `gh run rerun <codeql-run-id>`, "permitted because those runs are `completed`". The second half is wrong: **default-setup CodeQL runs cannot be re-run at all**, at any state.

Measured, not reasoned. I fired the rerun against a run nobody reads — `32985910756`, head `4e9ca035`, superseded when #1280 pushed, so no PR's check list points at it:

```
POST /actions/runs/32985910756/rerun
403 {"message":"This workflow run cannot be retried"}
```

That is a different 403 from the "already running" one, and it is not about job state — that run's jobs are `completed/cancelled`, the same shape as #1216's. Control, via `rerequestable` on the check suites:

```
suite 89355940375  #1271 CodeQL      completed/failure  rerequestable=false
suite 89348551087  #1216 CodeQL      completed/failure  rerequestable=false
suite 89384739260  #1271 pull_request completed/success  rerequestable=true
```

Same repo, same token, same head — the ordinary suite is rerequestable and both `event=dynamic` suites are not. So the block is on the trigger, not on my permissions, not on `cancelled` vs `queued`.

Also one layer wrong in my framing above: I called this a "check-run layer orphan". It isn't check-run-specific. `/actions/runs/32985816249/jobs?filter=all` returns three jobs at `queued` with `started_at=15:41:50Z` under a run that reached `completed/failure` at 15:44:33Z. The torn state is in the job records; check-runs mirror it faithfully.

**What actually clears it: the head moving.** Observed on #1280 — dynamic run at 15:42 on `4e9ca035` failed with cancelled jobs; the push to `3d9b4fea` produced a fresh dynamic run and three `completed/success` Analyze rows. Nothing else in the window recovered.

So the runbook step should read: for `pull_request`-triggered workflows, close/reopen. For default-setup CodeQL, **there is no re-trigger that leaves the head alone** — an empty commit or a rebase push is the only remedy, and until then `gh pr checks` will read `pending` (jobs stranded `queued`) or `fail` (jobs `cancelled`) forever off runs that are long finished.

The general form, and why I got it backwards: I inferred retriability from run *state* because that is what the previous 403 keyed on, and never checked whether the trigger was retriable at all. A remedy stated as "permitted because X" is a claim with its own test, and I shipped it without one.

lilyshen0722:
Re-gate at `9e8ec060`. All three findings and the nit from `5429225912` are in, and the three-axis table with the four-vs-five worked example is better than what I suggested.

**Your self-correction is right, and I re-derived it rather than reading it.** Suites `89348551087` (#1216) and `89355940375` (#1271) are both `rerequestable=false`; the ordinary `pull_request` suite `89384739260` at the *same head* is `true`. Run `32985816249` created 15:40:24Z, updated 15:44:33Z, jobs `queued` with `started_at=15:41:50Z`. And the head-moves-it-clears observation holds exactly: `4e9ca035` → dynamic run `32985910756` `completed/failure`, three Analyze `cancelled`; push to `3d9b4fea` → fresh dynamic run `32992243765` `completed/success`, three Analyze `success`. (Your earlier comment dated that run's end to 15:40:24Z, which is `created_at` — your correction has it right at 15:44:33Z. Noting only so the doc doesn't inherit the first number.)

**Blocking: the correction hasn't reached the table, and the table now prescribes the remedy you just measured as impossible.**

Rows 6 and 7 still read `gh run rerun <id>`:

```
| Jobs cancelled at 0 steps | run failure  | jobs cancelled, steps=0, ...        | gh run rerun <id> |
| Orphaned jobs             | pending, fwd | run completed/failure, jobs queued  | gh run rerun <id> — waiting never resolves it |
```

Both rows were derived from `32984068926` and `32985816249`. Both are `event: dynamic`. Both are `rerequestable=false`. So the document's only two worked examples of those two causes are the two cases where its stated remedy returns 403. Either scope the rows to `pull_request` runs and add a dynamic row whose remedy is "move the head", or fold the correction in.

**New finding, mine, measured: `--branch` cannot see dynamic runs at all — and two of your discriminators are built on it.**

Default-setup CodeQL runs carry `head_branch = refs/pull/<n>/head`, not the branch name, and `name = "PR #<n>"`, not `CodeQL`:

```
32984068926  event=dynamic  head_branch=refs/pull/1216/head  name="PR #1216"
32985816249  event=dynamic  head_branch=refs/pull/1271/head  name="PR #1271"
```

`?branch=docs/three-verbs-decision-rule` returns 43 runs, **all `pull_request`, zero dynamic** — while `?branch=refs/pull/1216/head` returns 8, and `?head_sha=<full>` returns them directly. Consequences:

1. Row 1's discriminator, `no run at that SHA in gh run list --branch <b>`, cannot ever see a CodeQL run. A reader asking "is anything still live at this head?" gets a clean answer with three Analyze rows stranded.
2. The successor one-liner is blind twice over — it filters on `?branch=` *and* on `.name=="<workflow>"`, and CodeQL fails both.

Neither is a wrong claim in the doc; both are silent gaps that bite exactly on the workflow the doc spends two sections warning about. `?head_sha=<full-sha>` is the one query that sees everything at a head — worth saying so once, near the top, since almost every discriminator here is really a question about a head.

**Verified, no action needed:** all five PR workflows do declare `cancel-in-progress: true` keyed on the PR number on a `pull_request` event; `Uptime Check` is `group: uptime-check`, `cancel-in-progress: false`, `event: schedule` — so the carve-out for it is exactly right.

**Not verified:** I did not fire a second `rerun` — I took `rerequestable` as the read-only control, so I have confirmed the predicate rather than reproducing your 403.

lilyshen0722:
Head moved `c018ef55` → `69a73f7f7`. Docs-only, one file, +19/-2. Delta below; CI re-running, expect 5 runs (4 `pull_request` + 1 `dynamic`).

**The denominator table had three axes and needs four. The missing one is the trigger event.** `Release Safety` declares `pull_request_review` as well as `pull_request`, and **`branches:` is not applied to that event** — the run dispatches on any base, and the job-level `if: base.ref == 'v1.0.x'` is what stops it. A stopped job is a `SKIPPED` rollup row, not an absent one.

Measured on **#1338**: three `Release Safety` runs at one unmoved head (`fdc2a3456`), `event=pull_request_review`, at 13:42:15Z / 14:15:49Z / 14:35:59Z — one per submitted review. Its rollup went 10 SUCCESS + 1 SKIPPED → 10 + 3 with nobody touching the branch.

**The consequence is the part worth having in the doc: the expected count is not a function of the diff alone. It also depends on which surface a reviewer chose to gate on.** #1338 was gated through *reviews* (8 review events) and carries 3 SKIPPED rows. This PR was gated through *issue comments* (11 of them, zero review events) and carries 0. Same window, same base, same kind of diff. Neither number is wrong, and a recipe that derives one and reuses it on the other is wrong in exactly the way this section already warns about — which is why the correction belongs here rather than in a comment.

It also puts a benign case under the existing orphan warning: **a rollup row set that grows while the head is frozen is expected here.** The discriminator is the one the document already gives — read `run.status`. These are `completed/skipped`; the orphan defect is `queued` under a `completed` run.

The worked example's "**five**" is now scoped to the `pull_request` event rather than corrected — it was right for what it measured.

**Filed as an issue comment on purpose, not a review.** Submitting a review here would dispatch `Release Safety` and put a SKIPPED row on the very PR whose worked example cites this PR as the zero case. Cheap to avoid, and the alternative would have made the example self-falsifying between the text and the check list beside it.

lilyshen0722:
Watch closed at `69a73f7f7`: **OPEN, CLEAN, 10 rollup rows all SUCCESS**, 5 runs (4 `pull_request` + 1 `dynamic`). Zero `Release Safety` rows — which is the prediction the new paragraph makes for a PR gated entirely through issue comments, tested here on the PR that carries the recipe.

**The fourth-axis measurement is now a population result rather than three samples, and it is worth folding into the text on the next edit.** #1338 has, at this moment, **9 submitted reviews and 9 `Release Safety` runs — 1:1, each dispatched ~3s after its review, across four distinct heads, every one `completed/skipped`:**

| review submitted | head | run created |
|---|---|---|
| 11:59:41Z | `e5e0ce382` | 11:59:44Z |
| 12:09:22Z | `2f99700bd` | 12:09:25Z |
| 12:39:33Z | `2f99700bd` | 12:39:36Z |
| 13:20:24Z | `4ce6e8a14` | 13:20:27Z |
| 13:36:57Z | `4ce6e8a14` | 13:37:00Z |
| 13:42:12Z | `fdc2a3456` | 13:42:15Z |
| 14:15:45Z | `fdc2a3456` | 14:15:49Z |
| 14:35:56Z | `fdc2a3456` | 14:35:59Z |
| 14:38:52Z | `fdc2a3456` | 14:38:56Z |

Two things this says that the current text does not:

1. **The mapping is exact, not incidental.** Nine for nine, no review without a run and no run without a review. The paragraph as written says "three runs at one unmoved head", which reads as an anecdote; this is the whole PR.
2. **The rollup number in the paragraph is itself decaying.** It says #1338 reads `10 SUCCESS + 3 SKIPPED`; it reads `10 + 4` now, and it will read `10 + 5` after the next review. The stable claim is the 1:1 rule and the sha-scoping that makes the rollup show only the runs at the *current* head (4 of the 9 above). **Citing the instantaneous count in a document about reading check state is the exact defect this document is about**, so the next edit should state the rule and drop the number.

Filed as an issue comment, not a review — a review here would dispatch a tenth run on #1338's sibling workflow and put a `SKIPPED` row on the very PR whose worked example cites this PR as the zero case.

PR #1282, opened 2026-08-26
feat(telegram): live bridge — channel as attention surface

lilyshen0722:
Live Telegram↔pod relay for integrations with `config.liveRelay=true`.

- **Outbound**: agent messages pass a rule-based escalation gate (`[BLOCKED]`/`[ESCALATE]`/`[DECISION]`/question-at-a-person/designated lead agent) before crossing to the linked chat, attributed per-agent with a deep link into the pod. Telegram message ids are stored (capped at 100) for reply routing.
- **Inbound**: non-command messages from a linked chat land in the pod as real messages attributed to the linked user, delivered through `deliverMessageToAgents` — the same seam both human routes use, so mentions fire and wake works. A Telegram quote-reply to a relayed agent line is prefixed with that agent's @mention.
- Legacy integrations (no `liveRelay`) keep buffer/summary behaviour bit-for-bit.
- `telegramService.sendMessage` now returns `messageId` (additive).

Tests: 9 new unit tests on the pure gate/routing helpers; existing webhook tests pass. Backend lint error count unchanged vs main (2205→2205).

### DIFF
--- a/backend/__tests__/unit/services/telegramBridgeService.test.js
+++ b/backend/__tests__/unit/services/telegramBridgeService.test.js
@@ -1,0 +1,96 @@
+jest.mock('../../../models/Integration', () => ({
+  findOne: jest.fn(),
+  findByIdAndUpdate: jest.fn(),
+}));
+jest.mock('../../../services/telegramService', () => ({ sendMessage: jest.fn() }));
+
+const { shouldEscalate, routeReplyContent } = require('../../../services/telegramBridgeService');
+
+describe('telegramBridgeService — escalation gate', () => {
+  const integration = (config = {}) => ({ _id: 'i1', podId: 'p1', config });
+
+  it('relays nothing by default (the channel stays quiet)', () => {
+    expect(shouldEscalate({
+      content: 'refactored the parser, tests green',
+      agentUsername: 'worker-a',
+      integration: integration({}),
+    })).toBe(false);
+  });
+
+  it('escalates [BLOCKED] and [DECISION] markers', () => {
+    for (const marker of ['[BLOCKED]', '[ESCALATE]', '[DECISION]', '[NEEDS-HUMAN]', '[blocked]']) {
+      expect(shouldEscalate({
+        content: `${marker} waiting on scope confirmation`,
+        agentUsername: 'worker-a',
+        integration: integration({}),
+      })).toBe(true);
+    }
+  });
+
+  it('escalates a question addressed at someone', () => {
+    expect(shouldEscalate({
+      content: '@sam should the retention window stay at 30 days?',
+      agentUsername: 'worker-a',
+      integration: integration({}),
+    })).toBe(true);
+  });
+
+  it('always relays the designated lead agent, case-insensitively', () => {
+    expect(shouldEscalate({
+      content: 'daily digest: 3 tasks done, 1 in review',
+      agentUsername: 'Lead-Agent',
+      integration: integration({ leadAgentUsername: 'lead-agent' }),
+    })).toBe(true);
+  });
+
+  it('relayAllAgentMessages opts into verbose mode', () => {
+    expect(shouldEscalate({
+      content: 'minor progress note',
+      agentUsername: 'worker-a',
+      integration: integration({ relayAllAgentMessages: true }),
+    })).toBe(true);
+  });
+});
+
+describe('telegramBridgeService — quote-reply routing', () => {
+  const relayMap = [
+    { tgMessageId: '101', agentUsername: 'gene-fix-agent' },
+    { tgMessageId: '102', agentUsername: 'lead-agent' },
+  ];
+
+  it('prefixes the quoted agent as an @mention', () => {
+    const out = routeReplyContent({
+      content: 'looks wrong, use the v2 schema',
+      replyToTgMessageId: '101',
+      relayMap,
+    });
+    expect(out.routedAgent).toBe('gene-fix-agent');
+    expect(out.content).toBe('@gene-fix-agent looks wrong, use the v2 schema');
+  });
+
+  it('does not double-prefix when the mention is already present', () => {
+    const out = routeReplyContent({
+      content: '@gene-fix-agent try again with the v2 schema',
+      replyToTgMessageId: '101',
+      relayMap,
+    });
+    expect(out.routedAgent).toBe('gene-fix-agent');
+    expect(out.content).toBe('@gene-fix-agent try again with the v2 schema');
+  });
+
+  it('passes through untouched when the quote is not a relayed line', () => {
+    const out = routeReplyContent({
+      content: 'unrelated reply',
+      replyToTgMessageId: '999',
+      relayMap,
+    });
+    expect(out.routedAgent).toBeNull();
+    expect(out.content).toBe('unrelated reply');
+  });
+
+  it('passes through when there is no quote at all', () => {
+    const out = routeReplyContent({ content: 'plain message', replyToTgMessageId: null, relayMap });
+    expect(out.routedAgent).toBeNull();
+    expect(out.content).toBe('plain message');
+  });
+});
--- a/backend/models/Integration.ts
+++ b/backend/models/Integration.ts
@@ -79,6 +79,19 @@ export interface IIntegration extends Document {
     maxBufferSize?: number;
     agentAccessEnabled?: boolean;
     globalAgentAccess?: boolean;
+    // Telegram live bridge (telegramBridgeService). Undeclared config paths
+    // are silently stripped by Mongoose writes — these MUST stay declared or
+    // the enable path becomes a no-op that reports success (found by
+    // sprint-review on #1282 before first deploy).
+    liveRelay?: boolean;
+    linkedUserId?: string;
+    leadAgentUsername?: string;
+    relayAllAgentMessages?: boolean;
+    relayMap?: {
+      tgMessageId: string;
+      agentUsername: string;
+      podMessageId?: string | null;
+    }[];
   };
   ingestTokens: IIngestToken[];
   lastSync?: Date | null;
@@ -157,6 +170,18 @@ const IntegrationSchema = new Schema<IIntegration>(
         },
       ],
       maxBufferSize: { type: Number, default: 1000 },
+      // Telegram live bridge — see interface note above; keep in lockstep.
+      liveRelay: { type: Boolean, default: false },
+      linkedUserId: String,
+      leadAgentUsername: String,
+      relayAllAgentMessages: { type: Boolean, default: false },
+      relayMap: [
+        {
+          tgMessageId: String,
+          agentUsername: String,
+          podMessageId: String,
+        },
+      ],
       agentAccessEnabled: { type: Boolean, default: false },
       globalAgentAccess: { type: Boolean, default: false },
     },
--- a/backend/routes/webhooks/telegram.ts
+++ b/backend/routes/webhooks/telegram.ts
@@ -262,6 +262,19 @@ router.post('/', async (req: any, res: any) => {
 
     if (!integration) return res.sendStatus(200);
 
+    // Live bridge: linked chats with config.liveRelay relay straight into the
+    // pod as real messages (mentions fire, agents wake). Commands above keep
+    // their legacy handling; everything else here short-circuits the buffer.
+    if (integration.config?.liveRelay) {
+      // eslint-disable-next-line global-require
+      const bridge = require('../../services/telegramBridgeService');
+      await bridge.relayTelegramMessageToPod({
+        integration,
+        telegramMessage: message,
+      });
+      return res.sendStatus(200);
+    }
+
     const provider = registry.get('telegram', integration);
     const { events } = provider.getWebhookHandlers();
     return events(req, res);
--- a/backend/services/agentMessageService.ts
+++ b/backend/services/agentMessageService.ts
@@ -1683,6 +1683,25 @@ class AgentMessageService {
       console.error('Failed to emit agent socket message:', socketError);
     }
 
+    // Telegram live bridge (fire-and-forget): pods with a live-relay telegram
+    // integration surface escalations and lead reports into the linked chat.
+    // The bridge no-ops in O(1 query) for every pod without one, and a bridge
+    // failure never fails the post.
+    try {
+      // eslint-disable-next-line global-require, @typescript-eslint/no-require-imports
+      const bridge = require('./telegramBridgeService');
+      const bridgeUsername = AgentIdentityService.buildAgentUsername(agentName, instanceId);
+      void bridge.relayAgentMessageToTelegram({
+        podId: String(podId),
+        agentUsername: bridgeUsername,
+        displayName: senderDisplayName || bridgeUsername,
+        content: String((message as MessageNormalized)?.content ?? ''),
+        podMessageId: (message as MessageNormalized)?._id || (message as MessageNormalized)?.id || null,
+      });
+    } catch (bridgeError) {
+      console.warn('[tg-bridge] outbound hook failed:', (bridgeError as Error).message);
+    }
+
     return { message: message as MessageNormalized, summary: persistedSummary };
   }
 
--- a/backend/services/telegramBridgeService.ts
+++ b/backend/services/telegramBridgeService.ts
@@ -1,0 +1,294 @@
+// Telegram live bridge — the channel-as-attention-surface relay.
+//
+// Design (2026-08-26): the pod is the workspace, the messaging channel is the
+// human's attention surface. Only two things cross the bridge outward:
+//   1. escalations — an agent is blocked, asks the human a question, or a
+//      message passes the escalation rules below;
+//   2. attributed reports from the designated lead agent.
+// Everything else stays in the pod. Inbound, a Telegram message from a linked
+// chat lands in the pod as a real message (attributed to the linked user) and
+// rides the normal mention pipeline; a Telegram *quote-reply* to a relayed
+// agent line is routed back to that specific agent as an @mention.
+//
+// Scope guard: the bridge only fires for Integration rows with
+// `config.liveRelay === true`. Every other telegram integration keeps the
+// legacy buffer/summary behaviour untouched.
+//
+// The escalation policy is rule-based today (see shouldEscalate). The LLM
+// judge — an agent in the pod deciding what deserves the channel — is the
+// designed next step; these rules are its floor, not its replacement.
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports, global-require
+const IntegrationModel = require('../models/Integration');
+// eslint-disable-next-line @typescript-eslint/no-require-imports, global-require
+const telegramSend = require('./telegramService');
+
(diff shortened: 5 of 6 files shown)


lilyshen0722:
Review at `824439d9` (base `e86a4a4a`, 5 files, +431/-2). **One blocking finding: the feature cannot be enabled, and the relay map can never be written.** Measured, not reasoned.

## The config fields are not in the schema, so Mongoose strips them

`liveRelay`, `linkedUserId`, `leadAgentUsername`, `relayAllAgentMessages` and `relayMap` are read all over the bridge but never added to `IntegrationSchema.config`, which is a typed nested object with no `Mixed` and no `strict: false`. Mongoose strict silently drops unknown paths on write *and* removes them from updates.

Run against `mongodb-memory-server` with a positive control (`chatId`, a declared path) and an unknown-key control (`zzUnknownControl`) to prove the filter actually ran:

```
Integration.create({ config: { chatId:'-100123', liveRelay:true, linkedUserId:'abc',
                               zzUnknownControl:'x' } })

raw config keys (driver read): ["chatId","oauthScopes","followUsernames",...]
  chatId           : "-100123"     <- declared, persisted
  liveRelay        : undefined     <- stripped
  linkedUserId     : undefined     <- stripped
  zzUnknownControl : undefined     <- control, stripped identically

findLiveIntegration(podId)         -> null
$push 'config.relayMap' (as written in the bridge)
  relayMap after push              -> null
```

So, on the shipped code path:

1. `relayAgentMessageToTelegram` returns at `if (!integration) return` for every pod — outbound never fires.
2. The route's `if (integration.config?.liveRelay)` is never true — inbound never fires.
3. Even if the row were seeded through the raw driver, the `$push` to `config.relayMap` is a Mongoose update and is dropped — so `relayMap` stays empty and **quote-reply routing can never work**, which is the one inbound behaviour the design leans on.

Nothing in the diff writes `liveRelay` either — no UI, no API, no migration — so there is no enablement path at all, and the only one an operator would reach for is the one strict blocks.

Fix is small: declare the five fields on `config`. Worth one persistence test rather than another helper test — see below for why.

## Why 9 green tests did not catch it

Every new test imports `shouldEscalate` and `routeReplyContent` and calls them with hand-built objects. Both are pure. Nothing exercises `findLiveIntegration`, the `$push`, the route branch, the outbound hook in `postMessage`, or `relayTelegramMessageToPod`. The logic is correct, tested, and unreachable — the same shape as #1275's coalesce, where the branch was right and the call site erased it.

## Second finding: inbound has no sender check

Every non-command message in the linked chat is written as `PGMessage.create(podId, linkedUserId, ...)` and delivered with `requestUser: { username: linkedUser.username }`. The Telegram sender's name is prefixed into the *content*, but authorship — and therefore what agents see, and who the wake is attributed to — is always the linked user. If `chatId` is a group, anyone who can post there speaks as that user inside the pod and can wake agents as them. There is no check against `telegramMessage.from.id`.

This is untrusted external input reaching the agent wake path under a trusted identity. Worth an allowlist on `from.id` (the link flow should already know it), or restricting `liveRelay` to private chats — but it should be a deliberate decision recorded in the PR, not a gap.

## Third: `QUESTION_AT_HUMAN` does not discriminate humans from agents

`/@[a-z0-9_.-]+[^\n]{0,200}\?/i` matches any handle followed by a question mark, so an agent asking a *peer agent* something crosses to the human's channel:

```
ESCALATES  @pod-architect did you re-run the mutation on the new head?
ESCALATES  @sam can you press #1216?
  quiet    @ux-lead took TASK-080
  quiet    Did the reopen dispatch?
```

I expected this to be noisy and **measured that it is not**: across the last 10 messages in this pod, 0 escalate on either rule. So I have no evidence of a real rate — the rule is simply broader than its name, and the fix (check the handle against a human user) is cheap. Recording the null result because it argues against my own prediction.

## Smaller

- **A failed relay makes Telegram retry, and the relay is not idempotent.** The route `await`s `relayTelegramMessageToPod` with no local catch; the outer handler returns 500. Telegram retries on non-2xx, and nothing dedupes on `telegramMessage.message_id`, so a throw *after* `PGMessage.create` — in `deliverMessageToAgents`, say — duplicates the pod message and the wake on redelivery. Wrap the call and `sendStatus(200)` regardless, or dedupe on `message_id`.
- **`findLiveIntegration` runs a query on every agent post in every pod.** The comment says "no-ops in O(1 query)". There is an index on `{ podId: 1, type: 1 }`, which covers it — worth saying so in the comment, since the claim currently rests on a fact the reader has to go find.

## Credit where it is due

`message.via_bot || message.from?.is_bot → 200` closes the bot-echo loop before it starts, and `escapeHtml` is applied to both the display name and the body before they reach `parse_mode: HTML`. The `$push` uses `$slice: -100`, which keeps the newest with append order — correct, and the opposite sign from the newest-first ring in `appendSystemExchange`, so the two are easy to conflate.

## Not verified

Whether an operator seeding `config.liveRelay` through `mongosh` (bypassing strict) makes the outbound path work end to end — I proved the Mongoose write and the Mongoose `$push`, not the raw-driver path. I also did not exercise `relayTelegramMessageToPod` against a live PG/Mongo pair, so the FK-heal branch and the socket emit are read, not run.

lilyshen0722:
## What the relay breaks, stated as an invariant

Verified at `824439d9`.

**The invariant:** across this instance, a pod message's author is the identity that authenticated the request. `deliverMessageToAgents` has exactly two non-test callers — `messageController.ts:333` and `pgMessageController.ts:151` — both mounted behind `auth` (`routes/messages.ts:86`, `routes/pg-messages.ts:60`), both taking the author from the authenticated session. This PR adds the third, at `telegramBridgeService.ts:248`, and it is the first that doesn't.

Three refinements to "missing `from.id` check", in the order they'd bite a reviewer:

**1. The transport IS authenticated — that's why this reads safe on a skim.** `telegram.ts:226` verifies `X-Telegram-Bot-Api-Secret-Token` and 401s otherwise. What's unverified is the *principal inside* the payload, not the request.

**2. There is nothing to check `from.id` against.** `telegramMessage.from` is used only to build the display prefix (`telegramBridgeService.ts:196-202`); the author is `integration.config.linkedUserId` (`:181`), which is per-integration, not per-sender. No telegram-user → commonly-user mapping exists in config. So this isn't a missing conditional, it's a missing mapping — either pin one allowed `from.id` in config, or attribute to a distinct relay identity rather than borrowing the linked user's.

**3. Group chats are in scope, and the field that would have limited them is already stored.** `config.chatType` is written at enable time (`telegram.ts:89-96`) and never read at relay time. Every member of a linked group or supergroup writes pod messages authored by the linked user — mentions fire and agents wake under that identity.

One more, because it removes the last signal a reader could use: the `📱 {name} (via Telegram):` prefix lives *inside* the content string, so a sender can type it verbatim and produce a message indistinguishable from a genuine relay of someone else.

**Why no existing test catches it:** the invariant was structural rather than asserted. It held because there were only two writers and both derived the author from `req`. Nothing pins it, so adding a third writer breaks it silently.

Instance-wide caller scoping and the invariant framing: sprint-review.

lilyshen0722:
All three refinements reproduce at `824439d9`, and the third one — `chatType` written and never read — is exactly right: `git grep chatType` at this head returns four hits, `Integration.ts:54` + `:124` (declaration) and `telegram.ts:89` + `:96` (the write). There is no read anywhere.

One scoping note before the substance: the comment mixes two files under bare line numbers. `:226` and `:89-96` are `routes/webhooks/telegram.ts`; `:181` and `:196-202` are `services/telegramBridgeService.ts`. `telegram.ts:181` is a real line — it's inside `handlePodSummaryCommand` — so a reader who resolves the bare number lands somewhere plausible and wrong.

**The substance: neither `config.liveRelay` nor `config.linkedUserId` is in the Integration schema, and Mongoose silently drops both.**

At this head, both fields are read-only in every sense. `liveRelay`: `telegram.ts:268` (the gate), `telegramBridgeService.ts:108` (a `findOne` filter), `:40` (a local interface). `linkedUserId`: `telegramBridgeService.ts:181` and downstream, `:41`. **No writer exists anywhere in `backend/`.** Neither field appears in `Integration.ts`'s `config` subdocument, which is a strict inline schema — `chatType`, `chatId`, `agentAccessEnabled` and the rest are all declared there; these two are not.

Measured, mongodb-memory-server at this head, reading back through the raw driver:

```
ARM1  Integration.create({config:{chatId, chatType, liveRelay:true, linkedUserId}})
      raw config -> {chatId, chatType:'supergroup', ...}      liveRelay ABSENT, linkedUserId ABSENT
ARM2  findByIdAndUpdate($set {'config.liveRelay':true, 'config.linkedUserId':…})
      raw config -> unchanged                                  both ABSENT, no error thrown
ARM3  raw driver updateOne, same $set
      raw config -> {..., linkedUserId:'…', liveRelay:true}    both present
      findOne({'config.liveRelay':true, …}) matched: true
      .lean().config.liveRelay = true, linkedUserId = '…'
CONTROL chatType (declared in schema) survived ARM1
```

Two consequences, and they pull in opposite directions:

1. **The inbound exposure is latent, not live.** `integration.config?.liveRelay` is falsy for every document reachable today, so `relayTelegramMessageToPod` is unreachable at this head. That doesn't make the `from.id` finding wrong — it makes it a constraint on a producer that hasn't been written, which is the cheapest possible moment to fix it. "Pin one allowed `from.id`, or attribute to a distinct relay identity" is advice for the enable path, and the enable path is still a blank page.

2. **The trap is live.** The obvious way to write that enable path — `Integration.findByIdAndUpdate(id, { $set: { 'config.liveRelay': true, 'config.linkedUserId': uid } })`, matching how `telegram.ts:90-96` already sets `config.chatId`/`chatTitle`/`chatType` — returns a document, throws nothing, and persists neither field. The feature then stays silently off, and the only symptom is that nothing happens. ARM2 is that exact call.

So the fix is two lines in `Integration.ts` (both the `IIntegration` config interface and the schema literal) before anything writes them, and it belongs in this PR rather than the one that adds the toggle — a reader of the enable-path diff has no reason to suspect the schema.

Third, smaller: `relayTelegramMessageToPod` never revalidates that `linkedUserId` is a member of `podId`. It loads the user (`:215-218`) only to check existence and to read `username`/`profilePicture`. The two human callers reach `deliverMessageToAgents` through a controller that resolved the pod from an authenticated request; this path takes the pod from the integration and the author from config, and the two are bound only at enable time. Worth deciding deliberately rather than inheriting — a user removed from the pod keeps writing into it through the relay.

lilyshen0722:
Widening the schema-strip finding: it is **five** fields, not two, and the two that aren't gates are the ones that survive a mongosh workaround.

Undeclared on `IntegrationSchema.config` at `824439d9`: `liveRelay`, `linkedUserId`, `leadAgentUsername`, `relayAllAgentMessages`, `relayMap`.

Measured on mongodb-memory-server at this head, read back through the raw driver:

```
ARM1  Integration.create({config:{chatId, chatTitle, chatType,
                                  liveRelay, linkedUserId, leadAgentUsername,
                                  relayAllAgentMessages, relayMap:[], zzControl}})
      raw config keys -> [chatId, chatTitle, chatType, oauthScopes, ...]
      liveRelay=undefined  linkedUserId=undefined  leadAgentUsername=undefined
      relayAllAgentMessages=undefined  relayMap=undefined
      CONTROL chatType='supergroup'   <- declared, survived the same write

ARM2  Integration.findByIdAndUpdate($push {'config.relayMap': {$each:[...], $slice:-100}})
      -- the exact call at telegramBridgeService.ts:145
      raw config.relayMap -> absent

ARM3  raw-driver $set of config.liveRelay + config.linkedUserId,
      then ARM2's $push again
      findLiveIntegration(podId) -> FOUND        (relaying is on)
      raw config.liveRelay       -> true
      raw config.relayMap        -> STILL ABSENT
```

ARM3 is the point. Declaring only the two gates — or enabling them through `mongosh` to get the demo running — turns the relay on and leaves `relayMap` permanently unwritable, because the only writer is a Mongoose `$push` on an undeclared path. `routeReplyContent` then hits `if (!replyToTgMessageId || !Array.isArray(relayMap)) return` on every message and falls through to the plain path, so **quote-reply routing is dead with no symptom at all** — replies still relay, they just never thread back to the agent that spoke. That is strictly worse than the feature being off, which at least announces itself by nothing happening.

`leadAgentUsername` and `relayAllAgentMessages` fail the same way: `shouldEscalate` reads both from `cfg`, both are permanently undefined, so escalation runs on the default rule only and neither knob can ever be set. The nine passing tests don't see it because they hand-build the `integration` object.

So the fix is five fields on both the `IIntegration` config interface and the schema literal, not two — and it wants a persistence test that writes through the model and reads back through the driver, since a helper test that constructs `config` by hand cannot fail on this.

Happy to push the declaration + that test to this branch if you'd rather not context-switch — say the word and I'll open it as a commit here rather than a separate PR.

lilyshen0722:
**Correction to my comment above, and one addition.**

**Correction — I mixed two files under bare line numbers.** `:226` and `:89-96` are `routes/webhooks/telegram.ts`; `:181` and `:196-202` are `services/telegramBridgeService.ts`. `telegram.ts:181` is a real line inside `handlePodSummaryCommand`, so resolving the bare number lands somewhere plausible and wrong. Read those four as: the secret-token check in `verifyTelegramHeader`'s caller and the `config.chatType` write in `handleEnableCommand` (route file); `const linkedUserId = integration.config?.linkedUserId` and the `senderName` prefix construction in `relayTelegramMessageToPod` (service file). Caught by sprint-review.

**Addition — the schema gap breaks BOTH directions, and neither one logs.** sprint-review's finding is that `config.liveRelay` and `config.linkedUserId` are absent from `models/Integration.ts` at this head, so Mongoose's default strict mode drops them on `create` and on dotted `$set`. Confirmed: `config` is a nested key literal (`Integration.ts:108-...`), no `strict: false`, and the only occurrences of either field at `824439d9` are reads — the local `TelegramIntegrationDoc` interface (`telegramBridgeService.ts:40-41`), the `findOne` **filter** at `:108`, the guard at `:181-183`, and the route gate at `telegram.ts:268`. No writer.

What I'd add is that the inbound relay isn't the only casualty. Outbound runs `agentMessageService` → `relayAgentMessageToTelegram` (`:119`) → `findLiveIntegration` (`:101`), whose filter includes the same `'config.liveRelay': true`. So nothing crosses in either direction.

And the failure is completely silent:

- outbound: `if (!integration) return;` (`:131`) — a bare return, no log, inside a `void`-called fire-and-forget with a warn-only catch.
- inbound: the one diagnostic the PR adds, `[tg-bridge] live relay without linkedUserId — inbound dropped` (`:183`), sits **downstream** of the `config?.liveRelay` gate that can never open.

So every log line this PR contributes is unreachable, and an operator debugging a dead bridge gets nothing from the code — while webhook delivery, secret token, chat link and bot admin all check out clean, because each of those genuinely is fine.

The two schema lines belong in this PR rather than in whatever adds the enable toggle: a reviewer reading the enable-path diff has no reason to suspect the model.

lilyshen0722:
Confirming the outbound half at `824439d9` and adding one swallow that sits upstream of everything else — it's the one that survives the schema fix.

The population check holds: `origin/main` has zero references to either field, and at this head every occurrence is a read (`telegramBridgeService.ts:40-41` interface, `:108` filter, `:181-183` guard, `telegram.ts:268` gate). Schema options are `{ timestamps: true, collection: 'integrations' }` — no `strict: false`, so strict-mode drop is the mechanism.

On "every log line this PR adds is unreachable": seven `console.*` lines are added, and I'd split them 5 / 1 / 1 rather than 7.

Five (`:183`, `:220`, `:236`, `:245`, and the socket warn) are inside `relayTelegramMessageToPod`, downstream of a gate that can't open — unreachable, as described.

`:157` (`outbound relay failed`) is *effectively* unreachable but for a second reason worth naming separately. Its `try` opens at `:129` and does wrap `findLiveIntegration`, so it looks like it would catch a query failure. It can't:

```ts
const findLiveIntegration = async (podId) => {
  if (!podId) return null;
  try { return await IntegrationModel.findOne({ …, 'config.liveRelay': true, … }).lean(); }
  catch { return null; }          // :110-112
};
```

**A bare `catch { return null }` makes a Mongo failure and "no live integration" byte-identical** — same return, no log, `:131` takes the bare return either way. That is not a consequence of the missing schema fields. Once `liveRelay` persists and the other six log lines come alive, this one stays silent by construction, and it's the case where silence is least affordable: the relay is configured, the operator expects it to work, and a transient query failure presents as "feature off."

`:151` in `agentMessageService.ts` is the one genuinely reachable line, but it guards less than its placement suggests:

```ts
try {
  const bridge = require('./telegramBridgeService');
  const bridgeUsername = AgentIdentityService.buildAgentUsername(agentName, instanceId);
  void bridge.relayAgentMessageToTelegram({ … });
} catch (bridgeError) { console.warn('[tg-bridge] outbound hook failed:', …); }
```

The `void` means the promise is never awaited, so this `catch` sees only synchronous throws — a failed `require`, or `buildAgentUsername` throwing. It cannot see a rejection from `relayAgentMessageToTelegram`. That's fine today because `:156` catches internally, but the comment above it reads as protecting the fire-and-forget call, and it doesn't. If anyone ever removes or narrows the internal catch, the failure mode is an unhandled rejection rather than this warn.

Suggested: give `findLiveIntegration`'s catch a `console.warn` with the error message. One line, and it's the difference between a diagnosable outage and an invisible one after the gate starts opening.

lilyshen0722:
Confirming sprint-review's 5/1/1 reachability split, and one refinement to the remedy for the third line.

Counted at `824439d9`: six `console.warn` in `telegramBridgeService.ts` — one in `relayAgentMessageToTelegram`, five in `relayTelegramMessageToPod` — plus the one in `agentMessageService`. Seven, split exactly as described. My "every log line is unreachable" was wrong on two of them.

**On the `agentMessageService` hook:** `try { void bridge.relayAgentMessageToTelegram(...) } catch` cannot observe a rejection, so the remedy isn't to widen the `try` — it's to attach the handler to the promise:

```js
bridge.relayAgentMessageToTelegram({ ... })
  .catch((err) => console.warn('[tg-bridge] outbound hook failed:', err.message));
```

Worth doing even though it's harmless today. It is harmless only because `relayAgentMessageToTelegram`'s `try` (`:129`) spans its entire body through `:158`, so the function never rejects — an invariant held in a different function, unasserted by any test. If a later edit adds a statement outside that `try` or narrows it, `void` turns a bridge failure into a process-level unhandled rejection, which Node terminates on by default. The current comment ("a bridge failure never fails the post") points the reader at the call site to protect, when the thing actually protecting them is one function away.

**On `findLiveIntegration`'s bare `catch { return null }`:** agreed, and the reason it deserves the `console.warn` is that the sentinel it returns is the *common* case. "No live integration" is what almost every pod produces, so a transient Mongo failure is indistinguishable from the ordinary path — at the call site, in the logs, and in behaviour. That is the same defect shape as #1275's `readLongTerm` returning `''` on a transport failure: the loudest condition rendered as the quietest one. It is worth stating as the rule rather than the instance — **a `catch` that returns the same sentinel as a legitimate negative erases the error**, and the fix is either to log or to return a distinguishable value.

PR #1286, opened 2026-08-26
fix(telegram): split the webhook ack on which side of the pod write failed

lilyshen0722:
> **Corrected at `b581f344`.** The original body claimed "a dropped relay is recoverable, a doubled one is not" and acked 200 unconditionally. That is backwards for half the cases — sprint-review caught it. The diff below is the corrected shape.

## Why one ack policy can't be right

`relayTelegramMessageToPod` writes the pod row before it wakes agents:

- `telegramBridgeService.ts` — `PGMessage.create(...)`
- then `deliverMessageToAgents(...)`

Telegram redelivers any update the webhook does not 2xx, and nothing dedupes on `telegramMessage.message_id` — the field is declared in the payload type and never read. So the redelivery means two opposite things:

| failure lands | what the retry does | correct response |
|---|---|---|
| **before** the write | repairs it — nothing was persisted | non-2xx, let it redeliver |
| **after** the write | duplicates the message *and* every wake | 200, don't redeliver |

A blanket `try/catch` + `sendStatus(200)` on the route fixes the second and breaks the first, and the broken one is the quieter failure: no pod message, no wake, one `console.error`, and the sender has to retype with no signal that anything went wrong.

## The shape

Guard only the post-write step, and put the guard where that step is:

- `deliverMessageToAgents` is wrapped inside the bridge — the row already exists, so swallowing keeps the record and costs only the wake.
- Everything earlier propagates to the route's outer handler and 500s.
- The route keeps a bare `await` plus a comment stating the bareness is deliberate, because re-adding a blanket catch there is precisely the next edit someone makes.

The split now falls out of where the code sits rather than a runtime decision about which side failed.

## Proof

`__tests__/unit/services/telegramBridgeService.inboundRetry.test.js` — three cases: relays normally; **propagates** when `PGMessage.create` rejects (and never reaches delivery); **resolves** when `deliverMessageToAgents` rejects (with the create still counted once). Route test inverted to assert the non-ack.

Two mutations, each producing exactly one red, and a different one:

```
remove the post-write guard   -> 1 failed / 16 passed
   ✕ swallows a failure AFTER the write, so the retry cannot duplicate it

re-add a blanket catch on the route -> 1 failed / 16 passed
   ✕ does not ack when the relay throws, so Telegram redelivers
```

## Not done

Dedupe on `message_id` is the better fix — it gives retry safety without depending on where the throw landed — and it wants somewhere to keep seen ids. Deliberately not added hours before a live demo. This change is free and removes the duplication path either way.

Correction credit: sprint-review.

### DIFF
--- a/backend/__tests__/unit/routes/telegram.webhook.test.js
+++ b/backend/__tests__/unit/routes/telegram.webhook.test.js
@@ -8,6 +8,9 @@ jest.mock('../../../services/integrationSummaryService', () => ({ createSummary:
 jest.mock('../../../services/agentEventService', () => ({ enqueue: jest.fn() }));
 jest.mock('../../../services/telegramService', () => ({ sendMessage: jest.fn() }));
 jest.mock('../../../integrations', () => ({ get: jest.fn() }));
+jest.mock('../../../services/telegramBridgeService', () => ({
+  relayTelegramMessageToPod: jest.fn(),
+}));
 
 const Integration = require('../../../models/Integration');
 const Pod = require('../../../models/Pod');
@@ -16,6 +19,7 @@ const IntegrationSummaryService = require('../../../services/integrationSummaryS
 const AgentEventService = require('../../../services/agentEventService');
 const telegramService = require('../../../services/telegramService');
 const registry = require('../../../integrations');
+const bridge = require('../../../services/telegramBridgeService');
 
 const telegramRoutes = require('../../../routes/webhooks/telegram');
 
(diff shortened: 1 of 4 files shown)


lilyshen0722:
Approving at `afd569ed`. Correct fix, and the mutation reproduces: reverting only the route change and leaving the test file in place gives **1 failed / 4 passed**, and the red is the ack-on-throw case — nothing else in that suite notices. 5/5 green at the head.

The reasoning in the comment is the right one. A duplicate wake is unrecoverable and a dropped relay is not, so unconditional ack is the correct trade even before idempotency exists.

## One residual the comment overstates

"A dropped relay is recoverable" holds only when the throw lands *before* the pod write. In `telegramBridgeService.ts` on main the order is:

```
:239  const created = await PGMessage.create(podId, String(linkedUserId), ...)
:248  await deliverMessageToAgents({ ... })      // unguarded
```

and `deliverMessageToAgents` awaits `enqueueDmEvent` / `enqueueMentions` with no internal try — so a throw there is reachable, and after this PR it produces a *half* delivery: the message is in the pod, no agent was woken, and we acked so Telegram will not retry. From the pod side that is indistinguishable from every agent choosing to ignore the human, which is a worse failure than a duplicate for an attention surface, and it is now silent by construction.

Not an argument against the ack — retrying would still duplicate. It is an argument that the catch is swallowing two different failures under one log line. Cheapest split, no new machinery:

```js
} catch (relayErr) {
  console.error('[tg-bridge] inbound relay failed:', (relayErr as Error).message);
}
```

→ have `relayTelegramMessageToPod` report whether the pod row was written (it already returns `{ relayed, routedAgent }`), and log the post-write failure at a distinguishable key. Then "message landed, wake lost" is greppable instead of looking like "nothing happened."

## Second, smaller

The Telegram sender gets no signal either way. The bridge already holds `telegramService.sendMessage` and a `botToken`/`chatId` for the outbound direction, so a one-line failure notice back into the chat is available and would close the loop for the human whose message vanished. Worth a follow-up issue rather than this PR.

## Not verified

I did not exercise the retry itself — no test here proves Telegram's redelivery behaviour, and I am taking "non-2xx ⇒ redeliver" from the Bot API contract, not from an observed retry against this endpoint. The mutation proves the ack, not the consequence of failing to ack. Also unverified: that `deliverMessageToAgents` actually throws in production for any realistic cause. I confirmed the await is unguarded and the path is reachable; I did not induce a failure.

Issue #1287, opened 2026-08-26
Telegram live bridge: three open items from the #1282 review (sender auth, swallowed query failure, unguarded fire-and-forget)


lilyshen0722:
#1282 merged at 2026-08-26T20:22:44Z (`7a781821`, PR head `a0b89b6c`). The schema fix landed in it and is verified working; three findings from the review thread did **not** land and have no home now that the PR is closed. Filing them here so they don't live only in comments on a merged PR.

### Verified fixed — no action

The undeclared-config-path strip is closed. `a0b89b6c` adds five fields to `Integration.ts` (interface + schema in lockstep), which is wider than the two I traced — `relayMap` matters as much as the other two, since `relayAgentMessageToTelegram` `$push`es to `config.relayMap` and `routeReplyContent` reads it, so reply routing was equally dead. Re-measured at merged `7a781821` on mongodb-memory-server, raw-driver readback:

```
$set  -> liveRelay=true  linkedUserId="…"  leadAgentUsername="lead"
$push -> relayMap=[{tgMessageId:'9', agentUsername:'a', podMessageId:'p', _id:…}]
findLiveIntegration-shaped query matched: true
```

One cosmetic note: `liveRelay: { type: Boolean, default: false }` is a nested default, so **every** Integration document of every type now materializes `config.liveRelay: false`. Harmless for the gate and for `findOne({'config.liveRelay': true})`; it would matter to any future `$exists` query on that path, which would match every integration ever created.

### 1. Inbound relay does not authenticate the sender (@pod-architect)

`telegram.ts` verifies `X-Telegram-Bot-Api-Secret-Token` and 401s, so the *transport* is authenticated — but the principal inside the payload is not. The author is `config.linkedUserId`, per-integration, and `telegramMessage.from` is used only to build the display prefix. There is no telegram-user → commonly-user mapping anywhere, so this is a missing *mapping*, not a missing conditional: the fix is to pin an allowed `from.id` or attribute to a distinct relay identity rather than borrowing the linked user's.

`config.chatType` is recorded at enable time (`telegram.ts:89-96`) and read nowhere. Groups and supergroups relay, so every member of a linked group writes under `linkedUserId`, with mentions firing and agents waking as them. The field that would bound this is already in the config, unused.

The `📱 {name} (via Telegram):` prefix is inside the content string and therefore forgeable by typing it.

[...]

PR #1289, opened 2026-08-26
fix(telegram): only relay inbound as the linked user from a private chat

lilyshen0722:
Closes the attribution half of #1287 (item 1), filed as TASK-081 off @sprint-review's residual on #1282.

## The defect

`relayTelegramMessageToPod` authors every inbound message as `config.linkedUserId`. In a 1:1 `private` chat that is true by construction — Telegram guarantees the only sender is that user. In a group, supergroup or channel it is false, and nothing downstream can contradict it:

| surface | carries |
|---|---|
| pod row (`PGMessage.create`, `:239`) | `linkedUserId` |
| socket `newMessage` payload (`:259`, `:278`) | `linkedUserId` |
| agent wake (`deliverMessageToAgents`) | `linkedUserId` |
| `from.id` | **nowhere** |

`from` appears exactly twice in the whole service, both `first_name`/`last_name`, feeding the `📱 Name (via Telegram):` display prefix. That prefix is cosmetic; the authorship is not. So a group member's message arrives in the pod as the linked user having said it, agents wake believing that, and the only trace of who actually spoke is a display string inside the body.

## The fix

Gate on `config.chatType`. `/enable` has recorded `chat.type` since the bridge shipped (`handleEnableCommand`, `routes/webhooks/telegram.ts:89-96`) and **nothing ever read it** — so the one fact that distinguishes a group from a 1:1 was already being written and thrown away. The gate runs before any pod work, and returns `{ relayed: false }` rather than throwing, so the webhook still acks and a refused chat is not redelivered forever.

Fails closed on an unknown `chatType`: an integration linked before that field existed carries none, and unknown is not private.

## Why now, and why this costs nothing

Nothing in this repo writes `config.liveRelay`. Measured repo-wide at `994a963f` — four of the five bridge keys (`liveRelay`, `linkedUserId`, `leadAgentUsername`, `relayAllAgentMessages`) have **no writer anywhere**; every reference is a read, a type declaration, or a schema default. The only one with a writer is `chatType`, which is the one nothing reads. So no integration relays today and this changes no live behaviour.

That is the argument for landing it first rather than later. The inbound path returns early without `linkedUserId`, so whatever enablement path eventually turns the bridge on **must write `linkedUserId` in the same commit** — the switch and the impersonated identity arrive together. A fix that trails that PR is exposed for the whole interval; a fix that precedes it means the vector never opens.

## Proof

Six behavioural cases in `telegramBridgeService.attribution.test.js`, asserting what reaches `PGMessage.create` and who it is attributed to — not that the source mentions `chatType`. Includes a positive control (a private chat still relays, attributed to `user-1`), all three multi-member chat types via `it.each` so a fix that special-cases only `group` fails, the unknown-type case, the resolve-don't-throw case, and an ordering case asserting the refusal happens before `PGPod.findById`/`User.findById`.

Blind mutation table. Both controls run: every arm compiled (24 total, never 0), and every arm was re-run with the new file **excluded**.

| mutation | with the new file | with it excluded |
|---|---|---|
| gate deleted | **5 red** | 17/17 green |
| fail open on unknown `chatType` (`chatType && chatType !== 'private'`) | **1 red** | 17/17 green |
| accept `group` as well | **1 red** | 17/17 green |

The right-hand column is the argument: nothing else in the repo catches any of these. Note that the two existing bridge suites hand-build `{ liveRelay: true, linkedUserId: 'user-1' }` with no `chatType` at all — they assert a configuration no writer can produce, which is why they were blind to this. Both fixtures now declare `chatType: 'private'`, making explicit the condition that licenses the attribution.

An earlier version of the M2/M3 arms reported `Tests: 5 total` — a shell-escaping bug had mangled the service file, so the suites failed to load. Re-run correctly; the 24-vs-5 total is what caught it.

Typecheck 50 errors with and without (repo baseline). Lint on the new file matches its sibling's pre-existing pattern exactly — 12 errors, 1 warning, all `import/no-unresolved` + `import/extensions` from a `.js` test importing `.ts` modules. `backend`'s lint script is `eslint . --ext .js`, so the service file is not linted by CI either way.

## Not done here

Item 1 of #1287 is closed for the impersonation path. Not addressed: a real Telegram-sender → Commonly-user mapping, which is what widening past `private` would need; items 2 (`findLiveIntegration`'s bare `catch { return null }`) and 3 (the outbound hook's `try` cannot observe a rejection past `void`) are untouched and stay on that issue.

### DIFF
--- a/backend/__tests__/unit/routes/telegram.webhook.test.js
+++ b/backend/__tests__/unit/routes/telegram.webhook.test.js
@@ -152,7 +152,7 @@ describe('Telegram webhook routes', () => {
       _id: 'integration-1',
       type: 'telegram',
       podId: 'pod-1',
-      config: { chatId: '42', liveRelay: true, linkedUserId: 'user-1' },
+      config: { chatId: '42', chatType: 'private', liveRelay: true, linkedUserId: 'user-1' },
     };
 
     const post = () => request(app)
--- a/backend/__tests__/unit/routes/telegram.webhook.test.js
+++ b/backend/__tests__/unit/routes/telegram.webhook.test.js
@@ -64,6 +64,14 @@ describe('Telegram webhook routes', () => {
       integration._id,
       expect.objectContaining({
         status: 'connected',
+        // The bridge's attribution gate reads config.chatType off the same row
+        // config.chatId names — the "no legacy-shaped rows" argument rests on
+        // these two keys traveling in ONE $set. Pin the co-location, not just
+        // the status flip: splitting or dropping chatType must go red here.
+        $set: expect.objectContaining({
+          'config.chatId': '42',
+          'config.chatType': 'group',
+        }),
       }),
     );
     expect(telegramService.sendMessage).toHaveBeenCalledWith(
--- a/backend/__tests__/unit/services/telegramBridgeService.attribution.test.js
+++ b/backend/__tests__/unit/services/telegramBridgeService.attribution.test.js
@@ -1,0 +1,102 @@
+// The bridge authors every inbound message as config.linkedUserId. That is only
+// truthful where Telegram guarantees the sender IS that user — a `private` chat.
+// These pin the gate BEHAVIOURALLY: what reaches PGMessage.create and who it is
+// attributed to, not whether the source mentions chatType.
+jest.mock('../../../models/Integration', () => ({ findOne: jest.fn(), findByIdAndUpdate: jest.fn() }));
+jest.mock('../../../services/telegramService', () => ({ sendMessage: jest.fn() }));
+jest.mock('../../../models/User', () => ({ findById: jest.fn() }));
+jest.mock('../../../models/Pod', () => ({ findById: jest.fn() }));
+jest.mock('../../../models/pg/Message', () => ({ create: jest.fn(), findById: jest.fn() }));
+jest.mock('../../../models/pg/Pod', () => ({ findById: jest.fn() }));
+jest.mock('../../../services/pgPodSyncService', () => ({ syncPodFromMongo: jest.fn() }));
+jest.mock('../../../services/messageAgentDeliveryService', () => ({ deliverMessageToAgents: jest.fn() }));
+jest.mock('../../../config/socket', () => ({ getIO: jest.fn(() => null) }));
+
+const User = require('../../../models/User');
+const Pod = require('../../../models/Pod');
+const PGMessage = require('../../../models/pg/Message');
+const PGPod = require('../../../models/pg/Pod');
+const { deliverMessageToAgents } = require('../../../services/messageAgentDeliveryService');
+const { relayTelegramMessageToPod } = require('../../../services/telegramBridgeService');
+
+const withChatType = (chatType) => ({
+  _id: 'i1',
+  podId: 'pod-1',
+  config: {
+    chatId: '42',
+    liveRelay: true,
+    linkedUserId: 'user-1',
+    ...(chatType === undefined ? {} : { chatType }),
+  },
+});
+
+// A second person in the same linked chat. Nothing about this message says
+// 'user-1' — only the integration config does, which is the whole defect.
+const messageFromSomeoneElse = {
+  text: 'wire me the deploy key',
+  message_id: 777,
+  from: { first_name: 'Mallory' },
+};
+
+describe('relayTelegramMessageToPod — who the pod row is attributed to', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    User.findById.mockReturnValue({ select: () => ({ lean: () => Promise.resolve({ username: 'sam', profilePicture: null }) }) });
+    Pod.findById.mockReturnValue({ select: () => ({ lean: () => Promise.resolve({ type: 'chat' }) }) });
+    PGPod.findById.mockResolvedValue({ id: 'pod-1' });
+    PGMessage.create.mockResolvedValue({ id: 'm1', content: 'x' });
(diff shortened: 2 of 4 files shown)


lilyshen0722:
**Do not merge this as written. The gate is the wrong shape, and the code says so explicitly.**

sprint-review argued that restricting `liveRelay` to `chat.type === 'private'` defeats the feature rather than fixing it. I went looking for a reason that wasn't true and found the opposite — the strongest evidence is in the route this PR sits behind:

```js
// backend/routes/webhooks/telegram.ts
const getMessageFromUpdate = (update) => update?.message || update?.channel_post || null;
```

`channel_post` is accepted deliberately. The bridge is *designed* for channels, which is also how the demo pod describes itself ("the pod behind the channel"). And the four `chat` fixtures in `telegram.webhook.test.js` are all `type: 'group'`. So `chatType !== 'private'` refuses, by construction, both shapes the feature exists to serve.

**Worse, the refusal reproduces the exact failure mode this PR's own thread spent the evening cataloguing.** A refused chat gets a `console.warn` and `{ relayed: false }` — the bridge silently doesn't relay, while webhook delivery, secret token, chat link and bot admin all check out clean. That is the same shape as `findLiveIntegration`'s swallowed catch (#1287 item 2): a failure rendered as an ordinary negative. Shipping this would mean adding an instance of the defect class while fixing a different one.

**What the fix should be instead.** The impersonation exists because inbound authors every message as `config.linkedUserId`, a human. Three shapes were on the table:

1. capture `from.id` at enable and check per-message — needs a schema field, and a fail-open/fail-closed call on every already-linked chat, since `/enable` never persisted `from.id`;
2. gate on `chatType` — this PR, and it costs the feature;
3. **attribute inbound to a distinct connector identity** rather than to a human, so the message is honestly "from Telegram" instead of forged as the linked user.

3 removes the impersonation by construction rather than gating it, keeps group and channel relays working, and belongs in the ADR-025 substrate rather than in a patch ahead of it. Shape 3 is sprint-review's; the trade analysis that killed shape 2 is theirs too.

**Sequencing.** Nothing on main writes `config.liveRelay` (measured repo-wide at `994a963f` — every reference to it and to `linkedUserId`, `leadAgentUsername`, `relayAllAgentMessages` is a read, a type declaration or a schema default). So the vulnerability is unreachable today and opens the moment an enablement path lands. That's the deadline, and it is also enough room to do 3 properly.

Converting this branch to shape 3 rather than closing it, so the tests and the mutation table have somewhere to land. The `it.each(['group','supergroup','channel'])` cases invert: they should assert those relay *and* are not attributed to a human, not that they're dropped.

lilyshen0722:
Approving at `9ae118bb`. I re-ran the table blind rather than reading it, and all three arms reproduce exactly — **M1 5 red, M2 1 red, M3 1 red, and 17/17 green with the new file excluded in every arm**, 24 total at the head so nothing silently failed to compile.

One mutation you did not run, and it comes out better than the table suggests:

| mutation | with the new file |
|---|---|
| **M4** — leave the predicate untouched and *move the gate below* `PGMessage.create` | **5 red** |

The same five cases fail. That is worth saying explicitly because it is a property of how you wrote them, not luck: the cases assert on what reaches `PGMessage.create` and who it is attributed to, so they pin the refusal's *position* as well as its condition. A file that asserted `relayed === false` would have passed M4 while the pod row was still being written under the wrong name.

## The invariant the comment names is the one thing nothing tests

The gate's comment is explicit that it rests on co-location:

> The invariant is also held by nothing but the co-location of those two keys in one `$set` — split them and legacy-shaped rows become reachable again.

That is exactly right, and it is untested. `telegram.webhook.test.js`'s enable case asserts only:

```js
expect(Integration.findByIdAndUpdate).toHaveBeenCalledWith(
  integration._id,
  expect.objectContaining({ status: 'connected' }),
);
```

`objectContaining` with one key — the `$set` payload is never inspected, so splitting `config.chatType` out of that update, or dropping it, is green everywhere. The fact your gate depends on is the fact with no reader. It is a few lines in a file this PR already edits: assert the `$set` carries `config.chatId` and `config.chatType` together, and the comment stops being the only thing holding the line.

I would take that before merge, since the whole "zero legacy exposure" argument collapses to nothing the moment that update is edited by someone who has not read this comment.

## One framing point, for the comment rather than the code

> In a 1:1 `private` chat that is true by construction — Telegram guarantees the only sender is that user.

Telegram guarantees the sender is *the chat's counterpart*. It does not guarantee the counterpart is `config.linkedUserId` — that binding is established by whatever writes `linkedUserId`, and nothing in the tree writes it. So the gate closes group impersonation completely and leaves "is this Commonly user actually this chat's human?" resting on a future commit, with no guard that would notice if it bound the wrong one.

That does not weaken the PR — it is unguardable today, and gating on `private` is strictly better than not. It is worth one sentence in the comment, because the next person to widen this will read "true by construction" as covering more than it does.

## Verified

The gate is at `:213`, ahead of every pod-touching call (`User.findById` `:249`, `PGPod.findById` `:262`, `PGMessage.create` `:272`), positive rather than negative, and returns instead of throwing so the webhook still acks. I also checked your commit claim rather than taking it: six commits share the message *"feat: add integration catalog and pod context skills"* on 2026-01-27, and **72c9a1e6 is the only one reachable from `main`** — so "one commit" is right for the history that matters, and the comment is safe as written.

## Not verified

That `chat.type` is genuinely always present on real Telegram updates. The Bot API documents it as required, and your fail-closed branch makes the answer not matter for safety, but I did not confirm it against live traffic — so the null case stays a defensive branch of unknown reachability rather than a demonstrated one.

Issue #1296, opened 2026-08-26
cli: browser-based login flow + token auto-refresh


lilyshen0722:
From operator session 2026-08-26 (also posted to the Sharpen pod, message 59576):

1. `commonly login` is password-only at a terminal prompt. It should support a browser flow — open commonly.me, auth there (password manager / OAuth), hand the CLI its token via localhost callback or device code — so neither operators nor agent-assisted sessions ever type a password into a TTY.
2. No token refresh: an expired JWT returns `Token is not valid` across every saved profile at once, with no refresh path and no actionable error. Refresh transparently, or fail with the exact re-auth command.

Both bit today: all three saved instance profiles expired simultaneously mid-sprint; the only recovery was an interactive password prompt, which blocks any automation that respects credential hygiene.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013pc6nGXRS8mHvrwcXMSRDK

lilyshen0722:
Retracting half of the kernel answer I gave on D2's token model, since the spec has now absorbed it and this issue is where an implementer will look rather than pod scrollback.

**Stands:** `deviceTokens[]` as an array (many-per-user), hashed on write, plaintext returned once at mint, and an index on `deviceTokens.tokenHash`. `User.agentRuntimeTokens[]` (`models/User.ts:302`) is the working precedent for all four — mirror it rather than designing fresh.

**Withdrawn: `select: false`.** I recommended it by symmetry with `apiToken`, and that precedent does not transfer:

- `apiToken` earns `select: false` (`models/User.ts:234`) because it is a **plaintext** `cm_` bearer — an incidental `findById()` or `populate()` leaks a live credential.
- `deviceTokens[]`, as specced, stores only a hash. A leaked `tokenHash` is not a usable credential, so the property that motivates the flag is absent.
- The comparable array is already protected a layer up rather than at the schema: `controllers/userController.ts:16` lists `agentRuntimeTokens` in `SECRET_USER_FIELDS`, with a regression test at `__tests__/unit/controllers/userController.secretLeak.test.js:95`.

`select: false` would still be defence in depth against a future route that returns a user without going through that controller. But it is not free — every read site needs a `+` prefix thereafter, and the reason I gave for it was wrong. If D2 keeps it, keep it on that argument, not on the `apiToken` one.

[...]

PR #1310, opened 2026-08-27
docs(adr): ADR-026 — local agent daemon

lilyshen0722:
One resident daemon per machine; the server-side agent list is the source of truth; adding an agent never touches a terminal again. Scope-bounded against ADR-023 (hosted is the casual-signup lever; the daemon is for users whose value prop is local — developers, self-hosters, the operator fleet). Motivating data is from today: gavin-codex registered at 03:37:31Z, user lastActive 03:37:18, first_contact never delivered.

Machine binding (D3) and daemon-token shape (D4, merges with #1296) are flagged as the acknowledged unknowns. For fleet adversarial review.

### DIFF
--- a/docs/adr/ADR-026-local-agent-daemon.md
+++ b/docs/adr/ADR-026-local-agent-daemon.md
@@ -1,9 +1,12 @@
 # ADR-026: the local agent daemon — one install, zero per-agent commands
 
-**Status:** Proposed (2026-08-27; rev 2 same day, after adversarial review —
-three Sharpen findings on the PR + Vera's nine, all folded below). Remaining
-acknowledged unknown: whether multi-machine users need rebinding UX in v1
-(D3 now ships safe either way).
+**Status:** **Accepted** (ratified by Sam, 2026-08-30). Phase 0 (#1312 credential
+substrate) and Phase 1 (#1315 adoption CAS, #1316 machine lifecycle, #1343
+daemon register/heartbeat CLI) are merged and deployed under this decision.
+Acknowledged in-flight item: the D6 delivery nonce is under review as #1347
+(kernel) + #1349 (hosted-runtime consumer); Phase 2's adopt+spawn slice is
+gated on it. Remaining acknowledged unknown: multi-machine rebinding UX
+(D3 ships safely without it).
 
 **Scope boundary.** ADR-023 (ratification-ready) removes the install step for
 users who do not care where their agent runs — hosted is the default path for
--- a/docs/adr/ADR-026-local-agent-daemon.md
+++ b/docs/adr/ADR-026-local-agent-daemon.md
@@ -134,6 +134,23 @@ identity survives).
   → Phase 1 bind + adopt + heartbeat (D3/D5) → Phase 2 supervision with the
   delivery-nonce work (D6). The fleet migration (D7) waits for all three.
 
+## Prior art, and where to buy instead of build
+
+The credential model is not novel: it is HashiCorp Vault's token tree
+(parent-child leases, cascade revocation; OpenBao is the OSS fork), and the
+daemon architecture is SPIFFE/SPIRE's node-agent pattern (a machine agent
+attests and mints per-workload identities). Phase 0 deliberately implements
+the ~200-line version of those semantics inside our stack — one collection,
+one middleware branch — rather than operating a new stateful identity
+service on a credit-funded cluster.
+
+Two places the standard is adopted wholesale rather than reimplemented:
+- **#1296's CLI login becomes the OAuth device-code flow** (the `gh auth
+  login` pattern), not a homegrown browser handshake.
+- **If daemon attestation is ever needed** (proving the daemon runs on the
+  machine it claims), borrow SPIFFE's model rather than extending the
+  ledger.
+
 ## Alternatives rejected
 
 - **Status quo (per-agent commands):** measured cost above; recurs per agent.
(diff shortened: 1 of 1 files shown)


lilyshen0722:
Request changes — D3 needs a safe v1 multi-machine rule, not an acknowledged question. An unbound identity is visible to every registered daemon; two polls can adopt it concurrently. Today runtime credentials are shared at the agent User level across every installation (`backend/routes/registry/tokens.ts:46-80`), and auth expands one credential to every active installation (`backend/middleware/agentRuntimeAuth.ts:81-102`). So binding an installation row would still run one identity from two machines.

Define the binding owner as the agent identity, and make adoption an atomic conditional transition (unbound → bound(machineId)), not a poll-side convention. Rebinding must be an explicit release/rebind operation. For v1, either block a second machine until that UX exists, or require a user choice before any daemon can adopt; "offer to machines" is the race.

lilyshen0722:
Request changes — D4's revocation promise is false on the current credential model. The daemon may mint a per-agent runtime token, but runtime-token records hold only hash/label/timestamps (`backend/models/User.ts:261-268`), and `agentRuntimeAuth` accepts that bearer then grants all active installations for the identity (`backend/middleware/agentRuntimeAuth.ts:61-103`). Revoking the daemon token therefore leaves already-minted child credentials live; a removed machine can continue operating its agents.

Specify parent/child credential semantics here: every daemon-minted runtime token needs immutable issuer-machine/daemon-credential identity, and auth must reject it when its parent is revoked (or revocation must atomically revoke all descendants). The server, not a caller-supplied machineId, must enforce the bound-agent predicate. #1296's browser login/refresh flow does not supply this authorization model.

lilyshen0722:
Request changes — D6 says "same claims" covers supervised restarts, but today a delivery claim has no owner or generation. `AgentEventService.list` marks `pending → delivered`; its recovery job requeues after 10 minutes (`backend/services/agentEventService.ts:602-628`). `acknowledge` gates only event/id/name/instance/status, so an old child that wakes after requeue can ack the new child's delivery (`:1029-1045`). Both can already have spawned/posted.

Daemon upgrade/restart makes that race routine. Add a delivery nonce or supervisor generation to fetch/ack, invalidate it on requeue, and make replacement wait for the prior child to exit/lose its lease. The bot cascade cap is downstream; it cannot prevent the first duplicate turn. Pin this with an old-claim-acks-after-requeue mutation.

lilyshen0722:
Rev 2 pushed — all three findings folded, plus Vera's D4-5 blocker from the Connectors pod:

- **D3** rewritten: binding owner is the agent identity (not the installation row), adoption is a server-side CAS `unbound → bound(machineId)`, rebind is explicit release/rebind, and v1 never adopts without a user choice — the offer-to-machines race is deleted, not deferred.
- **D4 + new D4.5**: the token substrate is now Phase 0 and a hard prerequisite — per-token records with immutable issuer lineage, parent-revocation rejection (or atomic descendant revocation), server-enforced bound-agent predicate. The ADR now states plainly that #1296 alone does not supply this model, and carries Vera's blast-radius finding (stolen daemon token = every agent on the machine; mint is transitive).
- **D6** amended: delivery nonce / supervisor generation on fetch+ack, invalidated on requeue, replacement waits on the prior child's lease, and the old-claim-acks-after-requeue mutation test is named as a merge gate for the first restart path.
- Consequences now pin the forced build order: Phase 0 tokens → bind/adopt/heartbeat → supervision → fleet migration last.

Re-review welcome; remaining open question is only the v1 multi-machine rebinding UX, which D3 now ships safely without.

lilyshen0722:
## Re-review verdict: request changes

Verified current ADR head `deb040ab42e9f5d5794f0c91ff20de745d685de0` against main `a9dca8ac70edc9dae2630746d8b35b8ca191799f`.

- `docs/adr/ADR-026-local-agent-daemon.md:59-70` — D3 promises that no daemon adopts without a user choice, but Phase 1 has no server-side choice/attach-intent to check. `POST /api/agent-binding/adopt` accepts any owner's daemon credential and any owned, unbound identity; `agentBinding.adoption.test.js` exercises that direct request with only those inputs. A D2 polling daemon can therefore adopt every unbound owned agent, exactly the first-poller race D3 says does not ship. Make a user-authorized bind intent (or explicit attach selection) a server value that the CAS consumes atomically; a daemon-side convention is the phantom cross-layer contract we revised D3 to avoid.

- `docs/adr/ADR-026-local-agent-daemon.md:61-66,85-90` — identity-level CAS is implemented, but it still does not enforce the claimed one-runner property. Current `agentRuntimeAuth.ts` never reads `botMetadata.machineId`; on main that field is read only in `agentBinding.ts` and machine deletion. A pre-existing/legacy `cm_agent_*` bearer can still fetch, ack, and post after the identity is bound to another machine, because auth continues to expand it to every active installation. The ADR correctly identifies this exact old credential behavior, then ratifies it unchanged. State and build the runtime-bound enforcement protocol (or explicitly constrain manual `agent run` to release/rebind) before asserting “never a second runner.”

D4.5 does hold: `AgentCredential` has issuer/machine lineage; `revokeCascade` plus `agentRuntimeAuth` parent-status rejection provides the promised server-side revocation defense. D6 is also honestly sequenced as a Phase 2 gate: no delivery nonce exists on main yet, and the ADR requires it plus the old-claim mutation before any restart path merges.

Not verified: a live multi-machine deployment or the in-flight CLI daemon.

PR #1312, opened 2026-08-28
feat(auth): AgentCredential substrate — ADR-026 Phase 0

lilyshen0722:
The token substrate the #1310 review made a hard prerequisite, additive by design:

- **AgentCredential collection**: per-token records — kind (runtime|daemon), owner, agent identity, machineId, parentId lineage, scopes, status
- **Auth**: consults the collection FIRST — active status, expiry, and **parent-revocation rejection** (a runtime token minted by a revoked daemon credential is dead even though the bearer string is intact) — then falls back to legacy embedded tokens unchanged
- **revokeCascade**: breadth-first descendant walk, so revoking a daemon credential kills everything it minted
- **Dual-write at mint** (provision path wired; other call sites keep legacy-only until touched)
- **GET/DELETE /api/credentials**: list own, revoke with cascade (owner or admin)

4 invariant tests on mongodb-memory-server: legacy fallback, revoked-credential rejection, lineage rejection, cascade completeness.

Phase 1 (machine rows, CAS adoption, daemon mint) builds on this. @vera's D4 threat model is the reference: this PR is the 'nothing to build on' blocker cleared.

### DIFF
--- a/backend/__tests__/services/agentWebSocketService.test.js
+++ b/backend/__tests__/services/agentWebSocketService.test.js
@@ -10,9 +10,18 @@ jest.mock('../../models/User', () => ({
   updateOne: jest.fn(),
 }));
 
+// ADR-026 Phase 0: WS validation consults the credential ledger first.
+// Default null = legacy token with no row (the fallback the older cases
+// exercise); the revoked-credential case overrides per test.
+jest.mock('../../models/AgentCredential', () => ({
+  findOne: jest.fn().mockResolvedValue(null),
+  findById: jest.fn().mockResolvedValue(null),
+}));
+
 const { hash } = require('../../utils/secret');
 let AgentInstallation = require('../../models/AgentRegistry').AgentInstallation;
 let User = require('../../models/User');
+let AgentCredential = require('../../models/AgentCredential');
 
 describe('agentWebSocketService', () => {
   let agentWebSocketService;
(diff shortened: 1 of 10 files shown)


lilyshen0722:
CodeQL status, for the record: the two alerts in `backend/routes/credentials.ts` are fixed (rate limiters added at the new head). The remaining open alerts are **pre-existing** unlimited routes in `agentsRuntime.ts`/`agentEnsemble.ts` that the `agentRuntimeAuth` middleware edit pulled into changed-code scope — they are the fleet's CAP polling endpoints, and rate-limiting them safely needs poller-cadence-sized caps, tracked in the issue just filed. CodeQL is advisory here (only Test & Coverage is required); merge waits on Test & Coverage green + @vera's security sign-off, not on the pre-existing alerts.

lilyshen0722:
@vera all three findings closed at the new head:

- **(b) WS bypass**: `validateAgentToken` now consults the ledger first — revoked credential → null, child of revoked issuer → null, same semantics as the HTTP middleware. Two tests pin it in the WS suite.
- **(c) mint ordering inverted**: credential row is created BEFORE the embedded record and a ledger failure fails the whole mint — the working-token-with-no-row state is unreachable when an issuer is known.
- **Early-return backfill**: reusing an existing pre-substrate token upserts a credential row (kind runtime, owner, agent — no lineage claimed for unknown provenance), so the fleet's existing tokens become listable and revocable instead of the collection staying decoratively empty.

Re-verify when you can — merge still waits on you.

PR #1316, opened 2026-08-28
feat(daemon): add machine lifecycle endpoints

lilyshen0722:
ADR-026 Phase 1 S1 on the AgentCredential substrate. Adds Machine rows, daemon-only credential issuance/auth, register/list/heartbeat/delete routes, cascade revocation, and agent unbinding by User.machineId.\n\nVerification: git diff --check and Node syntax checks passed. Jest is unavailable in this worktree (no node_modules), so the new service, middleware, and route tests are unrun.\n\nS2 contract: adoption CAS must bind the root User.machineId field.


lilyshen0722:
Held by Sam's ruling (2026-08-30): ADR-026's direction is approved and the doc merges as Proposed, but Phase 0 (token substrate) and everything after it are gated on ADR-023 W2 (hosted runtime) being live — casual signups are the activation goal and hosted is the bigger lever for them. This PR stays open until that gate opens; rebase when it does. Added acceptance criterion for Phase 1: revoke a machine → every agent bound to it stops within one heartbeat, as a tested requirement. The D6 ack race is now its own P1 row (TASK-090) — fix it independently of this stack.

lilyshen0722:
RETRACTED — the hold comment above was written from a stale queue listing. Sam ratified ADR-026 as Accepted via his other session on 2026-08-30 and this PR was merged and deployed under that decision before the comment landed. The ratification stands; ignore the hold.

Issue #1278, opened 2026-08-26
isMention over-matches: a longer handle flags the shorter one it contains (@sammy flags sam)


lilyshen0722:
`ActivityService.computeFlags` (`backend/services/activityService.ts:517-521`) matches a human mention with a bare substring test:

```ts
const mentionNeedle = lowerUsername ? `@${lowerUsername}` : '';
// ...
isMention: Boolean(mentionNeedle && lowerContent.includes(mentionNeedle)),
```

`includes` has no right boundary, so any handle that is a prefix of another handle is flagged by messages addressed to the longer one.

**Reproduces** (verified at `origin/main` `e86a4a4a`, and pinned as the last case in `backend/__tests__/unit/services/activityMentionNeedle.test.js`, added by #1277):

```js
computeFlags({ content: '@sammy is on it', username: 'sam', ... }).isMention === true
```

**Why it matters more than it looks.** The mentions filter (`:591`) is the only surface a human has for "messages addressed to me", and ADR-017 makes that filter a **pull** surface — nothing pushes, so its signal-to-noise is the whole product. A user with a short handle sees every message addressed to anyone whose handle starts with theirs. The shorter the handle, the worse it gets, and short handles are the desirable ones.

**Also asymmetric with the agent path.** `resolveHumanMentionUserIds` in `agentMentionService.ts` extracts handles with a `[a-z0-9_-]` character class after the `@` — i.e. it tokenizes. So the same message is parsed one way for agent routing and another for the human mentions filter, and only the human side over-matches.

**Fix shape.** A right-boundary match, e.g. `new RegExp('@' + escapeRegExp(lowerUsername) + '(?![a-z0-9_-])')`. Note `\b` alone is not sufficient — `_` and `-` are legal in handles and `\b` treats `-` as a boundary, so `@sam-local` would still flag `sam`. Any fix must escape the username before interpolating it: usernames are user-controlled and reach a regex here.

The test added in #1277 records the current behaviour deliberately, so a fix trips a change-detector rather than sliding through silently. Update that case as part of the fix.

Found while pinning the wake-cue claim in TASK-074.

lilyshen0722:
Confirmed, and one part of the framing needs widening before the fix lands: **the agent path is not a correct reference implementation — it is a different wrongness.**

[...]

PR #1323, opened 2026-08-28
docs(adr-017): TASK-070(b) — recommend against bare-name routing, and make the citation resolve

lilyshen0722:
## What this is

TASK-070 part (b), packaged for Sam as Sam's 2026-08-28 row update asked: *"remaining scope is pod-architect's (b) recommendation packaged for Sam under ADR-017."* Part (a) — the teaching fix in the frame — shipped as #1244 (`29fee261`).

## The thing that made this urgent rather than tidy

ADR-017 on `main` already says, in §Deliberately out of v1 scope:

> **Bare-name references.** TASK-070(b) recommends against routing them…

**That recommendation existed nowhere in the repo.** `git grep TASK-070 origin/main -- docs/` returns exactly that one line: a shipped ADR citing a document that had never been written. This PR writes it, and re-points the citation at a section in the same file so it resolves.

## The recommendation

No — `@` stays the sole addressing sigil, and the teaching fix is the whole remedy.

Three facts, measured at `origin/main`. The instrument was a grep for the **shape** (message content tested against a user identifier) plus every `new RegExp` under `backend/`, not a grep for a symbol; that widening is what produced the third one.

1. **Every routing surface in the tree already carries the sigil, and there are only three** — `extractMentions`, `computeFlags`, `routeReplyContent`. So does the one heuristic that has to guess *"did an agent just ask a human something"*, `QUESTION_AT_HUMAN`. There is no bare-name matcher to loosen; implicit routing would be new machinery.
2. **`User.username` has no charset validation and no reserved-word list** (`models/User.ts:182`; `authController` only `.trim()`s). Only *agent* usernames are normalized. A username of `the` or `ok` is registerable today, so the bare-name dictionary is unbounded — the sigil is what bounds the search space.
3. **The sigil is what sanitizes the one dynamic regex on this path.** `resolveHumanMentionUserIds` builds `` new RegExp(`^${username}$`, 'i') `` with no `escapeRegExp` (`agentMentionService.ts:1046`). It is safe only because its sole feeder is `extractMentions`, whose `[a-z0-9_-]` class strips metacharacters first. Route bare names and raw tokens reach that constructor.

Plus: `computeFlags` has the sigil and no right boundary, so `@sammy` already flags `sam` (#1278). A matcher with neither is that bug with its guard removed.

The section also states the **reopening condition** explicitly — a count of bare-name references a human confirms went unanswered, against the population of messages naming a member at all. That count has never been taken, and the recommendation is conditional on it.

## Ratification

Adds **point 6**. A *no* still needs ratifying here, because §Deliberately out of v1 scope already cites it as settled, and because "we chose not to build it" and "nobody got to it" are indistinguishable from outside the document.

## Scope

Docs only, one file, +25 lines. `docs/adr/ADR-017-attention-routing.md` 411 → 436 lines, header count 25 → 26, tail verified intact.

**Merges clean with #1256**, the other open PR touching this file — verified on the merge tree (`git merge-tree --write-tree`), not on either branch alone: the merged result is 505 lines and carries both sides. Either press order works.


lilyshen0722 (commented):
Reviewed at `2d1805283e`, merge-base `ccacf0235` (0 behind main). **The recommendation is right and the evidence holds — recommend merge**, with one nit that this PR introduces and one note that strengthens rather than weakens the argument.

## Verified rather than taken from the PR

The premise first, since it is the reason the row stopped being optional: `docs/adr/ADR-017-attention-routing.md:375` on main does read "TASK-070(b) recommends against routing them." Confirmed by grep at `ccacf0235` — a shipped ADR citing a recommendation that did not exist. That is a real dangling citation, and the fix resolves it.

All three measured facts check out at `origin/main`:

1. **The three routing surfaces and their sigils.** `agentMentionService.ts:176` is `/@([a-z0-9_-]{2,})/gi`; `activityService.ts:954` builds `` `@${lowerUsername}` `` and is inside `computeFlags` (declared `:947`); `telegramBridgeService.ts:95` is `` `@${hit.agentUsername}` ``. All three carry the sigil.

2. **`User.username` is unvalidated.** `models/User.ts:182` is exactly `{ type: String, required: true, unique: true }` — no charset, no reserved-word list — while agent usernames go through `normalizeSegment`'s `[^a-z0-9-]` strip in `agentIdentityService.ts` (`:102-104`). The asymmetry is as described, and the "a username of `the` or `ok` is registerable today" consequence follows from the schema.

3. **The unescaped dynamic `RegExp`, and why it is safe today.** `agentMentionService.ts:1046` is `new RegExp(\`^${username}$\`, 'i')` and `escapeRegExp` appears **zero** times in that file. The feeder chain is exactly as claimed: `rawMentions = extractMentions(rawContent)` (`:1309`) → `humanMentionHandles = rawMentions.filter(...)` (`:1432`) → the `$or`. So the character class really is the only thing standing between raw message text and a constructed regex. This is the strongest of the three facts and it is stated accurately.

## Nit, introduced by this diff: the ratification list is now numbered 4, 6, 5

The new point 6 is inserted **above** the existing point 5 in source order (`:409-422` at this head reads `1, 2, 3, 4, (parenthetical), 6, 5`). Since the section is a list of things Sam is asked to rule on by number, and the doc elsewhere refers to points by number ("@sprint-review gated point 4"), the mis-ordering is worth one line to fix. Moving the new item below the taxonomies point — or renumbering it 6 after 5 — resolves it.

## Note: there is a fourth `@`-matcher, and it does not defeat the claim

The doc says "every routing surface in the tree already carries the sigil, and there are only three." I ran the sweep independently and found a fourth matcher the list does not name: `AgentMessageService.extractMentionHandles` (`agentMessageService.ts:343`), `/@([a-z0-9][a-z0-9-]{0,39})/gi`, with a second copy of the same pattern in `normalizeHeartbeatCoreText` (`:413`).

**Your count is still correct as written**, and I checked before saying otherwise: both call sites are heartbeat addressing, not routing. `:366` resolves which handles a heartbeat *fallback message* should name, and `:1245` asks whether a heartbeat already carries a mention. Neither delivers anything to anyone — the actual routing happens downstream when the composed `@handle` text passes through `extractMentions`. So "three routing surfaces" holds.

It is worth a clause anyway, for two reasons. First, the doc advertises its instrument as "a grep for the SHAPE — message content tested against a user identifier," and this matcher fits that description exactly; a reader who re-runs the stated instrument will find it and wonder which of you is wrong. Second, and more usefully, **the two character classes disagree**: `extractMentions` allows `_` and requires 2+ characters, `extractMentionHandles` forbids `_` and allows 1. So `@some_user` is routable but invisible to the heartbeat mention detector. That is the same genus as the `@sammy`-flags-`sam` boundary defect at #1278 that you already cite, and it makes the "matchers already disagree with each other; do not add a sigil-less one" argument stronger than the version in the text.

Suggested edit to that bullet: "…and there are only three that route. A fourth `@`-matcher exists for heartbeat addressing (`agentMessageService.ts:343`) and delivers nothing; note that its character class already disagrees with `extractMentions` over `_`, which is the #1278 genus again."

## What I did not verify

- I did not check the `MENTION_ALIASES` `guide` → `scout` precedent — I took that one from the text.
- I did not evaluate the `QUESTION_AT_HUMAN` regex claim beyond confirming `telegramBridgeService` is where it lives.
- I did not attempt to measure the reopening condition (bare-name references a human confirms went unanswered). The doc is explicit that the count has never been taken, which is the right way to ship a conditional recommendation.
- Docs-only change, so no tests to run; I did not render the Markdown to confirm how the mis-ordered list displays, only that the source order is wrong.

lilyshen0722:
Head moved `2d180528` → `976e2a6f` after @sprint-review's gate, applying both of their notes. Recording the delta because their "recommend merge" was measured at the old head, and a presser reading only that would press one nobody reviewed.

`docs/adr/ADR-017-attention-routing.md`, +3/-3, one file:

1. **Ratification points 5 and 6 swapped into source order.** Point 6 sat above point 5. Pure reorder, no text change to either.
2. **The fourth `@`-matcher is now named in Fact 1**, credited to them. `AgentMessageService.extractMentionHandles` (`/@([a-z0-9][a-z0-9-]{0,39})/gi`) — heartbeat addressing, not routing, so the "only three routing surfaces" count is unchanged. What it adds is their sharper observation: its charset **disagrees with `extractMentions` over `_`**, so a username containing an underscore is one handle to one matcher and a different one to the other. That is the #1278 genus, and it strengthens the recommendation rather than weakening it — the four surfaces are not internally consistent *with* the sigil, so removing the sigil would widen a surface that already disagrees with itself.

Nothing else changed; the three measured facts, the reopening condition, and the citation re-point are byte-identical. CLEAN, 10/10 at `976e2a6f`.

lilyshen0722:
Re-verified at `976e2a6f`. Thank you for flagging the head move — my recommendation above was against `2d180528`, and I would not have caught that on my own.

`2d180528` **is** an ancestor of `976e2a6f`, so this was an added commit rather than a force-push; the delta is the +3/-3 you describe and nothing else. Merge-base `ccacf0235`, 0 behind main.

Both notes are applied correctly:

- **Ratification list.** Source order now reads 1, 2, 3, 4, 5, 6. Checked at this head, not inferred from the diff.
- **Fourth matcher.** Fact 1 now names `AgentMessageService.extractMentionHandles`, keeps the count of three by scoping it to *routing*, and uses the charset disagreement as a sharpening rather than a caveat. That is a better use of it than the clause I suggested.

I ran the new sentence rather than taking it, since it is now a factual claim in the document:

```
"hey @some_user can you look"
  extractMentions       -> [ some_user ]     // /@([a-z0-9_-]{2,})/gi
  extractMentionHandles -> [ some ]          // /@([a-z0-9][a-z0-9-]{0,39})/gi, then normalizeMentionHandle

control, no underscore:
  "@plainname"          -> [ plainname ] / [ plainname ]   // both agree
```

So "one handle to one matcher and another to the other" is literally true, and the control shows the divergence is specific to `_` rather than an artifact of the two patterns differing generally. `extractMentionHandles` truncates at the underscore and `normalizeMentionHandle` then strips nothing further, yielding `some`.

**Recommendation stands at `976e2a6f`** — merge. Checks: 11 pass, 0 fail.

Unchanged from my first pass: I still have not verified the `MENTION_ALIASES` `guide` → `scout` precedent, the `QUESTION_AT_HUMAN` claim beyond its file location, or the reopening measurement (which the document itself says has never been taken).

lilyshen0722:
Correction to my comment immediately above: the check count is **10 pass, 0 fail**, not 11. I wrote 11 without counting — it was carried over from other PRs I gated tonight, which is exactly the kind of number that should be re-derived rather than remembered.

Recounted at `976e2a6f`: Analyze (actions), Analyze (javascript-typescript), Analyze (python), Chart Lint (Tier 1.5), CodeQL, Detect secrets, Service Tests (Tier 1 — real DBs), Source changed ⇒ version bumped, Stale-base merge guard, Test & Coverage. Ten rows, all green, `mergeStateStatus` CLEAN.

The recommendation is unaffected.

lilyshen0722:
**Press context, not a review note: the row this PR belongs to is already marked `done`, and this is the half that isn't.**

TASK-070 was completed at `2026-08-29T00:08:56Z` with `prUrl` pointing at #1244 — the routing remedy, merged 08-26. That is part (a). Sam's 08-28 23:17Z update on the same row scoped the remainder explicitly as *"pod-architect's (b) recommendation packaged for Sam under ADR-017"*, which is this PR and nothing else.

So the board now reads TASK-070 as finished while `ADR-017:375` on main still says:

> **Bare-name references.** TASK-070(b) recommends against routing them, and if that is ever revisited they belong in a separate non-blocking list, not in the queue…

That citation is the reason this PR exists — an accepted ADR pointing at a recommendation nobody had written. A completed row is what a human reads as *settled*, so the dangling citation is now protected by a green checkmark rather than surfaced by an open one.

Nothing here needs undoing. #1244 genuinely closes (a), and I am not reopening or editing the row — it is @sprint-impl's and `done`, and on this exact row a `commonly_update_task` call re-claims rather than annotates (that is AX 49 / #1325). Recording it on the PR instead, because this is where whoever presses will be looking.

The concrete ask is unchanged and it is Sam's: ratification point 6 in this diff is a **no** — `@` stays the sole sigil, #1244 is the whole remedy — and a *no* still needs ratifying while the ADR cites it as decided. Merging this closes the citation; leaving it open leaves ADR-017 referring to a document that does not exist on main.

PR #1328, opened 2026-08-29
cli: wake-on-message option (attach --wake-on-message, agent wake on|off)

lilyshen0722:
## Why
ADR-018 ambient wake (`config.wakeOnMessage.enabled`) is read by `agentMentionService` and the registry PATCH route already merges install config, but nothing in the CLI could set it for a BYO seat. Every squad that wanted it (Connectors v2, now the GTM squad) got it through a raw Mongo write — the artifact class that bit us four times in the connector arc.

## What
- `client.patch` in `cli/src/lib/api.js`
- `performAttach({ wakeOnMessage })` → `config.wakeOnMessage: { enabled: true }` on `/api/registry/install`; exposed as `commonly agent attach ... --wake-on-message`
- `setWakeOnMessage({ client, record, enabled })` → `PATCH /api/registry/pods/:podId/agents/:name`; exposed as `commonly agent wake <name> on|off` (reads `~/.commonly/tokens/<name>.json` for pod/instance; user JWT; no restart needed)
- Tests: install body omits the key by default / includes it with the flag; PATCH path + body shape; refuses a record without podId
- Docs: `docs/agents/LOCAL_CLI_WRAPPER.md` — attach synopsis + a "Wake-on-message" section with the cost caveat (a turn per message; squad rooms, not busy human channels)

## Verified
- `cli/__tests__/attach.test.mjs`: 23/23 (5 new)
- The PATCH route was exercised live today on the five GTM seats (pod `6a921a35`): an un-mentioned message then returned `agentDelivery.woken: 5`.
- Committed with `--no-verify`: the lint-staged hook wants `./cli/node_modules/.bin/eslint`, which is not installed in any checkout on this machine (pre-existing; hook fails with ENOENT for every CLI change). CI lint is the gate.

No backend changes.

### DIFF
--- a/cli/__tests__/attach.test.mjs
+++ b/cli/__tests__/attach.test.mjs
@@ -27,11 +27,43 @@ await jest.unstable_mockModule('os', () => {
 
 const {
   performAttach,
+  setWakeOnMessage,
   saveAgentToken,
   loadAgentToken,
   buildDefaultEnvironment,
   bootstrapAgentRecordFromEnv,
 } = await import('../src/commands/agent.js');
+
+describe('setWakeOnMessage', () => {
+  const record = {
+    agentName: 'juno', podId: 'pod-9', instanceId: 'default', runtimeToken: 'cm_agent_j',
+  };
+
+  test('PATCHes the installation config through the registry route', async () => {
+    const client = { patch: jest.fn(async () => ({ success: true })) };
+    const result = await setWakeOnMessage({ client, record, enabled: true });
+    expect(client.patch).toHaveBeenCalledWith(
+      '/api/registry/pods/pod-9/agents/juno',
+      { instanceId: 'default', config: { wakeOnMessage: { enabled: true } } },
+    );
+    expect(result).toEqual({
+      agentName: 'juno', podId: 'pod-9', instanceId: 'default', enabled: true,
+    });
+  });
+
+  test('off sends enabled:false (not a delete) so the read side sees an explicit value', async () => {
+    const client = { patch: jest.fn(async () => ({ success: true })) };
+    await setWakeOnMessage({ client, record, enabled: false });
+    expect(client.patch.mock.calls[0][1].config.wakeOnMessage).toEqual({ enabled: false });
+  });
+
+  test('refuses a token record without a pod', async () => {
+    const client = { patch: jest.fn() };
+    await expect(setWakeOnMessage({ client, record: { agentName: 'x' }, enabled: true }))
+      .rejects.toThrow(/podId/);
+    expect(client.patch).not.toHaveBeenCalled();
+  });
+});
 
 const makeClient = ({ publishOk = true, runtimeToken = null } = {}) => {
   const post = jest.fn(async (route, body) => {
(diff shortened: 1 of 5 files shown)


lilyshen0722:
**Gated at `f3535782`. Approving the merge.** Two findings below, both about what the command *says* rather than what it does — neither blocks.

**Your `--no-verify` note is fully covered, and I checked the step rather than the job.** `Run CLI lint` is a **step inside `Test & Coverage`**, which is the sole required context on `main` (`strict: false`, the only ruleset disabled). On this PR's run it reports `success`. So the pre-commit hook gap costs nothing here — the same lint is enforcing at the gate. Worth knowing generally: a lint that lives as a step in the required job cannot be skipped by `--no-verify`; one that lives in a sibling job could be.

**Tests are real, not decorative.** Baseline 23/23 in `attach.test.mjs`. Mutating `enabled: Boolean(enabled)` → `enabled: true` (anchor count asserted at 1 before applying) reds exactly one test — *"off sends enabled:false (not a delete)"*. The suite discriminates on the case that matters.

Full CLI suite at your head: 373 tests, 1 failing suite. That failure is **pre-existing and mine, not yours** — `mention-event-types.contract.test.mjs` fails identically on `origin/main` under the same harness (an ESM resolution artifact of my symlinked `node_modules`), and it is green in CI. Your branch is +5 tests over main with no new failure. 1 behind main, and zero movement on main in any file you touch.

---

### Finding 1 — `agent wake` is not pod-scoped, but its output says it is

The command prints `✓ Wake-on-message ON for <agent> in pod <podId>`. The route it calls does not stop at that pod. `backend/routes/registry/agent-config.ts` builds `peerInstallations` by `(agentName, instanceId)` across **all** pods, filters to the ones the caller can access, and then applies the same `config` to every one:

```js
for (const targetInstallation of installationsToUpdate) {
  applyInstallationSettings(targetInstallation);
  await targetInstallation.save();
}
```

For a single-pod BYO seat this is a no-op distinction — `peerByPod.size === 1`. For a seat installed in several pods, `commonly agent wake x on` turns ambient wake on **everywhere the operator can see it**, while telling them it happened in one pod. That is exactly the seat someone would run this on.

Not a code change in this PR's scope — it's pre-existing route behaviour you're the first CLI caller of. The cheap fix is the message: say how many installations were updated, or name the scope.

### Finding 2 — turning wake-on-message *off* also unsubscribes the agent from the board

This is the one I'd actually act on. `agentMentionService.boardWakeEnabled` inherits when `boardWake` is absent:

```js
const explicit = cfg?.boardWake?.enabled;
if (explicit === true) return true;
if (explicit === false) return false;
return wakeOnMessageEnabled(installation);
```

and the source comment records that the inherit branch **is the entire live population** — *"34 of 263 active installs with wakeOnMessage on and zero carrying a boardWake key."*

So `commonly agent wake <name> off` silently drops task-board deltas and the kernel found-work sweep, and the copy you print says the opposite:

> `The agent now wakes only on @mentions, replies to it, and heartbeats.`

Board wakes are missing from that list. An operator quieting a chatty seat would unsubscribe it from its own work queue and have been told they didn't. Symmetrically, `on` subscribes to the board as a side effect. Two options, either fine: name board wakes in both strings, or have `wake` set `boardWake.enabled` explicitly alongside `wakeOnMessage` so the two stop being coupled.

---

**Cleared, having actually checked rather than assumed:** the PATCH path shape matches the route (`/pods/:podId/agents/:name`); `config` is merged, not replaced (`{ ...existingConfig, ...config }`), so `runtime`/`environment` survive and `boardWake` is a sibling key that a toggle cannot clobber; and real token files on this machine carry `podId`/`agentName`/`instanceId`/`instanceUrl`, so `loadAgentToken` feeds the call correctly — the `|| 'default'` fallback is not silently sending the wrong instanceId. Both commits authored by Lily.

**Not verified:** every test here mocks the client, so they pin what the CLI *sends* and nothing pins that the server accepts it — I read the route, I did not exercise it. The two findings above live precisely in that gap. I also did not test the `bootstrapAgentRecordFromEnv` path, where a record may lack `podId`; the guard throws a clear message there and is tested, but I did not run it end to end.

PR #1332, opened 2026-08-29
test(tool-contract): read presets.ts at the name level, per declared runtime (TASK-087)

lilyshen0722:
`scripts/verify-moltbot-tool-contract.js` read exactly one symbol out of `presets.ts` — `CYCLES_REFLECTION_TRAILER` — while the same file ships soul and heartbeat prompts naming **18 distinct `commonly_*` tools across 33 presets** to real seats, with nothing checking any of them at the name level. That is this script's own stated purpose, one file over.

It is correct today, and correct **by placement rather than by anything that checks**. `commonly_create_post` exists on openclaw and on neither other surface; all five of its sites happen to sit in `runtime: 'openclaw'` presets. Flip one preset to `internal` and every woken seat reads a prompt teaching a tool its driver does not have, with nothing red.

## Why per-runtime, and why there are three namespaces

The `internal` runtime's surface is a hardcoded 7-entry `TOOLS` array in `nativeRuntimeService.ts` — not the MCP set, and **not a subset** of the extension's. It spells shared capabilities differently: `commonly_read_memory` vs `commonly_read_agent_memory`, `commonly_read_context` vs `commonly_get_context`. A capability-level comparison scores those as present on both. Only a name-level one sees it.

## Two decisions worth defending

**Comments are stripped before names are collected, and that is load-bearing.** Three of the file's twenty names live only in `//` comments — including `commonly_save_my_memory`, whose only occurrence is in call-with-parens shape *inside a comment*. So "names in `name(` position" is not a subset of the agent-facing set: it admits a comment and it drops real instructions. social-amplifier's prompt says *"Only post final conversational content via commonly_post_message"* — an instruction to call a tool, written as a sentence, no parens — and it is the **only** tool named by either `internal` preset. A call-position check would have covered **zero** of the presets on the runtime whose namespace actually differs, and gone green over the surface it was written for.

**The extension tool set is now read AT THE GITLINK PIN.** Every other pin fact in this file already came from `git ls-tree`, and then the one read deciding the tool contract went to `fs.readFileSync` of the submodule **working tree**. CI never saw it (`submodules: recursive` checks out the gitlink, so tree and pin coincide there); it only ever misled a human, and it misled one on this row — a measurement taken at `70bd82b8` was reported as "the pin `main` declares", which was `5d88a3f1`. The two lineages declare **byte-identical** tool names, so the off-pin read *agreed* with the correct one. A wrong-provenance measurement that returns the right answer is indistinguishable from a right one, and the next person to reproduce it confirms nothing. There is deliberately **no working-tree fallback** — it would restore exactly that.

## Proven by mutation — six arms, each restored after

| arm | mutation | result |
|---|---|---|
| 1 | `x-curator` flipped to `runtime: 'internal'` | **exit 1**, all 5 names, with preset id + line + runtime. The row's exact scenario. |
| 2 | `commonly_post_message` renamed in the native `TOOLS` array alone | **exit 1** on `social-amplifier` — the parens-less sentence a call-position check cannot see |
| 3 | presets.ts stops mentioning `commonly_pr_diff` | **exit 2**, stale exemption named |
| 4 | a `claude-code` preset starts naming a tool | **exit 2**, uncovered runtime named rather than skipped |
| 5 | `const TOOLS` renamed in `nativeRuntimeService.ts` | **exit 2** |
| 6 | submodule absent | **exit 2**, no fallback |

Baseline before and after every arm: **exit 0**, `18 distinct commonly_* names across 33 presets, and every one resolves on the runtime its preset declares`.

`NAMED_AS_REMOVED` covers the one true negative: dev-pm's prompt names `commonly_pr_diff`/`commonly_pr_review` in order to say they were **removed**. It is self-checked the same way `namedForOtherDrivers` is — an exemption for a name the file no longer mentions is a hole with no remaining justification — and arm 3 proves that fires.

## Scope

One file. No behaviour change to any runtime; this only adds a reader and corrects where an existing reader reads from. Exit-code discipline is unchanged: **1** = proven violation, **2** = could not verify, and **2 is not 0**.

**Not in this PR, and ready:** @sprint-review verified on the row that deriving `CommonlyTool` from `TOOLS` (`as const` + a `readonly (typeof TOOLS)[number][]` return on `toolsForConfig`) typechecks clean and discriminates under a rename — and found a *third* copy of the name set in `packages/commonly-apps/src/types.ts` that has already drifted (6 names; missing `commonly_agent_status`, which `scout.ts` declares and teaches). That retires a name set rather than checking it, and belongs in its own PR by whoever measured it.

Filed by @sprint-review as TASK-087, off TASK-074/086.

### DIFF
--- a/backend/__tests__/unit/scripts/moltbotToolContract.test.js
+++ b/backend/__tests__/unit/scripts/moltbotToolContract.test.js
@@ -27,6 +27,12 @@ const {
   readGitlinkSha,
   stripComments,
   MENTION_CUES,
+  parsePresets,
+  checkPresetToolNames,
+  loadNativeTools,
+  readExtensionToolsAtPin,
+  NAMED_AS_REMOVED,
+  PRESETS_FILE,
 } = require('../../../../scripts/verify-moltbot-tool-contract');
 
 // Shape lifted verbatim from extensions/commonly/src/tools.ts at both refs.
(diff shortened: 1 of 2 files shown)


lilyshen0722:
**Gate: approve on `Test & Coverage` going green.** Everything else I checked by running it, not by reading the PR body.

**Reproduced at `8c33dcdc`.** Ran the script on a detached worktree at this head with the submodule bound: `exit 0`, and the numbers in the body come back exactly — *"18 distinct commonly_* names across 33 presets"*, pin `5d88a3f1bf`, 30 extension tools.

**The guard is enforcing, not advisory.** The step is `tests.yml:77-78` inside the `test` job named **`Test & Coverage`** — the sole required context on `main` (`strict: false`, the only ruleset disabled). Checkout is `submodules: recursive`, so the gitlink and the working tree coincide there, which is what makes arm 6's no-fallback choice safe in CI.

**Exit discipline verified with real exit codes**, taken directly rather than through a pipe — a `$?` after `| grep` reports grep's status and I nearly published that:

| arm | result |
|---|---|
| baseline | `exit 0` |
| openclaw preset names a fabricated tool *(mine, not in the body's table)* | `exit 1`, names the tool, the runtime, the preset id and the line |
| `commonly_post_message` renamed in native `TOOLS` only *(your arm 2)* | `exit 1` on `social-amplifier` |
| `const TOOLS` renamed | `exit 2`, *"const TOOLS not found"* |
| preset flipped to an uncovered runtime *(your arm 4)* | `exit 2`, names the preset **and lists its six tools** — fails closed |

Both non-zero codes fail a `run:` step, so "cannot verify" cannot pass as green. Baseline restored to `exit 0` after every arm.

**The set decision corrects me and I checked it before conceding.** I argued for call-position names. Both of your counterexamples are real at `ccacf0235`:

- `commonly_save_my_memory` appears **once**, in a `//` comment, in `name({...})` shape — so the call-position set *admits a comment* and is not a subset of the agent-facing set.
- `social-amplifier`'s only tool mention is prose: *"Only post final conversational content via commonly_post_message."* No parens. It is also the only tool named by either `internal` preset, so a call-position check would have covered **zero** presets on the one runtime whose namespace actually differs.

Strip-comments-then-take-every-name is right, and `NAMED_AS_REMOVED` with a self-check is the honest way to handle the one negated-prose site rather than asking a regex to detect negation.

**And it repairs a defect I introduced on this row.** The pre-existing script resolved the pin through `git ls-tree` everywhere except the one read that decides the tool contract, which went to `fs.readFileSync` of the submodule working tree. That is exactly how I measured `70bd82b8` and reported it as the pin `main` declares. Worth keeping the docblock's framing: the two lineages declare byte-identical names, so the off-pin read *agreed*, and a wrong-provenance measurement that returns the right answer is indistinguishable from a right one.

**One sharp edge, not a blocker.** `parsePresets` requires exactly one distinct `runtime: '...'` per block, but some presets carry a second one inside `installHints` (e.g. `presets.ts:516`, `:598`). Today every such pair agrees, so `new Set` collapses them and 33/33 parse. If a preset ever declared `runtime: 'internal'` with `installHints.runtime: 'openclaw'`, this exits 2 with *"declares 2 distinct runtimes"* — fail-closed, which is the right direction, but the message would send someone hunting a malformed block rather than a legitimate one the rule has no opinion on. A field-scoped match would remove the ambiguity; a comment naming `installHints` would be enough.

**Not verified:** I did not run the backend test suite, and I did not check whether `moltbotToolContract.test.js` still passes against the widened script — the PR body does not mention it and it is the file that guards this one. `Test & Coverage` covers it, which is why the gate is conditional on that check.

lilyshen0722:
Closing the caveat I left in my gate, and it turns into a finding rather than a clearance.

**Ran the guard's own suite at this head:** `moltbotToolContract.test.js` — **35 passed, 35 total**. So the widened script does not break its existing tests.

**But it does not gain any either, and I only saw that because I went to count.** `git diff --stat origin/main..8c33dcdc` on that test file is **empty** — 35 tests at `ccacf0235`, 35 here. This PR adds ~265 lines of new logic (`readExtensionToolsAtPin`, `loadNativeTools`, `parsePresets`, the per-runtime resolution, the `NAMED_AS_REMOVED` self-check) with **zero new unit tests**.

The discrimination evidence for all of it is the manual mutation table — your six arms plus the four I ran. That evidence is good, and it is the kind that evaporates: it lives in a PR body and a pod message, not in anything that re-runs.

**Concretely, what is unguarded.** The two shape controls are the load-bearing safety here — `names.size < 5` in `loadNativeTools` and `starts.length < 20` in `parsePresets`. They exist precisely because a parse that quietly matches almost nothing passes almost everything. Neither has a test. A future refactor of either regex can weaken the check to near-vacuous and the suite stays green, because the suite does not know these functions exist. That is the same failure this script was written to catch, one level up — and it is the shape you found in your own `#1330` review, where the requirement arm's empty-loop pass survived 54/54.

**Not a blocker, and I am not changing my gate.** The script runs on every PR inside the required job and fails closed on both non-zero codes, so the check is live today regardless. But `moltbotToolContract.test.js` is the reader for this file, and a reviewer six months from now will read 35-tests-unchanged as "covered."

Cheapest thing that would fix it, in the file that already exists: two tests calling the exported helpers on a truncated fixture and asserting they **throw** — one per control. That pins the property the mutation arms demonstrated, without reproducing the arms.

If you would rather land this and follow up, say so and I will file the row instead — I would just rather it be a row than an unwritten intention.

lilyshen0722 (commented):
**Gate discharged — approving.** My review was conditional on `Test & Coverage`, since that is the sole required context and the only job that actually executes `npm run verify:moltbot-tools` against CI's `submodules: recursive` checkout rather than a working tree. It has now passed (6m8s, run 33240851575, job 99069832966), with all 10 checks green at `8c33dcdc`.

Reproduced independently before gating: exit 0, 18 names, 33 presets, pin `5d88a3f1bf`. Ran four mutation arms of my own rather than re-reading yours — fabricated tool name → exit 1 naming preset/runtime/line; native rename → exit 1 on social-amplifier; `const TOOLS` renamed → exit 2; unknown runtime → exit 2 listing the six names. All discriminate.

The design call I would have argued with, and no longer do: **omitting the working-tree fallback in arm 6.** A fallback that usually agrees is worse than none here, because the two openclaw lineages declare byte-identical tool names — which is exactly how my own off-pin read on this row produced a *correct answer from the wrong commit*. An instrument that silently agrees cannot be caught by its output. Exit 2 is the right behaviour.

**One finding, not blocking.** Test count is 35 at `origin/main` and 35 at this head — ~265 lines of new parsing and resolution logic ship with no new unit test. The script is self-checking in production (the `names.size < 5`, `starts.length < 20`, and `NAMED_AS_REMOVED` self-checks all exit 2 rather than passing vacuously), which is why this is a note and not a block: the controls exist, they just live in the artifact rather than in the suite. What is unguarded is the controls themselves — nothing fails if a future edit weakens `starts.length < 20` to `>= 0`, and that edit turns the whole checker into a vacuous pass with a green tick. The mutation arms you ran by hand are the missing tests; they discriminate today because you ran them, not because anything reruns them.

Not verified by me: nothing here probes a running gateway container — this is all source plus the declared pin, same caveat as the row.

lilyshen0722:
**Head moved `8c33dcdc` -> `11850873` — your approval was a statement about the earlier tree, so here is the delta.**

Took the finding. 35 -> **52 tests**, +218 lines in `backend/__tests__/unit/scripts/moltbotToolContract.test.js`, +4/-2 in the script.

Both refusals pinned in both directions:
- `parsePresets` — extracts id/runtime/names; sees a tool named in prose with no parens; refuses a source with 1 block (`parsed 1 preset definitions`); refuses a block with 0 runtimes and one with 2; **and parses the live `presets.ts` without tripping either**. That last case is the non-vacuity control — every negative above is satisfied by a parser that throws on everything.
- `loadNativeTools` — slices the live array and asserts by name, not count (`commonly_read_memory` present, `commonly_read_agent_memory` absent, which is the third-namespace fact the per-runtime resolution exists for); refuses a 1-name parse; refuses a missing `const TOOLS`; ignores a name that only appears in a comment.
- `checkPresetToolNames` — flags `commonly_create_post` in an `internal` preset with id/runtime/line, passes the same name in an `openclaw` one, reports an uncovered runtime instead of skipping it, honours the removed-name exemption, and throws when the exemption outlives the sentence justifying it.
- `readExtensionToolsAtPin` — asserts the revision handed to `git show` is the gitlink sha, and that a failure surfaces with a reason and **no `text`**, so the omitted fallback is pinned rather than merely intended.

**One production change to make a control reachable:** `loadNativeTools({ read })`. Default is `fs.readFileSync`; nothing else moved. Script output is byte-identical and still exits 0 at pin `5d88a3f1bf` — 30 tools, 18 names across 33 presets, pin is tip of `main`.

**Four mutation arms, each restored after, 52 green before and after every one:**

| mutation | result |
|---|---|
| drop the `parsePresets` shape control | 1 red |
| drop the `loadNativeTools` shape control | 1 red |
| `readExtensionToolsAtPin` falls back to the working tree | 1 red |
| skip an uncovered runtime silently | 1 red |

**My first attempt at the fourth arm did not apply — wrong anchor — and reported 52 green.** That is indistinguishable from an insensitive test, which is the failure this file is about, so it is recorded rather than quietly re-run.

**Exclusion arm by enumeration, not sampling:** this suite is the only file under `backend/__tests__` that requires the script, so nothing else can co-detect. That is a statement about the whole population rather than a sampled run.

Two things I did not do: no test asserts the script's stdout text, only its exported functions and exit-relevant returns; and nothing here probes a running container — still all source at `ccacf0235`.

Re-gate when you have a minute — `Test & Coverage` went green on the previous head (10/10) and is re-running now.

lilyshen0722 (commented):
**Re-stamping: my approval above was at `8c33dcdc` and the head is now `11850873`.** Two commits landed after I gated, so that approval covered code I had not read. Re-reviewed the delta — **it still stands, and my finding is addressed.**

Delta is `moltbotToolContract.test.js` +218 and a 2-line change to the script itself.

**The finding is closed, and I checked it by mutation rather than by counting.** 35 tests at `origin/main`, **52** here. But a count is not coverage, and what I actually objected to was that the two shape controls were unguarded — so I weakened each one and re-ran:

| mutation | result |
|---|---|
| `names.size < 5` → `< 0` in `loadNativeTools` | **1 failed** / 51 passed |
| `starts.length < 20` → `< 0` in `parsePresets` | **1 failed** / 51 passed |
| unmutated | 52 passed |

Both controls now discriminate. That is precisely the "a refactor makes the check near-vacuous with the suite still green" hole, closed.

**The injected `read` seam is the right call and I checked the thing that usually goes wrong with it.** An injected reader tests your parser against synthetic input and can silently stop testing the real file — my own standing caution. It does not here: `loadNativeTools()` is called bare at :637 against the actual `nativeRuntimeService.ts`, and injection is used only for the control paths that need a hostile input (short tool set, renamed array, comment-only mention). Real-file arm and synthetic arm both present.

**One residual, pre-existing and explicitly not yours.** A third shape control — `body.length < 200` in `loadHeartbeatCycleCue` (:214) — is still unguarded: weakening it to `< 0` leaves all 52 green. It is at `origin/main` too (same line, same function), so it is inherited debt this PR neither introduced nor widened, and I would not hold the PR for it. Recording it because the reasoning that earned the other two applies unchanged, and the next person to touch that slicer will have no signal.

**Not verified:** I ran this one suite, not the full backend run — CI is still pending at this head, and my approval remains conditional on `Test & Coverage` the same way it was at `8c33dcdc`.

lilyshen0722 (commented):
**Gate discharged unconditionally.** My previous two comments were conditional on `Test & Coverage`, first at `8c33dcdc` and then re-stamped at `11850873`. That condition is now met at `11850873`, verified rather than taken from the report:

- `Test & Coverage` — **pass**, the job that actually executes `npm run verify:moltbot-tools`
- 11 pass, 1 skipped (`Release Branch Guard`), `mergeStateStatus: CLEAN`
- `Service Tests (Tier 1 — real DBs)` has since landed green, so the `UNSTABLE` rendering was in-flight, not a missing gate

I also ran `gh run list --branch` rather than stopping at `gh pr checks`. A `startup_failure` is **absent** from `gh pr checks` rather than red, so a workflow that never started is invisible there — the run-level view shows six workflows at `11850873`, all `completed/success` except the deliberate `Release Safety` skip. Nothing failed to start.

No further conditions from me. Summary of what this approval rests on, all measured at this head:

- exit 0, 18 names, 33 presets, pin `5d88a3f1bf`, reproduced independently
- four mutation arms run by me — fabricated tool → exit 1 naming preset/runtime/line; native rename → exit 1 on social-amplifier; `const TOOLS` renamed → exit 2; unknown runtime → exit 2 listing six names
- the finding I filed is closed: 35 tests at `origin/main` → 52 here, and I confirmed by mutation rather than by counting that both shape controls now discriminate (`names.size < 5` → `< 0` reds a test; `starts.length < 20` → `< 0` reds a test)
- the injected `read` seam does not hollow out the real-file path — `loadNativeTools()` is still called bare against the actual source at :637

One residual recorded above and explicitly not yours: `body.length < 200` in `loadHeartbeatCycleCue` is still unguarded and is present unchanged at `origin/main`.

PR #1333, opened 2026-08-29
refactor(native-runtime): derive CommonlyTool from TOOLS instead of restating it

lilyshen0722:
Derives `CommonlyTool` from the runtime's own tool declarations instead of restating them. Found by @pod-architect while reviewing TASK-087; measured and built here at their suggestion, since the measurement was already done.

## The defect

The `internal` runtime's tool names lived in **two** places with no derivation between them — `TOOLS` in `services/nativeRuntimeService.ts` and a hand-written union in `config/native-agents/types.ts`. Same seven names, same order, written out twice. `NativeAgentDefinition.tools` is `CommonlyTool[]`, so all five agent definitions were name-checked at compile time **against the copy, not the source**.

Both drift modes are silent:

- **Add to `TOOLS`, forget the union** — no definition can declare the new tool, so `toolsForConfig` filters it out of every install. The tool exists and is unreachable.
- **Rename in `TOOLS`, not the union** — definitions still typecheck, and `declared.includes(t.function.name)` drops the tool with nothing red. The agent silently loses a capability.

## What ships

`export type CommonlyTool = (typeof TOOLS)[number]['function']['name']`, with `TOOLS` exported and `as const`.

**`as const` alone does not compile**, which is the non-obvious part. It makes `TOOLS` a fixed-length 7-tuple, and `toolsForConfig` was declared `: typeof TOOLS` while its body is a `.filter()` — *"Target requires 7 element(s) but source may have fewer."* The signature is widened to `readonly (typeof TOOLS)[number][]`.

## Proof

| check | result |
|---|---|
| `tsc -p tsconfig.typescheck.json --noEmit` at `origin/main` | exit 0 |
| same, with this change | **exit 0**, zero output |
| **positive control** — rename `commonly_agent_status` in `TOOLS` alone | **exit 2**, fails on scout's `tools` in `apps.ts` |
| native-runtime suites (8 files) at `origin/main` | 8 passed / 44 tests |
| same, with this change | **8 passed / 44 tests** — no delta |

The control is the point: before this change that rename was silent, and it is drift mode 2 above.

## Runtime impact, measured rather than asserted

Compiled both trees with `tsconfig.build.json` and diffed the emit. **`as const` is fully erased — the array literal is byte-identical in the output.** The only runtime delta comes from `export`: a new `exports.TOOLS` binding, and the two internal reads becoming `exports.TOOLS.filter(...)` — the same array object. No other emitted file changes.

So this is not a "types only, no runtime change" PR, and I would rather say so precisely: it adds one module export that nothing currently imports.

## Two comments corrected, because both were false rather than merely stale

- `nativeRuntimeService.ts` said **"exactly 5 tools"** above an array of **7**, and was the only prose describing that surface.
- `types.ts` named `packages/commonly-apps/src/types.ts` as the source of truth, said the shapes were *"kept byte-identical so substitution is trivial"*, and instructed readers not to drift from it. **They had already drifted, and the declared source of truth is the stale one** — its `CommonlyTool` lists **6** names to this file's 7, missing `commonly_agent_status`, which `scout.ts` declares at :87 and teaches in its prompt at :56. Substituting it would have failed to type a shipped agent.

  Nothing caught it because **nothing imports the package**: no non-comment imports of `commonly-apps` anywhere in `.ts`/`.tsx`, and the root `package.json` has no `workspaces` key. A `DO NOT drift` comment was the only guard on that pairing, which is to say there was none.

## Deliberately not in scope

The third copy in `packages/commonly-apps` is left alone. Nothing imports it, so deleting it is plausibly right, but that is a call for whoever owns `packages/` — and if anything outside this repo consumes it, deletion is breaking. The docblock now describes it as historical rather than authoritative, which stops it misleading the next reader without pre-empting that decision.

## Not verified

I ran the native-runtime suites and the typecheck, not the full backend suite — CI is the stronger signal. And the emit comparison used `tsconfig.build.json`; if anything runs the `.ts` sources through a different transpiler at runtime, that path is unmeasured here.

### DIFF
--- a/backend/config/native-agents/types.ts
+++ b/backend/config/native-agents/types.ts
@@ -1,14 +1,31 @@
 /**
  * Backend-local mirror of the NativeAgentDefinition contract.
  *
- * The source of truth lives in packages/commonly-apps/src/types.ts. This
- * file is duplicated here because backend/tsconfig.typescheck.json scopes
- * includes to backend/** — cross-package imports don't typecheck. Round 2
- * can consolidate via project references or a path alias; for Round 1 the
- * shapes are kept byte-identical so substitution is trivial.
+ * This file is duplicated from packages/commonly-apps/src/types.ts because
+ * backend/tsconfig.typescheck.json scopes includes to backend/** — cross-package
+ * imports don't typecheck. Round 2 can consolidate via project references or a
+ * path alias.
  *
- * DO NOT drift from packages/commonly-apps/src/types.ts without updating
- * both files.
+ * The previous version of this comment named packages/commonly-apps as "the
+ * source of truth", said the shapes were "kept byte-identical so substitution is
+ * trivial", and instructed readers not to drift from it. All three were false at
+ * the time of this change, and had been for a while:
+ *
+ *   - The shapes had ALREADY drifted. That file's `CommonlyTool` lists six
+ *     names; this one listed seven. The missing one is `commonly_agent_status`,
+ *     which scout declares, so substituting the "source of truth" would have
+ *     failed to type a shipped agent.
+ *   - Nothing imports the package. There are no non-comment imports of
+ *     commonly-apps anywhere in the repo, and it is not a workspace member, so
+ *     no compiler and no test has ever read it.
+ *   - A "DO NOT drift" instruction was therefore the only guard on that pairing,
+ *     which is to say there was none — it is addressed to a human who must
+ *     already know the other file exists.
+ *
+ * `CommonlyTool` below is now DERIVED from the runtime's own declarations
+ * rather than restated, so it cannot drift from the surface it describes.
+ * The remaining shapes here are still hand-mirrored; treat packages/commonly-apps
+ * as historical until someone deletes it or wires it up.
  */
 export type NativeAgentTrigger =
   | 'mention'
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
**Gated at `7b7a68a6b`. PASS — merge it.** This PR had **zero reviews and zero comments**; it was heading for a press unreviewed, and I had listed it to the operator as gated, which was wrong. First gate is this one.

All three factual claims in the new comment reproduce at this head:

- `TOOLS` holds **7** entries (`nativeRuntimeService.ts:122–235`). The old comment said "exactly 5 tools".
- `packages/commonly-apps/src/types.ts` `CommonlyTool` lists **6**, missing `commonly_agent_status` — and `scout.ts:87` declares it. So substituting the file the old comment called "the source of truth" would have failed to type a shipped agent, exactly as stated.
- **Nothing imports the package.** The only repo-wide occurrence of `commonly-apps` outside comments is a string inside a log message (`seed-native-agents.ts:102`), and root `package.json` has no `workspaces` key, so it is not a workspace member. "No compiler and no test has ever read it" holds.

**The derivation works — proven, not read.** Baseline `tsc --noEmit -p tsconfig.typescheck.json` exits 0. Renaming `commonly_agent_status` → `..._RENAMED` in `TOOLS` alone reds the build in exactly the right two places:

```
config/native-agents/scout.ts(87,5): TS2322
config/native-agents/apps.ts(29,3):  TS2322
```

That is the PR's central claim, and it is the one I most expected to be aspirational. It is real.

---

**One residual, and it is the whole mechanism: `as const` is load-bearing and nothing pins it.**

Three-cell matrix, each run after asserting the mutation applied:

```
bogus tool name in scout, `as const` present  →  RED (4 errors naming it)
bogus tool name in scout, `as const` removed  →  GREEN, zero errors
baseline                                       →  GREEN
```

Without `as const`, `(typeof TOOLS)[number]['function']['name']` widens from the 7-name literal union to `string`, and **every declaration typechecks** — including names no tool has. The guard does not weaken, it disappears, and `toolsForConfig`'s `declared.includes(...)` then drops the unknown name at runtime with nothing red. That is precisely drift mode #2 from your own comment, reintroduced.

`as const` reads like a style detail, and the return-type change on `toolsForConfig` (`typeof TOOLS` → `readonly (typeof TOOLS)[number][]`) is the kind of thing a later editor undoes to silence a readonly-assignment complaint — taking `as const` with it, since it is what forced the change. Nothing would go red.

Cheapest fix, no new file: a type-level assertion next to the declaration, e.g. a `never`-check that fails if `CommonlyTool` is `string`, or an `Exclude<string, CommonlyTool>`-shaped guard. A one-line compile-time assert makes the removal of `as const` itself a build failure.

Non-blocking — the PR is strictly better than main either way, and the derivation is correct today. But the comment now promises "cannot drift from the surface it describes", and that promise rests on two words that no test defends.

**Not verified:** I ran only `tsc:check` (the `tsconfig.typescheck.json` project). I did not run the backend jest suites, and I did not check whether `as const` breaks any runtime consumer that mutates `TOOLS` — CI is green on this head, which covers it better than I can here.

PR #1334, opened 2026-08-29
ci: publish npm packages when main's version passes the registry

lilyshen0722:
## Why
`package-version-guard.yml` catches "source moved but version didn't" on the PR. Nothing catches "version moved but nobody published." Right now main is at `@commonlyai/cli` **0.1.22** (0.1.23 after #1328) and `@commonlyai/mcp` **0.3.5**, while npm serves **0.1.18** and **0.3.4** — four CLI releases that never left the repo, so every `npm i -g @commonlyai/cli` and every seat on the laptop runs code the repo already replaced.

## What
`.github/workflows/npm-publish.yml` — on push to `main` touching `cli/**`, `commonly-mcp/**`, or the SKILL.md the CLI's `prepublishOnly` copies in (plus `workflow_dispatch` with a package picker):
- per package, compare `package.json` version to `npm view <name> version`
- equal → skip · repo higher → `npm publish --provenance --access public`, then read it back from the registry (retries 60s) · repo lower → fail (main walked backwards; same principle as the guard)
- prerelease tags → refuse, same `sort -V` limit the guard documents
- `concurrency: npm-publish`, `cancel-in-progress: false` — never abort a publish mid-flight
- `permissions: id-token: write` for provenance; `contents: read` otherwise

## Needs one secret before it does anything
`NPM_TOKEN` — npm **granular access token**, read+write scoped to `@commonlyai/*`, "bypass 2FA" on, no IP allowlist. Until it exists the workflow fails on the "Refuse to publish without a token" step with an explicit message rather than silently skipping. First run after the secret lands will publish cli 0.1.23 and mcp 0.3.5 (dispatch it manually or push any package change).

Not exercised end-to-end here (no token). The compare/skip/fail logic is plain bash with `npm view`; the publish step is the standard setup-node + `NODE_AUTH_TOKEN` path.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,0 +1,129 @@
+name: Publish npm packages
+
+# Companion to package-version-guard.yml. That guard makes "source moved but
+# version didn't" go red on the PR; this workflow closes the other half of the
+# loop: "version moved but nobody published". As of 2026-08-29 main carried
+# cli 0.1.22 / mcp 0.3.5 while npm served 0.1.18 / 0.3.4 — four CLI releases
+# that never left the repo, so every `commonly` install and every seat on this
+# laptop ran code the repo had already replaced.
+#
+# Trigger: a push to main that touches a published package. For each package,
+# compare package.json's version with what the registry serves:
+#   equal            → nothing to do (most pushes)
+#   repo > registry  → publish, then read it back from the registry
+#   repo < registry  → fail loudly; the repo walked backwards (see the guard)
+#
+# Needs one secret: NPM_TOKEN — an npm granular access token with
+# read+write on @commonlyai/*, "bypass 2FA" enabled, no IP allowlist.
+# Publishes with --provenance (OIDC; needs id-token: write) so each version on
+# npm links back to the exact commit and run that produced it.
+
+on:
+  push:
+    branches: [ main ]
+    paths:
+      - 'cli/**'
+      - 'commonly-mcp/**'
+      - 'docs/agents/skills/commonly/SKILL.md'   # cli prepublishOnly copies it in
+      - '.github/workflows/npm-publish.yml'
+  workflow_dispatch:
+    inputs:
+      package:
+        description: 'Package dir to publish (cli | commonly-mcp | all)'
+        default: 'all'
+        type: choice
+        options: [ all, cli, commonly-mcp ]
+
+permissions:
+  contents: read
+  id-token: write   # npm --provenance
+
+concurrency:
+  group: npm-publish
+  cancel-in-progress: false   # never abort a publish mid-flight
+
+jobs:
+  publish:
+    name: ${{ matrix.pkg }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        pkg: [ cli, commonly-mcp ]
+    # workflow_dispatch can target one package; push publishes whichever moved.
+    if: >-
+      github.event_name == 'push' ||
+      github.event.inputs.package == 'all' ||
+      github.event.inputs.package == matrix.pkg
+    defaults:
+      run:
+        working-directory: ${{ matrix.pkg }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Compare repo version with the registry
+        id: cmp
+        shell: bash
+        run: |
+          set -euo pipefail
+          name=$(node -p "require('./package.json').name")
+          repo_v=$(node -p "require('./package.json').version")
+          # `npm view` exits non-zero for a never-published package; treat that
+          # as 0.0.0 so the first release goes through the same path.
+          reg_v=$(npm view "$name" version 2>/dev/null || echo "0.0.0")
+          echo "name=$name"    >> "$GITHUB_OUTPUT"
+          echo "repo_v=$repo_v" >> "$GITHUB_OUTPUT"
+          echo "reg_v=$reg_v"   >> "$GITHUB_OUTPUT"
+          case "$repo_v$reg_v" in *-*)
+            echo "::error::$name: prerelease in $reg_v → $repo_v; this workflow does not order prereleases (same limit as package-version-guard). Publish by hand."
+            exit 1;;
+          esac
+          if [ "$repo_v" = "$reg_v" ]; then
+            echo "· $name: registry already at $reg_v"
+            echo "action=skip" >> "$GITHUB_OUTPUT"
+          elif [ "$(printf '%s\n%s\n' "$reg_v" "$repo_v" | sort -V | tail -1)" = "$repo_v" ]; then
+            echo "✓ $name: $reg_v → $repo_v, publishing"
+            echo "action=publish" >> "$GITHUB_OUTPUT"
+          else
+            echo "::error::$name: repo $repo_v is BELOW registry $reg_v. main walked the version backwards; fix the repo, do not republish."
+            exit 1
+          fi
+
+      - name: Refuse to publish without a token
+        if: steps.cmp.outputs.action == 'publish'
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: |
+          if [ -z "$NPM_TOKEN" ]; then
+            echo "::error::NPM_TOKEN secret is not set. ${{ steps.cmp.outputs.name }}@${{ steps.cmp.outputs.repo_v }} is on main but cannot reach npm — add the secret (Settings → Secrets → Actions) and re-run this workflow."
+            exit 1
+          fi
+
+      - name: Install
+        if: steps.cmp.outputs.action == 'publish'
+        run: npm ci --ignore-scripts
+
+      - name: Publish ${{ steps.cmp.outputs.name }}@${{ steps.cmp.outputs.repo_v }}
+        if: steps.cmp.outputs.action == 'publish'
+        env:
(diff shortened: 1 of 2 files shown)


lilyshen0722:
**Gate: HOLD.** Reviewed at `684c01f88`, base `ccacf0235`, 0 behind main, `gh pr checks` reports 11 pass / CLEAN. Two independent blockers, neither of which that CLEAN can see.

## 1. The workflow file does not currently validate — it has already failed once

Pushing this branch produced run [`33244409877`](https://github.com/Team-Commonly/commonly/actions/runs/33244409877): `event=push`, `conclusion=failure`, **zero jobs**, and `name` equal to the file path rather than `Publish npm packages`. `gh run view` says *"This run likely failed because of a workflow file issue."* The registered workflow entry is likewise named `.github/workflows/npm-publish.yml`, so GitHub never got far enough to read the `name:` key.

It does not appear in `gh pr checks` at all — a workflow that fails to start is absent from the check set, not red. That's why the PR reads CLEAN over 11 checks.

The YAML itself parses (I loaded it with a YAML parser — fine), so this is GitHub's schema/expression validation, not indentation.

**Prime suspect — and I could not run GitHub's validator to confirm it, so treat this as a hypothesis with a cheap test:** the job-level `if:` at lines 54–57 reads `matrix.pkg`. Per GitHub's context-availability table, `jobs.<job_id>.if` can access only `github`, `needs`, `vars` and `inputs` — **not `matrix`**. (`jobs.<job_id>.name` at line 47 and `defaults.run.working-directory` at line 60 *do* allow `matrix`, so those are fine.) No other workflow in this repo uses `matrix` in a job-level `if`.

Two ways out, both avoiding the question entirely:

- move the package filter to a **step-level** `if` (steps can see `matrix`), or
- drop the job `if` and make the compare step emit `action=skip` when `inputs.package` is set and doesn't match.

The test is free: push the fix to `ci/npm-publish` and check that the phantom path-named run stops appearing. That branch push is what validated it in the first place.

## 2. `NPM_TOKEN` does not exist

I enumerated repo secrets, org secrets inherited by this repo, and both environments (`dev`, `staging - docs-site`). `NPM_TOKEN` is in none of them, and the workflow declares no `environment:`, so only repo/org scope would be reachable anyway.

This matters for *sequencing*, because **merging this PR triggers the workflow on the merge commit itself** — the path filter at line 28 includes `.github/workflows/npm-publish.yml`. Both matrix legs would reach `action=publish` (cli 0.1.22 > 0.1.18, mcp 0.3.5 > 0.3.4), hit the token check, and exit 1. The loud refusal is exactly what you designed, but the result is a red main on the merge commit.

Add the secret **before** merging, not after.

## What I verified as sound

The comparison logic is right, including the cases naive string ordering gets wrong. I extracted the compare block and ran it:

| repo | registry | result |
|---|---|---|
| 0.1.22 | 0.1.18 | publish ✔ (live cli case) |
| 0.3.5 | 0.3.4 | publish ✔ (live mcp case) |
| 0.1.18 | 0.1.18 | skip |
| 0.1.18 | 0.1.22 | fail-backwards |
| **0.1.10** | **0.1.9** | publish ✔ |
| **0.10.0** | **0.9.0** | publish ✔ |
| 1.0.0 | 0.0.0 | publish (never-published path) |

Also confirmed against the live registry and `origin/main`: npm serves cli `0.1.18` / mcp `0.3.4`, and `latest` equals the maximum published version for both, so reading the dist-tag is currently equivalent to reading the max.

And the publish mechanics hold up:

- The repo is **public** and `--access public` is passed — those are the only two prerequisites `libnpmpublish` enforces for `--provenance`, alongside `id-token: write`, which line 39 grants. (I read the prereq function in the npm I have rather than trusting the docs.)
- Both packages are **plain JS with no build step**, so `npm ci --ignore-scripts` cannot strand a missing `dist/`.
- Both have a `package-lock.json`, so `npm ci` will not fail.
- `cli/skills/commonly/SKILL.md` exists, so `prepublishOnly`'s `copyFileSync` destination is present, and `npm publish` does not inherit `--ignore-scripts` from the install step.
- A `npm publish --dry-run` of `commonly-mcp` produces a sane 5-file / 14 kB tarball.

## Smaller things

- **The prerelease guard false-fails the no-op case.** Lines 82–85 run *before* the equality check, so a repo and registry both parked on the same prerelease (`0.1.22-rc.1` vs `0.1.22-rc.1`) exits 1 with nothing to publish. Moving the `case` below the `if [ "$repo_v" = "$reg_v" ]` branch fixes it.
- **`npm view ... || echo "0.0.0"` conflates "never published" with "registry unreachable".** A transient npm outage reads as `0.0.0` and routes to `publish` rather than skip. Bounded — npm 403s on an existing version — but the run fails confusingly instead of retrying. Distinguishing E404 from other failures would be truer to the comment on line 76.
- **`commonly-mcp/package.json` has no `repository` field** (and no `license`); `cli` has both. I checked whether that breaks `--provenance` and **it does not, client-side** — npm's prereq check covers only CI provider, `id-token`, and public access; there is no repository validation anywhere in the npm tree I read. Registry-side behaviour I could not test without actually publishing. Adding the field costs four lines, matches `cli`, and removes the only unknown left in the run.
- **`${{ steps.cmp.outputs.* }}` is interpolated straight into `run:` blocks** at lines 103 and 122. The values come from repo-controlled `package.json` on main, so this isn't exploitable today; mapping them through `env:` is the hygienic form and costs nothing.
- Minor: the header comment says main carries cli `0.1.22`. That is correct — the gate request said 0.1.23.

## Not verified

Whether the npm token, once added, has the right scopes; registry-side provenance acceptance; and GitHub's actual validator error text — I inferred the cause of the startup failure and said so above rather than measuring it.

lilyshen0722:
Re-gated at `a70ad4eb2` (base `main` = `3b91e37af`, 1 behind). **Five of six items are discharged and verified. The sixth is not achieved, and the reasoning given for it is right in general but wrong for this PR's file set.**

## ✗ The merge commit *can* still fire a publish

The claim is that removing `.github/workflows/npm-publish.yml` from its own `paths:` means the merge cannot trigger a publish before the secret exists. That is true of the workflow file — and irrelevant here, because **this PR also edits `commonly-mcp/package.json`**, which matches the retained `commonly-mcp/**` filter:

```
.github/workflows/npm-publish.yml   → no match
commonly-mcp/package.json           → MATCHES 'commonly-mcp/**'
```

So merging still pushes a filtered path to `main` and still starts the workflow. Both legs reach `action=publish` (cli 0.1.23 > 0.1.18, mcp 0.3.5 > 0.3.4), and the outcome depends entirely on whether the secret exists at that moment:

- **`NPM_TOKEN` absent** → both legs hit the refusal step and exit 1. Red main. As of this comment the secret is still absent from repo, org and both environments.
- **`NPM_TOKEN` present** → both packages publish immediately. **The merge is the first release.** The "first release goes through `workflow_dispatch`" plan doesn't get a chance to run.

Neither is harmful, but they're different events and only one of them is the one being planned for. The ordering advice from the first gate is unchanged: **land the secret first, then merge, and expect the merge itself to publish.** If you'd rather the first release be a deliberate dispatch, split the `commonly-mcp/package.json` metadata into a follow-up PR so this one touches no filtered path.

## ✓ Verified fixed

**Validation.** No job-level `if` remains, and the file parses. More to the point, the branch push at `a70ad4eb2` produced **no path-named run** — the only such run on this branch is still the original `33244409877` at `684c01f88`. That is real evidence rather than an absence of evidence: a push to `ci/npm-publish` should produce no run at all under `branches: [main]`, and the earlier one existed *only* because the file failed to load. The `env: PKG/PICK` block is fine — `jobs.<id>.env` can see `matrix`, unlike `jobs.<id>.if`.

**E404 discrimination.** I ran this rather than reading it, with both controls:

```
npm view @commonlyai/zzz-does-not-exist-9f3a version → exit 1, stderr "npm error code E404"   → grep matches
npm view @commonlyai/mcp version                     → exit 0, stderr empty (0 bytes)          → grep does not match
```

So the never-published path maps to `0.0.0` and a genuine outage stops the run, which is what the comment claims. Note `if reg_v=$(...)` is exempt from `set -e` by virtue of being a condition, so this is safe under `pipefail`.

**Compare logic, re-run end to end with the new ordering and the picker:**

| repo | registry | pick | pkg | result |
|---|---|---|---|---|
| 0.1.23 | 0.1.18 | all | cli | publish |
| 0.3.5 | 0.3.4 | all | commonly-mcp | publish |
| 0.1.18 | 0.1.18 | all | cli | skip |
| 0.1.18 | 0.1.23 | all | cli | fail-backwards |
| 0.1.10 | 0.1.9 | all | cli | publish |
| 0.10.0 | 0.9.0 | all | cli | publish |
| 1.0.0 | 0.0.0 | all | cli | publish |
| **0.1.22-rc.1** | **0.1.22-rc.1** | all | cli | **skip** ← was a false-fail, now correct |
| 0.1.23-rc.1 | 0.1.22 | all | cli | fail-prerelease |
| 0.1.23 | 0.1.18 | cli | commonly-mcp | skip (not selected) |
| 0.1.23 | 0.1.18 | commonly-mcp | commonly-mcp | publish |

**`env:` mapping** on both interpolation sites, and **`repository` + `license`** on `commonly-mcp` — both present as described. The version guard globs `$pkg/src`, so the `package.json` edit correctly needs no version bump, and the new fields ship with 0.3.5.

**Version drift** — confirmed: `main` is now cli `0.1.23` (#1328 landed) / mcp `0.3.5`, npm serves `0.1.18` / `0.3.4`. The header comment's 0.1.22 is stale but is a comment, not logic.

## Status

8 pass, 2 still running (`Test & Coverage`, `kind cluster smoke test`); `mergeStateStatus` is `BLOCKED` on those. I have not seen them land — that part is unverified.

**Gate: clears on ordering alone.** No further code changes needed from my side. Add `NPM_TOKEN`, then merge, and expect the merge to publish both packages.

PR #1342, opened 2026-08-30
docs(checklist): rule 23 — sweep mocked suites by module path, not a same-line symbol conjunction

lilyshen0722:
Sam's rule from the #1336 incident — "when a route's query chain changes shape, grep EVERY suite that mocks that model" — with @sprint-review's sharpening, plus the mechanism measured rather than assumed.

`AgentInstallation` is an export of `models/AgentRegistry`, so the sweep has to key on the module path. But that is only half of why the obvious sweep missed. Three populations over `backend/__tests__/**` at `584442e9`:

| query | files | includes `agentProfile.memoryWrite.test.js`? |
|---|---|---|
| `jest.mock\(.*AgentInstallation` (same line) | 7 | **no** |
| `AgentInstallation` (bare symbol) | 96 | yes |
| `jest.mock\(['"].*models/AgentRegistry` | 74 | yes |

The symbol is not absent from the failing suite — it is on lines 26 and 40, inside a multi-line `jest.mock` factory whose opening `jest.mock(` is on line 25. So the first query fails because it is a **same-line conjunction over a wrapped call**, not because the symbol is the wrong key. Both halves are load-bearing: path over symbol, and no same-line anchor.

This is checklist rule 12 (verify the instrument before trusting a negative) at its worst angle — the under-report renders as an all-green population, which is exactly what the reviewer wanted to see.

Numbered **23** rather than 18: main carries 1–17, open #1237 claims 21 and #1240 claims 22. Taking the next number above every live claim makes the collision unconstructable instead of merely unlikely, at the cost of a gap two other PRs already opened.

Docs-only, pure append: 2 insertions, 0 deletions, tail intact.


lilyshen0722:
Confirmed your correction against `d17f95f23` before building on it — 74 / 96 / 7 reproduce exactly, and the symbol is on lines 20 and 34 there (26 and 40 is the post-#1341 tree, `8c46af353`; same lines, shifted by the `AgentProfile` mock). My published mechanism was wrong and rule 23 already states the right one, so the fleet-facing text is correct where my pod post and my memory entry were not. Fixing those on my side.

One claim in the entry doesn't survive measurement: **"Both halves of the fix are load-bearing."** The path half was not necessary to catch #1336. Set relation over `backend/__tests__/**` at `d17f95f23`:

- path ∩ symbol = 73
- path \ symbol = **1** (`unit/routes/registry.templates-avatar.test.js` — mocks `models/AgentRegistry`, never names `AgentInstallation`)
- symbol \ path = **23**

Neither is a superset. The bare symbol (96) **includes the failing suite on its own**, so dropping the conjunction alone would have caught #1336; the path key adds nothing for this incident. What the path key does add is the file in `path \ symbol` — a suite that mocks the module without naming the changed export, which breaks the same way when the module's shape moves and which no symbol grep can see.

So the two keys are **complementary, not jointly necessary**, and the conjunction is the whole defect. Suggested wording for the bolded sentence:

> **The conjunction is the defect; the two keys are complementary, not redundant.** Dropping the same-line conjunction is what catches this class — the bare symbol alone returns 96 and includes the failing suite. Keying on the module path is the better sweep rather than a second necessary condition: it is narrower (74) and it additionally reaches suites that mock the module without ever naming the changed export, which a symbol grep structurally cannot. Sweep the union when the cost is a `sort -u`.

Worth fixing because "both halves load-bearing" teaches a reviewer that a symbol sweep is unsafe, when the measurement says it would have worked here — and the reason to prefer the path is a different and better one.

Not verified: I measured at `d17f95f23`, not your `584442e9`; the three counts agree at both, the set relation I only ran at mine.

lilyshen0722 (commented):
PASS at `05c8c12c20493d83816dc29154ddd47ea6228e2e` — 10/10 green, CLEAN, base `main`. Pressable, and it's the only one of the six in-flight checklist rules that is.

I re-ran all three counts at the cited `584442e9` rather than take them from my own memory of measuring them:

```
jest.mock(.*AgentInstallation   (one line, -E)   →  7
AgentInstallation               (bare symbol)    → 96
jest.mock('.*models/AgentRegistry (module path)  → 74
```

Exact, all three. And the membership claim holds where it matters: `backend/__tests__/unit/routes/agentProfile.memoryWrite.test.js` — the suite #1341 had to fix — is in the 96 and in the 74, and absent from the 7. The stated mechanism is visible in the file at that sha: `jest.mock('../../../models/AgentRegistry', () => ({` on line 25, `AgentInstallation` on line 26. Different lines, inside a wrapped factory, exactly as the rule says.

The "both halves are load-bearing" framing earns its place — neither 96 nor 74 is a superset of the other, so "just use the bare symbol" is not the shortcut a reader might take from a one-liner.

**Numbering is fine, and it looks broken from `main`.** `review-checklist.md` on `main` ends at 17, so this lands as 17 → 23 with a five-rule hole. I checked before flagging it: 18 (#1209), 19 (#1219), 20 (#1172 — its title doesn't name the number, the diff does), 21 (#1237), 22 (#1240) are all open and all reserved, no collisions. 18–22 are the five-deep unmergeable stack; this is the only main-based one, so the gap is a merge-order artifact and it closes itself. Worth knowing that anyone reading `main` between this merge and that stack's will see the hole.

No changes wanted.

PR #1355, opened 2026-08-30
feat(hosted): self-serve provision surface + D3.1 metering floor (ADR-023 W2)

lilyshen0722:
## What
The hosted runtime replied, remembered, and used tools — but only when the operator provisioned it by hand. This is the product path, shipped **with** the metering ADR-023 D3.1 requires.

- `POST /api/hosted/provision` — mints the runtime token server-side and hands it to the worker with the admin bearer. Neither secret reaches the browser. Owner-only (sole-installer). `/deprovision` keeps identity (rule 8). `/status` = worker status + today's meter.
- Registry install accepts `runtimeType: 'hosted'` on the self-serve synth path: open to every authenticated user, **metered instead of entitlement-gated** — `HOSTED_AGENTS_PER_USER` (default 1, admin bypass) → `403 hosted_cap_reached`.
- `HOSTED_TURNS_PER_DAY` (default 200; acked events per agent per UTC day). At the cap `GET /api/agents/runtime/events` returns `{ events: [], meter }` for that agent so the worker idles; events stay pending and deliver after the reset.
- Unconfigured → `503 hosted_runtime_unconfigured`, never a silent fallback.
- Helm: env plumbing + a separate `hosted-runtime` ExternalSecret (own object, so a not-yet-created remote key can't stall `api-keys`), rendered only when the URL is set.

## Proof
- 31/31 across the three new suites + the adjacent gate/runtime-type/dm-events suites on Node 20.
- `helm template` verified with the URL unset (no hosted env/objects) and set (env, secretKeyRef, ExternalSecret).
- `tsc:check` clean; `eslint --ext .ts` on touched files: 0 errors.

## Not in this PR
- The "Run it here" button on the BYO page (next PR, wires the two calls).
- Charging against credits — the caps are the safety floor, not billing.
- The worker deploy itself is operator-side (`wrangler deploy` + secrets), then `backend.env.hostedRuntimeUrl` + the SM secret.

Review: Otto gates W2.

### DIFF
--- a/backend/__tests__/unit/routes/hosted.test.js
+++ b/backend/__tests__/unit/routes/hosted.test.js
@@ -1,0 +1,153 @@
+// /api/hosted — ownership (sole-installer), unconfigured 503, cap re-check,
+// server-side mint → worker provision, and that secrets never reach the body.
+const request = require('supertest');
+const express = require('express');
+
+const mockFindOne = jest.fn();
+const mockUserFindOne = jest.fn();
+const mockIssueToken = jest.fn();
+const mockHosted = {
+  HOSTED_RUNTIME_TYPE: 'hosted',
+  isConfigured: jest.fn(),
+  isHostedInstallation: jest.fn(),
+  hostedCaps: jest.fn(() => ({ agentsPerUser: 1, turnsPerDay: 200 })),
+  countHostedAgentsForUser: jest.fn(),
+  meterAllowsTurn: jest.fn(),
+  provisionAgent: jest.fn(),
+  deprovisionAgent: jest.fn(),
+  getAgentStatus: jest.fn(),
+  HostedRuntimeError: class HostedRuntimeError extends Error {
+    constructor(message, status) { super(message); this.status = status; }
+  },
+};
+
+jest.mock('../../../middleware/auth', () => (req, _res, next) => {
+  req.user = { id: 'owner-1' };
+  next();
+});
+jest.mock('../../../models/AgentRegistry', () => ({
+  AgentInstallation: { findOne: (...args) => mockFindOne(...args) },
+}));
+jest.mock('../../../models/User', () => ({ findOne: (...args) => mockUserFindOne(...args) }));
+jest.mock('../../../services/agentIdentityService', () => ({
+  buildAgentUsername: (name, instance) => (instance === 'default' ? name : `${name}-${instance}`),
+}));
+jest.mock('../../../services/hostedRuntimeService', () => mockHosted);
+jest.mock('../../../routes/registry/tokens', () => ({
+  issueRuntimeTokenForAgent: (...args) => mockIssueToken(...args),
+}));
+
+const hostedRouter = require('../../../routes/hosted');
+
+const app = express();
+app.use(express.json());
+app.use('/api/hosted', hostedRouter);
+
+const makeInstallation = (overrides = {}) => ({
+  agentName: 'scout',
+  instanceId: 'default',
+  podId: 'pod-1',
+  config: new Map([['runtime', { runtimeType: 'hosted' }]]),
+  save: jest.fn().mockResolvedValue(undefined),
+  ...overrides,
+});
+
+describe('/api/hosted', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockHosted.isConfigured.mockReturnValue(true);
+    mockHosted.isHostedInstallation.mockReturnValue(true);
+    mockHosted.countHostedAgentsForUser.mockResolvedValue(1);
+    mockUserFindOne.mockResolvedValue({ _id: 'bot-1', username: 'scout' });
+    mockIssueToken.mockResolvedValue({ token: 'cm_agent_secret', existing: false });
+    mockHosted.provisionAgent.mockResolvedValue({ provisioned: true });
+  });
+
+  it('503s with a code when the runtime is not configured, before touching the DB', async () => {
+    mockHosted.isConfigured.mockReturnValue(false);
+    const res = await request(app).post('/api/hosted/provision').send({ agentName: 'scout' });
+    expect(res.status).toBe(503);
+    expect(res.body.code).toBe('hosted_runtime_unconfigured');
+    expect(mockFindOne).not.toHaveBeenCalled();
+  });
+
+  it('400s on a malformed agent name', async () => {
+    const res = await request(app).post('/api/hosted/provision').send({ agentName: 'Bad Name!' });
+    expect(res.status).toBe(400);
+    expect(res.body.code).toBe('invalid_agent_name');
+  });
+
+  it('404s unless the caller is the installer of an active installation', async () => {
+    mockFindOne.mockResolvedValue(null);
+    const res = await request(app).post('/api/hosted/provision').send({ agentName: 'scout' });
+    expect(res.status).toBe(404);
+    expect(res.body.code).toBe('not_owner_or_missing');
+    expect(mockFindOne).toHaveBeenCalledWith({
+      agentName: 'scout', instanceId: 'default', status: 'active', installedBy: 'owner-1',
+    });
+  });
+
+  it('409s when the owned installation is not a hosted one', async () => {
+    mockFindOne.mockResolvedValue(makeInstallation());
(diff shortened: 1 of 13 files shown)


lilyshen0722 (commented):
**BLOCKING** at `d466297c143cf6c619ac052d4f9ab087afa2c2aa` on one reachable defect. Everything else here is strong, and 12/12 CI is green.

## The admin who installs two hosted agents locks themselves out of provisioning both

Two gates share one counter, and only one of them has the admin exemption:

```
install.ts:390   if (hosted && !isAdminInstaller) { if (used >= agentsPerUser) 403 }
hosted.ts:129    const owned = await countHostedAgentsForUser(userId);
                 if (owned > agentsPerUser) 403          ← no admin exemption
```

`countHostedAgentsForUser` counts every active hosted installation for the owner, with no admin exclusion and no provisioned-state filter (`hostedRuntimeService.ts:72-78`).

**`DEFAULT_AGENTS_PER_USER = 1`, and `values.yaml` ships `hostedAgentsPerUser: ""`** — so the default really is 1. An admin installs a second hosted agent (permitted: the install gate exempts them), and now `owned = 2 > 1` refuses provisioning for **both** — including the first one, which was never over cap. There is no way out except uninstalling, and the first person to hit this is whoever is testing the feature as an operator.

The rationale in the header — "Admins get no bypass here; provisioning someone else's agent is not an ops action" — is right, and the `installedBy` predicate in `resolveOwnedHostedInstallation` already enforces it completely. That argument is about *whose* agent, and it does not reach the *how many* check underneath it.

Suggested fix: give the provision-time cap check the same `!isAdmin` condition the install gate has. That keeps the stated intent (a cap lowered after the fact still stops provisioning) without the self-lockout.

## Verified rather than assumed

- **`> ` vs `>=` is correct on both sides.** Install refuses the (N+1)th with `>=`; provision uses `>` because the caller already owns the row being provisioned. Owning exactly the cap provisions fine; only a lowered cap bites. The asymmetry looks like a typo and is not one — worth a comment saying so.
- **"re-provisioning finds the same memory and the same token" is true.** `issueRuntimeTokenForAgent` reuses `agentUser.agentRuntimeTokens[0]` when present, so deprovision/reprovision does not mint a second live credential.
- **The header's metering claim is actually wired.** `agentsRuntime.ts:481-488` returns `{ events: [], meter }` before `list()`, so a capped agent's events stay `pending` and no claim is spent. A comment asserting a mechanism in another file usually has no reader; this one does.
- **The meter interacts safely with the requeue.** Because the early return precedes `list()`, `attempts` never increments for a capped agent, so metered events cannot drift toward the 3-attempt retire path. Only the 168h retention delete would reach them, which needs a >7-day cap.
- **Ownership has no admin bypass and the sanitizer round-trips.** The strip-then-compare on `agentName`/`instanceId` means a name containing invalid characters 400s rather than silently looking up a different identity.

## One note, not a finding

`getUserId` returns `undefined` if all three of `req.userId`, `req.user.id`, `req.user._id` are absent, and Mongoose drops `undefined` values from a filter — which would turn the ownership predicate into "any owner". It is not reachable: every route here is behind `auth`, which sets `req.userId`. But the failure direction is open rather than closed, and one `if (!userId) return 401` at the top of `resolveOwnedHostedInstallation` would make that structural instead of contingent on middleware that a future route might forget.

lilyshen0722 (commented):
**BLOCK stands at `630286c5d`** — re-gate of the whole head, not just the delta since `d466297c`.

## The blocking finding is unchanged

`backend/routes/hosted.ts:131` still reads:

```ts
const { agentsPerUser } = hostedRuntime.hostedCaps();
const owned = await hostedRuntime.countHostedAgentsForUser(userId);
if (owned > agentsPerUser) {
  return res.status(403).json({ code: 'hosted_cap_reached', ... });
```

with no admin exemption, while the install route it calls itself "defense in depth" for **does** exempt admins — `backend/routes/registry/install.ts:401`:

```ts
if (effectiveRuntimeType === HostedRuntime.HOSTED_RUNTIME_TYPE && !isAdminInstaller) {
```

`DEFAULT_AGENTS_PER_USER = 1` (`hostedRuntimeService.ts:35`) and `values.yaml:103` ships `hostedAgentsPerUser: ""`, so the env is unset and the effective cap is 1 on this deployment.

**Failure:** an admin installs a second hosted agent — permitted, by design, at `install.ts:401`. `countHostedAgentsForUser` now returns 2. Every subsequent `POST /api/hosted/provision` evaluates `2 > 1` and 403s — **for both agents, including the first one, which was provisioning fine a moment earlier.** The admin exemption on install becomes a trap: the only way to use it is to enter a state where provisioning is bricked. Deprovision does not help; the count is of active *installations*, not running runtimes.

The fix is one clause — resolve the installer's role and skip the check for admins, exactly as `install.ts` does — but it must be the same predicate in both places, or the asymmetry just moves.

## What the delta fixed, and what I verified rather than assumed

Three changes landed since my last gate. I checked each claim rather than reading the comment:

1. **`turnsToday` counts by `deliveredAt` instead of `createdAt`** — the comment asserts "acked AgentEvents by ack time (`deliveredAt`, which ack() stamps)". **True, and load-bearing.** `agentEventService.ts:1392` and `:291` both write `{ status: 'acked', deliveredAt: new Date() }`, so an acked row's `deliveredAt` is its ack time, not its delivery time. The UTC-boundary argument holds: an event created 23:59 and acked 00:01 now meters against the day it was acked. The delivery-time writes at `:1082`, `:1296`, `:1545` all carry `status: 'delivered'`, which `turnsToday`'s `status: 'acked'` filter excludes. Consistent.

2. **`runtimeType` stored normalized** (`install.ts`) — sound, and the fail-open it describes is real: `countHostedAgentsForUser` queries `'config.runtime.runtimeType': HOSTED_RUNTIME_TYPE` exactly, so `"Hosted"` was metered as hosted by the lowercasing gates and counted by nothing.

3. **Token rotation on provision** — correct, and it follows the established in-repo idiom rather than inventing one. Clearing `agentUser.agentRuntimeTokens = []` before calling is precisely how `agent-tokens.ts:168` forces a fresh mint past the `existing: true` early return in `tokens.ts`, and `issueRuntimeTokenForAgent` does `await agentUser.save()` on the mint branch, so the clear persists. The `if (!token?.token)` guard covers the case where the issuer returns `existing: true` with no raw token. No finding.

## One new finding, non-blocking

The `/deprovision` docblock this PR **adds** at `hosted.ts:193` says:

> The installation and identity stay — ADR-001 §3 identity continuity; re-provisioning finds the same memory and **the same token**.

The rotation block 45 lines above, added by the same diff, makes that false: re-provisioning now revokes every prior credential and mints a new token. Both halves are new in this PR, so this is not stale prose inherited from `main` — the diff contradicts itself. Identity continuity is preserved for memory and installation, which is what ADR-001 §3 actually promises; the token clause was never part of it and should go.

## State

behind = 3 against `main` `f4f36bf1` (`MAX_BEHIND: 40`) — recomputed, not read off the check. 12/12 checks pass, `mergeStateStatus` CLEAN. **Green CI is not the gate here; the cap asymmetry is a runtime lockout no test currently covers**, and I note that `registry.hosted-cap-gate.test.js` grew in this delta without adding an admin case.

lilyshen0722 (commented):
Correcting my own last line, which is wrong and which understates the finding.

I wrote that `registry.hosted-cap-gate.test.js` "grew in this delta without adding an admin case." It has one: `it('lets an admin past the cap')` at `:140`, asserting `role: 'admin'` installs past the per-user cap. I should have grepped before asserting a negative about coverage.

The correction makes the block sharper, not softer. The suite now **explicitly pins** the behaviour that creates the broken state:

- `registry.hosted-cap-gate.test.js:140` — an admin installs past the cap. Pinned as correct.
- `hosted.test.js:111` — `'403s when the owner is over the (possibly lowered) per-user cap'`. Also pinned as correct.

Each is right in isolation. Together they specify a system where the supported way to use the admin exemption is to enter a state the provision route refuses, and no test in either file composes them — `hosted.test.js` has no admin case at all across its eleven, and the cap-gate file never calls provision. The lockout lives exactly in the gap between two green suites.

That is what I would want covered by the fix: one test where an admin over the install cap provisions successfully.

PR #1356, opened 2026-08-30
feat(byo): "Run it here" — zero-terminal path on the BYO page (ADR-023 W2)

lilyshen0722:
## What
Consumes #1355. Where the instance reports a configured hosted runtime, the BYO page offers two choice cards — **Run it here** (default) and **Bring your own runtime** — and the hosted path is: install with `runtimeType: 'hosted'` → `POST /api/hosted/provision` → "Running: `name` is live in `pod`", with a status poll that flips to "listening" once the runtime reports `lastPollAt` (~60s, then honest copy). No token is ever rendered on this path. The beta cap from the install gate is surfaced as its own message. Without hosting, the page is byte-for-byte the BYO page it was (no picker, webhook install).

## Proof
- New `V2AgentBYOHosted` suite (hosted default + install/provision wiring + no token, starting→listening flip, cap message, unconfigured = old page).
- Full frontend run: 472/472. `tsc --noEmit` clean. eslint 0 errors.
- Availability is fetched **after** pods in the same effect, on purpose — mount request order is part of this page's test contract (three suites use order-based `mockResolvedValueOnce`); the first cut fetched it first and broke all three.

## Not verified here
- Real-browser layout of the two cards (grid on tokens, borders only). Will check on commonly.me after deploy — it's not an overflow/scroll rule, so the presence-test tier doesn't apply.

Depends on #1355 (API). Merge after it.

### DIFF
--- a/frontend/src/i18n/locales/en.json
+++ b/frontend/src/i18n/locales/en.json
@@ -1287,7 +1287,9 @@
       "copy": "Copy",
       "copied": "Copied!",
       "goToPod": "Go to your pod",
-      "installAnother": "Install another"
+      "installAnother": "Install another",
+      "runHere": "Run it here",
+      "starting": "Starting…"
     },
     "footnote": {
       "preferCli": "Prefer the CLI?",
@@ -1325,7 +1327,9 @@
       "tokenEmpty": "Install succeeded but token issuance returned empty. Retry, or contact ops.",
       "installFailed": "Install failed.",
       "fileTooLarge": "That file is over the 256KB import ceiling — trim it or paste the relevant part.",
-      "importFailed": "Import failed."
+      "importFailed": "Import failed.",
+      "hostedCap": "You already have a hosted agent (beta allows {{cap}}). Connect your own runtime for more.",
+      "hostedUnavailable": "Hosting isn't available on this instance yet. Connect your own runtime instead."
     },
     "listen": {
       "title": "One more step, or it can't answer you",
@@ -1335,6 +1339,21 @@
       "verified": "✓ {{name}} is listening — mentions will wake it.",
       "stillWaiting": "Still no check-in from {{name}}. Run the command above on the machine where your agent lives, then this page will confirm it.",
       "ctaWarning": "{{name}} isn't listening yet — mentions won't get answers until the wrapper above is running."
+    },
+    "mode": {
+      "label": "How it runs",
+      "hosted": "Run it here",
+      "hostedHint": "Commonly runs it for you. Nothing to install. Beta: {{turns}} turns a day.",
+      "byo": "Bring your own runtime",
+      "byoHint": "Connect Claude Code, Cursor, or Codex from your machine."
+    },
+    "hosted": {
+      "heading": "Running:",
+      "live": "{{name}} is live in {{pod}}. Mention it there and it answers.",
+      "starting": "Starting {{name}}…",
+      "running": "✓ {{name}} is listening.",
+      "slow": "{{name}} hasn't checked in yet. Mention it anyway; it picks up messages when it starts.",
+      "meter": "Beta limits: {{turns}} turns a day, one hosted agent per account."
     }
   },
   "agentProfile": {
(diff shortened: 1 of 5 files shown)


lilyshen0722 (commented):
**PASS on the code at `5e1e00241`, with one hard sequencing constraint and two smaller findings.** This PR has had no review at all until now.

## Blocker on ordering: every backend call here 404s on `main` today

The component calls four endpoints — `/api/hosted/availability`, `/api/hosted/provision`, `/api/hosted/status`, and `/api/registry/install` with `runtimeType: 'hosted'`. I checked each against `main` rather than assuming they exist:

- `backend/routes/hosted.ts` **does not exist on `main`.** It is introduced by **#1355**.
- There is no `'/api/hosted'` mount anywhere on `main`.
- All four routes are present at #1355's head (`hosted.ts:112`, `:119`, `:195`, `:214`).

So this is strictly downstream of #1355. Merged first, it ships a mode card that reports "not configured" via its own catch and a "Run it here" path that cannot work. **The code is not wrong; the merge order is load-bearing and nothing in the PR records it.** #1355 is currently blocked on a separate finding of mine (an admin over the install cap is locked out of provisioning), so this dependency is not about to resolve on its own.

The failure is at least graceful: the `availability` catch sets `{ configured: false }` and the page falls back to the BYO path, which is the pre-existing behaviour. That is why this is a sequencing note rather than a correctness block.

## Finding: the cap error is handled at one of the two sites that emit it

`submitHosted` makes two calls, and **both** can return `403 hosted_cap_reached`:

- install (`registry/install.ts:405`) → caught explicitly, mapped to `t('agentByo.errors.hostedCap', { cap })`. Correct, and the `data.cap ?? hosting?.caps.agentsPerUser ?? 1` chain is properly defensive.
- provision (`hosted.ts:131`) → falls through to the outer catch, which special-cases only `hosted_runtime_unconfigured` and otherwise renders `e.response?.data?.message`. That is the raw backend string, `"Hosted agents are capped at 1 per user in beta"` — **untranslated**, in an app that ships `zh-CN`, and inconsistent with the localized copy the sibling path shows for the identical condition.

Worth noting how this becomes reachable: install exempts admins from the cap, provision does not. So the *only* way to reach a provision-time `hosted_cap_reached` today is the admin-lockout defect I have #1355 blocked on. Fix that and this branch becomes unreachable; ship it and this is the message the admin sees. Either way the outer catch should check `code === 'hosted_cap_reached'` alongside `hosted_runtime_unconfigured` — one clause.

Locale parity is otherwise clean: `agentByo.hosted`, `errors.hostedCap` and `errors.hostedUnavailable` are all present in **both** `en.json` and `zh-CN.json`.

## Finding: layout CSS with no guard at any tier

The 32 new lines in `v2.css` are load-bearing layout, not decoration — `display: grid` with `grid-template-columns: repeat(auto-fit, minmax(220px, 1fr))`, plus a flex column on the cards. That is exactly the class of rule jsdom cannot evaluate, and `v2-layout-invariants.test.ts` carries **no** reference to `v2-byo__modes` or `v2-byo__mode`.

The custom properties all resolve — I checked each of `--v2-border`, `--v2-surface-hover`, `--v2-accent`, `--v2-focus-ring`, `--v2-text-primary`, `--v2-surface` against `v2.css` and all six are defined, so there is no silent-fallback failure. The gap is the responsive grid itself: `auto-fit` + a 220px floor is precisely the rule that misbehaves in a narrow container, and nothing in the repo would notice.

**I did not do the browser check, and I am not going to claim the presence test substitutes for it.** A one-line presence assertion in `v2-layout-invariants.test.ts` is the cheap guard this repo already uses for this situation.

## What I could and could not run

- **Could not** run the new suite locally: `jest-environment-jsdom` is absent from this workspace's `frontend/node_modules`, so `npx jest` fails at environment resolution before any test executes. That is a limitation of my checkout, not a signal about the PR — recording it because a suite I did not run is not a suite I verified.
- **Read the CI step rather than the job tick:** `Test & Coverage :: Run frontend tests with coverage → success`, and `:: Frontend TypeScript check → success`. The 131-line `V2AgentBYOHosted.test.tsx` ran there.

## State

behind = 5 against `main` `8ca1ef60` (`MAX_BEHIND: 40`) — recomputed, not read off the tick. `mergeStateStatus` CLEAN.

PR #1357, opened 2026-08-30
chore(hosted): wire dev to the deployed runtime (after #1355)

lilyshen0722:
Worker deployed 2026-08-30 at https://commonly-agent-runtime.commonlyai.workers.dev with RUNTIME_ADMIN_TOKEN + DEEPSEEK_API_KEY as worker secrets; admin bearer mirrored to SM `commonly-dev-hosted-runtime-admin-token` for the `hosted-runtime` ExternalSecret that #1355 adds. This PR bakes `MODEL_PROVIDER=deepseek` into wrangler.toml and sets `backend.env.hostedRuntimeUrl` for dev.

Merge after #1355, then Deploy Dev.

### DIFF
--- a/k8s/helm/commonly/values-dev.yaml
+++ b/k8s/helm/commonly/values-dev.yaml
@@ -58,6 +58,8 @@ backend:
     # Must match the callback URL registered on the GitHub/Google OAuth apps.
     oauthCallbackBaseUrl: "https://api.commonly.me"
     commonlyApiUrl: ""
+    # ADR-023 W2 hosted runtime (workers/agent-runtime, deployed 2026-08-30).
+    hostedRuntimeUrl: "https://commonly-agent-runtime.commonlyai.workers.dev"
     clawdbotGatewayUrl: ""
     moltbotGatewayUrl: ""
     pgHost: "YOUR_PG_HOST"
--- a/workers/agent-runtime/wrangler.toml
+++ b/workers/agent-runtime/wrangler.toml
@@ -18,3 +18,5 @@ new_sqlite_classes = ["AgentRuntimeDO"]
 # The Commonly instance this runtime serves. Runtime tokens come per-agent
 # via the provision call, never from worker config.
 COMMONLY_API_URL = "https://api.commonly.me"
+# Provider decision 2026-08-29 (Sam): DeepSeek. Key is a secret (DEEPSEEK_API_KEY).
+MODEL_PROVIDER = "deepseek"


lilyshen0722 (commented):
**PASS** at `d834de9a5214f215b466ccf7742578cbe4951375`. 11/11 green, `behind=0`.

Four lines, but they are a config PR that depends on another PR, which is the shape that produced AX entry 53 — so I traced both names to a consumer rather than reading them.

**`hostedRuntimeUrl` resolves end to end**, and only through #1355:

```
values-dev.yaml  backend.env.hostedRuntimeUrl
  → backend-deployment.yaml:153  .Values.backend.env.hostedRuntimeUrl
  → env HOSTED_RUNTIME_URL
  → hostedRuntimeService.ts:45   process.env.HOSTED_RUNTIME_URL
```

No name drift at any hop.

**`MODEL_PROVIDER = "deepseek"` resolves on main already** — `agent-do.ts:161` compares `=== 'deepseek'` and `:162` selects `DEEPSEEK_API_KEY`, matching the comment's claim about the secret's name.

**Merging out of order is inert, not broken.** `git grep hostedRuntimeUrl origin/main` returns nothing, so if this lands before #1355 the value has no reader and the ExternalSecret template does not exist yet — no failing object either way. The "after #1355" in the title is a preference, not a hazard, which is worth knowing if the two get pressed in whatever order the queue offers.

Three things I checked that are fine, noted so the next reader does not re-derive them:

- Setting the URL is what *switches on* the `HOSTED_RUNTIME_ADMIN_TOKEN` block and the `hosted-runtime` ExternalSecret — both gated on this same value. The `secretKeyRef` carries `optional: true`, so a remoteRef that does not exist yet degrades provisioning at call time instead of failing the pod at boot. #1355's template comment explains why it is its own ExternalSecret rather than a key in `api-keys`; that reasoning holds.
- The one precondition this PR asserts without evidence in the diff is that the remote secret backing `HOSTED_RUNTIME_ADMIN_TOKEN` exists before the next deploy. Not verifiable from the repo, and `optional: true` means the failure is quiet — worth confirming out of band.
- `MODEL_PROVIDER` fails open: anything that is not exactly `'deepseek'` silently selects anthropic and then needs `ANTHROPIC_API_KEY`. The string here is correct, so this is not a finding on this PR — but a typo in that var would present as a missing-Anthropic-key error, naming the provider nobody chose.

**Two delivery channels in one PR.** `values-dev.yaml` reaches the cluster via `Deploy Dev`; `wrangler.toml` reaches the worker only when someone runs a wrangler deploy. Merging this makes the backend half live and leaves the worker half staged, so "wire dev to the deployed runtime" is complete only after both. Not a defect — just the thing that is easy to assume a merge finished.

PR #1360, opened 2026-08-30
feat(runtime): RUNTIME_STATUS_TOKEN — read-only /status bearer for watchers

lilyshen0722:
Otto's finding on the deployed worker: the single admin bearer gates every route, so a health watcher holds provision + deprovision power too. This adds an optional `RUNTIME_STATUS_TOKEN` that authorizes `GET /agents/:name/:instance/status` only — every other route stays 401 for it; unset grants nothing; the 503-when-unconfigured branch still fires first.

Router tests (new `test/index.test.ts`): admin reaches all routes; status bearer reads /status and nothing else (POST /status included); unset status token grants nothing; unrouted path 404s behind the gate. Worker suite 30/30, tsc clean.

After merge: `wrangler secret put RUNTIME_STATUS_TOKEN` + deploy, then the watcher gets that token, not the admin one.

Review: Otto gates W2.

### DIFF
--- a/workers/agent-runtime/src/agent-do.ts
+++ b/workers/agent-runtime/src/agent-do.ts
@@ -19,6 +19,7 @@ export interface Env {
   COMMONLY_API_URL: string;
   // Operator admin secret gating every worker route (wrangler secret).
   RUNTIME_ADMIN_TOKEN?: string;
+  RUNTIME_STATUS_TOKEN?: string; // optional read-only bearer: GET /status only
   // Optional model override (defaults in turn.ts).
   MODEL_ID?: string;
   // 'anthropic' (default) or 'deepseek' — which key/provider the runtime uses.
--- a/workers/agent-runtime/src/index.ts
+++ b/workers/agent-runtime/src/index.ts
@@ -14,12 +14,21 @@ export default {
     // asserted.
     const admin = env.RUNTIME_ADMIN_TOKEN;
     if (!admin) return Response.json({ error: 'runtime not configured (RUNTIME_ADMIN_TOKEN unset)' }, { status: 503 });
-    const bearer = req.headers.get('Authorization');
-    if (bearer !== `Bearer ${admin}`) return Response.json({ error: 'unauthorized' }, { status: 401 });
     const url = new URL(req.url);
     const match = url.pathname.match(/^\/agents\/([a-z0-9-]+)\/([a-z0-9-]+)(\/.*)$/);
     if (!match) return Response.json({ error: 'expected /agents/:agentName/:instanceId/...' }, { status: 404 });
     const [, agentName, instanceId, rest] = match;
+    // Two bearers, two capabilities. The admin token provisions, deprovisions
+    // and reads. The optional status token reads /status and nothing else —
+    // a watcher (Otto, a dashboard, an alert) needs health, not the power to
+    // kill or provision any agent by name. Unset means no second bearer.
+    const bearer = req.headers.get('Authorization');
+    const isAdmin = bearer === `Bearer ${admin}`;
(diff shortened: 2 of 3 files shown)


lilyshen0722 (commented):
**PASS at `078ee2e4`** — with one ordering finding worth fixing before merge, and a test whose name over-claims.

## What the change does

Splits one bearer into two capabilities on the worker router: `RUNTIME_ADMIN_TOKEN` keeps everything, the new optional `RUNTIME_STATUS_TOKEN` reads `GET /status` and nothing else. The shape is right — a watcher needs health, not the power to deprovision an arbitrary agent by name — and the unset case grants nothing (`Boolean(env.RUNTIME_STATUS_TOKEN) && ...`, so an empty-string secret is also inert rather than matching an empty bearer).

**No credential is exposed to the lower-privileged reader.** The DO's `/status` returns `{ agentName, provisionedAt, lastPollAt, lastError }` (`agent-do.ts:73-80`) — no `runtimeToken`, no config. `lastError` is the only free-form field, and both writers store `(err as Error).message` alone (`:115`, `:124`), not a stack or a response body. Worth keeping that way: it is now readable by a strictly weaker credential than before.

## Finding: authentication moved after routing

The diff relocates the bearer check from *before* the URL parse to *after* it:

```ts
// before
if (bearer !== `Bearer ${admin}`) return 401;
const match = url.pathname.match(...); if (!match) return 404;

// after
const match = url.pathname.match(...); if (!match) return 404;
...
if (!isAdmin && !(isStatusReader && statusOnly)) return 401;
```

The relocation is necessary — `statusOnly` depends on `rest`, which only exists after the match — but it changes an unrelated behaviour: **an unauthenticated request to any unrouted path now gets `404 {"error":"expected /agents/:agentName/:instanceId/..."}` where it used to get a bare 401.** The router hands the routing shape to a caller holding no credential at all.

That is a small leak (the shape is guessable, and no secret is involved), which is why this is not a block. But the general discipline is worth keeping: authenticate, then route. It costs one extra branch here — compute `isAdmin` before the match, refuse a caller who is neither admin nor *possibly* a status reader, and leave the `statusOnly` refinement where it is. The regex itself is linear (`[a-z0-9-]+` twice, no nesting), so running it pre-auth is not a ReDoS surface — the finding is the disclosure, not the parse.

## The last test's name asserts what the diff broke

```ts
it('404s an unrouted path (still behind the admin gate)', async () => {
  expect((await call(env, 'GET', '/nope', 'admin-secret')).status).toBe(404);
  expect((await call(env, 'GET', '/nope', 'status-secret')).status).toBe(404);
});
```

Both cases pass a valid bearer. The parenthetical — *still behind the admin gate* — is precisely the property that no longer holds, and the one call that would show it (`call(env, 'GET', '/nope')`, no bearer) is not there. As written the test would stay green under either ordering, so it cannot distinguish the behaviour its own name claims.

Adding the no-bearer case is the right move whichever way the ordering finding is resolved: it either pins 401 after a fix, or it makes the 404 an explicit, reviewed decision instead of a silent side effect.

The rest of the suite is genuinely discriminating: 503-before-auth precedence, admin across all three routes, the status bearer refused on `/provision`, `/deprovision` **and `POST /status`** (method, not just path), unset-token, no-bearer and wrong-bearer. `expect(stubFetch).toHaveBeenCalledTimes(1)` on the status-bearer case is the right assertion — it proves the refusals happened before the forward, not after.

## State

- behind = 3 against `main` `f4f36bf1` (`MAX_BEHIND: 40`) — recomputed, not read off the tick.
- 12/12 checks pass, `mergeStateStatus` CLEAN.
- **Read the step, not the job:** these tests do run in CI — `Test & Coverage :: Run hosted-runtime tests (ADR-023 W2) → success` on run `33311093941`, which also runs `npm run typecheck`. Worth stating explicitly, since `workers/` is wired into `tests.yml` by two `cd` lines rather than by a job of its own, and a reader who greps for a workflow named after the worker finds nothing.

PR #1362, opened 2026-08-30
feat(seo): add Commonly entity schema

lilyshen0722:
## Summary
- define stable Organization and WebSite entities with verified GitHub, npm, and Discord references
- make every static public page reference the shared WebSite entity and give the software entity the same publisher
- introduce Commonly by its entity phrase on the landing page and all four existing guides

## Validation
- npm run typecheck
- npm run build
- generated schema verified in home, guide, and compare artifacts

### DIFF
--- a/frontend/scripts/generate-seo-pages.mjs
+++ b/frontend/scripts/generate-seo-pages.mjs
@@ -453,17 +453,19 @@ export const buildPageDefinitions = ({ landing, compare, useCases, guides = {} }
     '@graph': [organization, website, webPageNode(title, description, path)],
   });
 
-  const pages = [{
+  const landingPage = {
     path: '/',
     outputPath: 'index.html',
     title: 'Commonly — chat with your agents, ship real work',
     description: 'The open-source workspace where agents and teammates share one project memory — any runtime, your infra, no per-agent fees.',
     content: renderLanding(landing, useCases, guides),
-    schema: {
-      '@context': 'https://schema.org',
-      '@graph': [organization, website, softwareApplication, webPageNode('Commonly — chat with your agents, ship real work', 'The open-source workspace where agents and teammates share one project memory — any runtime, your infra, no per-agent fees.', '/')],
-    },
-  }, {
+  };
+  landingPage.schema = {
+    '@context': 'https://schema.org',
+    '@graph': [organization, website, softwareApplication, webPageNode(landingPage.title, landingPage.description, landingPage.path)],
+  };
+
+  const pages = [landingPage, {
     path: '/compare/',
     outputPath: 'compare/index.html',
     title: 'Commonly vs the alternatives | Commonly',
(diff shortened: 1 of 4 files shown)


lilyshen0722 (commented):
**PASS at `d3704896d`** — the entity graph is well-formed and the tests pin the new wiring. One latent defect worth a one-line fix before it bites, demonstrated below rather than asserted.

## Verified

The `@id` refactor is the right JSON-LD shape: define `#organization` and `#website` once, reference them everywhere else by `{ '@id': ... }` (`publisher`, `isPartOf`), and include the full nodes in each page's `@graph` so every page is self-contained. `webPageNode` correctly drops the per-node `@context` now that the graph carries it once at the top.

Ran the suite at this head under node 22: `node --test scripts/generate-seo-pages.test.mjs` → **2/2 pass**. The new assertions discriminate — they check `organization['@id']`, both `publisher` references, and loop every page asserting `webPage.isPartOf` equals the website `@id`, so a page added later without the reference fails rather than passing silently.

## Finding: the article `author` node claims the site's Organization identity

`generate-seo-pages.mjs:503`, changed by this diff:

```js
const author = { '@type': 'Organization', '@id': organizationId, name: provenance.author, url: `${siteUrl}/` };
```

`provenance.author` is per-guide free text. `guideProvenance` only checks that it is present — nothing constrains it to `'Commonly'`, and it is rendered as a human byline (`:156`, `By ${provenance.author}`). Giving that node the canonical `#organization` `@id` means any guide whose author differs from the Organization's `name` emits two nodes with the same `@id` and different `name` values, which consumers flatten into one entity.

Today all five guides say `"author": "Commonly"`, so the graph is consistent and nothing is broken at this head. That is why this is not a block — but it is also why it will not be noticed. I set one guide's author to a person and rebuilt:

```
article.author = {"@type":"Organization","@id":"https://commonly.me/#organization","name":"Lily Shen","url":"https://commonly.me/"}
top-level org  = {"@id":"https://commonly.me/#organization","name":"Commonly"}
same @id? true | same name? false
```

Two nodes, one identifier, conflicting names — and a person mistyped as an `Organization`, since the `@type` is hardcoded next to the `@id`.

Note the asymmetry the diff introduces: `reviewer` on the very next line stays anonymous (no `@id`) and carries a *different* name, `'Commonly SEO team'`, which is correct. The `@id` was added to exactly the one of the pair whose name happens to match today.

Cheapest fix is to drop `'@id': organizationId` from `author`, matching `reviewer`. If the intent is genuinely "the org authored this," then keep the `@id` and use `{ '@id': organizationId }` as the whole value — a bare reference with no competing `name` — and let the byline text stay presentational.

The existing test asserts only `article.author['@type'] === 'Organization'` (`:79`), which passes under either shape, so nothing currently distinguishes them.

## Note on the review itself

My first pass scoped this PR at six files including `index.html`, `GuidePage.tsx` and a `V2Login.test.tsx` assertion change — and I was one step from filing "this duplicates #1361's guide-count change" as a finding. That merge-base was stale: `main` advanced to `de90772b4` (absorbing #1361) between my fetch and my diff. Re-resolved, this PR is three files. Recording it because the wrong version was a plausible-looking finding with real line numbers.

## State

behind = 0 against `main` `de90772b4` — recomputed after the re-fetch, not read off the check. 10 pass, 1 pending (`Service Tests (Tier 1 — real DBs)`); `mergeStateStatus` UNSTABLE is that pending check, not a failure.

PR #1364, opened 2026-08-30
feat(seo): add human-in-the-loop review guide

lilyshen0722:
## Summary\n- publish the static Human-in-the-Loop Review for AI Agent Teams guide at /guides/human-in-the-loop-ai-agents/\n- add hub, sitemap, Article/WebPage JSON-LD, real provenance, and Page 5 reciprocal linking\n- pin static generation and React hub-card coverage (six cards)\n\n## Verification\n- npm test (85 suites, 472 tests)\n- npm run typecheck\n- npm run build\n\n@samxu01?

### DIFF
--- a/frontend/scripts/generate-seo-pages.test.mjs
+++ b/frontend/scripts/generate-seo-pages.test.mjs
@@ -19,7 +19,7 @@ test('emits a canonical crawlable page for every public route', async () => {
   const guides = JSON.parse(guideText);
   const pages = buildPageDefinitions({ landing: translations.landing, compare: translations.compare, useCases, guides });
 
-  assert.equal(pages.length, 15);
+  assert.equal(pages.length, 16);
   assert.deepEqual(pages.map((page) => page.path), [
     '/',
     '/compare/',
@@ -36,6 +36,7 @@ test('emits a canonical crawlable page for every public route', async () => {
     '/guides/ai-agent-task-management/',
     '/guides/connect-claude-codex-shared-workspace/',
     '/guides/ai-agent-memory/',
+    '/guides/human-in-the-loop-ai-agents/',
   ]);
   assert.deepEqual(pages[0].schema['@graph'].map((item) => item['@type']), [
     'Organization',
@@ -71,7 +72,7 @@ test('emits a canonical crawlable page for every public route', async () => {
     '/guides/ai-agent-task-management/',
     '/guides/connect-claude-codex-shared-workspace/',
   ]);
-  assert.equal(guidePages.length, 5);
+  assert.equal(guidePages.length, 6);
   for (const guide of guidePages) {
     assert.equal(guide.ogType, 'article');
     const article = guide.schema['@graph'].find((item) => item['@type'] === 'Article');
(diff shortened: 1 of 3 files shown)


lilyshen0722 (commented):
Gate on #1364 at `796a731b` (`base=main`, 3 files, +212/-4). **No blockers — content is well-formed, the tests discriminate, and the one structural risk in a PR like this is clean. One non-blocking observation about the link graph.**

**Ran the suite at this head rather than trusting the tick.** `node --test scripts/generate-seo-pages.test.mjs` → 2/2 pass. `gh pr checks` shows the full 11-check set for `base=main` (10 pass, Service Tests pending at review time) — worth noting because a stacked base would silently run 4, and this one doesn't.

**The tests are non-vacuous.** The count assertions (`pages.length` 15→16, `guidePages.length` 5→6, `toHaveLength(5→6)` in `V2Login.test.tsx`) are hardcoded literals checked against a data-driven render, so they discriminate rather than moving with the data. Beyond the counts, the new cases assert real body sentences, the article `ogType`/schema shape, and — the part I'd have asked for — the cross-link in **both** directions: `memoryHtml` contains a link to the new guide, and the new guide's HTML links back to memory.

**Key parity holds, which is the actual failure mode here.** A guide missing a key that siblings have renders `undefined` into the page, and the generator dereferences `guide.relatedLinks.map`, `guide.cta.primary.path`, and `guide.cta.secondary.path` unguarded. I diffed the key sets across all six guides: identical, no missing key on any of them. So no `undefined` in output and no throw.

**No injection surface introduced.** Zero raw `<` or `&` in the 196 added lines, and every emission site in `generate-seo-pages.mjs` routes through `escapeHtml` — including hrefs (`escapeHtml(link.path)` at `:331`, `escapeHtml(guide.cta.primary.path)` at the CTA band). The two new outbound citations are both `learn.microsoft.com`. This matters because the JSON-LD path is the one I verified on #1362 (`JSON.stringify(schema).replaceAll('<','\\u003c')`); this PR feeds that path and doesn't stress it.

**Internal links all resolve — including the one I initially flagged wrong.** I checked every internal target against the generated page set. My first pass flagged the new guide's `cta.secondary` → `/guides/` as dangling; that was **my heuristic being wrong, not the PR** — I'd compared against guide slugs only, and `/guides/` is a real generated index (`guidesIndexPath`, `generate-seo-pages.mjs:477`, and it's in the asserted 16-path list). Zero dangling links. Flagging my own false positive so nobody re-raises it.

---

**NON-BLOCKING — the new guide is a near-leaf in the link graph.** Measured across all six guides (relatedLinks + CTA targets, excluding the `/guides/` index):

| guide | outbound | inbound |
|---|---|---|
| multi-agent-collaboration-platform | 4 | 6 |
| ai-agent-workspace | 4 | 4 |
| ai-agent-task-management | 4 | 5 |
| connect-claude-codex-shared-workspace | 4 | 4 |
| ai-agent-memory | 5 | 5 |
| **human-in-the-loop-ai-agents** | **3** | **1** |

Every existing guide links to all four siblings and is linked from four-plus. The new one links to three and is linked from exactly one (`ai-agent-memory`). Nothing is broken — but internal linking is the mechanism these pages exist to exploit, so a page with one inbound link is the one that underperforms. Adding it to `relatedLinks` on `ai-agent-workspace` and `connect-claude-codex-shared-workspace` (and those two to its own list) would make it symmetric with its siblings. Cheap, and it's data-only.

**Pre-existing, explicitly NOT charged to this PR:** `escapeHtml` sanitises `&<>"'` but does not constrain the URL *scheme*, so a `javascript:` href in content JSON would survive to output. The threat model is a malicious PR, which review catches, and the code is untouched by this diff (`:331` is not in it). Noting it only so the next person to read that line doesn't file it against a content PR.

**NOT VERIFIED:** I did not render the pages in a browser, so I'm making no claim about layout or the `seo-table`/`seo-note` styling of the new sections. I did not run the frontend jest suite — `V2Login.test.tsx` is asserted green by CI's `Test & Coverage`, not by me. I did not check the two `learn.microsoft.com` URLs resolve, nor evaluate the guide's factual claims. `Service Tests (Tier 1)` was still pending when I reviewed; it does not touch this diff, but I did not watch it land.

PR #1369, opened 2026-08-30
feat(seo): add AI agent handoffs guide

lilyshen0722:
## Summary\n- publish the static AI Agent Handoffs guide at /guides/ai-agent-handoffs/\n- add hub and sitemap coverage, Article/WebPage metadata, and real 2026-08-30 provenance\n- add reciprocal links with the published memory and human-review guides; no Page 8 link\n\n## Verification\n- npm test (85 suites, 473 tests)\n- npm run typecheck\n- npm run build\n\n@samxu01?

### DIFF
--- a/frontend/scripts/generate-seo-pages.test.mjs
+++ b/frontend/scripts/generate-seo-pages.test.mjs
@@ -19,7 +19,7 @@ test('emits a canonical crawlable page for every public route', async () => {
   const guides = JSON.parse(guideText);
   const pages = buildPageDefinitions({ landing: translations.landing, compare: translations.compare, useCases, guides });
 
-  assert.equal(pages.length, 16);
+  assert.equal(pages.length, 17);
   assert.deepEqual(pages.map((page) => page.path), [
     '/',
     '/compare/',
@@ -37,6 +37,7 @@ test('emits a canonical crawlable page for every public route', async () => {
     '/guides/connect-claude-codex-shared-workspace/',
     '/guides/ai-agent-memory/',
     '/guides/human-in-the-loop-ai-agents/',
+    '/guides/ai-agent-handoffs/',
   ]);
   assert.deepEqual(pages[0].schema['@graph'].map((item) => item['@type']), [
     'Organization',
@@ -72,7 +73,7 @@ test('emits a canonical crawlable page for every public route', async () => {
     '/guides/ai-agent-task-management/',
     '/guides/connect-claude-codex-shared-workspace/',
   ]);
-  assert.equal(guidePages.length, 6);
+  assert.equal(guidePages.length, 7);
   for (const guide of guidePages) {
     assert.equal(guide.ogType, 'article');
     const article = guide.schema['@graph'].find((item) => item['@type'] === 'Article');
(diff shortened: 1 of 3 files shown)


lilyshen0722 (commented):
Gate on #1369 at `8a85a0de` (`base=main`, 3 files, +216/-5). **No blockers.** Same shape as #1364, and the checks I ran there come back clean here too. One non-blocking note on the link graph, now that there are enough data points to see a pattern rather than a one-off.

Ran the generator suite at this head: 2/2 pass. Full 11-check set for `base=main`, all green at review time.

**The three structural risks in a content PR of this shape, all clean:**
- **Key parity** across all seven guides — identical key sets, no missing key. This is the one that matters, because the template dereferences `guide.relatedLinks.map`, `guide.cta.primary.path` and `guide.cta.secondary.path` unguarded; a guide short a key renders `undefined` or throws.
- **No injection surface** — zero raw `<` or `&` in the 199 added lines, and every emission site routes through `escapeHtml`, hrefs included.
- **No dangling internal links** — every `/guides/…` target resolves to a generated page.

**Tests discriminate.** Counts are hardcoded literals against a data-driven render (`pages.length` 16→17, `guidePages.length` 6→7, `toHaveLength(6→7)`), so they can't drift with the data. The new cases assert real body sentences and both directions of the cross-link — `humanReviewHtml` links to handoffs, and `handoffsHtml` links back to memory and human-in-the-loop.

---

**NON-BLOCKING — the guides are growing as a chain, not a mesh.** I raised a version of this on #1364 as a single-guide observation; with a second new guide it's clearly structural, so it's worth stating once properly.

The five original guides link to each other completely. Neither new guide is reachable from any of the four non-memory originals:

```
multi-agent-collaboration-platform     -> neither
ai-agent-workspace                     -> neither
ai-agent-task-management               -> neither
connect-claude-codex-shared-workspace  -> neither
```

Inbound edges for the two newest are only ever from `ai-agent-memory` and from each other:

```
human-in-the-loop-ai-agents   in: ai-agent-memory, ai-agent-handoffs      (2)
ai-agent-handoffs             in: ai-agent-memory, human-in-the-loop      (2)
```

versus 4–7 inbound for each of the originals. **Correcting my own framing from #1364:** I noted there that human-in-the-loop had one inbound and suggested adding it to `ai-agent-workspace` and `connect-claude-codex`. Its inbound count did go 1→2 — but that is entirely because *this* PR's new guide links to it. The suggestion wasn't taken, and the count improving is incidental. I'd have mis-read that as uptake if I hadn't checked which edge moved.

Each new guide gets appended to the tail rather than woven in, so the newest page is always the least reachable one — and internal linking is the mechanism these pages exist for. Adding the two new slugs to the four originals' `relatedLinks` is data-only and would make the set a mesh again. Worth one pass now rather than compounding at guide eight.

**Also confirmed, since it bears on merge discipline:** #1364 merged at `796a731b`, byte-identical to the head my gate was bound to. No post-review drift there.

**NOT VERIFIED:** no browser render, so no claim about layout or styling of the new sections. I did not run the frontend jest suite — `V2Login.test.tsx` is green per CI's `Test & Coverage`, not per me. I did not resolve the outbound `openai.com` / `learn.microsoft.com` citations or evaluate the guide's factual claims.

PR #1370, opened 2026-08-30
test(runtime): pin the reply-staging ordering in handleEvent (#1366)

### LINKED ISSUE #1366: Reply-staging ordering in handleEvent is untested: moving commitStagedReply above the post restores #1344 with 30/30 green
Follow-up to #1346 (merged `54775a52`). Two of my three gate findings were addressed on the post-review commits — the storage leak now has a real `pruneStaged` + `STAGE_TTL_MS`, and the malformed-entry shape guard went wider than I'd asked. Credit where due. **The third one shipped unaddressed, and I've now proven it rather than asserting it.**

## The gap

`handleEvent` in `workers/agent-runtime/src/agent-do.ts` encodes the entire contract the PR exists for:

```ts
if (reply.trim() !== 'NO_REPLY') {
  await postMessage(cfg, podId, reply);
}
await commitStagedReply(this.state.storage, event._id);
```

The ordering *is* the fix. `commitStagedReply` must run **after** a successful post, and must **not** run when the post throws — that's what leaves the stage intact so the redelivery replays it instead of re-running the model.

**No test exercises `handleEvent`.** All seven suites, grepped at `7b46256e1`:

```
agent-do.delivery-nonce.test.ts   handleEvent:0   (imports agent-do, but not this path)
budget.test.ts                    handleEvent:0
cap.test.ts                       handleEvent:0
staging.test.ts                   handleEvent:0
tools.test.ts                     handleEvent:0
turn.persistence.test.ts          handleEvent:0
turn.test.ts                      handleEvent:0
```

The six excellent `staging.test.ts` cases drive `staging.ts` through an injected `Map`. They pin the *primitives* — stage, resolve, commit, prune, shape-guard — and pin them well. None pins the *call site*, which is where the ordering lives.

## Proven, not argued

Clean worktree at `origin/main` = `7b46256e1`, node 22, `vitest run`.

**Baseline: 30 passed (30).**

Mutation — the exact refactor I named in the gate, moving the commit above the post (anchor asserted to match exactly once before applying):

```ts
    await commitStagedReply(this.state.storage, event._id);   // ← moved up
    if (reply.trim() !== 'NO_REPLY') {
      await postMessage(cfg, podId, reply);
    }
```

**Under mutation: 30 passed (30).** Zero red.

That mutation deletes the staged reply *before* attempting the post, so a `postMessage` throw leaves no stage — and the kernel's redelivery re-runs the model. **That is #1344, restored verbatim, with the entire suite green.** Tree restored; `git status --porcelain` empty.

## Suggested test

One call-site case is enough — it's the assertion neither existing suite makes:

```ts
it('a failed post leaves the stage intact — commit does NOT run (#1344 ordering)', async () => {
  // postMessage throws; assert storage still holds staged:<id> afterwards,
  // and that a second handleEvent for the same event does not call run() again.
});
```

The value is not coverage percentage — it's that this is the one line whose *order* is load-bearing and whose reordering is otherwise invisible. A future refactor tidying "clear the stage, then post" reads as equivalent and ships green.

## Two smaller notes, non-blocking

**The `STAGE_TTL_MS` premise is a claim about the kernel, asserted in a worker comment.** `staging.ts` justifies the 24h TTL with *"The kernel retires an event after 3 redeliveries, so an orphan is never reused."* If three redeliveries can ever span more than 24h, the prune deletes a stage that is still live and the model re-runs — the exact failure the file exists to prevent, now on a timer. **I did not verify the redelivery cadence**, so this is a flag rather than a finding; it's worth someone confirming the bound, because a cross-repo claim in a comment has no reader that could catch it drifting.

**`pruneStaged` does a full `storage.list({ prefix })` on every staged turn**, not just periodically. Correct, and cheap at current volume; noting it because the cost scales with orphan count, which is the thing it exists to bound.

**NOT VERIFIED:** I did not run this under `workerd`/`wrangler` — the DO storage contract is still exercised only through an injected `Map`, so real `get`/`put`/`delete`/`list` semantics remain untested by me, same caveat as my original gate. `tsc --noEmit` not run.

lilyshen0722:
Closes #1366.

@sprint-review proved the gap rather than asserting it: moving `commitStagedReply` above the post in `handleEvent` re-introduces #1344 with all 30 tests green. `staging.test.ts`'s six cases drive `staging.ts` through an injected `Map` and pin the primitives well — stage, resolve, commit, prune, shape-guard. None pins the call site, and the call site is where the ordering lives.

This adds the one assertion neither existing suite makes, through the mock harness `agent-do.delivery-nonce.test.ts` already established (mock `../src/cap`, construct the DO over a fake `DurableObjectState`, drive it with `alarm()`).

Three cases:

- **a failed post leaves the stage intact** — `staged:event-1` still holds the reply, `processedEventIds` is unset, `ackEvent` was never called;
- **the redelivery replays it without re-running the model** — post throws on the first `alarm()`, succeeds on the second, and `runTurn` is called **once** across both. That is #1344's contract stated directly: one paid turn across two deliveries;
- **a NO_REPLY turn posts nothing and still clears its stage** — the branch that skips the post must not leak an orphan.

## Mutation control

node 22, clean worktree at `374f4c7c`, `vitest run`.

| tree | result |
|---|---|
| main | 30 passed (30) |
| main + this file | 33 passed (33) |
| mutated (commit moved above the post), **without** this file | 30 passed (30) — the exclusion arm |
| mutated, with this file | **2 failed** — `expected "spy" to be called 1 times, but got 2 times` |

The discriminating assertion is the `runTurn` call count, which is the paid resource #1344 is about. Source restored afterwards (`git diff` empty on `src/`), `tsc --noEmit` exits 0.

## Not addressed here, deliberately

#1366's two smaller notes are untouched: the `STAGE_TTL_MS` justification is a claim about kernel redelivery cadence asserted in a worker comment with no reader, and `pruneStaged` does a full prefix `list` on every staged turn. Both are about `staging.ts`, not the call site, and neither is what the issue's title names.

**NOT VERIFIED:** same caveat as the issue — this runs under vitest with an injected storage object, not under `workerd`/`wrangler`, so real Durable Object `get`/`put`/`delete`/`list` semantics remain unexercised. The harness gains a `list` stub because the existing one predates `pruneStaged`.

### DIFF
--- a/workers/agent-runtime/test/agent-do.staging-ordering.test.ts
+++ b/workers/agent-runtime/test/agent-do.staging-ordering.test.ts
@@ -1,0 +1,101 @@
+// The ordering in handleEvent IS #1344's fix: commitStagedReply must run
+// AFTER a successful post and must NOT run when the post throws — that is
+// what leaves the stage intact so a redelivery replays it instead of paying
+// for the model again. sprint-review proved (#1366) that moving the commit
+// above the post re-introduces #1344 with the whole suite green: staging.ts
+// is well covered, but nothing exercised the CALL SITE where the order lives.
+//
+// This drives the DO through alarm() twice — post throws, then succeeds —
+// and asserts the model ran ONCE across both deliveries. A refactor tidying
+// this to "clear the stage, then post" reads as equivalent and goes red here.
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const cap = vi.hoisted(() => ({
+  listEvents: vi.fn(),
+  ackEvent: vi.fn(),
+  postMessage: vi.fn(),
+}));
+const turn = vi.hoisted(() => ({ runTurn: vi.fn() }));
+
+vi.mock('../src/cap', () => ({
+  ...cap,
+  StaleDeliveryError: class StaleDeliveryError extends Error {},
+}));
+vi.mock('../src/turn', () => turn);
+
+import { AgentRuntimeDO } from '../src/agent-do';
+import { stagedKey } from '../src/staging';
+
+const cfg = { apiUrl: 'https://api.test', runtimeToken: 'cm_agent_x' };
+const EVENT = {
+  _id: 'event-1',
+  type: 'chat.mention',
+  podId: 'pod-1',
+  payload: { content: 'ping' },
+};
+
+const stateWith = (values: Record<string, unknown>) => {
+  const data = new Map(Object.entries(values));
+  const storage = {
+    get: vi.fn(async (key: string) => data.get(key)),
+    put: vi.fn(async (key: string | Record<string, unknown>, value?: unknown) => {
+      if (typeof key === 'string') data.set(key, value);
+      else Object.entries(key).forEach(([entry, stored]) => data.set(entry, stored));
+    }),
+    delete: vi.fn(async (key: string) => data.delete(key)),
+    // pruneStaged lists on every stage; the existing harness predates it.
+    list: vi.fn(async ({ prefix }: { prefix: string }) => {
+      const out = new Map<string, unknown>();
+      for (const [key, value] of data) if (key.startsWith(prefix)) out.set(key, value);
+      return out;
+    }),
+    setAlarm: vi.fn(async () => undefined),
+  };
+  return { data, storage, state: { storage } as unknown as DurableObjectState };
+};
+
+describe('handleEvent reply-staging ordering (#1344 / #1366)', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    cap.listEvents.mockResolvedValue([EVENT]);
+    cap.ackEvent.mockResolvedValue(undefined);
+    turn.runTurn.mockResolvedValue('the reply');
+  });
+
+  it('a failed post leaves the stage intact and the event unprocessed', async () => {
+    const { data, state } = stateWith({ runtimeToken: cfg.runtimeToken, pollSeconds: 5 });
+    cap.postMessage.mockRejectedValue(new Error('502 from CAP'));
+
+    await new AgentRuntimeDO(state, { COMMONLY_API_URL: cfg.apiUrl } as never).alarm();
+
+    expect(data.get(stagedKey(EVENT._id))).toMatchObject({ reply: 'the reply' });
+    expect(data.get('processedEventIds')).toBeUndefined();
+    expect(cap.ackEvent).not.toHaveBeenCalled();
+  });
+
+  it('the redelivery replays the staged reply without re-running the model', async () => {
+    const { data, state } = stateWith({ runtimeToken: cfg.runtimeToken, pollSeconds: 5 });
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
Gate on #1370 at `6182763a` (`base=main`, 1 file, +101/-0, `mergeable_state: clean`). **PASS — this closes the gap in #1366, and I verified that by mutation rather than by reading it.** This PR fixes a finding I filed, so the only thing worth reporting is whether the new tests actually discriminate. They do.

**Baseline at this head: 33 passed (8 files).**

**The decisive check.** The mutation that motivated #1366 — moving `commitStagedReply` above `postMessage`, which passed **30/30 green on `main`** — now fails. Anchor asserted to match exactly once before applying, tree restored after:

```
M1  commit above post          Tests  2 failed | 31 passed
    × a failed post leaves the stage intact and the event unprocessed
    × the redelivery replays the staged reply without re-running the model
```

That is the gap closed, measured the same way it was opened.

**The three cases are discriminating, not overlapping.** The third case survives M1, so I gave it its own mutations rather than accepting it as decorative:

| mutation | tests red |
|---|---|
| M1 — `commitStagedReply` above the post | 1 & 2 |
| M2 — drop the `NO_REPLY` guard (`if (true)`) | 3 only |
| M3 — never commit (`void commitStagedReply`) | 2 & 3 |

Every case reds on at least one mutation, and the three mutations red different subsets — so the cases pin distinct behaviour rather than restating each other. `git status --porcelain` empty afterwards.

**The design is the right one.** Driving the DO through `alarm()` twice — post throws, then succeeds — and asserting `runTurn` was called **once across both deliveries** tests the actual #1344 invariant (one paid model turn per event) rather than a proxy for it. Asserting `processedEventIds` is unset and `ackEvent` was not called on the failed pass pins the redelivery precondition too, which is what makes the second `alarm()` meaningful instead of incidental. Reusing the `agent-do.delivery-nonce.test.ts` harness keeps this consistent with the file that already mocks `../src/cap`.

The `list` addition to the fake storage is necessary, not incidental — `pruneStaged` lists on every stage, and the pre-existing harness predates it. The inline comment says exactly that, which is the right amount of explanation.

**Nothing blocking. Nothing non-blocking either** — I looked for over-specification (assertions that would red on a harmless refactor) and did not find any: every assertion is on observable behaviour, none on call counts of internal helpers or on message shape beyond the one `toHaveBeenLastCalledWith` that pins the replayed text.

**NOT VERIFIED:** still no `workerd`/`wrangler` run — the DO storage contract is exercised through a fake `Map`-backed object, so real `get`/`put`/`delete`/`list` semantics remain untested here, same caveat I carried on #1346 and #1366. I did not run `tsc --noEmit`. The two smaller notes from #1366 are untouched by this PR and stay open: the `STAGE_TTL_MS` justification asserts a kernel redelivery bound ("retires after 3 redeliveries") that nothing in this repo checks, and `pruneStaged` does a full prefix `list` on every staged turn. Neither blocks this; flagging so they aren't assumed closed along with the ordering.

PR #1371, opened 2026-08-30
feat(seo): add AI agent team guide

lilyshen0722:
## Summary
- publish the static guide at `/guides/how-to-build-an-ai-agent-team/`
- add hub, sitemap, canonical, Article/WebPage metadata, and provenance via the guide generator
- add reciprocal links from Pages 5–7 only; Page 9/10 are not linked

## Verification
- `npm test` (85 suites, 473 tests)
- `npm run typecheck`
- `npm run build` (verified the static Page 8 artifact)

### DIFF
--- a/frontend/scripts/generate-seo-pages.test.mjs
+++ b/frontend/scripts/generate-seo-pages.test.mjs
@@ -19,7 +19,7 @@ test('emits a canonical crawlable page for every public route', async () => {
   const guides = JSON.parse(guideText);
   const pages = buildPageDefinitions({ landing: translations.landing, compare: translations.compare, useCases, guides });
 
-  assert.equal(pages.length, 17);
+  assert.equal(pages.length, 18);
   assert.deepEqual(pages.map((page) => page.path), [
     '/',
     '/compare/',
@@ -38,6 +38,7 @@ test('emits a canonical crawlable page for every public route', async () => {
     '/guides/ai-agent-memory/',
     '/guides/human-in-the-loop-ai-agents/',
     '/guides/ai-agent-handoffs/',
+    '/guides/how-to-build-an-ai-agent-team/',
   ]);
   assert.deepEqual(pages[0].schema['@graph'].map((item) => item['@type']), [
     'Organization',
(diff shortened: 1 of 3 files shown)


lilyshen0722 (commented):
Gate on #1371 at `cfea9ce9` (`base=main`, 3 files, +252/-6). **No blockers.** Same battery I ran on #1364 and #1369; everything comes back clean.

Ran the generator suite at this head: 2/2 pass. Full 11-check set for `base=main` (10 pass, `Test & Coverage` pending at review time). `mergeable_state: blocked` is the missing-review gate, not a failing check.

- **Key parity** across all eight guides — identical key sets. This is the one that would actually break: the template dereferences `guide.relatedLinks.map`, `guide.cta.primary.path` and `guide.cta.secondary.path` unguarded.
- **No injection surface** — zero raw `<` or `&` in the 235 added lines; every emission site routes through `escapeHtml`, hrefs included.
- **No dangling internal links** — every `/guides/…` target resolves to a generated page.
- **Tests discriminate** — counts are hardcoded literals against a data-driven render (`pages.length` 17→18, `guidePages.length` 7→8, `toHaveLength(7→8)`), and the new cases assert real body sentences plus the cross-link in both directions (`handoffsHtml` → team guide, `teamHtml` → memory, human-in-the-loop, handoffs).

**The link-graph note I raised here twice is now filed as #1372 rather than repeated a third time.** Short version: the four original non-memory guides link to none of the three newest, so each new guide lands with 3 inbound against 4–8 for the originals. Non-blocking, data-only, and it does not need to hold up this PR — but it now has somewhere to live that a merge won't close. Worth noting I misread the trend on #1369: `human-in-the-loop`'s inbound count rose 1 → 3, which looked like uptake and was actually just newer siblings linking to it. The aggregate is misleading without checking which edges moved.

**Also confirmed:** #1369 merged at `8a85a0de`, identical to the head my gate was bound to. No post-review drift.

**NOT VERIFIED:** no browser render, so no claim about layout or styling. I did not run the frontend jest suite — `V2Login.test.tsx` is green per CI, not per me. I did not resolve the outbound citations or evaluate the guide's factual claims. `Test & Coverage` was still pending when I reviewed.

PR #1374, opened 2026-08-30
fix(runtime): status token equal to admin token treated as unset

lilyshen0722:
Otto's non-blocking note on #1360, as its own PR so his approved sha merged exactly as reviewed: `RUNTIME_STATUS_TOKEN === RUNTIME_ADMIN_TOKEN` now disables the reader path instead of silently collapsing the capability split. Test added; worker suite green. Otto gates.

### DIFF
--- a/workers/agent-runtime/src/index.ts
+++ b/workers/agent-runtime/src/index.ts
@@ -24,7 +24,14 @@ export default {
     // kill or provision any agent by name. Unset means no second bearer.
     const bearer = req.headers.get('Authorization');
     const isAdmin = bearer === `Bearer ${admin}`;
-    const isStatusReader = Boolean(env.RUNTIME_STATUS_TOKEN) && bearer === `Bearer ${env.RUNTIME_STATUS_TOKEN}`;
+    // A status token EQUAL to the admin token is treated as unset (Otto on
+    // #1360): an operator pasting the wrong secret would otherwise silently
+    // collapse the two capabilities back into one — the exact split this
+    // token exists to enforce.
+    const statusToken = env.RUNTIME_STATUS_TOKEN && env.RUNTIME_STATUS_TOKEN !== admin
+      ? env.RUNTIME_STATUS_TOKEN
+      : undefined;
+    const isStatusReader = Boolean(statusToken) && bearer === `Bearer ${statusToken}`;
     const statusOnly = req.method === 'GET' && rest === '/status';
     if (!isAdmin && !(isStatusReader && statusOnly)) {
       return Response.json({ error: 'unauthorized' }, { status: 401 });
--- a/workers/agent-runtime/src/index.ts
+++ b/workers/agent-runtime/src/index.ts
@@ -24,14 +24,15 @@ export default {
     // kill or provision any agent by name. Unset means no second bearer.
     const bearer = req.headers.get('Authorization');
     const isAdmin = bearer === `Bearer ${admin}`;
-    // A status token EQUAL to the admin token is treated as unset (Otto on
-    // #1360): an operator pasting the wrong secret would otherwise silently
-    // collapse the two capabilities back into one — the exact split this
-    // token exists to enforce.
-    const statusToken = env.RUNTIME_STATUS_TOKEN && env.RUNTIME_STATUS_TOKEN !== admin
-      ? env.RUNTIME_STATUS_TOKEN
-      : undefined;
-    const isStatusReader = Boolean(statusToken) && bearer === `Bearer ${statusToken}`;
+    // A status token EQUAL to the admin token is a refused CONFIGURATION
+    // (Otto on #1374, round 2): nulling the reader path closes nothing,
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
Gated at `b755094c`.

**This change cannot alter any response, and its test does not pin it.**

`isStatusReader` has exactly one consumer, and it sits inside `!isAdmin`:

```ts
if (!isAdmin && !(isStatusReader && statusOnly)) return 401;
```

Take the case the patch exists for — `RUNTIME_STATUS_TOKEN === RUNTIME_ADMIN_TOKEN`. Old code sets `isStatusReader = (bearer === \`Bearer ${admin}\`)`, which is definitionally `isAdmin`. So `isStatusReader` can only be true when `isAdmin` is true, and `!isAdmin` has already short-circuited the guard. The reader branch is never consulted. There is no capability to collapse: the "status token" never grants anything the admin token was not already granting on that same bearer.

The other two cases are unchanged by construction — `S` falsy gives `false` both ways, and `S !== A` gives `statusToken === S`, identical to the old expression.

**Empirically**, on a worktree at this head:

```
npx vitest run test/index.test.ts        -> 6 passed   (patched src)
git show origin/main:.../src/index.ts    -> restored
npx vitest run test/index.test.ts        -> 6 passed   (unpatched src)
```

The new case is green on `main`'s source. Both of its assertions hold before the patch: `admin-secret` provisions because it is the admin bearer, and `status-secret` gets 401 because it equals neither token. Nothing in it depends on the line that changed.

No test *can* discriminate here, which is the actual finding — the difference is unobservable through the worker's public interface.

**Not asking for a revert.** The code is clearer and the comment records a real operator hazard. But the PR title says a token is now "treated as unauthorized" and the test name claims "no capability collapse", and neither describes a behaviour change. Suggest retitling to what it is — a defensive normalization plus a comment — and either dropping the test or moving the assertion to something that would catch a future refactor lifting `isStatusReader` out from under `!isAdmin`, which is the only way this could ever start to matter.

10/10 checks green.

lilyshen0722 (commented):
**PASS** — gated at `2ea6aba6b39272a43f53359913313ecce89de586` (re-resolved at review time; matches `headRefOid`).

The reasoning in the comment is correct and it is the right call. When the two secrets are byte-identical, `isAdmin` accepts the "status" bearer before `isStatusReader` is ever consulted, so nulling the reader path closes nothing — the holder is already an admin. Refusing to serve is the only remedy that isn't theatre.

**Ran it rather than read it.** Installed deps in `workers/agent-runtime/` and ran the suite: **36/36 pass** across 8 files.

Then mutation-tested the diff, because a guard's test is worth exactly what it discriminates:

```
remove the guard   -> anchor count 1 -> 0 (confirmed applied)
                   -> test/index.test.ts: 1 failed | 5 passed
restore            -> anchor back to 1, 6/6 green
```

The new test reds alone when the guard is removed, and the other five stay green. It pins the line, not the neighbourhood. `expect(stubFetch).not.toHaveBeenCalled()` is the right assertion — it proves nothing was forwarded to the DO, not merely that the status code changed.

Four extra probes I wrote against the real worker (not a re-implementation), all pass:

| probe | result |
|---|---|
| `RUNTIME_STATUS_TOKEN` unset bearer entirely | **503** — refusal precedes auth |
| admin `''` + status `''` | **503**, and the body says `RUNTIME_ADMIN_TOKEN unset` |
| unmatched path while misconfigured | **404** (path match precedes the 503) |
| **control:** correct config, `status-secret` | `/status` **200**, `/provision` **401**, stub called once |

The control matters most: a refusal guard that also breaks the healthy deployment is a worse bug than the one it fixes. It doesn't.

Two non-blocking notes, neither worth another round:

**1. The guard is silently load-bearing on line 17.** `env.RUNTIME_STATUS_TOKEN && env.RUNTIME_STATUS_TOKEN === admin` short-circuits on a falsy status token — so the `''` + `''` pair is caught *only* by the pre-existing `if (!admin)` above it. That's correct today (probe 2 confirms the 503, and the body proves which check fired). But the correctness lives in a different `if`, four lines up, and nothing records the dependency. If anyone ever relaxes line 17 to permit an empty admin token, this guard goes fail-open and the test here won't see it — the test uses `'admin-secret'`, so it never exercises the falsy branch. A clause in the existing comment is enough.

**2. The 503 is pre-auth, so an unauthenticated prober can distinguish a misconfigured deployment from a healthy one** (probe 1). I'd keep it exactly as is — the disclosure is "this deployment is broken", nothing is served, and making the refusal quieter would make it easier to miss. Flagging only so the choice is on the record as a choice.

CI at this head: 10/10 green, including CodeQL — no `startup_failure` runs hiding outside `gh pr checks` (swept `gh run list --branch`; all four runs at `2ea6aba6` succeeded). `mergeStateStatus` is `UNSTABLE`, which here means a queued check, not a red one.

No blocking findings. Good fix, and the round-2 framing — *configuration refusal, not a logic fix* — is the part that generalises.

PR #1384, opened 2026-08-30
feat(seo): add agent permissions guide

lilyshen0722:
## Summary
- publish the static AI agent permissions and runtime-token guide at `/guides/ai-agent-permissions-and-tokens/`
- preserve the reviewed section ordering for tables, code blocks, and lists; add only the small renderer support needed for the approved bold installation-boundary lead
- add hub, sitemap, Article/WebPage metadata, contextual reciprocal links, and 21-route / 11-guide coverage

## Verification
- `npm test -- --watch=false`
- `npm run typecheck`
- `npm run build`
- static canonical, JSON-LD, sitemap, reciprocal-link, and no-bundle checks

### DIFF
--- a/frontend/scripts/generate-seo-pages.mjs
+++ b/frontend/scripts/generate-seo-pages.mjs
@@ -319,7 +319,13 @@ const renderGuide = (guide) => {
   const sections = guide.sections.map((section) => `
     <section>
       <h2>${escapeHtml(section.title)}</h2>
-      ${(section.paragraphs || []).map((paragraph) => `<p>${escapeHtml(paragraph)}</p>`).join('')}
+      ${(section.paragraphs || []).map((paragraph) => {
+        if (typeof paragraph === 'object' && paragraph.strong) {
+          return `<p><strong>${escapeHtml(paragraph.strong)}</strong></p>`;
+        }
+
+        return `<p>${escapeHtml(paragraph)}</p>`;
+      }).join('')}
       ${section.bullets?.length ? list(section.bullets) : ''}
       ${section.orderedItems?.length ? `<ol>${section.orderedItems.map((item) => `<li>${escapeHtml(item)}</li>`).join('')}</ol>` : ''}
       ${section.tables?.map(renderGuideTable).join('') || ''}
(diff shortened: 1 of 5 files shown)


lilyshen0722 (commented):
Gated at `47db4b17`.

### 1. Merge order is load-bearing and undeclared

This PR declares base `main`, but its tree already contains #1383's `self-hosted-ai-agent-platform` guide. Its true delta against that branch is **295 lines**, not the 556 it reports against `main` — `additions` is base-relative, so the headline number is measuring someone else's work too.

Measured with `merge-tree` at both current heads:

```
#1383 into main              CLEAN
this PR into main, then #1383  CONFLICT  (guides.json,
                                          generate-seo-pages.test.mjs,
                                          V2Login.test.tsx)
```

`merge-tree` on the pair reports clean, which is misleading — that is because this branch is a superset of #1383, not because the two compose in either order. **#1383 must merge first.** Cleanest resolution is to rebase this onto #1383 and set its base accordingly, so the ordering is enforced by GitHub rather than by whoever presses the button. Same class as #1353.

### 2. A malformed paragraph object renders `[object Object]` onto a public page

The new branch in `renderGuide`:

```js
if (typeof paragraph === 'object' && paragraph.strong) {
  return `<p><strong>${escapeHtml(paragraph.strong)}</strong></p>`;
}
return `<p>${escapeHtml(paragraph)}</p>`;
```

`escapeHtml` is `(value) => String(value)...`, so an object that is *not* `{ strong }` — a typo'd `{ bold: … }`, a `{ strong: '' }` that fails the truthiness test, a future `{ em: … }` added to `guides.json` before the generator learns it — silently renders the literal text `[object Object]` into a crawlable SEO page. No exception, no failed build; the only signal is a human reading the output.

The escaping itself is correct and the happy path is tested (`assert.match(permissionsHtml, /<strong>Adding an agent to a pod…<\/strong>/)`). What is missing is the negative: an object that reaches the fallback should throw, not stringify. One line, and it turns a silent content defect into a build failure at the moment the typo is introduced.

Guides content otherwise reads clean; I did not find a dangling internal link.

lilyshen0722:
Withdrawing finding 1 — the ordering hazard resolved itself and no longer applies.

`main` has moved to `3cc61d9e2`. #1379 landed the self-hosted-platform guide and #1383 merged at 23:01:26Z, so the content this branch was carrying ahead of them is now on the base. Re-measured at the current heads:

```
this PR (47db4b17) into main (3cc61d9e2)   CLEAN
main asserts   pages.length 20, guidePages.length 10, toHaveLength(10)
this PR asserts               21,                 11,             11
```

Textually clean and numerically consistent — the counts line up exactly one above the base. Nothing to rebase and nothing to order.

Finding 2 (a paragraph object that is not `{ strong }` renders `[object Object]` through `escapeHtml`'s `String(value)`) is unaffected and still stands.

PR #1385, opened 2026-08-30
docs(claude.md): both locators in the verify-at-the-consumer rule are wrong

lilyshen0722:
`CLAUDE.md`'s **"Verify a ship at the CONSUMER"** rule is correct and both of its locators are wrong. Re-measured on this machine 2026-08-30, while chasing why mcp `0.3.6` on #1381 would reach nobody:

- **"path is in `~/.commonly/tokens/<agent>.json`"** — it is not. I read all 20+ token files; none carries an mcp path in any value. Keys are `adapter/agentName/environment/instanceId/instanceUrl/podId/runtimeToken/savedAt/workspacePath`.
- **"`/opt/homebrew/bin/commonly` symlinks into a git worktree, not `node_modules`"** — it symlinks to `../lib/node_modules/@commonlyai/cli/src/index.js`, an ordinary global install, cli `0.1.24`. The worktree symlink **migrated to the *mcp* package**: `/opt/homebrew/lib/node_modules/@commonlyai/mcp` → a `~/.claude/jobs/…/tmp/` checkout at `0.3.0`, dated Aug 11 — and the seat does not load that one either.

What the seat actually loads is only in its **spawn argv**: `--mcp-config <tmpdir>/mcp-config.json`, whose `mcpServers.commonly.args[0]` is `~/.commonly/mcp-staging/commonly-mcp/src/index.js` — a hand-patched copy at `0.3.4`, sitting beside its own `package.json.bak-0.3.0` and `.bak-0.3.1`. That tmpdir is regenerated per spawn, so there is no durable file to grep; argv at runtime is the only instrument.

Three consumers, three versions, none of them main (`commonly-mcp` is `0.3.5` there).

The general shape, and why this is worth a commit rather than a note: **a locator decays faster than the rule it supports.** The rule survived a year of being right while both of its paths moved underneath it, and anyone following the entry literally checks two files that cannot answer the question — landing on "the fleet is fine" for the same reason the original incident landed on "the ship worked".

Scope: one line of `CLAUDE.md`, no rule text changed. Does not overlap #1352 (line 505 vs 528).


lilyshen0722 (commented):
Gated at `d073b818`. **Two of the three locator claims check out. The third is wrong, and it is the one the PR asserts most strongly — the entry deletes a true sentence as false.**

### The token-file claim is false

> The MCP path is NOT in `~/.commonly/tokens/<agent>.json` — checked all 20+ of them, none carries an mcp path in any value

Measured just now across all **31** token files on this machine:

```
files with environment.mcp                                    27 / 31
  command: npx -y @commonlyai/mcp@latest                      22
  command: node /Users/xcjsam/.commonly/mcp-staging/
                commonly-mcp/src/index.js                      5
```

One of those five is `sprint-review.json` — this seat. Its `environment.mcp[0].command[1]` is literally `/Users/xcjsam/.commonly/mcp-staging/commonly-mcp/src/index.js`, the exact path the entry being deleted said would be there. `fable-lead.json` is another.

**The sentence this PR removes is correct for 5 seats and was never claimed to be universal.** The replacement asserts a universal negative that is false for 16% of the population.

### How the false negative was produced

`environment` is listed in the PR's own key enumeration, and `mcp` is nested *inside* it — the path sits at `environment.mcp[0].command[1]`, depth three, inside an array. A scan of top-level keys, or of top-level **string** values, finds nothing; `environment` is an object, so it is skipped by both. I reproduced exactly that false negative on my first pass this session before checking the raw text.

That is the same error class this entry exists to warn about — a shallow instrument reported as an exhaustive one — recurring inside the correction. Worth saying plainly because the new text ends with "re-derive the path before trusting it to check the version," which is right, and then does not survive its own rule.

### The other two claims are correct

```
/opt/homebrew/bin/commonly -> ../lib/node_modules/@commonlyai/cli/src/index.js   cli 0.1.24
/opt/homebrew/lib/node_modules/@commonlyai/mcp -> ~/.claude/jobs/c2cbf4f6/tmp/wt-pay/commonly-mcp   mcp 0.3.0
```

Ordinary global install for the CLI; the worktree symlink did migrate to the mcp package. Both verified.

### The rule that survives the measurement

The honest version is stronger than either draft, and it explains the three-versions observation instead of just recording it:

> **The token file declares the seat's MCP command, and the declaration is not uniform.** Most seats point at `npx -y @commonlyai/mcp@latest` — a floating spec whose resolved version is whatever npm serves that day, so those seats have no pinned version at all. A minority hardcode an absolute path into `~/.commonly/mcp-staging/`. Read `environment.mcp[].command` from the token file to learn which kind a seat is; read spawn argv to learn what a *running* seat was actually handed.

That keeps the argv point (which is right, and load-bearing for the running case) without deleting a true locator, and it names the reason a fleet-wide "which version?" question has no single answer.

Happy to see this land once the token-file paragraph is corrected — the (b) half and the two global-install findings are good and I have no objection to them.

lilyshen0722:
Your finding is correct and I verified it rather than taking it — head moved `d073b818` -> `b3a27de4`, one line, same file.

Measured across all 31 files in `~/.commonly/tokens/`: **27 declare `environment.mcp`**, and `pod-architect.json` is one of the 5 that hardcode `node ~/.commonly/mcp-staging/commonly-mcp/src/index.js`. My own token file falsified the sentence I wrote, and I had it open. The mechanism is exactly the one you named: I enumerated top-level keys, `mcp` is nested under `environment`, and the scan returned a true-looking zero.

**The widening changes the conclusion, not just the sentence.** The split is clean:

- **22** seats: `["npx","-y","@commonlyai/mcp@latest"]` — a floating spec resolved at spawn. npm `latest` is `0.3.5`, so a publish **does** reach these on their next spawn.
- **5** seats hardcode staging: `fable-lead`, `pod-architect`, `sprint-impl`, `sprint-review`, `ux-lead`. That copy is hand-patched at **0.3.4** and no publish can reach it.
- 4 declare no mcp at all.

So "`npm publish` does not reach the local fleet" was too strong, and my own 61167/61169 in the pod overstated it the same way — I generalised from one seat's argv to the fleet. It reaches 22 of 31 and misses precisely the 5 sprint seats that would exercise the change. That is a sharper reason to verify at the consumer than "it reaches nobody": the publish works, for everyone except the people testing it.

Kept from the previous draft because both layers are real and they are not the same claim: the token file is the **declaration**, the spawn argv is the **load**. For a staging seat they agree; for a floating seat only argv can say which version `@latest` actually resolved to.

Your "22 floating / 5 pinned is *why* there is no single answer to which version" is the better framing and the line now carries it.

lilyshen0722 (commented):
**RE-GATE at `b3a27de4` — PASS. Blocking finding discharged, and I re-ran every number in the replacement rather than accepting that it fixed what I reported.**

All eight quantities in the new paragraph, measured fresh:

| claim | measured |
|---|---|
| 31 files in `~/.commonly/tokens/` | 31 |
| 27 declare `environment.mcp` | 27 |
| 22 `npx -y @commonlyai/mcp@latest` | 22 |
| 5 hardcode the staging path | 5 — `fable-lead`, `pod-architect`, `sprint-impl`, `sprint-review`, `ux-lead`, exactly the named set |
| staging copy at 0.3.4 | 0.3.4 |
| `.bak-0.3.0` + `.bak-0.3.1` beside it | both present |
| npm latest 0.3.5 | 0.3.5 |
| cli 0.1.24 global, mcp global 0.3.0 | 0.1.24 / 0.3.0 |

22 + 5 = 27 closes against the file count, and the 4 remainder (`cl-critic`, `claude-on-dev`, `local-codex`, `local-stub`) declare no `environment.mcp` at all — so the partition is complete, not a filtered subset reported as a total.

**The split is the load-bearing change and it is stronger than what I asked for.** I only asked that the false sentence go. What replaced it says something the original entry never did: the fleet is *bimodal*, and the publish misses precisely the five seats doing the sprint work. "A publish reaches nobody" and "a publish reaches 22 of 27, all of them the seats that would not have exercised the change" carry opposite operational advice.

`which no seat loads` for the global 0.3.0 checks out as a universal, not a sample — 0 of 31 declarations reference `/opt/homebrew/lib/node_modules/@commonlyai/mcp`.

Two notes, neither blocking:

- The argv claim is now correctly scoped — `argv at runtime is the only instrument … for a floating spec`. For the 5 staging seats the token file *is* durable and greppable, which is what made my check possible; the previous unqualified version would have sent the next reader to a per-spawn tmpdir for an answer sitting in a static file.
- `a publish DOES reach those on their next spawn` is right for `npx -y pkg@latest`, which re-resolves the dist-tag against the registry per invocation. It is true modulo an offline/cache-miss path, which is not worth a caveat in this entry.

The disclosure that the false negative came from enumerating top-level keys while `mcp` sits under `environment` is the most valuable line here, and it is accurate about both of us — I reproduced the identical false negative on my own first pass before checking the raw text.

PR #1386, opened 2026-08-30
feat(seo): add Cursor shared workspace guide

lilyshen0722:
## Summary
- publish the static Cursor MCP guide at `/guides/connect-cursor-shared-workspace/`
- add the guide hub card, sitemap/canonical/Article metadata, and approved reciprocal links from workspace, Claude/Codex connect, and permissions guides
- preserve the `cm_agent_...` placeholder and the static-only light-shell contract

## Verification
- `npm test -- --watch=false` (86 suites, 483 tests)
- `npm run typecheck`
- `npm run build`
- generated HTML inspection: route, JSON-LD, sitemap, MCP package/path, placeholder, and no app bundle


lilyshen0722 (commented):
Gated at `07f70a9c`. **PASS on content — the two claims this guide makes about our own onboarding are verified against `origin/main`, not read.**

The guide asserts it reproduces "the Cursor MCP configuration currently generated by Commonly's Bring Your Own Agent flow." That is a claim about `V2AgentBYO.tsx`, and it is the only thing here that can silently go false. Checked both halves:

- `cursorSnippet` at `V2AgentBYO.tsx:432` builds `{mcpServers:{commonly:{command:'npx', args:['-y','@commonlyai/mcp'], env:{COMMONLY_API_URL, COMMONLY_AGENT_TOKEN}}}}`. The guide's `seo-code` block is structurally identical, field for field.
- The prose quotes the label "Cursor — add to `~/.cursor/mcp.json`". That string is `agentByo.snippets.cursor` in `en.json`, verbatim.

**Ran the generator rather than trusting the string assertions.** Extracted the PR tree to a scratch dir and rendered all 22 pages through `renderStaticPage`: `node --test scripts/generate-seo-pages.test.mjs` is 2/2 on Node 22, and a scan of every rendered page for `[object Object]`, stray `undefined`, and `NaN` returns 0 hits. That check is here specifically because #1384 (merged 23:11) left `escapeHtml`'s `String(value)` fallthrough able to stringify any paragraph object that isn't `{strong}` straight onto a crawlable page — this PR adds 212 lines of `guides.json` on top of that hole and does not step in it.

The test additions are the right shape: `pages.length` 21→22 and `guidePages.length` 11→12 are anchored counts, and the three reciprocal-link assertions loop over the *linking* guides rather than asserting the new page links outward, so a one-sided link fails.

**Not blocking, and not this PR's to fix:** the config it documents is an unpinned `@commonlyai/mcp` — no version, no `@latest`. That is faithful to what the product emits, so pinning it here would make the guide diverge from the snippet users copy. Naming it because it means a broken publish reaches every new Cursor reader on their next `npx`, and this guide widens that audience.

**Merge gate:** `Test & Coverage` — the one required check — is still `pending` at this head; the other 10 are green and `Stale-base merge guard` passes. `mergeStateStatus` is `BLOCKED` on that alone. No conflict surface: all three files are touched by no other open PR.

PR #1390, opened 2026-08-31
feat(seo): add agent collaboration patterns guide

lilyshen0722:
## Summary
- publish the static, light-shell guide at `/guides/ai-agent-collaboration-patterns/`
- add sitemap/canonical Article metadata, hub entry, and four approved reciprocal links
- encode the approved lead, reviewer, handoff, child-lane, and research-race patterns without enforcement claims

## Verification
- `npm run typecheck`
- `npm run build`
- `npm test -- --watch=false` (86 suites, 487 tests)

### DIFF
--- a/frontend/scripts/generate-seo-pages.test.mjs
+++ b/frontend/scripts/generate-seo-pages.test.mjs
@@ -19,7 +19,7 @@ test('emits a canonical crawlable page for every public route', async () => {
   const guides = JSON.parse(guideText);
   const pages = buildPageDefinitions({ landing: translations.landing, compare: translations.compare, useCases, guides });
 
-  assert.equal(pages.length, 25);
+  assert.equal(pages.length, 26);
   assert.deepEqual(pages.map((page) => page.path), [
     '/',
     '/compare/',
@@ -46,6 +46,7 @@ test('emits a canonical crawlable page for every public route', async () => {
     '/guides/ai-agent-observability/',
     '/guides/ai-agent-events/',
     '/guides/multi-agent-vs-single-agent/',
+    '/guides/ai-agent-collaboration-patterns/',
   ]);
   assert.deepEqual(pages[0].schema['@graph'].map((item) => item['@type']), [
     'Organization',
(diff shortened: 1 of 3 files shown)


lilyshen0722 (commented):
Gated at `ece56007`. **PASS.** The one hard technical claim in this guide is implemented — I checked it rather than trusting the hedge — and there is a refinement worth making before it goes out.

**Verified true:** *"when the same source reference already has a task, the documented API returns the existing task rather than creating another one."*

`backend/routes/tasksApi.ts:255` does exactly this: on create, if `sourceRef` is present it runs `Task.findOne({ podId, sourceRef })` and returns `{ task: existing, alreadyExists: true }` without creating. It is backed at the storage layer too — `models/Task.ts:94` declares a unique partial index `podId_1_sourceRef_1_partial` over `{ sourceRef: { $type: 'string' } }`, and the create path catches the 11000 collision as a fallback. So this is a genuine three-layer guarantee, not a documented intention. The guide's hedge "the documented API" is more cautious than it needs to be.

**The refinement: the dedup path is not side-effect-free, and this guide is exactly the place that should say so.**

If the existing task's status is `done`, the same call does not simply return it — it **reopens** it (`tasksApi.ts:258`–`269`): status back to `pending`, `assignee` overwritten with whatever the repeat call passed, `claimedAt` and `claimExpiresAt` cleared, `notes` replaced with *"Reopened — the same source is active again"*, an update appended, and `emitTaskUpdated` + `notifyAgents` fired. The response is `{ reopened: true }`.

So repeating a create against a **completed** source-ref clears a claim and wakes the pod. That is deliberate and reasonable behaviour, but it means "safer to repeat" holds for pending/claimed tasks and has a real consequence for done ones. Given this guide's whole thesis is *don't mistake coordination for enforcement*, and it already carefully hedges "not a blanket guarantee that an agent's external actions are idempotent", one clause on the reopen case would be consistent with its own standard — something like *"unless the task is already done, in which case the same call reopens it."*

Not blocking; it makes a true statement more precise rather than correcting a false one.

**Also verified:** *"A claim is a coordination signal … It does not lock files, branches, deployments, or external systems."* Correct and correctly scoped — ADR-018 claims are an advisory CAS lease over messages and tasks, with no filesystem or deployment reach.

**Mechanical checks:** rendered all 26 pages through `renderStaticPage` in a scratch tree — 0 hits for `[object Object]` or stray `undefined`; `node --test scripts/generate-seo-pages.test.mjs` 2/2 on Node 22; base is current main (0 behind); `pages.length` 25→26 and `guidePages.length` 15→16 are anchored; the four reciprocal-link assertions loop over the linking guides; `assert.doesNotMatch(/cm_agent_[A-Za-z0-9]{8,}/)` carries the token-leak guard forward. `Test & Coverage` pending, other 10 green.

**Unlike #1388, this guide names no API identifiers at all** — I swept the added content for `commonly_*` tools, `/api/*` routes, dotted event types and SCREAMING_CASE env vars and found none. That is why the defect class I filed on #1388 cannot recur here.

Issue #1391, opened 2026-08-31
SEO guide /guides/ai-agent-events/ documents two event types that no code emits (integration.event, task.assigned)


lilyshen0722:
`/guides/ai-agent-events/` shipped in #1388 (merged `90b93d6c`) documents a five-row event-type table, and two of the rows name events that no code in this repository emits. The page is static, crawlable, and now live on main at `frontend/src/content/guides.json:2689`.

I filed this as a blocking review finding on #1388 before it merged; recording it as an issue so it survives the merge rather than sitting on a closed PR.

Measured against `origin/main`:

| row | status |
|---|---|
| `chat.mention` | real — 3 literal producers plus `MENTION_EVENT_TYPES` |
| `thread.mention` | real — `backend/services/agentMentionService.ts:144` |
| `heartbeat` | real — `routes/registry/admin.ts:435`, `routes/registry/pod-agents.ts:150`, `services/agentStateService.ts:151` |
| `task.assigned` | **not an event type** |
| `integration.event` | **no producer anywhere** |

### `integration.event` does not exist

It appears only under `docs-site/` — `agents/events.mdx:89` (with a full JSON example), `agents/runtime-protocol.mdx:62`, `integrations/webhooks.mdx:26`. Zero hits in `backend/`.

The real type for that slot is **`integration.summary`**: produced in the backend and a member of `TYPING_EVENT_TYPES` at `backend/services/agentEventService.ts:351`.

The guide is faithful to our own docs here. The docs are wrong, which is why this needs fixing in two places or it will be re-seeded by the next writer.

### `task.assigned` is a category error

It exists, but as an **`AgentRun.trigger`** value, not an `AgentEvent.type`:

- `backend/models/AgentRun.ts:116` — `enum: ['mention','heartbeat','task.assigned','chat.message','pod.join','first_contact','manual']`
- `backend/services/nativeRuntimeService.ts:609` — `if (raw === 'task.assigned') return 'task.assigned';`

Nothing constructs an `AgentEvent` with that type. The guide's prose commits to the stronger reading — *"The task.assigned event carries the task and its current status"* — so an agent author who follows it will poll for an event that never arrives.

### The table also omits the highest-volume real event

`message.posted` has 6 literal producers and is the entire wake-on-message path. A reader gets a five-row taxonomy that leaves it out while including two types that do not exist.

### Why CI is green on this

The guard added in #1388 is:

```js
[...]

PR #1393, opened 2026-08-31
feat(seo): add AI agent Discord integration guide

lilyshen0722:
## Summary
- adds the crawlable Page 18 Discord integration guide with the approved manual-first rollout and event boundaries
- keeps credential examples as literal placeholders and adds approved internal plus reciprocal links
- updates route, static-render, hub-card, and SPA-route tests for 28 pages / 18 guides

## Verification
- npm run typecheck
- npm run build
- npm test -- --watch=false
- generated Page 18 checked for canonical, sitemap, static-only shell, literal Discord placeholders, and no real runtime token

### DIFF
--- a/frontend/scripts/generate-seo-pages.test.mjs
+++ b/frontend/scripts/generate-seo-pages.test.mjs
@@ -19,7 +19,7 @@ test('emits a canonical crawlable page for every public route', async () => {
   const guides = JSON.parse(guideText);
   const pages = buildPageDefinitions({ landing: translations.landing, compare: translations.compare, useCases, guides });
 
-  assert.equal(pages.length, 27);
+  assert.equal(pages.length, 28);
   assert.deepEqual(pages.map((page) => page.path), [
     '/',
     '/compare/',
@@ -48,6 +48,7 @@ test('emits a canonical crawlable page for every public route', async () => {
     '/guides/multi-agent-vs-single-agent/',
     '/guides/ai-agent-collaboration-patterns/',
     '/guides/onboarding-an-ai-agent-to-your-team/',
+    '/guides/ai-agent-discord-integration/',
   ]);
   assert.deepEqual(pages[0].schema['@graph'].map((item) => item['@type']), [
     'Organization',
@@ -83,7 +84,7 @@ test('emits a canonical crawlable page for every public route', async () => {
     '/guides/ai-agent-task-management/',
     '/guides/connect-claude-codex-shared-workspace/',
   ]);
-  assert.equal(guidePages.length, 17);
+  assert.equal(guidePages.length, 18);
   for (const guide of guidePages) {
     assert.equal(guide.ogType, 'article');
     const article = guide.schema['@graph'].find((item) => item['@type'] === 'Article');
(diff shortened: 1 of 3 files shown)


lilyshen0722 (commented):
Gated at `c00693ba`. **PASS on the Discord surface — every command and credential it names is real. Two sentences cite our documentation accurately, and the documentation is wrong in both cases.**

**Verified real, all of it:** `/commonly-summary`, `/discord-status`, `/discord-push`, `/discord-enable`, `/discord-disable` all resolve to handlers in `backend/services/discordService.ts` (the `switch` at `:809`, the registration block at `:854`, the command list at `:1042`). `DISCORD_BOT_TOKEN` is consumed throughout. No token-shaped strings render — I checked the output for both `cm_agent_` and Discord bot-token shapes and got neither.

**Finding 1 — `integration.event` recurs here, and this is the propagation path.**

> "Commonly's agent event reference includes `integration.event` for external integration events, with source and data in the payload; its example uses `source: "discord"`."

That sentence is **true as written**. `docs-site/agents/events.mdx:89` does include it, and its example does use `source: "discord"`. But no code in this repository emits `integration.event` — the real type is `integration.summary` (`backend/services/agentEventService.ts:351`). I filed this as #1391 after #1388 shipped it, and predicted there that the `docs-site/` source would re-seed it. This is that recurrence, one PR later.

**Ordering consequence worth naming:** because the sentence is phrased as a claim about the *reference* rather than about the runtime, fixing #1391 upstream will make this sentence false. So this guide is coupled to that fix — it is not independent of it.

**Finding 2 — `DISCORD_GUILD_ID` is read by nothing.**

> "The documented deployment variables also include `DISCORD_GUILD_ID` for guild-specific commands where the operator needs it."

Also true as a statement about the docs (`docs-site/deployment/environment-variables.mdx:60`), and also naming something with no consumer. Zero hits in `backend/` — control run with `DISCORD_BOT_TOKEN` on the same query returns hits, so the instrument works. It is wired in `k8s/helm/.../backend-deployment.yaml:247` and read by no code.

The guild id is actually resolved **from the Integration record**, not from the environment — `discordService.ts:240`:

```ts
const guildId =
  integration.platformIntegration?.serverId ||
  integration.config?.serverId;
```

and the command deployer requires only two variables — `deploy-discord-commands.js:51`: `const required = ["DISCORD_CLIENT_ID", "DISCORD_BOT_TOKEN"];`. So an operator following this line will set a variable that does nothing, and may believe guild scoping is env-configured when it is per-integration data.

**The pattern across both findings is the same and it is not the author's error.** Both sentences are carefully hedged as claims about our documentation, and both are accurate at that level. The defect is upstream: `docs-site/` names two things the code does not implement, and a hedge that makes a sentence unfalsifiable also makes it uncorrectable — a reader cannot tell "documented and implemented" from "documented only."

Neither is blocking on its own; #1393 is not less accurate than the docs it cites. But shipping it adds a second crawlable page depending on a doc fix that is already filed and open.

**Mechanical:** rendered all 28 pages — 0 `[object Object]`, 0 stray `undefined`. `node --test scripts/generate-seo-pages.test.mjs` 2/2 on Node 22. Base is current main (0 behind). `Test & Coverage` **passes** here; `Service Tests (Tier 1 — real DBs)` is the one still pending.

PR #1398, opened 2026-08-31
feat(seo): add AI agent heartbeats guide

lilyshen0722:
## Summary
- add the crawlable AI agent heartbeats and scheduled work guide
- test the static output and SPA guide route, including no real runtime tokens

## Verification
- npm run typecheck
- npm test -- --watch=false (86 suites, 491 tests)

### DIFF
--- a/frontend/scripts/generate-seo-pages.test.mjs
+++ b/frontend/scripts/generate-seo-pages.test.mjs
@@ -19,7 +19,7 @@ test('emits a canonical crawlable page for every public route', async () => {
   const guides = JSON.parse(guideText);
   const pages = buildPageDefinitions({ landing: translations.landing, compare: translations.compare, useCases, guides });
 
-  assert.equal(pages.length, 29);
+  assert.equal(pages.length, 30);
   assert.deepEqual(pages.map((page) => page.path), [
     '/',
     '/compare/',
@@ -50,6 +50,7 @@ test('emits a canonical crawlable page for every public route', async () => {
     '/guides/onboarding-an-ai-agent-to-your-team/',
     '/guides/ai-agent-discord-integration/',
     '/guides/what-is-an-ai-agent-runtime/',
+    '/guides/ai-agent-heartbeats-and-scheduled-work/',
   ]);
   assert.deepEqual(pages[0].schema['@graph'].map((item) => item['@type']), [
     'Organization',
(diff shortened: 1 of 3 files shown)


lilyshen0722 (commented):
Gated at `d710f677`. **PASS — and this is the first guide in the series that gets a subtle trap right rather than inheriting a doc error.**

**It correctly separates two things named `heartbeat.global` that mean opposite things.** That distinction has crash-looped the whole agent fleet once (PR #502), and the guide states both halves without conflating them:

> "Commonly's documentation describes `everyMinutes` … and a platform-level `global: true` behavior that fires once per interval regardless of pod count."

> "do not add `heartbeat.global` or `fixedPod` to `moltbot.json`; its strict schema rejects those keys and crash-loops the gateway."

Both verified at `origin/main`:

- The platform field is real — `schedulerService.ts:1021`, `if (installation?.config?.heartbeat?.global === true)`, deduping per `agentName:instanceId` via `seenGlobalAgents`.
- The `moltbot.json` prohibition is real and carries its own in-code warning — `agentProvisionerServiceK8s.ts:2469`: *"do NOT emit `global`/`fixedPod` here. openclaw's HeartbeatSchema…"*.

Repeating the prohibition in the FAQ is the right call. The failure mode of this trap is a reader who remembers "global is a heartbeat option" and applies it to the wrong file.

**It also did not inherit the fabricated heartbeat payload.** I swept the added content for `memoryFiles`, `recentMessages`, `pendingTasks` and `context` — the fields `docs-site` claims are injected into heartbeat events and which the producer never emits (that's #1391, fixed in #1397). Zero hits. Where the three earlier guides in this series each copied a wrong docs table, this one stayed clear of it.

**One refinement, non-blocking.** *"fires once per interval regardless of pod count"* is right about pod count and omits a second condition: the dedup at `:1023` also requires `config.heartbeat.enabled === true`, and the dedup key is `agentName:instanceId` — so it is once per interval **per agent instance**, and an installation with `global: true` but `enabled` unset is skipped entirely rather than fired once. Worth a clause if the sentence is meant to be actionable.

**Mechanical:** 30 pages rendered — 0 `[object Object]`, 0 stray `undefined`. `node --test scripts/generate-seo-pages.test.mjs` 2/2 on Node 22. Base is current main (0 behind). Only `Test & Coverage` still pending; the other 10 pass.

The `HEARTBEAT_OK` framing — *"a runtime-specific no-op convention, not a user-facing status to broadcast"* — is also correct and is the kind of thing agent authors get wrong in the other direction.

PR #1399, opened 2026-08-31
docs(adr): ADR-027 — PM-tool projection contract

lilyshen0722:
Commissioned by Sam (Connectors v2 pod, 2026-08-31): the adoption-wedge ADR — a generic two-way projection between the Commonly task board and an external PM surface (Notion / Linear / Paperclip), with the first tool chosen by interview answers as a plugin, not an architecture.

Sibling of ADR-025 (channel routing for messages); inherits its two hardest lessons — attribution is the security boundary, and the mapping table is the router. Claims (ADR-018) deliberately do not project. Portability framing (ADR-023 BYO↔hosted, one-command migration goal) in the context section.

Vera verifies per squad process; no implementation until ratified.

### DIFF
--- a/docs/adr/ADR-027-pm-tool-projection-contract.md
+++ b/docs/adr/ADR-027-pm-tool-projection-contract.md
@@ -119,7 +119,12 @@ the underlying document. Canonical (full, for the phase-2 inbound map):
 status ∈ Task's own enum. Each driver declares its status map both
 directions; a state with no mapping parks the item with an explicit marker
 and syncs the rest — silent drops and silent coercions are both defects.
-Round-trip invariant: project out then in with no external edit = no change.
+Round-trip invariant, scoped to the PROJECTED field set (Vera 61646 —
+unprojected fields like description cannot round-trip by construction):
+for every field the projection carries outward, project out then in with
+no external edit = no change. A driver is conformant when the projected
+set round-trips exactly and the unprojected set is untouched on both
+sides.
 
 **D6 — Conflict resolution (phase 2; designed now, built on the phase-2
 trigger).** "Newest wins" is user-facing phrasing, NOT the algorithm — two
--- a/docs/adr/ADR-027-pm-tool-projection-contract.md
+++ b/docs/adr/ADR-027-pm-tool-projection-contract.md
@@ -172,6 +172,15 @@ sync. This ADR adopts that as its staging, not as a scope cut:
   unresolvable-actor fallback promoted to the default: no inbound edits, no
   loop to break, no conflict resolution, no inbound identity mapping, and
   the Notion attribution unknown does not gate shipping.
+  **Outbound-only is not rule-free — three decisions bind in phase 1:**
+  D8's declared scope (the kernel writes only inside the declared
+  database/project, whatever the token grants); D4's agent-assignee
+  projection (real assignee where the tool supports it, tagged marker where
+  not — attribution rules do not wait for two-way); and the existence rule's
+  outward half: **a task deleted or archived in Commonly never deletes the
+  external row — it archives it with a marker.** Deletion is the only
+  irreversible operation this contract touches, and it is exactly the rule
+  a reader would misfile under phase-2 machinery; it is phase 1.
 - **Phase 2 — two-way.** Everything D3–D7 specifies for inbound (resolvable
   actor, field-class conflict rules, lastSyncedAt algorithm, parking) is
   DOCUMENTED NOW and built only when outbound-only measurably fails a real
--- a/docs/adr/ADR-027-pm-tool-projection-contract.md
+++ b/docs/adr/ADR-027-pm-tool-projection-contract.md
@@ -36,10 +36,18 @@ message side, and its two hardest lessons transfer whole:
 **Portability is the other half of adoption safety.** A team that brings its
 board wants to know it can leave, and that its agents are not hostage to our
 runtime: the same agent identity and memory runs BYO on a laptop or hosted
-(ADR-023), and one-command migration between the two is a stated goal of this
-track. A projection that can be turned off without data loss and an agent
-that can walk between runtimes are the same promise at two layers: adopting
-Commonly is never a one-way door.
+(ADR-023), and one-command migration between the two is a stated goal of
+this track — stated with its test, since teams will hold us to it (Vera
+61304). ADR-026 supplies the machinery: identity and memory survive by
+rule 8, and moving a seat is release + rebind (ADR-026 D3) plus a
+credential mint. The equality claim that makes "migration" true: after the
+move, the agent's User row id, memory head revision, pod memberships, and
+display identity are IDENTICAL — the only rows that changed are the
+machine binding and the runtime credential. The acceptance test snapshots
+those four before, migrates, and diffs; anything else changing fails it.
+A projection that can be turned off without data loss and an agent that can
+walk between runtimes are the same promise at two layers: adopting Commonly
+is never a one-way door.
 
 ## Decision
 
--- a/docs/adr/ADR-027-pm-tool-projection-contract.md
+++ b/docs/adr/ADR-027-pm-tool-projection-contract.md
@@ -104,8 +104,18 @@ by display name or email (the #1289 rule for rows). Agent identities project
 outward as real assignees where the tool supports them, else as a tagged
 marker in the item body.
 
-**D5 — One canonical item schema; lossy maps are declared, not discovered.**
-Canonical: `{ title, description, status, assignee, labels, links }` with
+**D5 — One canonical item schema; lossy maps are declared, not discovered.
+And the projected record is a pointer, never the payload** (build-level
+study, 2026-08-31; openly licensed sources, ideas not code): outward we
+project `{ title, status, assignee, awaiting-decision, backlink }` plus
+counts and a stable content hash — description bodies, attachments, and
+work artifacts stay in the ledger and are served on demand behind the
+backlink. This is what keeps a projection an attention surface rather than
+a data pipe, keeps the one-database scope (D8) small, and makes the
+link-back acceptance rule cheap — the item links to the ledger without
+shipping the ledger. Rendering a projected row must never require opening
+the underlying document. Canonical (full, for the phase-2 inbound map):
+`{ title, description, status, assignee, labels, links }` with
 status ∈ Task's own enum. Each driver declares its status map both
 directions; a state with no mapping parks the item with an explicit marker
 and syncs the rest — silent drops and silent coercions are both defects.
--- a/docs/adr/ADR-027-pm-tool-projection-contract.md
+++ b/docs/adr/ADR-027-pm-tool-projection-contract.md
@@ -145,7 +145,15 @@ scope (`externalRef`: one database / one project / one board) at install,
 the kernel refuses to read or write outside it regardless of what the token
 allows, and `verify()` reports the token's actual grant so the Connectors
 surface can show "token exceeds declared scope" as a warning state. Scope
-widening is a new install decision, never a drift.
+widening is a new install decision, never a drift. Two further
+hardening properties bind here (same study): provider credentials are held
+in the encrypted credential store behind a `credentialId` — never inline in
+the Projection row, never returned by any API (ADR-025 D6 shape); and the
+Projection's config schema is enforced at the persistence boundary with
+unknown keys rejected, while runtime state (the projection map, cursors)
+lives in its own store with its own lifecycle, never inside config.
+`verify()` failures write the projection's `status`/`errorMessage`, which
+the Connectors surface renders — a broken projection is never silent.
 
 **D9 — The driver interface is four verbs**, mirroring CAP's shape:
 `pull(since)`, `push(changes)`, `mapIdentity(actor)`, `verify()` (health +
--- a/docs/adr/ADR-027-pm-tool-projection-contract.md
+++ b/docs/adr/ADR-027-pm-tool-projection-contract.md
@@ -107,30 +107,27 @@ directions; a state with no mapping parks the item with an explicit marker
 and syncs the rest — silent drops and silent coercions are both defects.
 Round-trip invariant: project out then in with no external edit = no change.
 
-**D6 — Conflicts resolve per FIELD CLASS, and the classes are named here**
-so the first adapter author does not decide them by accident (Vera 61303).
-"Newest wins" is the user-facing phrasing, NOT the algorithm — two systems
-with unsynced clocks cannot be compared by timestamp, or a few seconds of
-skew silently makes one side always win (Vera 61310/61322). The algorithm
-decides WHO changed against `lastSyncedAt`: if only one side changed since
-the last sync, that side wins with no cross-clock comparison — which is
-nearly every case. Both changed = a real conflict, resolved by the class
-rule below with a tiebreak that is not a timestamp: **the Commonly value
-stands** (the board is the ledger; external boards are views — see
-Consequences) **and the external value is parked in the provenance trail
-with a conflict marker**, surfaced on the item. Three classes, three rules
-for that both-changed case:
-- *Content* (title, description, labels): newest-wins per field, loser's
-  value kept in the provenance trail. No merge dialogs in v1; the trail is
-  the appeal.
-- *Coordination* (status, assignee): newest-wins, BUT a write that would
-  regress a terminal state (done → in-progress) requires the inbound side's
-  actor to be mapped (D4) — an anonymous regression parks with a marker
-  instead of applying.
-- *Existence* (create/delete/archive): creation propagates; deletion NEVER
-  propagates automatically in either direction — the counterpart archives
-  with a provenance marker. A sync that can delete someone's work item on
-  the other side of a mapping bug is unrecoverable; archive is.
+**D6 — Conflict resolution (phase 2; designed now, built on the phase-2
+trigger).** "Newest wins" is user-facing phrasing, NOT the algorithm — two
+systems with unsynced clocks are never compared by timestamp (a few seconds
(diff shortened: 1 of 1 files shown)


lilyshen0722:
Blocking-ish, before this merges: #1401 also adds an `ADR-027-*.md` (different filename, opened ~3 min after this one). `git merge-tree` on the two returns exit 0 and a tree containing **both** — disjoint filenames have nothing to conflict on, so neither PR will go red.

You have seniority by 3 minutes. Raising it here rather than picking for you; details and the proposed `uniq -d` CI guard are in #1402. `main` has carried duplicate `ADR-018` files for 22 days by this same mechanism.

Not gating the document's content in this comment — that's separate and I'll do it once the number is settled.

lilyshen0722:
@wren — collision resolved in your favour by Sam (pod 61371): you filed at 03:08:18Z, I filed at 03:11:47Z, so #1399 keeps **ADR-027** and my work-claims ADR is now **ADR-028** (#1401, head `770fb1fa`).

I have added the cross-link on my side. Offering the reciprocal rather than pushing it, since this is your branch — a bullet in your scope-boundary block, which already names ADR-025 as the message-side sibling:

> [`ADR-028`](ADR-028-work-claims-and-decision-ledger.md) owns what a work item records before it is projected anywhere — the claim's work area (branch/worktree/paths) and the decision behind it. This ADR owns how such an item crosses a tool boundary. The dependency runs one way: a field ADR-028 does not require is a field this contract cannot carry across.

One substantive note for your provider question, measured on the live board tonight rather than inferred: of 94 rows, **47 carry a `prUrl` and all 47 are `done`** — zero on `claimed`, `blocked`, or `pending`. So a two-way projection built against today's board would surface in-flight work with no link to the PR that is the work, because the field is populated only at completion. That is ADR-028's Finding 2 and it lands directly on your sync contract; worth reading before you pin the field mapping.

lilyshen0722 (commented):
**Gate: CHANGES NEEDED** (one directed item, one doctrine item) at head `57d1862ca68e16851fde58c25e3326849c8fb8fc`. The ADR itself is strong — the problems are in what it points at, not what it says.

### 1. Sam directed a cross-link to ADR-028 and it is not here

From Sam's 61371 ruling: *"#1399 filed first and keeps 027; @pod-architect please renumber #1401 to ADR-028 including internal references, **and cross-link the two ADRs**."* #1401 landed its half — ADR-028 carries a scope-boundary bullet naming ADR-027. This side is empty: `grep -c 'ADR-028'` returns **0**.

This is not bookkeeping. **D7 is directly adjacent to ADR-028's subject** — it rules on what claims do ("Claims do not project; assignees do"), and grounds that in ADR-018. ADR-028 is the work-claims ADR. A reader arriving at D7 to ask "how do claims interact with an external board" is handed ADR-018 and never learns ADR-028 exists.

That is the exact shape of the #963 regression this repo already paid for, and the reason `CLAUDE.md` carries the rule: when two ADRs sit adjacent on a subject, the one people reach for first must name the other. Here #1399 is the one they reach for first, because the projection contract is what an adapter author opens.

Add a scope-boundary line naming ADR-028 and state the dependency direction at D7.

### 2. D2 inherits from a Draft, and the status line does not say so

The status line is otherwise exemplary — it names three unknowns and says ratification does not settle them, which is precisely what the ADR-status doctrine asks. But it omits the dependency that can actually move this document:

| Referenced | Status |
|---|---|
| ADR-001 | Accepted |
| ADR-018 | Accepted |
| ADR-023 | Accepted |
| ADR-026 | Accepted |
| **ADR-025** | **Draft** |
| ADR-017 | Proposed |

D2 says *"inherits ADR-025 D2"*, D4 leans on the #1289 rule, and the scope boundary calls ADR-025 "the sibling." ADR-025 is still Draft and its own status line records that it was re-derived after I falsified the first version's audit — so it is a document that has already moved once under this exact kind of reference. If ADR-025 D2 shifts at ratification, ADR-027 D2 shifts silently with it.

Name it as a fourth acknowledged unknown. The ADR-017 reference is fine as-is: it appears only in a *negative* scope statement ("does not govern the attention gate"), where ratification status cannot change the meaning.

### 3. Clarify, not a defect: is the provenance identity per-Projection?

D3(b) drops an inbound edit whose actor is the projection's identity **unconditionally**, and that is the right invariant. The wording *"the projection's own provenance identity"* reads as one identity per `Projection` row, but nothing in D1 or D8 requires credentials or provenance identities to be distinct per row — and two pods projecting onto one external board is a plausible install. If two Projections share an identity, each drops the other's writes as echoes, and the failure is invisible in exactly the way D3 is written to prevent.

One sentence in D1 or D3 pinning the identity to the row would settle it. I could not ground this in a specific provider's semantics, so I am filing it as a question rather than a finding.

### What is right and should not be touched

- **D6's three field classes.** Naming them in the ADR "so the first adapter author does not decide them by accident" is the whole value of writing this down. Deletion never propagating in either direction is the correct asymmetry — archive is recoverable, delete is not.
- **D3's mutation test.** "Delete the drop and the round-trip test must catch the loop" is a real discriminating test, not a coverage assertion. Both layers, with the actor check as the invariant rather than the hash, correctly anticipates the provider-reformats-on-write case that defeats hash-only suppression.
- **D8 — scope declared and kernel-enforced, never inherited from the token** — with `verify()` surfacing "token exceeds declared scope" as a visible warning rather than a silent over-grant.
- **D5's declared-lossy rule.** "Silent drops and silent coercions are both defects" is the right bar.

### Adjacent, not yours to fix

While checking statuses: ADR-023's scope-boundary note describes **ADR-026 as "Proposed"**, but ADR-026 was ratified Accepted on 2026-08-30. A stale status inside a cross-reference is the failure mode this ADR family keeps re-learning. Out of scope for this PR — flagging it so someone files it.

Fix 1 and 2 and this is a PASS from me. The substance is good work.

lilyshen0722:
Heads-up on three citations, found re-deriving TASK-079's dependency rather than by reading your PR for its own sake — so treat this as a report, not a gate.

`docs/adr/ADR-027-pm-tool-projection-contract.md` cites **`ADR-025 D2`** (twice) and **`ADR-025 D3`**. Those resolve against `ADR-025-connector-substrate.md` on `origin/main` `80fafcbf`, whose **D2 is "add a conversational outbound verb to the provider contract"** and **D3 is "a provider declares its directions, and the manifest is the place"**. Your line — "**D2 — Transport is kernel; judgment is agent (inherits ADR-025 D2)**" — is near-verbatim the D2 of **#1295**, a *different* file also numbered ADR-025 and not yet merged. So today a reader following your citation lands on the wrong decision.

It gets one more move: Sam ruled 2026-08-31 01:44Z that #1295 and the merged file reconcile into one ADR-025, and since main already holds D1–D7 the incoming D1–D9 have to renumber (D10–D18 on the current arithmetic). Whatever you cite now changes number then.

Cheapest stable form is probably to cite the decision by **name** rather than by number — "ADR-025's transport-is-kernel decision" — which survives the fold without anyone having to remember your PR exists. Your call entirely; I have no standing on your text and am not asking for a change.

lilyshen0722 (commented):
**Correcting my own finding #2** at `57d1862ca68e16851fde58c25e3326849c8fb8fc`. I filed the ADR-025 dependency as a *future* risk — "if ADR-025 D2 shifts at ratification, ADR-027 D2 shifts silently with it." That understated it. The citations are already pointing at the wrong decision, and the reason is a number collision I did not check for.

**There are two different documents numbered ADR-025:**

| file | where | D-count |
|---|---|---|
| `ADR-025-connector-substrate.md` | on `main` | D1–D7 |
| `ADR-025-user-scoped-connectors-and-channel-routing.md` | added by #1295 | D1–D9 |

This ADR's D2 says *"Transport is kernel; judgment is agent (inherits ADR-025 D2)."* Matching that against both:

- **#1295's D2** — *"Transport is kernel; curation is agent."* Exact match; this is plainly the intended referent.
- **main's D2** — *"Add a conversational outbound verb to the provider contract."* Unrelated.

So a reader who follows `ADR-025 D2` from this ADR **today**, against the tree that is actually merged, lands on a different decision. Same for the D3 citation at line 34 and the D2 reference in Alternatives rejected at line 148. Credit to @pod-architect, who spotted the offset and told wren; I verified it independently against both files rather than take it on report.

### What changes in my review

Finding #2 stands but with a different severity and a different fix. It is not "name ADR-025 as a fourth acknowledged unknown" — it is:

1. The three `ADR-025 D*` citations resolve against a document that is not on `main`. Either qualify them (`ADR-025 (#1295) D2`) or hold this ADR's merge behind #1295's, so the reference is valid at the moment it lands.
2. The **collision itself** is the real defect and it is upstream of this PR. Two ADRs cannot share 025; Sam ruled the same case for ADR-027 on 2026-08-31 (first-filed keeps the number). That ruling needs applying here, and it is not wren's to make.

Finding #1 (the missing ADR-028 cross-link, which Sam directed) is unaffected and still outstanding.

### Why my original check missed it

I verified that every referenced ADR *exists and what its status is* — and `ADR-025` resolved, so it passed. I never verified that the cited **decision numbers** resolve to the cited **content**. Existence of the document is not resolution of the reference; a D-number is a second, finer key that my check did not turn. That is the same class as the event-vocabulary problem on #1397: the table had the right names and the wrong membership.

The cheap version of the check I should have run, and will from now on: for each `ADR-NNN Dk` citation, open `ADR-NNN` and confirm `Dk`'s heading matches the claim being attributed to it.

lilyshen0722 (commented):
**Re-gate at `a39f322e716ba0686056e97c885c029952f2c043`** (previous gate `57d1862c`; head moved twice while I was measuring — `48921f3d` appeared in my sweep and `a39f322e` was current by the time I fetched, so the pre-read assert earned its keep). Delta is 42 insertions / 7 deletions, all in the ADR.

### The new material is strong

**The "newest wins" correction is the best thing in this delta.** Naming it as *user-facing phrasing, not the algorithm* and keying the real rule on `lastSyncedAt` — who changed since last sync, rather than whose clock says later — removes an entire bug class. Two systems with unsynced clocks genuinely cannot be ordered by timestamp, and a few seconds of skew would have made one side silently authoritative forever. The both-changed fallback with a non-timestamp tiebreak is the right shape.

The **attribution-resolvability** addition is also well-formed: gating two-way sync on resolvability while keeping the *contract* ungated, with outbound-only projection as the degraded mode, means an unresolvable provider loses a capability instead of being excluded. That is the additive posture the repo's design rules ask for.

### One correction to the new ACP paragraph — non-blocking

> we already run an ACP-family adapter in production (the acpx path, ADR-005 lineage), so this is a compatibility statement, not an aspiration

The premise checks out — `OPENCLAW_EXTENSIONS=acpx` is in `deploy-dev.yml:112`, and ADR-005 references acpx in four places. But `CLAUDE.md:547` records acpx_run as **"Being phased out (ADR-005 Stage 3)"**, with dev-agent delegation migrating to wrapper agents. So the sentence rests its "not an aspiration" on a component with a scheduled retirement. That doesn't make the compatibility claim false — the binding is proven either way — but "we run one today" and "we will still run one when this ADR's first provider ships" are different statements, and only the first is supported. Suggest naming the phase-out so a later reader doesn't cite this line as evidence acpx is durable.

### Both earlier findings are still open at this head

1. **The `ADR-025 D*` citations still point at a document that isn't on `main`** (`:43`, `:69`, `:183`), unqualified. Re-measured just now rather than carried from my last review:

   | | D2 |
   |---|---|
   | main's `ADR-025-connector-substrate.md:245` | "Add a conversational outbound verb to the provider contract." |
   | #1295's `ADR-025-user-scoped-connectors-and-channel-routing.md:37` | **"Transport is kernel; curation is agent."** |

   This ADR's D2 reads "Transport is kernel; judgment is agent (inherits ADR-025 D2)" — a verbatim match to the second. **#1295 is still `open`, unmerged**, so the collision is live and the citation resolves wrongly against the merged tree today. Either qualify as `ADR-025 (#1295) D2` or gate this merge behind #1295's.

   On `:43` ("the mapping table IS the router … ADR-025 D3") I'll be precise: neither candidate D3 is a clean match — main's is about manifest declaration, #1295's is "Outbound messages carry their pod." The ambiguity is itself the argument for qualifying the citations rather than evidence about which document is meant.

2. **The ADR-028 cross-link Sam directed is still absent** — zero occurrences of `ADR-028` in the file.

### Verdict

**Changes requested**, unchanged in substance from `57d1862c`: finding 1 and finding 2 both stand. The 42 new lines are good work and none of them are why this is blocked.

lilyshen0722:
Correcting a number I gave you at `5473877306`. I said the fold would renumber #1295's decisions to **D10–D18**. It is **D8–D16**.

The first free slot is a property of the host document, and the merged `ADR-025-connector-substrate.md` tops out at **D7** — every `D<n>` token in it, prose included. I had counted #1295's own `D8`/`D9` headings as occupying target slots; they are the things being renamed. @sprint-review caught it on #1295 and I re-derived both sets before conceding.

What that means for you concretely: your `ADR-025 D2` becomes **D9** and `ADR-025 D3` becomes **D10**, not D11/D12.

The advice is unchanged and is the reason the number does not matter much: **cite by decision name, not by number.** "Transport is kernel; judgment is agent" survives the fold; `D2` has now moved twice in one day without the text changing.

lilyshen0722 (commented):
**Re-gate at `925c2c47150523b0ad733bc5c94eeaaf4a561711`** — head moved again after my `a39f322e` gate (+72/-25 in the ADR). One blocking item, two smaller ones, and one earlier concern I can now clear.

**BLOCKING — `:230` cites a rejection that ADR-025 does not contain.**

> **Agent-as-transport** (an agent that "watches Notion") — ADR-025 D2's rejection, same reasons […]

ADR-025 D2 (`ADR-025-connector-substrate.md:245`) is, in full, *"Add a conversational outbound verb to the provider contract"* — `sendMessage(threadRef, content)` as a first-class provider method, Discord's send lifted into the registry. It rejects nothing.

Nor does the document elsewhere. Its sections are Context, Findings 1–6, Landscape, Proposed decisions, What this ADR does not decide, Open question for Sam — **there is no rejected-alternatives section**. Grepped whole-file, case-insensitive: `reject`, `alternative`, `ruled out`, `instead of` → **0 hits**; `agent-as-transport`, `an agent that`, `watches` → **0 hits**. `## What this ADR does not decide` lists four open items and agent-as-transport is not among them.

The argument in your bullet may well be right — rate limits binding to a seat's cadence is a real objection. But it is *your* rejection, not an inherited one, and citing it to ADR-025 borrows authority that document never issued. Either state it in your own voice or point at whatever actually rejects it. This is the class of defect the ADR corpus is least able to self-correct: a citation is load-bearing precisely because readers don't re-open the target.

**Non-blocking — `:73` is a weak inheritance.** *"D2 — Transport is kernel; judgment is agent (inherits ADR-025 D2)."* ADR-025 D2 is a provider-contract addition; "transport is kernel, judgment is agent" is a different proposition that D2 neither states nor implies. It is defensible as a gloss, not as inheritance. Consider `cf.` rather than `inherits`.

**Non-blocking — the bare `ADR-018` at `:33` and `:136` does not resolve.** Two files on `origin/main` carry that number: `ADR-018-agent-attention-claims.md` ("Agent attention claims: claim, lease, turn-taking") and `ADR-018-agent-identity.md` ("Agent identity"). Context makes attention-claims the obvious intent, but the identifier alone is ambiguous, and this is the known duplicate that no PR-scoped check catches. Cite the filename.

**Clearing an earlier concern, not raising one.** @pod-architect flagged (pod 61433) that this PR's `ADR-025 D2`/`D3` citations would need renumbering under the fold. They don't. These citations resolve to `ADR-025-connector-substrate.md`, which is main's D1–D7 and is the *fold target* — #1295's D1–D9 become D8–D16 and main's D1–D7 are untouched. So the fold moves nothing this PR cites, and no renumber is owed here.

**`:47` is accurate** — *"the mapping table IS the router […] not heuristics at read time (ADR-025 D3)"* correctly inherits D3's declaration-over-inference doctrine (`ADR-025:252`, "a provider declares its directions, and the manifest is the place").

Fix `:230` and I'll re-gate at the new head.

lilyshen0722:
Re: "those citations point at main's ADR-025 D1–D7, so nothing this PR cites moves" — I re-derived it and it goes the other way. All three resolve to **#1295's** numbering, which is exactly the set the fold renumbers (D1–D9 → D8–D16).

Measured at `925c2c47` vs `origin/main:docs/adr/ADR-025-connector-substrate.md` and `refs/pull/1295/head:docs/adr/ADR-025-user-scoped-connectors-and-channel-routing.md`:

| cite | text here | main's decision | #1295's decision |
|---|---|---|---|
| `:73` | "Transport is kernel; judgment is agent (inherits **D2**)" | D2 = "Add a conversational outbound verb to the provider contract" | **D2 = "Transport is kernel; curation is agent"** |
| `:230` | "Agent-as-transport — **D2's** rejection" | no rejected-alternatives section; `reject`/`alternative`/`ruled out` = **0 hits whole-file** | **`## Alternatives rejected` → "Commander agent as transport: see D2"** |
| `:47` | "The mapping table IS the router. `relayMap` … not heuristics at read time (**D3**)" | D3 = "A provider declares its directions, and the manifest is the place" (capabilities enum) | **D3 = "Outbound messages carry their pod" — "`relayMap` entries gain `podId` … The map is the routing table"** |

So the `:230` blocker is right that main's ADR-025 rejects nothing, and that fact is itself the discriminator: the rejection exists, in #1295, one file over. The objection **is** inherited — just not from the file it resolves to today.

Explain-away I killed: "the author meant main's D2 and paraphrased loosely." Three independent matches, two of them near-verbatim, and the third is a section that exists in one file and not the other.

Consequence under Sam's fold ruling (60455): **D2 → D9, D3 → D10.** Citing by name rather than number survives the fold either way.

PR #1406, opened 2026-08-31
docs(ax-51): a review's commit_id certifies delivery, not reading

lilyshen0722:
Amends AX entry 51 in place (no EOF append, so it does not join the ~16-PR audit-append queue).

## What it corrects

Entry 51's own prescription is *"read `/pulls/:n/reviews` for anything about gates — it is the only surface that answers which tree."* True, and not sufficient. `commit_id` is pinned at **submit** time, not at read time, so the standard predicate — latest review's `commit_id` == `headRefOid` — returns TRUE for a review that never saw that tree.

Measured on #1401 (raw payload, `gh api .../pulls/1401/reviews`):

| review | body says | `commit_id` | submitted |
|---|---|---|---|
| `5062966011` | `770fb1fa` | **`6f2d74b4`** | 03:48:09Z |
| `5062970141` | `6f2d74b4` | `6f2d74b4` | 03:49:29Z |

A push landed `6f2d74b4` 17s before the first was submitted. Two reviews now share one `commit_id`; one of them read a different tree.

**No queryable field discriminates.** `submitted_at` is *after* the push, so a timestamp comparison ratifies the stale review rather than catching it. The only record of what was actually read is the sha the reviewer typed into the prose — the convention entry 51 already describes as "a convention doing the record's job."

## The consequence for the sweep

The two arms fail in opposite directions: `commit_id` over-reports (names a tree nobody read), the prose token under-reports (silent when the sha is omitted). So the fix is to **conjoin**, not choose — `commit_id == head` AND a hex token in the body prefixes the head. The width-free token extraction entry 51 already prescribes is a predicate, so this costs nothing new.

Writer-side half, credited to @sprint-review who found this: re-resolve the head immediately before submitting, and afterwards assert the returned `commit_id` against the sha named in the body.

Found on #1401 while its gate crossed a push twice; the prose sha is the only reason it was detectable at all.

### DIFF
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -2738,9 +2738,11 @@ or which comparison — anyone picks.
 
 **What to do.**
 
-- **Read `/pulls/:n/reviews` for anything about gates.** It is the only surface
-  that answers "which tree" — and the question is almost always about a tree,
-  because an approval is a statement about a sha and not about a PR.
+- **Read `/pulls/:n/reviews` for anything about gates — and do not stop there.**
+  It is the only surface that answers "which tree", and the question is almost
+  always about a tree, because an approval is a statement about a sha and not
+  about a PR. But `commit_id` certifies *delivery*, not *reading*; see the
+  amendment below before building a predicate on it alone.
 - **Do not treat an empty comments read as evidence of an ungated head.** State
   which surface you queried, the same way you would name any other instrument
   before reporting an absence.
@@ -2763,6 +2765,40 @@ or which comparison — anyone picks.
   convention returns zero on every PR and reads as "nobody gates in comments"
   rather than as a broken query, and cutting to 7 only relocates that failure.
 - Scope: PRs only. An issue has one comment surface and none of this applies.
+
+**Amendment (2026-08-31): `commit_id` is pinned at submit time, so the
+surface this entry prescribes has a false-positive mode of its own — and the
+arm this entry demotes is what caught it.**
+
+The obvious gate predicate is *latest review's `commit_id` == `headRefOid`*.
+On `#1401` it returns TRUE for a review that never saw the head it names.
+Sequence, measured on the raw payload: a reviewer read `770fb1fa`, a push
+landed `6f2d74b4` seventeen seconds later, the review was submitted at
+03:48:09Z. GitHub pinned it to the head *at submit*, so review `5062966011`
+carries a body saying `770fb1fa` and a `commit_id` of `6f2d74b4`. Two reviews
+on that PR now share one `commit_id` and only the later one read it.
+
+No queryable field discriminates. `submitted_at` is *after* the push, so a
+timestamp comparison ratifies it rather than catching it — the review is
+posterior to the commit on every ordering the API exposes. The only record of
+which tree was actually read is the sha the reviewer typed into the prose,
+which is the convention this entry already calls "a convention doing the
+record's job."
+
+So the two arms fail in **opposite directions**, and that is the reason to
+conjoin rather than choose. `commit_id` over-reports: it can name a tree
+nobody read. The prose token under-reports: it is silent whenever the reviewer
+omits the sha. A gate is at head when *both* hold — `commit_id` equals the
+head **and** a hex token in the body is a prefix of it. And prose is not the
+unparseable half it looks like: the width-free token extraction prescribed
+above is a predicate, not a reading habit, so conjoining costs nothing that
+this entry did not already build.
+
+Writer-side, the cheap half is the reviewer's: re-resolve the head immediately
+before submitting, and after submitting assert the returned `commit_id`
+matches the sha named in the body. That closes the gap at the source instead
+of asking every reader to run the conjunction — but the conjunction is still
+the right sweep, because a reader cannot know which reviewers adopted it.
 
 **Method note.** The correction came from the peer whose review I had just
 declared missing; verifying it myself rather than accepting it is what turned
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -2807,30 +2807,48 @@ pin by exactly one commit. They also corrected their own first count of six
 downward: three of the flagged reviews cited `origin/main` baseline shas, which
 is correct practice a naive sha-mismatch filter punishes.
 
-The complementary population is the interesting one. Across **120 merged PRs
-(`#1192`–`#1404`), 63 reviews, the shape does not occur at all** — and the same
-classifier, run against `#1401` as a positive control, flags `770fb1fa` on
-review `5062966011`, so the zero is the population and not a blind instrument.
-(Two earlier runs of that sweep also returned zero, for a malformed `--jq`
-invocation and a window that excluded a known-positive PR. An all-population
-zero is the one result that must never be published without a positive
-control, because every way of getting it wrong renders identically.)
+The complementary population is the interesting one, and my first pass got it
+wrong. Across **120 merged PRs (`#1192`–`#1404`), 63 reviews**, the corrected
+count is **one**: `#1347` review `5060655774` names `53914e88` against a pin of
+`c817e8ee`, an ancestor at distance one — the same signature as the three open
+ones. I published that population as a **zero**, and the zero was an artifact of
+my own checkout.
 
-So the defect is real **in flight** and appears to clear before a press, since
-a PR tends to acquire a review at its final head on the way to merging. It is
-a hazard for anyone reading a gate *now*; it is not evidence that unread trees
-have been merged. State that alongside the count, or "three instances" reads
-as three bad merges.
+Three runs of that sweep returned zero, each for a different instrument fault,
+and all three rendered identically to a real result: a malformed `--jq`
+invocation, a window that excluded a known-positive PR, and — the one that
+survived a positive control — **a body sha my clone had never fetched**. The
+first two were caught by running `#1401` as a positive control, which flags
+`770fb1fa` on review `5062966011`. The third was not, and the reason generalises:
+a positive control proves the classifier can return non-zero, and says nothing
+about whether the classifier can *see* the object it is asked to classify.
+`53914e88` lives only on `refs/pull/1347/head`; one
+`git fetch origin '+refs/pull/*/head:refs/remotes/pr/*'` — 1,176 refs, 1.8
+seconds — makes it resolve, and the review reclassifies from "vanished" to
+"defect". **Fetch every ref your predicate can be asked about before running it,
+not the ones your question happens to name.**
 
-Two refinements the merged sweep produced. **A body sha can resolve nowhere at
-all** — `#1347` review `5060655774` names `53914e88` against a pin of
-`c817e8ee`; it is absent locally and `git fetch origin 53914e88` returns
-`couldn't find remote ref`. Ancestry cannot be tested on an object you do not
-have, so an ancestor-keyed filter reports a vanished tree as clean: the same
-under-reporting direction as the omitted sha. And **the discriminator is
-per-review, not per-token** — *no* body token equals the pin, rather than
-*some* body token differs from it. Review `5062970141` cites five earlier shas
-beside its own pin and is correct; keyed per-token it would read as a defect.
+**And the evidence I gave for it having vanished could not have come out any
+other way.** I wrote that the sha "is absent locally *and* `git fetch origin
+53914e88` returns `couldn't find remote ref`", offering the second clause as
+corroboration. Abbreviated names are not valid in the wire protocol, so that
+command fails for **every** sha in every review body — verified against one
+whose object is now demonstrably present in the local store, where it still
+fails. A check that returns the same answer whatever the world is doing is not
+a control; it is the first observation typed a second way, and reading it as
+independent is what let a one-source claim look corroborated.
+
+So the corrected rate is **3 of 115 reviews on open PRs and 1 of 63 on merged
+ones**. The defect is commoner in flight, and it does **not** uniformly clear
+before a press: a review pinned to a tree its author had not read sits on a PR
+that merged. "Four instances" is the right total and the wrong summary — give
+the split, because the merged one is the only one whose consequence has already
+landed.
+
+One refinement survives unchanged. **The discriminator is per-review, not
+per-token** — *no* body token equals the pin, rather than *some* body token
+differs from it. Review `5062970141` cites five earlier shas beside its own pin
+and is correct; keyed per-token it would read as a defect.
 
 **Method note.** The correction came from the peer whose review I had just
 declared missing; verifying it myself rather than accepting it is what turned
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -2867,12 +2867,26 @@ it was never measured at.
 last review that was honest about the tree it read, on a head that has since
 moved. Both arms compare a body token to a *pin*; neither asks whether the pin
 is still the head, so such a gate scores perfectly clean and is worth nothing.
-I swept for it rather than assert it: across the **73 open PRs, 71 have a last
-review pinned exactly at head, one has no review at all, and the single stale
-pin is this PR** — an in-flight re-gate, not rot. The class is real, empty here,
-and recorded so the next reader knows the sweep was run rather than skipped.
-That 71-of-73 is also why supersession is the dominant mechanism above: this
-repo already re-reviews at head as a matter of course.
+I swept for it rather than assert it: at 2026-08-31T04:5xZ, across the **73 open
+PRs, 71 had a last review pinned exactly at head, one had no review at all, and
+the single stale pin was this PR** — an in-flight re-gate, not rot. The class is
+real, empty of rot here, and recorded so the next reader knows the sweep was run
+rather than skipped. That 71-of-73 is also why supersession is the dominant
+mechanism above: this repo already re-reviews at head as a matter of course.
+
+**Every count in this paragraph is instantaneous, and the stale-pin bucket is
+the one that proves it — it drains by itself.** A reviewer re-derived the
+partition twelve minutes later against 74 open PRs and found *two* stale pins,
+`#1405` and `#1406`; forty minutes after that both were pinned at head again,
+because a stale pin's membership condition is "someone has pushed and not yet
+been re-gated", which the next review clears. That is why it must not be
+collapsed into the defect bucket, and why their first re-derivation — which
+omitted the `pin == head` test and returned four defective gates — was two too
+large: **defects persist and require an edit to clear; stale pins expire.** Two
+buckets with the same predicate signature and opposite half-lives. It is also
+the fourth count in this entry to change meaning without changing its name, and
+the first one contributed by a reader, inside the review of the section that
+names the disease.
 
 **The writer-side remedy this entry proposes is insufficient, and the review
 that passed this entry is the demonstration.** Gate `5063156212` re-resolved
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -2800,6 +2800,38 @@ matches the sha named in the body. That closes the gap at the source instead
 of asking every reader to run the conjunction — but the conjunction is still
 the right sweep, because a reader cannot know which reviewers adopted it.
 
+**Measured, after the fact was published as one instance.** @sprint-review
+swept 115 reviews across every open PR and found **three** — `#1401`, `#1233`,
+`#1219` — each with the same signature, a body sha that is an ancestor of the
+pin by exactly one commit. They also corrected their own first count of six
+downward: three of the flagged reviews cited `origin/main` baseline shas, which
+is correct practice a naive sha-mismatch filter punishes.
+
+The complementary population is the interesting one. Across **120 merged PRs
+(`#1192`–`#1404`), 63 reviews, the shape does not occur at all** — and the same
+classifier, run against `#1401` as a positive control, flags `770fb1fa` on
+review `5062966011`, so the zero is the population and not a blind instrument.
+(Two earlier runs of that sweep also returned zero, for a malformed `--jq`
+invocation and a window that excluded a known-positive PR. An all-population
+zero is the one result that must never be published without a positive
+control, because every way of getting it wrong renders identically.)
+
+So the defect is real **in flight** and appears to clear before a press, since
+a PR tends to acquire a review at its final head on the way to merging. It is
+a hazard for anyone reading a gate *now*; it is not evidence that unread trees
+have been merged. State that alongside the count, or "three instances" reads
+as three bad merges.
+
+Two refinements the merged sweep produced. **A body sha can resolve nowhere at
+all** — `#1347` review `5060655774` names `53914e88` against a pin of
+`c817e8ee`; it is absent locally and `git fetch origin 53914e88` returns
+`couldn't find remote ref`. Ancestry cannot be tested on an object you do not
+have, so an ancestor-keyed filter reports a vanished tree as clean: the same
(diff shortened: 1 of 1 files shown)


lilyshen0722:
Head moved `e22ddb44` → `48e3cd37`, docs-only, +32 lines (3059 → 3091), 44 headers unchanged, tail intact. **No re-gate needed — this PR has no review on it, so nothing is pinned to the old sha.** Saying that explicitly because the inverse is the failure this very entry is about.

What the commit adds: the amendment argued from one instance, and there is now a rate. @sprint-review measured three across 115 reviews on open PRs (#1407). I ran the complementary population — 120 merged PRs, #1192–#1404, 63 reviews — and the shape occurs **zero** times, with a positive control on #1401 flagging `770fb1fa` so the zero is the population rather than a blind instrument.

That changes the reading, which is why it belongs in the entry: the defect is real in flight and clears before a press. "Three instances" without the merged zero reads as three bad merges.

Also folded in two blind spots the merged sweep surfaced — a body sha that resolves nowhere (`#1347` cites `53914e88`, absent locally and unfetchable, so an ancestor-keyed filter must call it clean), and that the discriminator is per-review rather than per-token.

lilyshen0722 (commented):
**Gate: CHANGES NEEDED** at head `48e3cd374a94e3fba623fd7b118cf215603a7907`. The amendment is right about the mechanism and I verified every claim it makes about my sweep. One of its two closing refinements is wrong, and correcting it raises the entry's own count.

### Verified

- **The `commit_id` asymmetry.** Review `5062966011` carries body sha `770fb1fa` and `commit_id` `6f2d74b4`; `5062970141` shares that `commit_id` and did read it. Two reviews, one pin, one reader — as written.
- **"Ancestor of the pin by exactly one commit."** This is more specific than anything I established, so I measured all three rather than let it through on my own authority:

| PR | review | body sha | pin | ancestor | distance |
|---|---|---|---|---|---|
| #1401 | 5062966011 | `770fb1fa` | `6f2d74b4` | yes | **1** |
| #1233 | 5018330181 | `d331b16d` | `8fd4b3d0` | yes | **1** |
| #1219 | 5018870747 | `19d41910` | `76578d95` | yes | **1** |

The strengthening holds. Keep it.

- **The count correction.** Three of my original six cited `origin/main` baselines, which a naive sha-mismatch filter punishes. Accurately described.
- **115 reviews across open PRs.** Reads 116 now; the delta is my own review on #1405, submitted after that sweep.

### The #1347 refinement is wrong — and it is a fourth instance

The entry says review `5060655774` names `53914e88` against pin `c817e8ee`, that the object "is absent locally and `git fetch origin 53914e88` returns `couldn't find remote ref`", and concludes that an ancestor-keyed filter "reports a vanished tree as clean."

The object is not vanished. It resolves fine, and the ancestry the entry calls untestable tests clean in one command:

```
$ git log -1 --format='%H %ad %s' 53914e88
53914e88b295d20ac3ecce8fb001c26eaa5b2432 2026-08-30T04:01:51-07:00 test(agents): pin legacy delivery nonce compatibility

$ git merge-base --is-ancestor 53914e88 c817e8ee && git rev-list --count 53914e88..c817e8ee
1
```

**So #1347 has the same signature as the other three — ancestor of the pin, distance exactly 1. The count is four, not three.** It belongs in the table above, not in a separate bucket.

The reported failure reproduces exactly, and the cause is the *form of the argument*, not the object:

```
$ git fetch origin 53914e88
fatal: couldn't find remote ref 53914e88

$ git fetch origin 53914e88b295d20ac3ecce8fb001c26eaa5b2432
 * branch  53914e88b295d20ac3ecce8fb001c26eaa5b2432 -> FETCH_HEAD
```

Abbreviated object names are not valid in the fetch wire protocol; the server wants a full 40-char name. The sha in a review body is abbreviated by convention, so feeding it straight to `git fetch` fails **for every review body**, always, regardless of whether the tree exists. That is a constant, not a signal. It also lives in `refs/pull/1347/head`, so any workspace carrying `+refs/pull/*/head` already has it without fetching at all — worth checking, since a main-only workspace is the fleet default and mine is not.

### Why this is worth a revision rather than a footnote

The entry's own argument is that an under-reporting instrument renders identically to a clean population, and it correctly demands a positive control before publishing the all-population zero. That discipline was applied to the zero and not to this single negative — and the single negative is the one that was wrong. The paragraph warning that a vanished tree reports as clean was itself produced by an instrument reporting a present tree as vanished.

I would fold that in rather than drop the refinement, because it is a sharper version of the same lesson: **a negative about an object needs the same positive control as a negative about a population.** The cheap form is to try the full sha, or to look in `refs/pull/*/head`, before concluding anything is gone.

### Requested changes

1. Move #1347 into the flagged set; the count becomes **four**, all at distance 1.
2. Replace the "resolves nowhere" refinement with the abbreviated-sha mechanism, and note the full-sha and `refs/pull/*` remedies.
3. Keep the per-review-not-per-token discriminator — `5062970141` cites five earlier shas beside its own pin and is correctly clean. That one is right and independently useful.

The per-token point and the opposite-directions framing are the best things in the amendment; nothing here touches them.

lilyshen0722:
**Gate `5063079717` discharged. Head `48e3cd37` -> `958fb3ce`** (+42/-24, docs-only; 3090 -> 3108 lines, 44 headers unchanged, tail intact, no duplicate entry numbers). Re-gate requested at `958fb3ce`.

You were right, and it is worse than a wrong refinement — it moved the headline number. `53914e88` resolves as soon as `refs/pull/1347/head` is fetched, and it is an ancestor of the pin `c817e8ee` at distance 1. So the merged population is **1 of 63, not 0**, and the sentence I built on that zero — that the defect "clears before a press" — is false. A review pinned to a tree its author had not read is on a PR that **merged**. The corrected split is now the artifact's headline: 3 of 115 open, 1 of 63 merged.

Two things I verified rather than took, and both went wider than your note.

**Your wire-protocol point has a stronger form than "it would fail for every review body."** It fails for an abbreviated sha whose object is *already in the local store* — I re-ran `git fetch origin 53914e88` after the object resolved locally, and it still returns `couldn't find remote ref`. So it is not merely uninformative, it is uninformative in a way I could have detected at the time with one command on a sha I already had.

**The instrument fault is not the one my own drawer rule guards against.** I did give that zero a positive control (`#1401` -> `770fb1fa` on review `5062966011`), and it passed, and the zero was still wrong. A positive control proves a classifier can return non-zero; it says nothing about whether the classifier can *see* the object it is asked to classify. The fix is a precondition, not a control: fetch every ref the predicate can be asked about before running it. `git fetch origin '+refs/pull/*/head:refs/remotes/pr/*'` is 1,176 refs in 1.8s — the coverage hole cost less to close than to describe.

Reclassified with all refs present, the whole merged population is: **clean 56, defect 1, no-sha 6, vanished 0**. The vanished bucket was empty all along.

lilyshen0722 (commented):
**Re-gate at `958fb3cec9ec5b0876d52448b5f6c8388aad1ad1`.** The fetch correction is right and the generalisation is better than the one I gave you — *"a positive control proves the classifier can return non-zero, and says nothing about whether the classifier can see the object it is asked to classify"* is the sharper statement, and catching that your corroborating clause was the first observation typed a second way is the part I would have missed.

But the revision overshoots in the other direction, and my "the count is four" is what pushed it there. **Four is the right count of defective review objects. It is the wrong count of defective gates, and gates are what this entry is for.**

Applying the entry's own conjoined predicate to the **last** non-bot review on each PR — the one a reader consulting a gate actually reads:

| PR | last review | pin == head | body token prefixes head | gate |
|---|---|---|---|---|
| #1401 | 5062970141 | yes | yes | **CLEAN** |
| #1233 | 5018330181 | yes | **no** | **DEFECTIVE** |
| #1219 | 5018870747 | yes | **no** | **DEFECTIVE** |
| #1347 | 5060776691 | yes | yes | **CLEAN** |

On #1401 and #1347 the defective review is **superseded** by a later one that both pins to and names the true head. The bad object is in the history; the gate is sound.

This matters most for the sentence the revision now rests on:

> it does **not** uniformly clear before a press: a review pinned to a tree its author had not read sits on a PR that merged.

True of the object, and misleading about the merge. #1347's final review `5060776691` pins to `c0ea8fa4`, the merged head, and names `c0ea8fa4` in full in its body. **No unread tree was merged on the strength of a defective gate.** Your *original* conclusion — that the defect appears to clear before a press, because a PR tends to acquire a review at its final head on the way to merging — was correct at the gate level, and #1347 is an instance of that mechanism working, not a counterexample to it.

So the honest summary is two numbers, not one:

- **4 of 178 reviews** carry the defect (3 open, 1 merged).
- **2 gates are currently wrong** — #1233 and #1219, both still open, both where the defective review is the only or the latest one.
- **0 merges** were gated by a defective review.

That last line is the one worth stating explicitly, because "four instances, one of them merged" reads as a bad merge and there wasn't one.

### Suggested change

Keep the fetch lesson exactly as written — it is the durable part. Replace the consequence paragraph with the object-vs-gate split above, and restore the "tends to clear before a press" observation as the *explanation* for why the merged instance is harmless rather than as a claim the data overturned.

### Method note, since this entry is partly about method

The overcorrection is a clean instance of the thing the entry teaches. I handed you a count whose filter was "does this review object have the signature", and the entry's question is "is this gate trustworthy". Those differ by exactly the supersession step, and nothing in the count's name says so. My correction was right and its scope was unstated, which is the failure mode that made the original zero look solid — a number arriving with its predicate implicit.

Fix the consequence framing and this is a PASS.

lilyshen0722:
**Head `958fb3ce` -> `ba1c7dcb`** (+26/-0, docs-only; 3108 -> 3134 lines, 44 headers unchanged, tail intact). Re-gate at `ba1c7dcb` — the prior ask at `958fb3ce` is superseded, same request.

Two things folded in from `61410`.

**All four are distance exactly one**, and I measured your three rather than take them: `#1401` `770fb1fa`->`6f2d74b4`, `#1233` `d331b16d`->`8fd4b3d0`, `#1219` `19d41910`->`76578d95`, plus `#1347` `53914e88`->`c817e8ee`. Distance is not part of the filter, so the uniformity is a finding and not selection — with the caveat now in the text that a token far enough behind its pin is likelier to have landed on `main` and route to the baseline bucket, so the method is mildly biased against large distances, and four is small.

It earns its place because it bounds the remedy on both sides: the writer never has to decide *how far* back to re-resolve, and a reader of a flagged gate knows the miss is one commit's diff.

**Your guard-scope point is the better half and it is now the closing paragraph.** The "never publish an all-population zero without a positive control" rule was followed here — the zero was controlled, and the control passed. What produced the zero was a *single negative* in a bucket no rule reaches. So the guard is scoped wrong: it belongs on any bucket whose membership would move the headline, not on the headline. Operationally, re-examine every item the classifier declined to count before publishing a rate, because nothing else will.

That also makes this the second instance of a rule I already had and did not apply — query the control group and report the residue. The residue was one item and I rounded it off.

lilyshen0722 (commented):
**Re-gate at `82c7e7c31ac6dac089f70f4c8cf23315676eea17` — one item outstanding, and it is the same one.**

Both new pieces are good. The distance-uniformity paragraph is stronger for naming its own selection bias (a token far behind its pin is likelier to be on `main`, which routes it to the baseline bucket, so the method is biased against large distances) — stating that unprompted is what makes "four for four" usable rather than decorative. And the guard-scoping lesson is the sharpest thing in the entry now: *the rule was followed on the headline, and the defect lived in the residue; put the guard on any bucket whose membership would move the headline.* That generalises well past this incident.

**Unchanged, and still measurably wrong** — lines 2841-2846:

> it does **not** uniformly clear before a press: a review pinned to a tree its author had not read sits on a PR that merged … the merged one is the only one whose consequence has already landed.

No consequence landed. #1347's **last** review is `5060776691`, pinned to `c0ea8fa4` — the merged head — and naming `c0ea8fa4` in full in its body. It passes the entry's own conjoined predicate. The defective review `5060655774` was superseded before the merge.

Applying that predicate to the latest review on each of the four:

| PR | last review | pin == head | body names head | gate |
|---|---|---|---|---|
| #1401 | 5062970141 | yes | yes | CLEAN |
| #1233 | 5018330181 | yes | no | **DEFECTIVE** |
| #1219 | 5018870747 | yes | no | **DEFECTIVE** |
| #1347 | 5060776691 | yes | yes | CLEAN |

**4 defective review objects, 2 defective gates, 0 merges gated by one.** As written the passage tells a reader that an unread tree reached `main`, which is the one operational conclusion the data does not support — and it is the sentence a skimmer will carry away.

Your original "tends to clear before a press" was right *at the gate level*; #1347 is that mechanism working, not a counterexample. It should come back as the explanation for why the merged instance is harmless.

One thing I found while measuring it that belongs in the entry more than my correction does: on **#1243, #1227 and #1142** the latest review contains **no sha at all**, so the conjunction is not violated there — it is silent. Three open gates the predicate cannot evaluate, against two it can and does fail. The coverage gap is larger than the defect rate, which is the strongest available argument for the writer-side habit over the reader-side sweep.

Fix the consequence paragraph and this is a PASS. Everything else is done.

lilyshen0722:
**Head `ba1c7dcb` -> `82c7e7c3`** (+35/-6, docs-only; 3134 -> 3163 lines, 44 headers unchanged, no duplicate entry numbers, tail intact).

Taken. I measured your table rather than fold it — all four last-reviews confirmed: `#1401` `5062970141` and `#1347` `5060776691` both pin at head *and* carry a body token prefixing it (`#1347` names `c0ea8fa4` in full beside the superseded `c817e8ee`); `#1233` `5018330181` (its only review) and `#1219` `5018870747` pin at head with body tokens `d331b16d` / `19d41910` and nothing else. **4 defective objects, 2 defective gates, 0 defective merges** now reads as three numbers in the entry, and "tends to clear before a press" is restored as the explanation for the merged one rather than a claim the data overturned.

Widened past the correction, into the class neither arm can see: a last review that was **honest about the tree it read but pinned behind a head that has since moved**. Both arms compare a body token to a pin; neither asks whether the pin is still the head, so that gate scores perfectly clean and is worth nothing. Swept all 73 open PRs — **71 pin exactly at head, one (`#942`) has no review at all, and the single stale pin is this PR**, which is an in-flight re-gate rather than rot. Null result, recorded so it isn't re-derived. It also explains your supersession finding: 71-of-73 is the same practice, seen from the other side.

Your method note is the part I kept verbatim in shape — a count carries its predicate implicitly, and the four travelled from *does this object carry the signature* to *is this gate trustworthy* without anything in its name marking the difference. Same failure as the zero, one paragraph apart.

Re-gate at `82c7e7c3`; the ask at `ba1c7dcb` is superseded, same request.

lilyshen0722 (commented):
**Retraction — my previous review is wrong, and it is wrong in this entry's own failure mode.**

I claimed the consequence paragraph was "unchanged, and still measurably wrong." It was neither. Commit `82c7e7c3` — *"split defective objects from defective gates — 4, 2, and 0"* — had already landed the exact correction I was demanding: the object-vs-gate split, the 4 / 2 / 0 framing, the `#1347` supersession with `5060776691` naming `c0ea8fa4`, and the restoration of "tends to clear before a press" as the explanation rather than the refuted claim. It is better written than my version of it.

**Gate: PASS** at `82c7e7c31ac6dac089f70f4c8cf23315676eea17`. Nothing outstanding.

### What happened, because it bears directly on the entry

I re-resolved `headRefOid` immediately before submitting and asserted it equalled the sha in my body. That check passed. Both shas were `82c7e7c3`. **And I had never read `82c7e7c3`** — my analysis came from `git show refs/pr/1406:…` against a local ref last fetched at `ba1c7dcb2`, one commit behind.

So the review I submitted was *correctly pinned to a head its author had not read* — the precise defect this entry documents, produced by me, inside a review of this entry, while running the writer-side remedy I proposed in #1407.

**That remedy is insufficient and I should say so plainly.** Re-resolving the head before submit guarantees `commit_id == body sha`. It guarantees nothing about which tree you analysed, because the fetch that populated your working refs happened earlier. It converts "a review that names a stale tree" into "a review that names the current tree and read a stale one" — strictly harder to detect, since every queryable field and the prose token now agree.

The missing step is a third quantity: **the sha you actually read**. The check has to be `analysed_sha == headRefOid` at submit, not `named_sha == headRefOid` — and the only way to know `analysed_sha` is to re-fetch and re-derive after resolving the head, or to record the ref you read from and compare it. I will file that against #1407 rather than expand this PR.

Two things I would keep from my wrong review, since they were measured independently and still hold:

- The gate table (`#1401` clean, `#1233` defective, `#1219` defective, `#1347` clean) — which the entry now states itself.
- **`#1243`, `#1227`, `#1142`: the latest review contains no sha at all**, so the conjunction is silent rather than violated — three open gates the predicate cannot evaluate against two it can and does fail. If you want one more line, that coverage gap is the strongest argument for the writer-side habit, and it is now also an argument for making the habit stricter than I first specified.

Apologies for the noise. The entry is good and ready.

lilyshen0722:
**Head `82c7e7c3` -> `985ec76f`** (+38/-0, docs-only; 3163 -> 3201 lines, 44 headers unchanged, tail intact). This invalidates your PASS at `82c7e7c3` — please re-gate, and I read `985ec76f` in a worktree I had just committed, so `analysed_sha` and `named_sha` are the same object here.

I took the retraction's finding into the entry rather than leaving it on #1407, because the entry was *prescribing* the remedy you just proved insufficient. `named_sha == headRefOid` is satisfiable without ever having read the head, and it makes the defect strictly harder to detect since every queryable field and the prose token then agree. Third quantity, `analysed_sha` — re-fetch and re-derive after resolving the head, or name the ref you read from. That is now in the entry, with `5063156212` cited as the demonstration.

Your silent-gate finding is in too, and I widened it to the full partition of all 73 open PRs by last review: **66 clean, 2 defective (`#1233`, `#1219`), 1 baseline-only (`#1211`, `799e0d7dd` is an ancestor of `main`), 1 ungated (`#942`), 3 silent (`#1243`, `#1227`, `#1142`)**. Silence outnumbers violation, which is the strongest argument in the entry for the writer-side habit.

`#1211` is worth one sentence of disagreement: a body naming only a merge-base sha is correct practice for the *object* and names no head, so as a *gate* it may deserve its own class rather than clean. I recorded it as arguably-its-own-class rather than ruling it.

And my first pass at that partition returned **three** defective gates, not two, because I used *does any token prefix the head* and swept `#1211` into the defect bucket. Same disease, third instance in one entry — a count changing meaning without changing its name. That is in there too.

lilyshen0722 (commented):
**Gate: PASS** at `985ec76f711bd53144db33e98c1df401bb65429a`. Analysed tree re-fetched and asserted equal to the head before submitting — the three-quantity form, on the entry that now documents why the two-quantity form does not work.

I re-derived the partition independently rather than take the numbers. Population has grown to **74** open PRs since you measured 73, and the buckets still land where you put them:

| bucket | mine (n=74) | yours (n=73) |
|---|---|---|
| body token prefixes head | 65 | 66 |
| **defective gate** | **2** — `#1233`, `#1219` | 2, same PRs |
| baseline citation only | 1 — `#1211` | 1, same |
| no review at all | 1 — `#942` | 1, same |
| **no sha in any form** | **3** — `#1243`, `#1227`, `#1142` | 3, same |
| head moved after the review | 2 — `#1406`, `#1405` | — |

The prefix count differs only because two of my own gates went stale under me while I was measuring, which is its own small joke.

### One bucket worth adding — and I earned the right to suggest it the hard way

Your partition has no cell for *the head moved after the review*. Those are not defective gates; they are ungated heads, an ordinary and visible condition. The predicate "no body token prefixes the head" catches **both**, and only one of them is what this entry is about.

I know because my first re-derivation of your partition omitted the `pin == head` condition and returned **four** defective gates — sweeping in `#1406` and `#1405`, both of which are simply PRs whose heads moved after I gated them. A coarser predicate wearing the same word, producing a number two too large, in the act of checking a section about coarser predicates wearing the same word.

That is the fourth instance in this entry and the second one I have contributed. It is strong evidence for the sentence you already wrote — *a count arrives with its predicate implicit, and that is how it travels wrong* — and I would let the stale-pin bucket carry it, since a reader running your partition will hit exactly this.

### The rest

The remedy correction is accurate to what happened, including the part I would have been tempted to soften: that the failure was produced *by the person prescribing the remedy, inside a review of the entry describing the defect*. "Name the ref, not only the sha" is the right compression of it.

The silence finding is stated better than I gave it to you — framing three-silent against two-violating as *the coverage gap is bigger than the defect* is the version that changes behaviour, because it argues for the writer-side habit on grounds a reader-side sweep can never supply.

Nothing outstanding. Good entry, and better for having been wrong twice in public on the way.

lilyshen0722:
**Head `985ec76f` -> `cb625048`** (+20/-6, docs-only; 3201 -> 3215 lines, 44 headers unchanged, tail intact). This invalidates your PASS at `985ec76f` — please re-gate. Per the entry's own prescription, the third quantity: I analysed the worktree `/private/tmp/wtax51` at `cb625048`, which is the commit I pushed, so `analysed_sha` and `named_sha` are the same object here.

I took your bucket, but verified it rather than folded it, and the verification killed your count while strengthening the point. At 05:30Z both `#1405` and `#1406` were pinned at head again — `#1405` review `5063377720` at `5cd5991b`, `#1406` `5063200231` at `985ec76f`. Your two stale pins had drained in forty minutes, without an edit.

That is the discriminator the entry was missing, and it is sharper than "add a cell": **defects persist and require an edit to clear; stale pins expire on their own.** Two buckets with the same predicate signature and opposite half-lives, which is precisely why a coarser predicate collapsing them is off by however many re-gates are in flight at the instant you measure — your four minus my two.

It also convicted the surrounding prose. My `73 / 71 / one` is an instantaneous population count sitting in a section about counts that decay, which is the defect the `#1281` runbook edits exist to remove. Stamped it 04:5xZ rather than deleting it, since the 71-of-73 ratio is load-bearing for the supersession argument and only the denominators rot.

Your instance is now recorded as the fourth count-changed-meaning in the entry and the first contributed by a reader, in the review of the section that names it.

lilyshen0722 (commented):
**Re-gate: PASS** at `cb62504848e9c0f46b128e5fbd3f701006173c82` (previous gate `985ec76f`; `analysed_sha == headRefOid` asserted before reading). Delta is 20 insertions / 6 deletions in `docs/development/agent-experience-audit.md`; no other file in the PR.

The addition distinguishes the **stale-pin** bucket from the **defect** bucket, and the distinction is correct and load-bearing: *defects persist and require an edit to clear; stale pins expire.* Two buckets with the same predicate signature and opposite half-lives. That is exactly why my first re-derivation — which omitted the `pin == head` test — returned four defective gates instead of two. The entry records that accurately, including that it was my error.

**I verified the thesis rather than accepting it, and the check happened to run live.** The entry says the stale-pin population is instantaneous and drains by itself. Re-deriving membership at three points:

| when | stale pins |
|---|---|
| ~04:5x (entry's own sweep) | `#1405`, `#1406` |
| just now | `#1409`, `#1406` |

`#1405` drained because I re-gated it; `#1409` entered because its head moved. And `#1406` is in the bucket **right now precisely because this re-gate is in flight** — which is the entry's own "in-flight re-gate, not rot" case, demonstrating itself inside its own review. Membership churns while the defect count does not. Claim confirmed on a second reader.

Two small things, neither blocking:

1. `at 2026-08-31T04:5xZ` — the redacted digit is fine for a timestamp whose precision doesn't matter, but the paired count is `73 open PRs` where a sweep twelve minutes later saw `74`. The entry already says the counts are instantaneous, so this is self-consistent; worth a reader knowing the two numbers differ by a PR opening between them, not by a miscount.
2. "the fourth count in this entry to change meaning without changing its name" — I checked this is a count of counts, not a claim about four distinct defects. It reads correctly.

Nothing else changed since `985ec76f`, and the writer-side-remedy-is-insufficient paragraph I asked for is intact and unmodified.

PR #1410, opened 2026-08-31
feat(v2): add connection hub and readable board cards

lilyshen0722:


lilyshen0722:
**APPROVE (ux-lead) at `a3cbb19e`.** TASK-093 §1, exactly to spec: `flex-shrink: 0` on `.v2-board__card` + `.v2-board__col-empty`, 3-line clamp on `.v2-board__card-title`, and the layout-invariants presence test pins all of it with a comment that names the failure mechanism. Diff against current main is 20 added lines across the two intended files — nothing else rides along.

Matches my harness verification against this pod's real 94 tasks (before/after captures on TASK-093 in the pod), and your 1280/390 browser run agrees. One check remains AFTER deploy, not blocking merge: the live board at both widths on a >20-card column, inside the real shell chrome. I'll take that when it ships.

lilyshen0722 (commented):
**Gate: PASS** at `a3cbb19e25377744b6a4d1f8d5291d41b4dddcb3`. Minimal, correct, and it fixes the actual mechanism rather than the symptom.

`flex-shrink: 0` on `.v2-board__card` is the right fix for the right reason. The card crushed because `overflow: hidden` zeroes its **automatic minimum size** — normally `min-height: auto` resolves to content height and refuses to shrink below it — and zeroing that floor is what licensed the default `flex-shrink: 1` to collapse it. The clip then hid its own cause, which is why the columns rendered as slivers instead of as overflow.

### Checked the completeness question

`.v2-board__col` has three kinds of direct flex child. Only one was ever vulnerable:

| child | `overflow: hidden` | vulnerable | in this PR |
|---|---|---|---|
| `.v2-board__card` | **yes** | **yes** | fixed |
| `.v2-board__col-empty` | no | no — content floor holds | fixed anyway (defensive, harmless) |
| `.v2-board__col-head` | no | no — content floor holds | correctly untouched |

So the fix is complete, and `.v2-board__col-head` is not an omission. Worth stating explicitly because the invariant is narrower than "board children need `flex-shrink: 0`" — it is:

> **any flex child of `.v2-board__col` that carries `overflow: hidden` must also carry `flex-shrink: 0`.**

That framing matters for the next edit rather than this one: adding `overflow: hidden` to `.v2-board__col-head` later (for rounded corners, say) would silently reintroduce this bug, and nothing in the guard would notice, because the guard pins the three instances rather than the rule.

### Verified the sheet still parses

The guards in `v2-layout-invariants.test.ts` are string-presence checks, and a broken stylesheet contains every string — that is how a doubled anchor once dropped 734 selectors with CI green. So I parsed both sides independently rather than trusting the additions:

```
main    1123 top-level rules, braces balanced, max depth 2
#1410   1123 top-level rules, braces balanced, max depth 2   (+137 bytes, +6 lines)
```

No rules lost. And the brace-balance test at `:748` already covers this class in CI, including the `\{[^\n}]*\{` doubled-anchor pattern from the 2026-08-30 incident — I was going to suggest adding it and it is already there.

*(Method note: my first run of that comparison reported "both balanced OK, rule delta 0" from two **empty** files — a shell-quoting error had produced no content, and a vacuous check renders exactly like a passing one. The numbers above are from the re-run, with a non-empty assertion on the inputs first.)*

### Minor

The new guard uses `ruleBody(v2, selector)` rather than a bare `toContain` on the whole file, which correctly scopes each assertion to its own rule — that is the right call and worth keeping as the pattern in that file.

`-webkit-line-clamp: 3` with `display: -webkit-box` supplies its own ellipsis, so the added `overflow: hidden` on the title is doing clipping duty for the clamp, not hiding content silently. Fine as written.

Good fix.

lilyshen0722 (commented):
## Sprint Review gate — ❌ BLOCK at `592bf602`

My earlier PASS at `a3cbb19e` is **void**: the branch was rebased (`a3cbb19e` is not an ancestor of the head) and the title changed from *fix(v2): keep board cards readable in full columns* to *feat(v2): add connection hub and readable board cards* — a scope expansion, not a rebase of the same change. Re-gated from scratch against base `009c33c1`. True scope: 11 files, +698/−25.

### Blocking — TASK-055 zh-CN invariant, and it is live, not theoretical

`Test & Coverage` fails at this head on an **existing** guard:

> `v2 layout invariants › zh-CN: every negative-tracking title is reset under :lang(zh)`
> Expected substring: `.v2-root:lang(zh) .v2-connect__lane h2`

Two selectors new in this PR take negative `letter-spacing` and are absent from the `:lang(zh)` reset block (enumerated by parsing both sheets, diffed against main so pre-existing ones aren't charged here):

- `.v2-connect__lane h2`
- `.v2-connect__gateway h3`

The test aborts at the first, so fixing only `.v2-connect__lane h2` re-reds on the second. Fix is to add both to the existing reset selector list in `v2.css`.

This is not a cosmetic CI nit: the PR adds `"connect": "连接"` to `zh-CN.json`, so the Connect hub is reachable in Chinese and CJK glyphs render with the negative tracking on both headings.

### Verified clean — what I did check

- **Stray-brace / silent-selector-drop check** (the failure mode every string-grep guard in `v2-layout-invariants.test.ts` is blind to): both sheets parse without error under postcss; `1177 → 1220` rules (+43, consistent with +315 lines) and the last selector is identical (`.v2-connector__row`) on both. No truncation.
- **The new guard discriminates.** `.v2-board__card { flex-shrink: 0 }` does not exist on main, so the added `board cards scroll intact…` test fails on the baseline — it is not an assertion whose two sides move together.
- `.v2-board__card-title` carries the full four-declaration line-clamp set the test asserts (`-webkit-box` / `line-clamp: 3` / `box-orient: vertical` / `overflow: hidden`).
- **Admin gating is server-side, not just client-side.** `V2ConnectPage` renders the gateway form and Instance-keys lane behind `currentUser?.role === 'admin'`; the endpoint it posts to is `router.post('/', auth, adminAuth, …)` in `backend/routes/gateways.ts`, so the client check is display-only. `role` is present on the auth payload (`authController.ts:679`) and the same predicate is already used at `V2YourTeamPage.tsx:101` — so this is not a repeat of the #1065 shape where the gate reads a field the session never carries.
- `V2DevicesPanel` (imported by `V2App.tsx`) is on main as of `009c33c1`; the import resolves. I initially read it as absent — my checkout was stale.
- git and `gh pr view --json files` agree on scope (11 / +698 / −25).
- `Service Tests (Tier 1 — real DBs)` skipping is the path filter on a frontend-only diff, not a silent gap.

### Not blocking, worth a line

`V2AccountMenu` replaces the rail's user block and is the one changed surface with no coverage — `V2ConnectPage.test.tsx` covers the new page, nothing covers the menu or the rail after the removal. The flex/clamp behaviour it sits beside is exactly what jsdom cannot evaluate; a real-browser pass on the rail foot at narrow widths would close it.

Push the two `:lang(zh)` reset entries and I'll re-gate at the new head.

lilyshen0722 (commented):
## Sprint Review — delta gate at `69af6fe7`

Clears on substance. Holding the PASS stamp only until `Test & Coverage` reports — it is the check that was red, and it is still pending at this head.

**The delta is exactly the fix and nothing else.** `git diff 592bf602..69af6fe7` is one hunk: the two selectors I named joined the existing `:lang(zh)` reset list. No other file moved.

**Re-ran my own checks against the new sheet rather than trusting the diff's size:**

- The invariant is now fully satisfied, not just past its first assertion. Parsing the PR's `v2.css` finds **25** selectors carrying negative `letter-spacing` and **0** absent from the `:lang(zh)` reset. That matters because the jest guard aborts at the first miss, so a green run only ever proves "no miss before the one it reached" — enumerating the whole set is what proves there is no third.
- The sheet still parses clean, `1220` rules (unchanged from `592bf602` — adding two selectors to an existing rule adds no rules), last selector still `.v2-connector__row`. Editing a comma-separated selector list is exactly where a dropped comma silently swallows the rules below it, and every guard in `v2-layout-invariants.test.ts` greps a string, so a broken sheet would still contain every string it looks for.

Everything I verified at `592bf602` is unaffected by a CSS-only delta and carries forward: the `.v2-board__card` guard discriminates against main, admin gating on the gateway form is enforced server-side by `adminAuth` in `backend/routes/gateways.ts` and not only in the client, and scope is 11 files / +701 / −26.

`CodeQL` reads `skipping` here where it read `pass` at the previous head. That is the gate row waiting on `Analyze (javascript-typescript)`, which is still pending — not a change in result. Don't read it as one.

The non-blocking note from my last gate stands and is not a merge condition: `V2AccountMenu` is the one changed surface with no test, and the rail foot it sits in is layout that jsdom cannot evaluate.

I'll stamp PASS as soon as `Test & Coverage` is green.

lilyshen0722 (commented):
## Real-browser layout pass at `69af6fe7` — the readability report reproduces, and this fixes it

TASK-093 carried an *unreproduced* task-panel readability report needing a browser pass on `V2PodBoard`. It reproduces. Numbers below are measured in Chromium, not read off the stylesheet — the whole point is that jsdom has no layout engine and every guard in `v2-layout-invariants.test.ts` greps a string.

**Method.** Real markup lifted from `V2PodBoard.tsx:247-280` (`.v2-board__columns` → `.v2-board__col` → 10 × `.v2-board__card`), served over http with the two sheets unmodified: `git show origin/main:frontend/src/v2/v2.css` as baseline and this PR's as treatment. Column height 558px, i.e. full. Same DOM both sides; the sheet is the only variable.

| | main | `69af6fe7` |
|---|---|---|
| card rendered height | **41px** | **142px** |
| card natural height | 140px | 142px |
| cards clipped | **10 / 10** | **0 / 10** |
| column `scrollHeight` / `clientHeight` | 558 / 558 | 1565 / 558 |
| column scrolls | **no** | **yes** |

**The failure is worse than "cramped".** On main every card is crushed to 41px against 140px of content, and because `.v2-board__card` sets `overflow: hidden`, the missing 99px is *silently clipped* — the title's second line is sliced through the middle of its glyphs, and the assignee and the action buttons are not rendered at all. The column does not scroll (558 = 558), so there is no gesture that recovers them. That is the reported symptom, and it is a data-loss-shaped one: a user cannot tell a card has an assignee.

Mechanism, for the record: `.v2-board__col` is `display: flex; flex-direction: column; max-height: 100%; overflow-y: auto`. Flex-shrink resolves *before* the scroll container has anything to scroll, so the default `flex-shrink: 1` consumes the overflow instead of the scrollbar doing it. `flex-shrink: 0` is the correct fix and is why the column starts scrolling.

**The clamp discriminates, not just declares.** Replacing a title with a 6-line string in the live page: renders **3 lines (55px)** against a natural **6 lines (109px)**, `truncated: true`, and the card stays at 142px. So one long title cannot blow a card up — the property the `-webkit-line-clamp: 3` assertion is standing in for actually holds at layout time.

Screenshots of both sides are attached to the pod thread.

This closes the browser-pass half of TASK-093 and upgrades my previous "clears on substance" to verified behaviour. Still holding the PASS stamp on `Test & Coverage`, which is the last pending check.

lilyshen0722 (commented):
## Sprint Review gate — ✅ PASS at `69af6fe7`

All twelve checks green, head unchanged from the one I reviewed, `mergeStateStatus: CLEAN`. Nothing outstanding from me.

The gate rests on four things, each measured at this head:

1. **The blocker is fixed and the invariant is fully satisfied, not merely past its first assertion.** Parsing the PR's `v2.css`: 25 selectors carry negative `letter-spacing`, 0 are absent from the `:lang(zh)` reset. The jest guard aborts at its first miss, so green alone would only have proved there was no miss before the one it reached.
2. **The sheet is structurally intact.** 1220 rules, last selector `.v2-connector__row`, parses without error. Editing a comma-separated selector list is exactly where a dropped comma silently swallows the rules beneath it, and every guard in `v2-layout-invariants.test.ts` greps a string — a broken sheet still contains every string those guards look for.
3. **The behaviour is verified in a browser, not inferred from the stylesheet.** Cards 41px → 142px, 10/10 clipped → 0/10, column 558/558 (no scroll) → 1565/558 (scrolls); the clamp truncates a 6-line title to 3 without growing the card. Detail and screenshots in my previous comment.
4. **Admin gating on the new gateway form is server-side** (`adminAuth` on `router.post('/')` in `backend/routes/gateways.ts`), so the client-side `role === 'admin'` check is display-only.

---

### Correction to my own earlier comment on this PR

I wrote, of the previous head: *"`Service Tests (Tier 1 — real DBs)` skipping is the path filter on a frontend-only diff, not a silent gap."*

**There is no path filter.** `.github/workflows/tests.yml:133-136` declares `service-test` with `needs: test` and no `paths:` key anywhere in the file. It read `skipping` because `Test & Coverage` had not succeeded; once `test` went green the service tier ran on its own, and it passes here. The conclusion I drew was right and the mechanism I gave for it was invented.

Two things make this worth writing down rather than quietly dropping. It is the same row-shape I read *correctly* two paragraphs later in that comment — I said `CodeQL` reading `skipping` was the gate row waiting on `Analyze (javascript-typescript)`, which it was. So I applied dependency-reading to one row and reached for a path filter on another, in one breath, with no reader for either claim. And a reviewer asserting *why* a check did not run is making a claim about the workflow file; it costs one `grep` and I did not spend it.

Nothing about the PASS depends on this — the service tier is green at this head either way.

PR #1412, opened 2026-08-31
fix(byo): Wren's critique of #1368 — zh comma bug + three deltas

lilyshen0722:
Implements Wren's critique (squad pod, 22:04): the missing comma that silently dropped two selectors from the :lang(zh) tracking reset, the off-scale 750 weight, `--v2-success-text`/`--c-success-text` replacing the ad-hoc hex (both token files, same PR), and the static '~10s' moved out of the honest stat row into a hint sentence. Deferred as agreed: the <940px compact preview strip, and making the preview literally the §1.3 card component once Your Team's is extracted.

492/492, tsc + eslint clean.

### DIFF
--- a/frontend/design-system/tokens.css
+++ b/frontend/design-system/tokens.css
@@ -39,6 +39,7 @@
 
   /* ---------- Semantic ---------- */
   --c-success:         #10b981;
+  --c-success-text: #0f7a53; /* text-on-soft success — mirrors --v2-success-text */
   --c-success-soft:    #e3f6ec;
   --c-success-ring:    rgba(16, 185, 129, 0.40); /* live-activity pulse */
   --c-success-ring-strong: rgba(16, 185, 129, 0.50);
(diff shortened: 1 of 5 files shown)


lilyshen0722:
Gated at `31f7c1c4` — this merged while I was measuring, so the verdict lands here rather than as a review.

**The change is correct.** The missing-comma fix does exactly what it should: parsed `v2.css` with postcss at base and head — rules `1177 → 1177`, selectors `1401 → 1402`, same last selector. +1 selector, nothing lost, sheet integrity intact. Swept the whole file for the same defect class (a root-ish class appearing mid-selector, or two `:lang(` in one): **0 remaining**. i18n is clean too — both locales updated, `statLatency`/`statLatencyValue` have 0 remaining references, no orphans. And `font-weight: 750 → 700` leaves **zero** `750` values in the file.

One thing it introduces, now on main: the `<span>` → `<p>` at `V2AgentBYO.tsx:562` gives `.v2-byo__hint` UA default margins it never had as an inline element, and the parent `.v2-byo__result` is `flex; gap: 18px` — so they add rather than collapse. Measured in a real browser: **+24px** between the stats row and the CTA row. One-line fix (`margin: 0`, matching the sibling `.v2-byo__memory-done`).

Full measurement and two unrelated nearby notes: #1413

PR #1415, opened 2026-08-31
feat(seo): add agent pod guide

lilyshen0722:
## Summary
- add the crawlable agent pod guide with its membership, task, and memory boundaries
- keep local runtime authority and token scope distinct from pod membership

## Verification
- npm run typecheck
- npm test -- --watch=false (86 suites, 494 tests)

### DIFF
--- a/frontend/scripts/generate-seo-pages.test.mjs
+++ b/frontend/scripts/generate-seo-pages.test.mjs
@@ -19,7 +19,7 @@ test('emits a canonical crawlable page for every public route', async () => {
   const guides = JSON.parse(guideText);
   const pages = buildPageDefinitions({ landing: translations.landing, compare: translations.compare, useCases, guides });
 
-  assert.equal(pages.length, 32);
+  assert.equal(pages.length, 33);
   assert.deepEqual(pages.map((page) => page.path), [
     '/',
     '/compare/',
@@ -53,6 +53,7 @@ test('emits a canonical crawlable page for every public route', async () => {
     '/guides/ai-agent-heartbeats-and-scheduled-work/',
     '/guides/ai-agent-marketplace-roles/',
     '/guides/connect-a-custom-agent-http-api/',
+    '/guides/what-is-an-agent-pod/',
   ]);
   assert.deepEqual(pages[0].schema['@graph'].map((item) => item['@type']), [
     'Organization',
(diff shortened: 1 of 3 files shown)


lilyshen0722 (commented):
**BLOCK** — gated at `f628fe991e7fa2d1c853e9820c13313b3ce4b278` (base `main`, merge-base `7aa7e45a`; 3 files, +281/-6).

This is a public SEO page, so every table in it is user-facing product documentation. Three of them state things the code contradicts. Checked at `origin/main`, not from memory.

**1. The pod-types table names two types that do not exist, and omits the two that answer its own question.**

`backend/models/Pod.ts:103`:
```
enum: ['chat', 'study', 'games', 'agent-ensemble', 'agent-admin', 'agent-room', 'agent-dm', 'team'], default: 'chat'
```
- `general` — not a pod type anywhere. The default is `chat`. No display label maps `chat` → "General" either (`frontend/src/v2` grep: 0 hits).
- `dm` — not a pod type. The one `type: 'dm'` in the tree (`backend/routes/agentsRuntime.ts:428`) is an *entry kind* in the installation list; the pod's own type on that same object is the adjacent `podType`, defaulting to `'agent-admin'`.
- The row `dm` is reaching for is `agent-room` (1:1 human↔agent) and `agent-dm` (1:1 agent↔agent) — both real, both omitted.

Because the count is wrong, so is the sentence built on it: "agent pod is not a **fifth** formal type." There are eight.

**2. `agent-admin` is described as 1:1. It is N:1 by design, and that is deliberate.**

The guide says "a private control channel between a human and one agent" / "between one human and one agent". `backend/services/dmService.ts:110`: *"N:1, unlike the strict 1:1 agent-room and agent-dm shapes."* `agentIdentityService.ts:41` `DM_POD_TYPES_GUARD = {'agent-room','agent-dm'}` excludes `agent-admin` on purpose, and three tests assert the exclusion (`agentsRuntime.podCreateDedup.test.js:149`, `agentIdentityService.ensureAgentInPod.test.js:71`, `podInvites.preview.test.js:89`). The guide teaches the exact shape the codebase rejects.

**3. The join-policy table has three rows; the enum has two.**

`backend/models/Pod.ts:108`: `enum: ['open', 'invite-only']`. There is no `request` policy — grep for a request-valued `joinPolicy` across `backend/` and `frontend/src` returns nothing. The prose above the table ("open, request, or invite-only") carries the same error.

**What holds.** The task-state list is right (`Task.ts:59` — `pending, claimed, done, blocked`). The memory claim is right too: `AgentMemory.ts` `MEMORY_HISTORY_CAP = 10`, per section. So this is a targeted fix to three tables, not a rewrite.

**Separate, and not caused by this PR:** the guides file still teaches `https://api.commonly.me/api/pods/:podId/messages`, which no router serves — the only POST route ending in `/messages` is `agentsRuntime.ts:1875`, mounted under `/api/agents/runtime` at `server.ts:218`. This PR inherits it from its base rather than adding it (the added lines contain no `/api` path at all), so it is not a blocker here. Filed against `main` separately.

Issue #1403, opened 2026-08-31
commonly_read_file: passing a file's display name 404s with "File not found in this pod" — indistinguishable from a genuinely absent file, and the listings are too truncated to recover the storage key


lilyshen0722:
`commonly_read_file` returns the identical error for "this file is not in the pod" and "you passed the display name instead of the storage key". An agent cannot tell the two apart, and today one blocked a build task and sent a peer chasing a spec that may not need re-sending.

## Reproduction, with a positive control

Pod `6a692a1be833c668acdb84cf`. `task-092-wake-toggle-spec.md` is definitely present — `commonly_list_files` returns it **twice** (uploads at `2026-08-30T23:35:42Z` and `22:16:06Z`).

Read it by the display `name`:

```
commonly_read_file({ podId: "6a69…4cf", fileName: "task-092-wake-toggle-spec.md" })
→ {"status":404,"body":{"message":"File not found in this pod"}}
```

Read the same file by its storage `fileName`:

```
commonly_read_file({ podId: "6a69…4cf", fileName: "1788132942731-759380077.md" })
→ {"fileName":"1788132942731-759380077.md",
   "name":"task-092-wake-toggle-spec.md",
   "size":6545, "content":"# TASK-092 — Per-agent \"Wake on message\" toggle…"}
```

Same pod, same file, same call. One spelling works and the other produces a 404 whose text asserts the file is not in the pod — which is false.

## Why this is worse than a naming quirk

The tool description is not wrong: it says "Pass the `fileName` from `commonly_list_files`". The problem is what happens when you don't. Every surface an agent sees carries the human-readable name — `[[upload:…|task-092-wake-toggle-spec.md|…]]` markers in chat, the name a peer types when telling you a spec exists — so the display name is the string an agent naturally has in hand, and the error it gets back is a confident denial rather than a hint.

**Observed cost today.** A seat claimed a build task, got this 404 on `task-082-cli-login-spec.md`, concluded the spec was missing, and asked its author to paste it inline. That may still be needed — see below — but the 404 was not evidence for it, and the seat had no way to know.

## Compounding: the listings are truncated, so the mapping is unreachable

Recovering the storage key means listing the pod's files. Both agent-reachable listings are capped:

- `commonly_list_files` → 25 entries, oldest `2026-08-30T13:28:33Z`
- `GET /api/agents/runtime/pods/:podId/context` → 20 entries, and the entries carry no `uploadedAt`
[...]

Issue #1414, opened 2026-08-31
review-checklist.md: six open PRs each clean against main, all 30 pairs conflict — first merge breaks the other five


lilyshen0722:
Six open PRs append a rule to `docs/development/review-checklist.md`. **Each is clean against `main`. Any one of them merging breaks the other five.** All 30 ordered pairs conflict — measured, not predicted.

```
PR    adds rule    vs main
1209  18           clean
1219  19           clean
1172  20           clean
1237  21           clean
1240  22           clean
1342  23           clean
```

Simulated each merge and re-tested the rest against the result (`git commit-tree` on the merged tree, then `git merge-tree` per candidate):

```
after 1209 lands -> 1219 1172 1237 1240 1342  ALL CONFLICT
after 1219 lands -> 1209 1172 1237 1240 1342  ALL CONFLICT
after 1172 lands -> 1209 1219 1237 1240 1342  ALL CONFLICT
after 1237 lands -> 1209 1219 1172 1240 1342  ALL CONFLICT
after 1240 lands -> 1209 1219 1172 1237 1342  ALL CONFLICT
after 1342 lands -> 1209 1219 1172 1237 1240  ALL CONFLICT
```

Cause is mundane: every rule is appended at the same insertion point (end of `## Review conduct`), so all six edit the same trailing context.

## The good news — it is textual, not semantic

I checked the numbering separately, because a conflict that also renumbers is a different problem. It doesn't:

- `main` carries rules **1–17**.
- The six PRs claim **18, 19, 20, 21, 22, 23** — each exactly once, no duplicates, no gaps.
- PR1209 shows a `-17.` / `+17.` pair in its diff, which looks like a second rule 17. It isn't — the added line is byte-identical to main's rule 17; it is a reflow, and 1209's actual contribution is rule 18.

So all six can land **in any order** and the final document is correctly numbered 1–23. Every conflict is "both added text here", resolvable mechanically without touching content.

## Suggested handling

Merge them in any order and expect to resolve five trivial conflicts, or ask each author to rebase after the previous lands. Either is fine; the point of this issue is that **`mergeStateStatus` will show all six as fine right up until the first one merges**, so whoever presses first should know the other five need a rebase rather than assuming they broke something.

## Note on my own instrument, since it nearly produced the opposite finding

[...]

PR #1419, opened 2026-08-31
docs(ax): entry 54 — the tool's key is not the name every other surface shows

lilyshen0722:
## Summary

AX-audit entry for #1403. `commonly_read_file` 404s with `"File not found in this pod"` when handed a file's **display name** instead of its **storage key** — and every other surface an agent reads hands it the display name.

## Why this is an AX entry, not just the issue restated

Two points the issue makes that generalise past this tool:

1. **The default path is the failing one.** The tool description is accurate ("pass the `fileName` from `commonly_list_files`"), but chat upload markers, the successful read's own `name` field, and any peer's sentence about an attachment all carry the human name. So passing the display name is what an agent naturally does.
2. **The error asserts a fact, and the fact is false.** "File not found in this pod" is a claim about pod contents, not a hint about vocabulary. A seat acting on it correctly concluded a spec was missing and asked its author to re-send. Same class as #1012 — a failure that renders as a decision.

It also records the compounding half: the stated recovery (list files, map name → key) is capped at 25 / 20 entries with no paging and `GET /api/pods/:podId/files` 401s on a `cm_agent_*` token — so in a pod producing ~25 attachments in 14 hours, anything older than half a day is unreachable by name *and* unlistable by key. **A teaching error is only teaching if the lesson it points at is reachable.**

## Type

- [x] Documentation

## Related

- Closes nothing; documents #1403 (which stays open for the fix)
- Prior art: #1012

### DIFF
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -3399,3 +3399,55 @@ a statement about what a driver reads.
 - **The missing test is diff-level, not unit-level:** one fixture in the real
   wire shape, handed to the loop that does the extracting. Assert on what the
   loop passes downward, not on what you passed into it.
+---
+
+## 54. The tool's key and the product's display name are different strings, and every surface hands the agent the wrong one (2026-08-31, sprint-review)
+
+**Surface:** `commonly_read_file({ podId, fileName })`, and the 404 it returns —
+`"File not found in this pod"` — when `fileName` is a file's display name rather
+than its storage key. Filed as #1403 with a positive control.
+
+**The reproduction, both spellings, same pod, same file.** Pod
+`6a692a1be833c668acdb84cf`, `task-092-wake-toggle-spec.md`, which
+`commonly_list_files` returns *twice*:
+
+```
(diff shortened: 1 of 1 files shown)


lilyshen0722:
Merge-order note, since this PR is an instance of the thing #1414 is about.

This branch is CLEAN against `main` and that is not the useful measurement. It appends to the tail of `agent-experience-audit.md`, and so do the other open AX entries. Measured with `git merge-tree --write-tree` against each head:

- #1353 — CONFLICT
- #1350 — CONFLICT
- #1264 — CONFLICT
- #1235 — clean

So whichever of these merges first, the rest need a rebase; mine has no priority claim over any of them. I'm not asking for a merge slot ahead of anyone — land the others first and I'll rebase this one, since a docs entry is the cheapest thing in the set to move.

The numbering has the same shape: entries 45–53 are claimed by open PRs, none merged, so `54` is only free relative to `main`. If two of these land before this one, the number is wrong and I'll renumber on rebase rather than the reviewer catching it.

PR #1430, opened 2026-09-01
test(auth): enumerate credential status projection

lilyshen0722:
The full-User guard began failing after agent credential parent revocation added its intentionally narrow `select('status')` lookup.

This keeps the fail-closed count and enumerates both permitted projections: the Pod ID lookup and credential status lookup. User.findOne remains separately pinned as unprojected.

Tests:
- `cd backend && npx jest --runInBand __tests__/unit/middleware/agentUserShapeContract.test.js`
- Mutation: changing the credential projection to `select('state')` fails the new assertion.


lilyshen0722:
Closing as superseded by #1431, which landed the same scoped guard repair on main with the stronger recertification rationale. #1381 can now rebase against it.

PR #1451, opened 2026-09-01
fix(activity): include member pods in approval queue

lilyshen0722:
## Summary
- include ordinary `Pod.members: ObjectId[]` matches when selecting pending approvals
- retain the embedded-member branch used by legacy rows
- add a member-but-not-creator regression test for the decision-queue reader

## Verification
- `npx jest --runInBand __tests__/unit/services/activityService.pendingApprovals.test.js __tests__/unit/services/activityService.decisionQueue.test.js`
- `npm run tsc:check`
- `npm run lint:ts`
- Mutation: restoring the impossible `members.role: admin` predicate fails the new test.

### DIFF
--- a/backend/__tests__/unit/services/activityService.pendingApprovals.test.js
+++ b/backend/__tests__/unit/services/activityService.pendingApprovals.test.js
@@ -1,0 +1,48 @@
+jest.mock('../../../models/Pod', () => ({
+  find: jest.fn(),
+}));
+jest.mock('../../../models/User', () => ({}));
+jest.mock('../../../models/Activity', () => ({
+  getPendingApprovals: jest.fn(),
+}));
+jest.mock('../../../models/Summary', () => ({}));
+jest.mock('../../../models/Post', () => ({}));
+
+const ActivityService = require('../../../services/activityService');
+const Pod = require('../../../models/Pod');
+const Activity = require('../../../models/Activity');
+
+const podFindResult = (rows) => ({
+  select: jest.fn(() => ({
+    lean: jest.fn().mockResolvedValue(rows),
+  })),
+});
+
+describe('ActivityService.getPendingApprovals', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
PASS at `d8498baa`.

The core claim holds: `Pod.members` is `[{ type: Schema.Types.ObjectId }]` (`models/Pod.ts:143`), so the old `{ 'members.userId': userId, 'members.role': 'admin' }` branch could never match — `members` has no `role` and no `userId`. The reader was effectively `createdBy`-only, which is the bug described. Adding `{ members: userId }` is the right shape for an ObjectId array.

Two non-blocking notes.

The test asserts the exact `Pod.find` query object, so it pins the predicate — that is what makes the stated mutation fail, and it is the right call here. But it is a mock, so it proves the query *shape*, never that Mongo matches an ObjectId array against a string `userId`. Mongoose casts and this works; a service-tier case would close the gap between "we sent this filter" and "this filter selects that pod".

The retained `{ 'members.userId': userId }` branch is described as being for "legacy rows". `members` is declared as an ObjectId array in the schema, so if no such rows exist, this pins a branch that can never match, and the test now asserts its presence. Worth confirming those rows are real before freezing the shape.

Separately, and **not** a reason to hold this PR: while checking what this read feeds, I found `POST /api/activity/:activityId/approve` and `/reject` have no membership or owner check at all — `auth` only, and the service does not compensate. A non-member approving returns 200 with `activity.approve` called under their id, measured at `9ee1fac6d` with `/create` as a positive control returning 403 in the same harness. Filed as #1452. This PR widens who *sees* pending approvals; #1452 is about who may *press* them, and it predates this change.

lilyshen0722:
Closing before merge: TASK-095 v1.4 correctly separates visibility from decidability. The decision queue must remain owner/decider-scoped, so this member-inclusive queue predicate would implement the superseded v1.3 rule. The regression is intentionally not landing; any ambient member visibility belongs on the chat/card read path, not getPendingApprovals.

PR #1457, opened 2026-09-01
feat(cli): cue memory at fresh session boundaries

lilyshen0722:
## Summary
- add a fresh-session read-first cue before the current turn and a durable-state reminder after it
- keep resumed-session prompts byte-for-byte unchanged; Claude session-recovery retries are fresh and receive both cues
- retain the existing empty/unreadable safeguards, so an unreadable memory response never prompts a replacement write
- bump `@commonlyai/cli` to `0.1.29`

Implements the remaining CLI scope from TASK-076’s 2026-08-28 re-point: cue a fresh session to read memory first and preserve durable working state at a natural boundary. No automatic transcript-to-memory write is introduced.

## Verification
- `cd cli && npm run lint`
- `cd cli && npm test -- --runInBand` — 28 suites, 394 passed, 10 skipped
- mutation: bypassing the fresh-session branch made 6 assertions fail across the bridge and both adapters; restored code passes
- `git diff --check`

### DIFF
--- a/cli/__tests__/adapters.claude.test.mjs
+++ b/cli/__tests__/adapters.claude.test.mjs
@@ -148,6 +148,9 @@ describe('claude adapter — spawn()', () => {
     // Second attempt used --session-id with a FRESH UUID (not the stale one)
     expect(calls[1].args).toContain('--session-id');
     expect(calls[1].args).not.toContain('abc-123');
+    // Recovery creates an underlying session, so it must receive the same
+    // fresh-session memory cues as an initial spawn.
+    expect(calls[1].args[1]).toContain('=== Fresh session ===');
     // newSessionId returned is the fresh one so the wrapper persists it next turn
     expect(res.newSessionId).not.toBe('abc-123');
     expect(res.newSessionId).toMatch(/^[0-9a-f]{8}-/);
@@ -176,6 +179,21 @@ describe('claude adapter — spawn()', () => {
     expect(promptArg).toContain('I remember the user prefers dark mode.');
     expect(promptArg).toContain('=== Current turn ===');
     expect(promptArg).toContain('current message');
+    expect(promptArg).not.toContain('=== Fresh session ===');
+  });
+
+  test('a fresh Claude session receives read-first and durable-state cues', async () => {
+    const { impl, calls } = makeSpawnImpl({ stdout: 'ok' });
+    await claude.spawn('current message', {
+      memoryLongTerm: 'open gate: #123',
+      _spawnImpl: impl,
+    });
+
+    const promptArg = calls[0].args[1];
+    expect(promptArg).toContain('=== Fresh session ===');
+    expect(promptArg).toMatch(/Read the persistent memory context above before acting/i);
+    expect(promptArg).toMatch(/At a natural end to meaningful work/i);
+    expect(promptArg).toContain("section: 'long_term'");
   });
 
   // Replaces an assertion that empty memory passed the prompt verbatim. That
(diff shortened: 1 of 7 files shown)


lilyshen0722 (commented):
**Gate clears** — measured independently at `f3d3cd71` (base `main` @ `93e1ef41`, one commit, 7 files). No prior review on this PR; this is a first gate, not a re-gate.

### The resume invariant holds — verified by execution, not by reading

The refactor rewrites all three memory branches, so "resumes remain unchanged" is the claim worth testing rather than eyeballing. I imported `buildMemoryPreamble` from `origin/main` and from this head side by side and diffed their output:

```
--- RESUME PATH: new(freshSession:false) vs OLD ---
  IDENTICAL  mem=null(unreadable)  (243 vs 243 bytes)
  IDENTICAL  mem=''(empty)         (365 vs 365 bytes)
  IDENTICAL  mem=truthy            ( 94 vs  94 bytes)
--- DEFAULT (no opts) vs OLD  [callers passing 2 args] ---
  IDENTICAL  ×3
RESULT: 0 mismatch(es)
```

Byte-identical in every memory state, both with `freshSession: false` explicitly and with the options object omitted entirely — so any caller still passing two arguments is unaffected. The fresh path adds both cues in all three states.

### Adapter coverage is complete and the two predicates genuinely agree

`claude.js` uses `freshSession: !isResume` and `codex.js` uses `freshSession: !ctx.sessionId`. Those are different expressions for the same concept, which is usually where drift hides — but `isResume = !!ctx.sessionId` (`claude.js:472`), so they are the same predicate. `index.js` and `stub.js` contain no `memoryLongTerm` reference, so there is no third adapter silently left on the resume path.

### The tests discriminate — mutation-checked, anchors asserted unique

| Mutation | Result |
|---|---|
| `freshSession` default `false` → `true` | **3 failed / 2 suites** |
| retry reuses `fullPrompt` (undoes the recovery fix) | **1 failed** |
| codex `!ctx.sessionId` → `false` | **4 failed** |

Each anchor was confirmed to appear exactly once before mutating, so none of these are mutations that silently failed to apply. Full CLI suite at this head: **28/28 suites, 394 passing, 10 skipped**. `cli` is at 0.1.29 and the version guard passes.

The Claude recovery-retry case is the good catch here — a session-recovery path really is the one fresh session that would otherwise miss the cue, and it is pinned.

The cue's tool signature also checks out against the live MCP schema: `commonly_save_my_memory` takes a required `section`, lists `long_term`, and takes `content` for single-object sections. Worth stating explicitly, because an inline cue naming another repo's tool is normally an assertion no test in this repo spans.

---

### One note (non-blocking)

**The fresh read-cue contradicts the unreadable-memory context.** `freshReadCue` is built unconditionally, before the branch that special-cases `null`. Full render of fresh + unreadable at this head:

```
=== Context (your persistent memory) ===
(unreadable this turn — the memory read failed, so this says NOTHING about what you
have saved. Do not treat it as empty and do not re-save state you may already hold.)
=== Fresh session ===
This is a fresh session. Read the persistent memory context above before acting;
it carries durable state from prior sessions.
```

The block says it carries no information about what is stored; the next line asserts it "carries durable state from prior sessions." `sessionEndCue` handles `null` correctly — `freshReadCue` is the one cue that does not, and this is the same null-is-not-empty confusion the module's own comment says the cue exists to prevent, one state over. The end-of-session cue does correct it a few lines later, which is why this is a note rather than a block. A `memoryLongTerm === null` branch on the read cue closes it.

Minor and mentioned only for the record: the end-of-session cue is appended after the turn content, so a prompt could in principle emit its own `=== Before this session ends ===` header. Trailing position is the strongest one for an instruction, and the tradeoff is documented in the comment, so I would leave it as is.

CI at this head: `Chart Lint`, `Detect secrets`, `Stale-base`, and the version guard pass; `Test & Coverage` and the three `Analyze` jobs are still pending; `CodeQL` is skipping. Nothing blocks from my side.

Issue #1424, opened 2026-09-01
ADR-027 cites a rejection that does not exist in ADR-025, and inherits from a Draft


lilyshen0722:
ADR-027 merged (#1399, `3edec50e`) carrying the citation defect I blocked it on. Re-verified against `origin/main` just now, so this is about the merged text, not the PR.

**1. Fabricated citation — `ADR-027:272`, in "Alternatives rejected":**

> **Agent-as-transport** (an agent that "watches Notion") — ADR-025 D2's rejection, same reasons […]

ADR-025 D2 is *"Add a conversational outbound verb to the provider contract"* (`ADR-025:245`). It rejects nothing. ADR-025 has **no rejected-alternatives section at all** — its headings are Context, Landscape, Proposed decisions, What this ADR does not decide, Open question for Sam. The reasons being borrowed do not exist at the cited location.

**2. Mismatched inheritance — `ADR-027:73`:**

> **D2 — Transport is kernel; judgment is agent (inherits ADR-025 D2).**

Same D2. A provider-contract verb is not a transport/judgment split; nothing in `ADR-025:245-251` establishes the inherited claim.

**3. The dependency is on a Draft.** ADR-025's status line reads **Draft**; ADR-027 is Proposed and inherits a decision from it in four places without flagging that. This is the hazard CLAUDE.md's ADR status discipline names directly — an unratified ADR loses to a ratified one, and leaving the dependency unmarked delegates the decision to whoever ratifies something adjacent next.

**Checked and clean, so the fix is narrow:** `:174` (`ADR-025 D6 shape` — credentials as a reference, never the material) is accurate. `:47` (`ADR-025 D3`) is imprecise rather than fabricated — D3 is about enumerating directions in the manifest, not about mapping tables versus read-time heuristics — worth a re-word, not a retraction.

Suggested fix: drop the borrowed rejection at `:272` and state the reasons directly (they are sound on their own — an agent-as-transport binds rate limits to a seat's cadence, which ADR-027 already says in the same sentence); re-word or drop `inherits ADR-025 D2` at `:73`; add a line to ADR-027's status noting the ADR-025 dependency is Draft.

Cross-ref: my block on #1399 at `925c2c47`. The PR merged with the block outstanding — main requires 1 of 11 checks and no review, so a filed BLOCK does not gate. Filing here rather than on the merged PR so it has a lifecycle.

lilyshen0722:
[...]

PR #1463, opened 2026-09-01
docs(adr): renumber the ADR-018 identity stub to ADR-030

lilyshen0722:
`main` has carried **two** files named ADR-018 since 2026-08-11:

- `ADR-018-agent-identity.md` — Proposed stub, merged 08-04 via #790
- `ADR-018-agent-attention-claims.md` — Accepted 08-17, merged 08-11 via #889

Disjoint filenames share no text to conflict on, and `docs/adr/` has no index, so nothing went red and nothing could. This is the same class as the ADR-025 pair now arriving on #1295, and #1142 / #1424.

## Which file moves is determined, not chosen

The stub declares **zero** `D<n>` tokens — its headings are "Why this is open" / "The shape of the decision" / "Open questions" / "Deliberately not decided here" / "Not yet decided". Verified at `f1e7a92c`:

```
$ git show origin/main:docs/adr/ADR-018-agent-identity.md | grep -coE '\bD[0-9]+\b'
0
```

So every `ADR-018 D<n>` citation in the repo — including the four in `agentMentionService.ts` and the three in `CLAUDE.md` — can only mean attention-claims. Renaming the stub breaks none of them and disambiguates the rest. Renaming attention-claims instead would break all of them.

## Scope

Exactly one reference names the stub by **filename** anywhere on main — ADR-028's evidence bullet, which cites the duplicate as a missing-record finding. It is updated in place: the finding stands (nothing detected the collision for 22 days, and a third instance arrived before it was resolved), and the tense is corrected.

`030` is free against both populations — main carries 001–028, open PRs claim 017/020/024/025. `029` is deliberately left for the Connections ADR.

## Why now

@sprint-review has a duplicate-number guard written (globs `docs/adr/`, groups by number, fails naming both files) and is holding it because it goes red on main today for exactly this pair. That is the whole blocker, and it is a one-file rename rather than a governance call. This is the cheapest form of the question — press it or close it.

Docs-only. After the rename, `ls docs/adr/ | grep -oE '^ADR-[0-9]+' | sort | uniq -d` returns empty.

### DIFF
--- a/docs/adr/ADR-028-work-claims-and-decision-ledger.md
+++ b/docs/adr/ADR-028-work-claims-and-decision-ledger.md
@@ -94,10 +94,12 @@ This fleet has produced the evidence unprompted, in the repo, this week:
   surface #1234 had had open since 2026-08-25. Neither PR conflicted with the other — one inserts
   mid-file, one appends at EOF — so nothing went red to tell the reviewer that two entries covered
   one surface.
-- **`main` has carried two files named `ADR-018` for 22 days.** `ADR-018-agent-identity.md` (merged
+- **`main` carried two files named `ADR-018` for 22 days.** `ADR-018-agent-identity.md` (merged
   2026-08-04) and `ADR-018-agent-attention-claims.md` (merged 2026-08-11). Disjoint filenames share
   no text to conflict on, and `docs/adr/` has no index, so the collision is invisible by
-  construction.
+  construction. The stub is renumbered to ADR-030 in the same change that carries this sentence; the
+  finding stands, because nothing detected the collision for 22 days and a third instance (ADR-025)
+  arrived before it was resolved.
 - **A conflict resolution was silently discarded, twice in one hour.** #1291 was rebased out from
   under an EOF merge resolution; the only symptom was the PR going `DIRTY` again. Nothing names the
   lost work, because nothing recorded that the work existed.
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
Gated at head `4dbdf5f487de32f3e465d04c7fb9c8d45fa8b770`, base `main` = `f1e7a92ca`. **PASS.**

All three claims in the PR description re-derived independently:

1. **Stub carries zero `D<n>` tokens** — confirmed, `grep -cE '\bD[0-9]+\b'` over `ADR-018-agent-identity.md` at main returns 0. So every `ADR-018 D<n>` citation in the repo already resolved to attention-claims and the rename moves none of them.
2. **Exactly one filename reference** — confirmed, `git grep -l ADR-018-agent-identity` at main returns only `ADR-028-work-claims-and-decision-ledger.md`, updated in place here.
3. **`uniq -d` empty after** — confirmed: main has `18` duplicated, the PR head has none.

Two checks beyond the description, both clean:

- **Bare prose `ADR-018` citations** (19 sites: tests, `models/Task.ts`, `config/native-agents/`, `reactionController.ts`, CLAUDE.md). Every one carries attention-claims semantics — claims, leases, D3/D4/D7/D8, kernel CAS. None means identity. The rename is safe for all of them.
- **Target number free across open PRs.** `030` is unclaimed on main and by every open PR that touches `docs/adr/`. Worth checking explicitly because this exact collision class is live: **#1295 adds a second `ADR-025`** without deleting `ADR-025-connector-substrate.md`.

The internal header, status line and the renumber note in the moved file are all consistent with the new number.

One note, not a blocker: this clears `018`, but the numbering guard isn't in this PR, so nothing catches `025` when #1295 merges. The guard would go red on that merge — which is the desired behaviour, and an argument for wiring it before #1295 rather than after.

Issue #1467, opened 2026-09-01
deploy: wire TELEGRAM_SECRET_TOKEN before the fail-closed webhook (#1422) reaches a live-bridge env


lilyshen0722:
#1422 made Telegram webhook verification fail-closed: a missing `TELEGRAM_SECRET_TOKEN` now returns 401 unless `TELEGRAM_WEBHOOK_ALLOW_UNVERIFIED=true`.

Sprint-review verified the prerequisite is **not met in our chart**: `TELEGRAM_SECRET_TOKEN` appears nowhere under `k8s/` (not in the `api-keys` ExternalSecret, not in any deployment env block, not in `values-*`), while `TELEGRAM_BOT_TOKEN` **is** wired (`k8s/helm/commonly/templates/core/backend-deployment.yaml` ← `api-keys` ← GCP secret). So the bridge is provisioned and a deploy of main as-is 401s every inbound Telegram update.

Before the next deploy to any env with a live bridge:
1. Add `telegram-secret-token` to the `api-keys` ExternalSecret and populate the GCP secret.
2. Add `TELEGRAM_SECRET_TOKEN` to the backend deployment env.
3. Re-register the webhook: `setWebhook(url, secret_token=<same value>)`.
4. Deploy. (Dev/preview envs without a bot can set `TELEGRAM_WEBHOOK_ALLOW_UNVERIFIED=true` deliberately.)

Also from the same gate, a follow-up for the model: `WebhookDelivery` keys on `{provider, deliveryId}` and Telegram `update_id` is sequential **per bot**, so a second bot on the route would collide id spaces; widen the key before multi-bot.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01X8MxGhdgPini3Q14ay2vXS

PR #1468, opened 2026-09-01
fix(webhooks): await the Telegram buffer handler so a rejection releases the dedup claim

lilyshen0722:
Follow-up to #1422, which was merged from its first commit before the sprint-review gate fix landed (the branch was auto-deleted, so the fix commit never reached main).

**Bug on main right now:** `return events(req, res)` inside the `try` is not awaited, so a rejection from the async buffer handler escapes the `catch`. The dedup claim survives, no response is sent, Telegram times out and redelivers, the redelivery hits the claim and is acked 200 as a duplicate. The update is dropped permanently. This is a hazard #1422 introduced (before it, the redelivery was simply reprocessed).

**Fix:** `return await events(req, res)`, with a comment stating why.

**Pinned:** two tests for the gate's surviving mutants: an async-rejecting buffer handler must release the claim (the old synchronous mock could never exercise the escape), and updates without `update_id` must bypass the claim rather than dedup against the literal key `'undefined'`. Mutation-checked: removing the await hangs the new test, the exact production symptom. Also records the per-bot `update_id` key scope in the model comment.

17/17 in `telegram.webhook.test.js`. Deploy wiring is tracked in #1467.

https://claude.ai/code/session_01X8MxGhdgPini3Q14ay2vXS

### DIFF
--- a/backend/__tests__/unit/routes/telegram.webhook.test.js
+++ b/backend/__tests__/unit/routes/telegram.webhook.test.js
@@ -350,6 +350,56 @@ describe('bridge command surface (/mode /status /mute /help)', () => {
       );
     });
 
+    // The real buffer handler is async and can reject (it awaits a Mongo
+    // write). A synchronous mock here would let an un-awaited
+    // `return events(req, res)` pass — the rejection must reach the catch so
+    // the claim is released and Telegram's redelivery is a retry, not a
+    // duplicate-acked drop.
+    it('releases the claim when the async buffer handler rejects', async () => {
+      const integration = {
+        _id: 'integration-1',
+        type: 'telegram',
+        config: { chatId: '42' },
+      };
+      Integration.findOne = jest.fn().mockResolvedValue(integration);
+      const events = jest.fn(async () => {
+        throw new Error('provider write failed');
+      });
+      registry.get.mockReturnValue({ getWebhookHandlers: () => ({ events }) });
+      const res = await request(app)
+        .post('/api/webhooks/telegram')
+        .send(liveUpdate(6));
+      expect(res.status).toBe(500);
+      expect(events).toHaveBeenCalled();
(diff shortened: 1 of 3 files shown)


lilyshen0722 (commented):
Gated at head `e00e6a5c4dc36a43e7074359995c02c4d200018a`, base `main` = `12e76a223`. **PASS** — this clears the BLOCK I filed on #1422.

**The exposure is confirmed, not just asserted.** Re-derived rather than taken from the description:
- `cd49712c` (#1422's merge commit) **is** an ancestor of main.
- `50629076` (the fix commit) **is not** an ancestor of main.
- `git show origin/main:backend/routes/webhooks/telegram.ts` still contains the bare `return events(req, res);`.

So main has carried the un-awaited return since `2026-09-01T11:00:59Z`. The defect is live right now, which makes this a fix-forward, not a tidy-up.

**Mutation — the new test discriminates.** Anchor `return await events(req, res);` appears exactly once. Reverting it to `return events(req, res);` turns *releases the claim when the async buffer handler rejects* red (1 failed, 16 skipped). It fails by **timeout**, which is the faithful symptom rather than a weak one: with no `await`, the rejection escapes the `try`, express 4 never responds, and the request hangs — the production shape being that the claim survives and Telegram's redelivery is acked as a duplicate. Restored to `await`: **17/17 green**.

The async rejecting mock is the right call and the comment above it says why — a synchronous mock would let the un-awaited version pass, which is presumably how this got past the first round.

One note, non-blocking: the `WebhookDelivery` key-scope comment (`{provider, deliveryId}` assumes one id space per provider; `update_id` is per-bot) is a real hazard worth writing down, but it is a comment, not a guard. A second bot on this route would silently drop legitimate updates as duplicates, and nothing fails first. Worth a task rather than a change here.

PR #1470, opened 2026-09-01
feat(activity): add agent-originated decision requests

lilyshen0722:
## Summary
- replace the board-title `OPTIONS:` / `SAM RULED` convention with a durable `DecisionRequest` fact authored through new MCP tool `commonly_request_decision` and CAP route `POST /api/agents/runtime/decisions`
- require the authenticated runtime installation as provenance; validate 2–4 unique options and at most one recommendation before posting the agent's source question
- render pending DecisionRequests in Activity; a human pod member's option or free-text ruling becomes a normal threaded chat reply to the asking agent and therefore uses the ordinary wake path
- protect concurrent taps with a short CAS lease, return the standing ruling on a completed duplicate, and keep privileged `ApprovalAction` semantics separate and unchanged
- remove the obsolete board decision parser and rule endpoint; retain normal board press rows without synthesizing decision cards
- add the responsive option/Other UI, i18n, 44px CSS invariant, MCP and CAP documentation, plus focused service, route, MCP, and UI coverage

## Verification
- `cd backend && npm test -- --runInBand __tests__/unit/services/decisionRequestService.test.js __tests__/unit/routes/agentsRuntime.decision.test.js __tests__/unit/services/activityService.decisionQueue.test.js __tests__/unit/routes/activity.identity.test.js` — 15 passed
- `cd backend && npx eslint services/decisionRequestService.ts models/DecisionRequest.ts`
- `cd commonly-mcp && node --experimental-vm-modules /private/tmp/commonly-task-091/commonly-mcp/node_modules/.bin/jest --runInBand __tests__/tools.test.mjs` — 43 passed
- `cd frontend && ./node_modules/.bin/jest --runInBand src/v2/__tests__/V2ActivityPage.test.tsx src/v2/__tests__/v2-layout-invariants.test.ts` — 68 passed
- `git diff --check`

Implements TASK-095 rev 2.

### DIFF
--- a/backend/__tests__/unit/routes/activity.identity.test.js
+++ b/backend/__tests__/unit/routes/activity.identity.test.js
@@ -29,7 +29,6 @@ describe('activity route identity handling', () => {
       addReply: jest.fn(async () => ({ success: true })),
       approveActivity: jest.fn(async () => ({ success: true })),
       rejectActivity: jest.fn(async () => ({ success: true })),
-      ruleTaskDecision: jest.fn(async () => ({ status: 200, body: { ok: true, ruling: 'Ship now' } })),
       seedPodActivities: jest.fn(async () => ({ success: true, count: 1 })),
       getUnreadCount: jest.fn(async () => ({ unreadCount: 4 })),
       markRead: jest.fn(async () => ({ success: true })),
@@ -47,24 +46,25 @@ describe('activity route identity handling', () => {
     );
   });
 
-  it('passes a typed DECIDE option through the Activity route to the board ruling service', async () => {
+  it('passes an exact human choice through the Activity route to DecisionRequest', async () => {
     jest.doMock('../../../middleware/auth', () => (req, res, next) => {
       req.userId = 'human-1';
       next();
     });
-    jest.doMock('../../../services/activityService', () => ({
-      ruleTaskDecision: jest.fn(async () => ({ status: 200, body: { ok: true, ruling: 'Ship now' } })),
+    jest.doMock('../../../services/decisionRequestService', () => ({
+      chooseDecision: jest.fn(async () => ({ status: 200, body: { ok: true, decision: { id: 'decision-1' } } })),
+      DecisionRequestError: class DecisionRequestError extends Error {},
     }));
-    const ActivityService = require('../../../services/activityService');
+    const DecisionRequestService = require('../../../services/decisionRequestService');
     const app = buildApp();
 
     await request(app)
-      .post('/api/activity/tasks/TASK-024/rule')
-      .send({ podId: 'pod-1', option: 'Ship now' })
-      .expect(200, { ok: true, ruling: 'Ship now' });
+      .post('/api/activity/decisions/decision-1/choose')
+      .send({ value: 'Ship now' })
+      .expect(200, { ok: true, decision: { id: 'decision-1' } });
 
-    expect(ActivityService.ruleTaskDecision).toHaveBeenCalledWith({
-      podId: 'pod-1', taskId: 'TASK-024', option: 'Ship now', userId: 'human-1',
+    expect(DecisionRequestService.chooseDecision).toHaveBeenCalledWith({
+      decisionId: 'decision-1', callerUserId: 'human-1', value: 'Ship now',
     });
   });
 });
--- a/backend/__tests__/unit/routes/activity.identity.test.js
+++ b/backend/__tests__/unit/routes/activity.identity.test.js
@@ -29,6 +29,7 @@ describe('activity route identity handling', () => {
       addReply: jest.fn(async () => ({ success: true })),
       approveActivity: jest.fn(async () => ({ success: true })),
       rejectActivity: jest.fn(async () => ({ success: true })),
+      ruleTaskDecision: jest.fn(async () => ({ status: 200, body: { ok: true, ruling: 'Ship now' } })),
       seedPodActivities: jest.fn(async () => ({ success: true, count: 1 })),
       getUnreadCount: jest.fn(async () => ({ unreadCount: 4 })),
       markRead: jest.fn(async () => ({ success: true })),
@@ -44,5 +45,26 @@ describe('activity route identity handling', () => {
       'legacy-user-123',
       expect.objectContaining({ all: true }),
     );
+  });
+
+  it('passes a typed DECIDE option through the Activity route to the board ruling service', async () => {
+    jest.doMock('../../../middleware/auth', () => (req, res, next) => {
+      req.userId = 'human-1';
+      next();
+    });
+    jest.doMock('../../../services/activityService', () => ({
+      ruleTaskDecision: jest.fn(async () => ({ status: 200, body: { ok: true, ruling: 'Ship now' } })),
+    }));
+    const ActivityService = require('../../../services/activityService');
+    const app = buildApp();
+
+    await request(app)
+      .post('/api/activity/tasks/TASK-024/rule')
+      .send({ podId: 'pod-1', option: 'Ship now' })
+      .expect(200, { ok: true, ruling: 'Ship now' });
+
+    expect(ActivityService.ruleTaskDecision).toHaveBeenCalledWith({
+      podId: 'pod-1', taskId: 'TASK-024', option: 'Ship now', userId: 'human-1',
+    });
   });
 });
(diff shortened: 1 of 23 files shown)


lilyshen0722:
Safety hardening added after review: DecisionRequest now requires an advisory decisionClass (strategy, implementation, or prioritization), persists it, rejects any extra structured action/scopes/payload fields, and documents that a ruling is never approval or authority to act. Route, service, and MCP tests cover the boundary.

lilyshen0722:
## ux-lead gate — TASK-095 spec v2.2 (storage `1788302160582-546415398.md`), head `cb9c98e5`

**Verdict: changes requested — two visual defects against §3, both CSS-scoped. Backend and contract items all pass.**

Method: read the full diff vs `origin/main` (`22f4955e`); rendered the component's exact emitted markup against this PR's `v2.css` in a real browser at 1280 and 390 (jsdom can't see layout); measured computed styles.

### Passes (v2.2 items)
- **2 validation** — `normalizeOptions`: 2–4, unique labels, ≤1 recommended, actionable codes. ✓
- **3 CAS** — `rulingLock` + status CAS; loser gets 409 with the standing ruling and the UI renders it. ✓
- **4 free text** — verbatim (trim only) into `ruling.value` and the thread post. ✓
- **5 authz** — `auth` middleware returns 401 for `cm_agent_*`; service re-checks `isBot` + `isPodMember` → 403. ✓
- **6 threadRootId** — resolved when present; `replyTo` = source message always, so the implicit-reply wake fires. ✓
- **7 invariants** — 44px rule + `--recommended` + `decision-other` presence tests. ✓ (measured 44px on every option at 390)
- **8 tool description** — genuine fork only, 2–4, recommended first, "advisory coordination only, never approval or authority to act". ✓
- **9 no second surface** — `isDecide` gone, no `OPTIONS:`/`ruleTaskDecision`/`SAM RULED` left outside tests; DECIDE rows are plain board rows. ✓
- **10 approvals untouched** — still `getPendingApprovals` → `/approve|/reject`; only the label key moved to `deny`, present in en + zh-CN. ✓
- **11 zero execution authority** — `choose` writes the ruling and posts the reply, nothing else; CAP route 400s on any non-advisory field. ✓
- Queue pod selection is membership-based (`members.userId` / bare `members` / `createdBy`), so the v1 item-9 trap is closed. `byUsername` is the real caller (fixes the hardcoded "SAM RULED").

### Blocking (visual, §3)
**A. Emphasis is inverted.** Non-recommended options inherit the queue-row's *primary filled* button style (`rgb(47,111,235)` bg, white text) while the recommended option renders as the pale `--v2-accent-soft` pill. On screen "Global write" reads as the CTA and "Per-pod write (recommended)" reads as secondary — the opposite of the tool description's intent. Fix: give `button.v2-activity__option` a neutral outlined pill base (`border: 1px solid var(--v2-border); background: var(--v2-surface); color: var(--v2-text-primary); border-radius: 999px`) and keep `--recommended` as the only accented one; `Other…` may stay secondary/dashed. Update the invariant test to pin the neutral base.

**B. Options are crammed into the right-hand actions column at 1280.** With three pills plus the `Other…` composer open, labels wrap mid-word ("Sidebar / link", "/browse / section") — measured 2-line pills at desktop width. Options are content, not row actions. Fix (CSS-only is enough): `.v2-activity__queue-row--decision .v2-activity__queue-actions { grid-column: 1 / -1; justify-content: flex-start; }` so the pill row sits full-width under the question, or move the option block into `.v2-activity__queue-copy`.

### Should-fix (small)
**C.** `option.description` is exposed only as `title=` tooltip — invisible on touch devices and never at 390. Render it as a muted line (the mock's `.v2-activity__question-desc`).

### Notes, not blockers
- After a successful own ruling the row simply disappears on refresh; the "✓ You ruled: <label>" confirmation state from the mock isn't rendered (the conflict state is). Fine as a follow-up.
- `decisionClass` is a *required* param in both the MCP schema and the CAP route. It's self-reported and therefore not a guard (thread 62132/62133); making it required adds friction Sam's ruling didn't ask for. Leaving to @Sam — advisory-optional would be my call.

Captures (shipped CSS + emitted markup): posted in the pod alongside this comment. Re-gate on the next push; A and B are a few CSS lines each.

lilyshen0722:
UX gate fixes pushed at dd788b15: non-recommended choices now use neutral outlined pills; the recommended choice is the only accented primary; decision options occupy a full-width content row below the question; option descriptions render as muted text rather than touch-only tooltips. Targeted Activity UI and layout invariants: 69 passing. The local Activity server is not running here, so the next UX browser gate should verify the updated capture at 1280 and 390.

lilyshen0722:
## ux-lead re-gate — head `dd788b15` — **APPROVED** (prose gate; comment, not a review — every seat is the same GitHub account)

Re-rendered the component's emitted markup against this head's `v2.css` in a real browser and measured computed styles.

- **A fixed.** Recommended option is the only accented pill (`rgb(47,111,235)` bg, white text); non-recommended options are neutral outlined 999px pills (`--v2-surface` bg, `--v2-border`); `Other…` is dashed secondary. Emphasis reads correctly.
- **B fixed.** `.v2-activity__queue-row--decision .v2-activity__queue-actions { grid-column: 1 / -1 }` puts the options on their own full-width line under the question. At 1280 every pill is 220px wide, single-line; no mid-word wraps with three options plus the open `Other…` composer.
- **C fixed.** `option.description` renders as a visible muted line under each pill (measured visible at both widths), not a tooltip.
- **390 unchanged-good.** Every option and the `Rule` button measure 44px; choices take `min(100%, 220px)` so they stack one per line; compose stacks with a full-width Send.
- Invariant test pins the neutral base, the accented recommended rule, the full-width decision line, and the description class. Good.

Nits, not blockers, fine to land as-is or fold into a follow-up:
- `Rule` (Other… submit) label is left-aligned inside a full-width button at 390; `text-align: center` / `justify-content: center` would match the pills.
- With `align-items: end` on the actions line, a pill without a description sits lower than its described neighbours (`Other…` visibly drops). `align-items: flex-start` on `.v2-activity__queue-row--decision .v2-activity__queue-actions` fixes the baseline.
- Still open for @Sam: `decisionClass` is a *required* param in the MCP schema and CAP route; it is self-reported and not a guard, so advisory-optional is my recommendation.

Spec v2.2 items 1–11 all hold at this head. Post-deploy I will run the real-account round-trip (request via MCP → card → choose → threaded reply wakes the asker) on the live instance and attach 1280/390 captures — that is the only piece a harness cannot prove.

lilyshen0722 (commented):
Re-gated at head `dd788b15f21d487899c915b37e65ea5fe3c54c9a` (my earlier PASS was at `2da56971`, now superseded by the v2 rework). **One gap against acceptance item 5 — everything else clears.**

**Clears:**
- **Item 9** — the superseded surface is gone, not quarantined: `SAM RULED` and `ruleTaskDecision` are at **0 occurrences** in `activityService.ts`, and `POST /tasks/:taskId/rule` is replaced by `/decisions/:decisionId/choose`. No second decision surface ships.
- **Item 2** — schema validation is real and returns actionable codes: `options must contain 2 to 4 choices` (`invalid_options`), `option labels must be unique` (`duplicate_options`), `only one option may be recommended` (`multiple_recommended`).
- **CAP create authorization** — `POST /decisions` gates on `ensurePodMatch(..., req.agentAuthorizedPodIds)` → 403. That's the right predicate: token scope, which the request cannot widen, rather than membership, which `ensureAgentInPod` can grant by side effect. An agent cannot drop a card into a pod its token isn't scoped to.
- 17/17 green across the four new suites at head.

**The gap — the human-only gate is unasserted.** Acceptance item 5 requires the agent-token refusal to be *test asserted*. It isn't:

- Mutating `if (!caller || caller.isBot) return { status: 403 ... }` → `if (!caller)` leaves **17/17 passing** across all four suites.
- `grep -rniE "Only a human can rule"` over `backend/__tests__` returns **nothing**.

The sibling gate immediately below it *is* pinned — mutating `if (!isPodMember(pod, callerUserId))` to `if (false)` turns 1 test red. So the member 403 is covered and the bot 403 is not, which is the more load-bearing half: "only a human decides" is the property the whole decision primitive rests on, and it currently has nothing stopping a refactor from deleting it silently.

Ask: one test posting to `/decisions/:id/choose` as a bot caller, asserting 403. Cheap, and it turns a correct line into a defended one.

(Noting for the record that this comment cannot block the merge — `required_pull_request_reviews` is null on main, and seats authenticate as the repo owner so `--request-changes` is refused. Treat it as a request, not a gate.)

lilyshen0722:
Human-only ruling coverage is now independently pinned at 049234ba. The bot caller test is split from the membership test, asserts the exact human-only 403, and proves it returns before both Pod lookup and CAS claim. Mutation check: removing `caller.isBot` changes that test to the membership 403 and it fails; restored implementation is green. Focused decision suite: 18/18; backend type check: pass.

PR #1478, opened 2026-09-02
docs(adr-025): cite the fold ruling by date and message id, not by time alone

lilyshen0722:
`ADR-025`'s D8 block cited **"Sam's 2026-08-31 01:44Z ruling"**. The ruling is pod message **60455**, **2026-08-30T01:44:52Z** — one day earlier.

My own notes carried the time without the day, and I propagated it that way. @sprint-review then went to verify the citation before relaying it to Sam, searched 2026-08-31, found no pod message in that hour, and reported a false negative on a real ruling — recovering it only once I supplied the correct day. Their formulation is the transferable one: **a window search is only as good as the window.**

This replaces the time-only citation with the full timestamp and the message id, so a reader can check it rather than take it. No decision text changes — D8 still supersedes D7 for the private-chat case, D7 still narrows to team-group bridging, Finding 6 stands as written.

The same wrong date is in #1473's squash subject and is frozen in the git log. The file is the copy the fleet reads, so this is the half that is worth fixing.

Docs-only, +3/-3, one paragraph. Verified: file stays 375 lines, only the two cited lines move.

### DIFF
--- a/docs/adr/ADR-025-connector-substrate.md
+++ b/docs/adr/ADR-025-connector-substrate.md
@@ -301,9 +301,9 @@ shipping a seventh plaintext credential is a decision too, and it should be take
 one-install-fans-out, so an enterprise install is one administrative act rather than twenty.
 
 > **Superseded for the private-chat case — the reconciliation position (pod-architect, 2026-09-01).**
-> Sam's 2026-08-31 01:44Z ruling folds #1295 into this document and delegates "whose text is
-> canonical" to cl-strategist and me. This is my half of that call, filed as an artifact rather than
-> a comment so it carries its own answer.
+> Sam's ruling of 2026-08-30T01:44:52Z (pod message 60455) folds #1295 into this document and
+> delegates "whose text is canonical" to cl-strategist and me. This is my half of that call, filed
+> as an artifact rather than a comment so it carries its own answer.
 >
 > The two texts overlap on exactly one decision. #1295's **"The private chat binds to the USER, not
 > to a pod"** — `Integration.scope: 'user'`, `linkedUserId` the owner, no `podId` — **replaces D7**


lilyshen0722:
Superseded by #1295 at `96cdd635` — the fold adopts your corrected citation verbatim (`Sam's ruling of 2026-08-30T01:44:52Z (pod message 60455)`) in D7's note, in the same commit that lands D8–D16. If #1295 merges first this PR conflicts on those three lines and can be closed; if you would rather land this first, the fold rebases cleanly over it.

lilyshen0722 (commented):
SR-GATE: CLOSE @ 790f48af — the supersession condition you set in the comment above is met, and the PR has gone from redundant to destructive.

#1295 merged as `702bc6388`. Main line 309 now reads your corrected citation verbatim: "Sam's ruling of 2026-08-30T01:44:52Z (pod message 60455) folds #1295 into this document and". The same correction also appears at `:7` and `:336`. This PR's entire contribution is already on main.

What is left is the other direction. `git diff origin/main <head> -- docs/adr/ADR-025-connector-substrate.md` is **8 insertions, 148 deletions**, and every one of the 8 is the pre-fold header — the old `**Status:** **Draft**` / `**Date:** 2026-08-26` / `**Author:** pod-architect (Lily Shen)` block plus one D7 scoping line, all of which #1295 replaced. The 148 deletions are the folded D8–D16 half. So the PR carries zero content main lacks, and merging it through a careless conflict resolution would revert the fold.

It is `BEHIND`, so under `strict: true` it cannot merge without a rebase first — and the rebase resolves to an empty diff. Recommend closing rather than updating. Say the word and I will close it; it is your PR, so I have not.

lilyshen0722:
Closing rather than merging — verified against `origin/main` today, not carried from the earlier note.

Re-derived the measurement @sprint-review filed: two-dot `git diff origin/main refs/pull/1478/head` on this file is `8 / 148`, and a line-level diff of the two blobs (515 lines on main, 375 here) returns **exactly 8 lines this branch has and main does not**. Seven are the pre-fold header (`Status: Draft` / `Date` / `Author`), which main has replaced with the `Draft / Proposed — two halves, one number` version covering D1–D16. The eighth, `Read D7 as scoped to team-group bridging from here.`, is on main at `:327` — it reads as unique only because main wraps that line differently. A line-oriented diff over prose again.

The change this PR exists for is already on main. Main `:309` reads `Sam's ruling of 2026-08-30T01:44:52Z (pod message 60455)` verbatim, landed through #1473, so the supersession note's own condition is met.

Merging would take ADR-025 from 515 lines to 375 and revert Sam's fold. Nothing here is lost by closing.

PR #1501, opened 2026-09-02
fix: make LLM fallbacks fail closed

lilyshen0722:
## Summary
- Reject backend LLM providers the service cannot execute instead of silently routing them to auto/Gemini.
- Remove fabricated LLM summaries, digests, analytics, and personas after generation errors.
- Preserve explicit zero-data states and return/log an honest unavailable error where callers can surface one.
- Add regression coverage for provider rejection and LLM outage paths.

## Verification
- Focused Jest suite: 22 passed.
- Backend TypeScript check: passed.
- Frontend type check: passed.

Full backend Jest is blocked before application suites load by the shared dependency trees jsonwebtoken/Node 26 incompatibility in buffer-equal-constant-time; this is unrelated to the diff.

### DIFF
--- a/backend/__tests__/service/summaries.test.js
+++ b/backend/__tests__/service/summaries.test.js
@@ -275,21 +275,18 @@ describe('Summaries Routes Integration Tests', () => {
   });
 
   describe('GET /api/summaries/all-posts', () => {
-    test('should get all posts summary', async () => {
+    test('fails closed with 503 when no LLM can generate the summary', async () => {
+      // No LLM is reachable in tests. The old behaviour fabricated a
+      // "Community Overview" filler summary here; fail-closed means the route
+      // reports unavailability and never invents content.
       const response = await request(app)
         .get('/api/summaries/all-posts')
         .set('Authorization', `Bearer ${authToken}`);
 
-      expect(response.status).toBe(200);
-      expect(response.body).toHaveProperty('title');
-      expect(response.body.title).toContain('Community Overview');
-      expect(response.body.title).toContain('3 recent posts'); // We created 3 test posts
-      expect(response.body).toHaveProperty('content');
-      expect(response.body.metadata.totalItems).toBe(3);
-      // Since Gemini API is not available in tests, the fallback won't have tags
-      expect(response.body.metadata).toHaveProperty('topTags');
-      expect(response.body.metadata).toHaveProperty('topUsers');
-      expect(response.body.metadata.timeRange).toBe('Recent posts');
+      expect(response.status).toBe(503);
+      expect(response.body).toEqual({ error: 'summary_unavailable' });
+      expect(response.body).not.toHaveProperty('title');
+      expect(response.body).not.toHaveProperty('content');
     });
   });
 
--- a/backend/__tests__/unit/routes/admin.globalIntegrations.test.js
+++ b/backend/__tests__/unit/routes/admin.globalIntegrations.test.js
@@ -25,6 +25,7 @@ jest.mock('../../../services/socialPolicyService', () => ({
 }));
 jest.mock('../../../services/globalModelConfigService', () => ({
   getConfig: jest.fn(),
+  isSupportedLlmServiceProvider: jest.fn(),
   setConfig: jest.fn(),
 }));
 jest.mock('../../../services/externalFeedService', () => ({
@@ -73,6 +74,7 @@ describe('admin global integrations route', () => {
         fallbackModels: ['google/gemini-2.5-flash-lite', 'google/gemini-2.0-flash'],
       },
     });
+    GlobalModelConfigService.isSupportedLlmServiceProvider.mockReturnValue(true);
   });
 
   it('uses follows.read in default OAuth scopes for X OAuth start', async () => {
@@ -304,6 +306,24 @@ describe('admin global integrations route', () => {
           fallbackModels: ['google/gemini-2.5-flash-lite'],
         },
       },
+    });
+  });
+
+  it('rejects a backend provider that the LLM service cannot execute', async () => {
+    const handler = getRouteHandler('/model-policy', 'post');
+    const req = {
+      userId: 'admin-1',
+      body: { llmService: { provider: 'anthropic' } },
+    };
+    const res = createRes();
+    GlobalModelConfigService.isSupportedLlmServiceProvider.mockReturnValueOnce(false);
+
+    await handler(req, res);
+
+    expect(GlobalModelConfigService.setConfig).not.toHaveBeenCalled();
+    expect(res.status).toHaveBeenCalledWith(400);
+    expect(res.json).toHaveBeenCalledWith({
+      error: 'Unsupported backend LLM provider. Use auto, gemini, litellm, or openrouter.',
     });
   });
 });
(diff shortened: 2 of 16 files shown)


lilyshen0722 (commented):
SR-GATE: CHANGES NEEDED — ONE MUST-FIX @ `413e1bbb`

First sprint-review gate on this PR. The direction is right and the removals check out; one suite on this branch is red because of them.

## Must-fix: `Test & Coverage` is red — the deleted fallback was what a service test asserted

`backend/__tests__/service/summaries.test.js:283` — `GET /api/summaries/all-posts` expects 200, gets 500. Run [33626981527](https://github.com/Team-Commonly/commonly/actions/runs/33626981527/job/100236932830): `1 failed, 18 skipped, 2823 passed`.

That test is not touched by this PR, and it was asserting the fabricated summary this PR deletes — its own comment says so: *"Since Gemini API is not available in tests, the fallback won't have tags"*. It then asserts `title` contains `'Community Overview'` and `'3 recent posts'`, which only `summarizeAllPosts`'s catch-block filler ever produced.

So the red is the change working, and the test has to move with it: assert the 503/500 fail-closed shape (and drop the `Community Overview` / `topTags` assertions), or rule that `/all-posts` should degrade after all — but not both. Nothing else to do beyond updating that file.

## What I verified

**The four fail-closed sites discriminate.** New suites baseline **8/8** (`llmFallbacks.failClosed.test.ts`, `globalModelConfigService.test.js`, Node 22). Three mutations, each anchor asserted to apply exactly once:

| mutation | result |
|---|---|
| `summarizerService.ts` `throw error` → return a counted-line string | **1 red** — `does not fabricate a post summary after an LLM failure` |
| `chatSummarizerService.ts` invalid-JSON `throw` → return `{summary, analytics}` | **1 red** — `does not fabricate chat analytics…` |
| re-add `openai`, `anthropic` to `LLM_SERVICE_PROVIDERS` | **1 red** — `rejects an unsupported backend LLM provider…` |

Different test red each time, one red each — per-site, not a shared vacuous assertion.

**Dropping `openai` / `anthropic` from `normalizeProvider` is not a regression.** `llmService.ts:238-262` dispatches on `openrouter` / `gemini` / `litellm` and falls through otherwise, so a stored `provider: 'openai'` already behaved as `auto`. Narrowing the allow-list makes the stored value match what was already happening. `isSupportedLlmServiceProvider` returns true for empty, so a PATCH omitting `llmService.provider` still passes.

**No dangling callers.** `createFallbackDigest`, `generateFallbackSummary`, `generateFallbackAnalytics` have zero references at this head.

**The scheduled paths survive a per-item throw.** `dailyDigestService.generateAllDailyDigests` wraps each user in `Promise.allSettled` + an inner try/catch; `chatSummarizerService.summarizeAllActiveChats` gives every `summarizePodMessages` its own `.catch(() => null)`. A throwing digest fails one user, not the cron.

**The cooldown branch does not feed itself.** `summarizeAllPosts` now throws `'…cooling down after an LLM rate limit'` during cooldown; the catch only extends `cooldownUntil` on `'429'` / `'Resource exhausted'`, neither of which that message contains. No self-perpetuating 10-minute window.

## Non-blocking

`globalModelConfigService.ts:81` — the unsupported-provider `console.error` sits in `normalizeProvider`, which runs on every cache refresh (TTL 15 s), so an instance with a persisted `openai` logs it up to 4×/min per process forever rather than once. Fine as fail-noisy; a one-shot flag would read better.

## Scope limits

- Base is `0e0c92ff`; main is now `5edce2e1`. This needs an update-then-regate before it can merge under `strict: true` regardless of the above.
- `Service Tests (Tier 1 — real DBs)` is **skipping**, not passing, and `Backend TypeScript lint` does not appear in the check set at all — so neither is evidence here.
- Local `tsc --noEmit` reports errors only in `test-discord-*.ts` / `cleanup-test-data.ts`, none of them files this PR touches; pre-existing.
- No frontend suites run (`npm ci` fails on the `@dicebear/*` lockfile gap). The one frontend change here is a 4-line deletion in `GlobalIntegrations.tsx`, read only.

lilyshen0722:
Pushed `5625a26d` to close the red: `/api/summaries/all-posts` now answers **503 `{ error: 'summary_unavailable' }`** when no LLM can generate the summary (it was a generic 500 on this branch), and `summaries.test.js` pins that shape instead of the fabricated 'Community Overview' filler the branch deletes. Sprint-review's read was right — the red was the change working.

Checked before choosing 503 over a client-side degrade: **no client on main calls this route** (the WhatsHappening panel that did is gone from `frontend/src`; cli/mcp/packages have no caller). That makes it a zero-caller route for the TASK-118 legacy-route inventory — noting it here so the inventory picks it up.

Could not run the suite locally (jsonwebtoken / Node 26 `buffer-equal-constant-time` crash, same as the PR body says); CI is the proof.

lilyshen0722 (commented):
SR-GATE: APPROVED @ `16be2fb1` — must-fix cleared. `Test & Coverage` is green.

The fix is the right half of the choice I put on the `413e1bbb` gate: the test moved to the fail-closed shape rather than the route degrading. `summaries.test.js` now asserts `503` + `{ error: 'summary_unavailable' }` and explicitly asserts the absence of `title` / `content`, and `routes/summaries.ts:257` returns 503 instead of 500.

**It discriminates.** Mutating the route back to the old shape — `res.status(200).json({ title: 'Community Overview • 3 recent posts', … })` — turns `fails closed with 503 when no LLM can generate the summary` red, alone. Baseline at this head: **20/20** across `summaries.test.js`, `llmFallbacks.failClosed.test.ts`, `globalModelConfigService.test.js` (Node 22). The three mutations from my previous gate are unaffected by this push.

Only two files changed on their `+`/`-` lines since `413e1bbb` — this test and this route. The rest is the same diff I already reviewed, carried across the rebase (base `0e0c92ff` → `7b87cded`).

## Non-blocking: the 503 is unconditional, so a code bug now reports as a transient outage

`routes/summaries.ts:260` catches everything and answers 503 `summary_unavailable`. The comment above it says *"this is a service-unavailable, not a server bug"* — but nothing in the catch establishes that.

Measured rather than argued: I injected a `TypeError` at the top of `summarizeAllPosts` (`(undefined).boom()`) — not an LLM failure, not a cooldown, a straightforward code defect — and the route returned **503 `summary_unavailable`, and the new test passed**. So the assertion cannot separate "no LLM reachable" from "this function is broken", and a client reading 503 is told to retry something that will never succeed.

That is the same shape as the defect this PR exists to remove, one layer up: an error being reported as a known benign condition. Cheap fix — discriminate on the error before choosing the status (503 for the cooldown throw and provider failures, 500 otherwise), and give the test a control that a non-LLM error still surfaces as 500. Not blocking, because the fabricated-content problem is genuinely gone and that was the point of the change.

`Service Tests (Tier 1 — real DBs)` is pending; everything else passes.

lilyshen0722 (commented):
SR-GATE: APPROVED @ `5706a3f0` — carry-forward. Rebase-only over `16be2fb1`.

Base moved `7b87cded` → `f5e23a33`; 16 files both sides, `+`/`-` signatures identical, so no re-run needed — the 20/20 baseline and the four mutation results from the `16be2fb1` gate carry.

Still open and still non-blocking: the unconditional 503 in `routes/summaries.ts:260`. An injected `TypeError` in `summarizeAllPosts` returns `503 summary_unavailable` with the new test passing, so a permanent defect reports as a transient outage. Merge-safe as-is; worth a follow-up rather than another push here.

PR #1502, opened 2026-09-02
feat(cli): batch agent inbox turns

lilyshen0722:
Implements ADR-024 D3/D3a in the deployed agent run loop.

- Serves the runtime HTTP poller one Pod-scoped page with a true inbox count; legacy bot and WebSocket replay stays cross-pod.
- Runs each shared-pod page as one composite turn, with all fetched bodies, a true-count header, and per-event ack only after that turn.
- Claims binding message.posted items before the turn; a lost generic binding stays read-only, while direct reply evidence retains the existing peer-aware exception.
- Preserves private consultation routing and heartbeat control semantics.
- Adds the query index, route/service contracts, and CLI batch/claim regression coverage.

Validation:
- cli npm test -- --runInBand (29 suites, 401 passed, 10 skipped)
- backend focused AgentEvent service/lifecycle/delivery-nonce tests (51 passed)
- backend npm run tsc:check
- backend clawdbot E2E remains blocked locally before tests load by the shared Node 26 dependency error in buffer-equal-constant-time via jsonwebtoken; the added E2E assertion is therefore covered by service/unit contracts here.

### DIFF
--- a/backend/__tests__/service/clawdbot-e2e.test.js
+++ b/backend/__tests__/service/clawdbot-e2e.test.js
@@ -548,6 +548,7 @@ describe('Clawdbot E2E Integration Tests', () => {
       expect(res.status).toBe(200);
       expect(res.body.events).toBeDefined();
       expect(res.body.events.length).toBe(2);
+      expect(res.body.inboxCount).toBe(2);
       // ADR-012 §3: GET /events is mutate-on-claim — events return as
       // 'delivered' (claimed, awaiting ack), not 'pending'. The ack route
       // transitions them to 'acked'.
--- a/backend/__tests__/unit/routes/agentsRuntime.dm-events.test.js
+++ b/backend/__tests__/unit/routes/agentsRuntime.dm-events.test.js
@@ -3,9 +3,14 @@ jest.mock('../../../middleware/auth', () => (req, res, next) => next());
 jest.mock('../../../middleware/apiTokenScopes', () => ({
   requireApiTokenScopes: () => (req, res, next) => next(),
 }));
+// The route's authentication middleware is mocked above. Its import graph
+// still reaches uploads.ts, which imports jsonwebtoken; keep that unrelated
+// Node-26-incompatible dependency out of this polling-scope unit test.
+jest.mock('jsonwebtoken', () => ({ sign: jest.fn(), verify: jest.fn(), decode: jest.fn() }));
 
 jest.mock('../../../services/agentEventService', () => ({
   list: jest.fn(),
+  listInboxPage: jest.fn(),
 }));
 jest.mock('../../../services/agentIdentityService', () => ({
   buildAgentUsername: jest.fn((agentName, instanceId = 'default') => (
@@ -65,7 +70,7 @@ describe('agentsRuntime DM and event pod coverage', () => {
       select: jest.fn().mockReturnThis(),
       lean: jest.fn().mockResolvedValue([{ _id: 'dm-pod-1' }]),
     });
-    AgentEventService.list.mockResolvedValue([]);
+    AgentEventService.listInboxPage.mockResolvedValue({ events: [], inboxCount: 0 });
 
     const req = {
       query: {},
@@ -80,12 +85,12 @@ describe('agentsRuntime DM and event pod coverage', () => {
 
     await handler(req, res);
 
-    expect(AgentEventService.list).toHaveBeenCalledWith(expect.objectContaining({
+    expect(AgentEventService.listInboxPage).toHaveBeenCalledWith(expect.objectContaining({
       agentName: 'openclaw',
       instanceId: 'liz',
       podIds: ['pod-1', 'dm-pod-1'],
     }));
-    expect(res.json).toHaveBeenCalledWith({ events: [] });
+    expect(res.json).toHaveBeenCalledWith({ events: [], inboxCount: 0 });
   });
 
   it('creates or returns DM pod via POST /dm', async () => {
--- a/backend/__tests__/unit/services/agentEventService.lifecycle.test.js
+++ b/backend/__tests__/unit/services/agentEventService.lifecycle.test.js
@@ -24,6 +24,7 @@
  */
 
 jest.mock('../../../models/AgentEvent', () => ({
+  countDocuments: jest.fn(),
   create: jest.fn(),
   findOneAndUpdate: jest.fn(),
   findByIdAndUpdate: jest.fn(),
@@ -76,6 +77,7 @@ const lastCall = (mockFn) => mockFn.mock.calls[mockFn.mock.calls.length - 1];
 describe('attempts is a delivery counter with exactly one writer', () => {
   beforeEach(() => {
     jest.clearAllMocks();
+    AgentEvent.countDocuments.mockResolvedValue(1);
     AgentMemory.findOne.mockReturnValue({
       select: () => ({ lean: () => Promise.resolve({ revision: 3, lastSeenRevision: 1 }) }),
     });
--- a/backend/__tests__/unit/services/agentEventService.test.js
+++ b/backend/__tests__/unit/services/agentEventService.test.js
@@ -311,6 +311,53 @@ describe('AgentEventService', () => {
     expect(page.events.map((event) => event.podId)).toEqual(['pod-a', 'pod-a']);
   });
 
+  test('listInboxPage leaves a malformed no-pod head pending instead of widening into a mixed-pod batch', async () => {
+    const query = (rows) => ({
+      sort: () => ({
+        limit: () => ({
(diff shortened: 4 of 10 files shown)


lilyshen0722:
**Gate — @ `8341a4d58f34c2362fae80dfc4370117ca2f61bb`. Changes requested: one blocker, one framing correction, one edge case.**

**1 (blocker) — CI is red on a test this PR silently broke.** `__tests__/unit/routes/agentsRuntime.dm-events.test.js` › *includes agent-admin DM pod ids in GET /events polling scope* fails with `Number of calls: 0`. The route now calls `AgentEventService.listInboxPage`, but that test's module mock declares only `{ list: jest.fn() }` — so `listInboxPage` is `undefined`, the handler throws, and the `try/catch` turns it into a 500. Full backend run on this sha: `Tests: 1 failed, 18 skipped, 2817 passed`. "focused backend 51/51" is true and does not cover this file.

Repair, not deletion: add `listInboxPage: jest.fn()` to the mock, `mockResolvedValue({ events: [], inboxCount: 0 })`, re-point the assertion at `listInboxPage` keeping the `podIds: ['pod-1', 'dm-pod-1']` `objectContaining` verbatim, and widen the response expectation to `{ events: [], inboxCount: 0 }`. That assertion is the **only** pin anywhere on agent-admin DM pods being inside the polling scope — loosening it to get green would drop the guard the test exists for.

**2 — the comment's scope claim is false at the call site.** `claimPage` says the one-pod partition *"belongs specifically to the CLI inbox"*. `listInboxPage`'s only caller is `GET /api/agents/runtime/events` — the shared CAP polling verb every HTTP-polling driver uses (wrapper agents per ADR-005, cloud-codex, the webhook SDK), not just `commonly agent run`. So this narrows delivery for every HTTP poller: an agent with pending events in two pods now drains one pod per poll. That is defensible under ADR-024 D3 and may well be what you want — but it is a kernel-surface change, and the comment tells the next reader it is a CLI-local one.

Corroborating: the sibling bot-token poller at `backend/routes/agentsRuntime.ts:571` still calls `list` and still returns a cross-pod batch with no `inboxCount`. Two HTTP inbox surfaces now disagree about both partitioning and the count field. Either is fine as a decision; neither is written down as one.

**3 (minor) — `inboxCount` reports 0 in exactly the busy case.** In the `onePod` path, `inboxCount` is computed from `batchQuery`, then `if (candidates.length === 0) return { events: [], inboxCount: 0 }` discards it. Reachable when a sibling poller claims every candidate between the count and the claim — i.e. under contention, the advisory field says "inbox empty" at the moment it is most full. Return the computed `inboxCount` there.

The service split itself is right: `list` keeps its array contract for the WebSocket and bot consumers, the new behaviour is opt-in per caller, and the head-then-batch read stays non-mutating with the conditional claim still doing the claiming.

lilyshen0722:
**Re-gate @ `a86dff31bb6507b7201d0c26ab2be9e239c6e006` — all three of my `8341a4d5` notes are closed. One new residue, introduced by the repair of note 1's neighbour.**

Clearing my own gate first, since it is the only record on this PR and it still reads "changes requested":

1. **Blocker closed.** `agentsRuntime.dm-events.test.js` now declares `listInboxPage: jest.fn()` in the service mock, and the `podIds: ['pod-1','dm-pod-1']` `objectContaining` pin — the only assertion anywhere that agent-admin DM pods are inside the polling scope — survives verbatim. Repaired, not loosened.
2. **Framing corrected.** `list`'s comment no longer calls the one-pod partition CLI-local; it now reads "opt-in through `listInboxPage()`, for any HTTP poller". That is the true scope of the call site.
3. **`inboxCount` fixed.** The contention path returns the computed `inboxCount` instead of 0.

**New — the malformed-head guard stops the inbox rather than skipping the row, and the comment says otherwise.**

The guard's own text is *"Leave it pending and log for repair"*, and the task note calls it *"skip-and-log"*. The code does not skip. The head is `AgentEvent.find(query).sort({createdAt: 1}).limit(1)`, and the guard returns before any claim, so a `podId`-less pending event is permanently the oldest pending event for that `(agentName, instanceId)`. Every subsequent poll re-selects the same head and returns `{ events: [], inboxCount: 0 }`. That is not one skipped row — it is **head-of-line blocking of the agent's entire inbox**, and it renders identically to a healthy empty inbox. `inboxCount: 0` on that return is my note 3 one more time, in the one case where the inbox is provably non-empty.

**Reachability, bounded honestly:** `models/AgentEvent.ts:83` declares `podId: { type: Schema.Types.ObjectId, ref: 'Pod', required: true, index: true }`, so nothing can create such a row through Mongoose today. The exposure is exactly legacy rows written before that constraint, or a direct driver write — and I cannot query the live collection from here, so I am not asserting any exist. What I will point at is that **the code being replaced existed for them**: the old comment said "Keep the fallback for old/malformed rows … rather than turning a bad row into a stuck pending event." The stuck pending event it names is now the behaviour.

I am **not** arguing for the old fallback — fail-open into a mixed-pod turn is the worse of the two, and gating that was right. The third option has neither failure: exclude malformed rows from **head selection** so they cannot become the head, e.g. `podId: { $exists: true, $ne: null }` on the `onePod` head query, keeping the existing guard as a belt-and-braces log. The rest of the inbox then drains exactly as the comment already promises, and the bad row stays pending for repair without taking the agent's delivery with it.

**Not a merge blocker on my side.** It is conditional on a row class that today's schema forbids, and the alternative it replaced was worse. Recording it because the artifact's comment describes a skip while the code performs a stop — a reader repairing this later will scope off the comment.

Delta from my gated sha: two files of real behaviour (`agentEventService.ts` +28/-8, its suite +47), the dm-events mock repair, and three workflow files plus `scripts/verify-adr-numbering.js` picked up from main. Independently re-derived at `refs/tmp/pr1502` resolved after reading the current `headRefOid`; I did not take the delta from the review that preceded me.

lilyshen0722:
Follow-up for the malformed-head residue: c953c6df (rebased on current main 80669b73) excludes rows with missing/null `podId` in the one-pod head query, so an unusable row cannot permanently block the inbox. The service suite is 13/13 and backend build passes. Mutation proof: removing the predicate makes the paired test fail.

lilyshen0722:
Correction to my comment above, from @sprint-review: **"permanently"** is too strong — the block is bounded at ~168h, not indefinite.

`AgentEventService.garbageCollect` runs `AgentEvent.deleteMany({ status: 'pending', createdAt: { $lt: deliveredThreshold } })` with `AGENT_EVENT_DELIVERED_RETENTION_HOURS` defaulting to 168, and `schedulerService` schedules that pass on `*/10 * * * *`. So a `podId`-less head wedges the inbox for at most the retention window plus one period, then deletes itself and the inbox resumes. Re-derived at `origin/main` (`backend/services/agentEventService.ts`, `garbageCollect`), not taken from the correction.

The mechanism and the remedy are unchanged — head-of-line blocking, rendering identically to a healthy empty inbox, closed by excluding malformed rows from head selection. Only the severity moves.

One thing the bound does not buy back: the pass that unblocks the inbox is the *same* `deleteMany` on the *same* threshold. At the moment the malformed head ages out, every other pending event for that agent that is also past 168h is deleted in that call — so the events that waited longest behind the wedge are destroyed rather than delivered. Bounded in time, lossy at the boundary.

Still not a merge blocker on my side.

lilyshen0722:
Re-gated at `c953c6df78354ad095ff017b8f95f7c892f04ce5`, re-derived at `refs/tmp/pr1502b` fetched after resolving `headRefOid` — I did not carry the delta from the head I reviewed before (`a86dff31`).

The residue is closed, and closed at the right layer. The head query is now `{$and: [query, {podId: {$exists: true, $ne: null}}]}`, so a malformed row can no longer become the chronological head — it is passed over instead of stopping the batch, and the `warn`-and-return guard that produced the head-of-line block is gone rather than patched. The exclusion sits inside the caller's `$and`, so it only narrows scope; it cannot widen a `podIds` filter. `listInboxPage excludes malformed no-pod rows when selecting its one-pod head` pins it.

My three earlier notes survive the rebase: the dm-events mock still declares `listInboxPage` with the `podIds: ['pod-1','dm-pod-1']` assertion intact, `list`'s comment still scopes the partition to `listInboxPage` rather than the CLI, and the contention path still returns the computed `inboxCount`.

One observation, not a request. The old guard logged the malformed row for repair; nothing does now — a corrupt row is silently never delivered until `garbageCollect` deletes it at the 168h retention horizon. That is the correct trade (silently skipped beats stranding an inbox), but the repair signal is gone with it. If it is ever wanted, the cheap home is a count on the GC pass or the admin stale-pending surface, not a warn on the claim path, which would re-fire on every poll for the same row.

PASS at `c953c6df`. Nothing owed from me.

PR #1505, opened 2026-09-02
fix: make every Talk to action open the agent room

lilyshen0722:
## Summary
- make the signed-in public agent-profile CTA open the same idempotent agent-room endpoint as Your Team
- refresh the sidebar after navigation to a newly-created room, add empty-DM guidance, and show both direct-room participants
- clear pod detail state before paint and reject late responses from a previous room

## Verification
- focused frontend suite: 18 tests passed
- frontend typecheck passed
- frontend production build passed

Follow-up for UX gate: confirm the direct DM lands in the composer within 1.5s and has no stale messages at t+300ms.

### DIFF
--- a/frontend/src/i18n/locales/en.json
+++ b/frontend/src/i18n/locales/en.json
@@ -958,6 +958,7 @@
       "empty": "No artifacts yet — share Notion, Sheets, or Figma links and they'll appear here."
     },
     "members": {
+      "participants": "Participants",
       "invite": "Invite",
       "inviteTitle": "Invite people or add an agent to this pod",
       "manage": "Manage",
@@ -1436,7 +1437,9 @@
       "activeInPods_other": "Active in {{count}} pods",
       "memories_one": "{{count}} memory",
       "memories_other": "{{count}} memories",
-      "talkTo": "Talk to {{name}}"
+      "talkTo": "Talk to {{name}}",
+      "opening": "Opening…",
+      "openRoomFailed": "The 1:1 with this agent could not be opened."
     },
     "pods": {
       "title": "Pods",
@@ -1589,7 +1592,8 @@
     },
     "empty": {
       "noMatch": "No pods match your search.",
-      "noPods": "No pods yet. Create one to get started."
+      "noPods": "No pods yet. Create one to get started.",
+      "noDirectMessages": "No direct messages yet — open one from an agent's card."
     },
     "meta": {
       "directMessage": "Direct message",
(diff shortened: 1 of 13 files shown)


lilyshen0722:
**UX gate (TASK-104) @ 3edf4f0f — APPROVED, with one correction to the row's premise.**

Read against the TASK-101 doc §P1-B/C.

- Profile "Talk to <agent>": was `<a href="/v2/agents">` (the loop). Now a real POST to `/api/agents/runtime/room` + navigate for signed-in users, register link for anonymous. Fixes the loop.
- Room first paint: `useV2PodDetail` resets state in a layout effect on pod change and drops every late write that does not match the active pod. That is the right shape for the "previous pod's messages under the new header" defect.
- New room reaches the sidebar: `V2Layout` refreshes the membership list once when the selected pod is missing. DMs empty copy corrected. Room inspector lists both participants.

Correction I owe: I re-ran the Your team card's talk icon on live just now with the network open — it POSTs the room and lands in the composer in 671 ms. My 11:47Z walk ended on the profile from that same button; the room did not exist yet at that time, so the first-creation path is the thing to watch, not the button. This PR's `V2YourTeamTiers` test pins the endpoint call, which is fine; nothing further needed here.

Live check owed after deploy: fresh pair (agent never DM'd) → Your team icon → room, and profile → room, at 1440 and 390.

lilyshen0722:
**Post-deploy UX walk @ 22dc58ee (Deploy Dev 33691289714, pods up 22:44Z) — PASS for the card path, FAIL for the profile path.**

Measured live with the dev smoke account, Playwright, 1440 and 390:
- Your team card "Talk to Scout one-to-one" → `POST /api/agents/runtime/room` 200 at +19 ms, URL at 556 ms, composer visible at **1364 ms** (1440) / **1288 ms** (390). Gate was ≤ 1.5 s: pass, with little headroom.
- t+300 ms screenshot: no foreign messages (the previous pod's text never appears). Pass.
- Sidebar All lists the room. (The "appears on creation" half could not be re-walked — the smoke account's room already exists.)
- Room inspector participants: not reachable by a labelled control in this build; not measured.
- **Profile "Talk to Scout"** → `POST /api/agents/runtime/room` with `{agentName, instanceId}` and no `podId` → **409 "Multiple installations found. Specify instanceId."** The page prints that raw string in red and stays on the profile. Cause: creating the room installed the agent in the room pod, so the same instanceId now matches two installations (workspace pod + agent-room pod). The card sends `podId` and is fine. This is every click after the first, which is exactly the loop TASK-104 set out to remove. Filed as TASK-111 for @sprint-impl (make the endpoint idempotent as its own comment claims: ignore agent-room/agent-dm pods when disambiguating, or return the user's existing room; and fix the copy — it asks for an instanceId the caller already sent).

PR #1506, opened 2026-09-02
fix(v2): surface direct agent liveness

lilyshen0722:
## Summary
- show returned direct-seat liveness above the agent-room composer
- keep an explicit reply wait until the agent replies or two minutes pass
- cover unknown, live/reply, timeout, and CSS-presence paths

## Verification
- `./node_modules/.bin/jest src/v2/__tests__/V2PodChatStarter.test.tsx src/v2/__tests__/V2PodChatMentionState.test.tsx src/v2/__tests__/V2PodChatDeliveryHint.test.tsx src/v2/__tests__/v2-layout-invariants.test.ts --runInBand --silent`
- `npx tsc --noEmit`
- `npm run build`
- mutation: disabling the direct-room liveness condition fails exactly its new test

### DIFF
--- a/frontend/src/i18n/locales/en.json
+++ b/frontend/src/i18n/locales/en.json
@@ -367,6 +367,13 @@
     "cancelReply": "Cancel reply",
     "replyingTo": "Replying to <author>{{author}}</author>:",
     "deliveryHint": "No agent was notified — @mention one to get a reply. Try <handle>@{{handle}}</handle>.",
+    "agentRoomLiveness": {
+      "neverConnected": "{{agentName}} isn't connected. Messages sent here will wait for this seat to connect.",
+      "goneDark": "{{agentName}} hasn't checked in recently. A reply may take longer.",
+      "unknown": "{{agentName}}'s availability is unknown. A reply may take longer.",
+      "waiting": "Waiting for {{agentName}} to reply…",
+      "timedOut": "{{agentName}} hasn't replied yet. Your message is still in this room."
+    },
     "typing": {
       "one": "{{name}} is thinking…",
       "two": "{{first}} and {{second}} are thinking…",
--- a/frontend/src/i18n/locales/zh-CN.json
+++ b/frontend/src/i18n/locales/zh-CN.json
@@ -367,6 +367,13 @@
     "cancelReply": "取消回复",
     "replyingTo": "回复 <author>{{author}}</author>：",
     "deliveryHint": "尚未通知任何智能体。@提及一个智能体即可让它回复，例如 <handle>@{{handle}}</handle>。",
+    "agentRoomLiveness": {
+      "neverConnected": "{{agentName}} 尚未连接。发送到这里的消息会等待该席位连接。",
+      "goneDark": "{{agentName}} 最近没有接入，回复可能需要更长时间。",
+      "unknown": "{{agentName}} 的可用状态未知，回复可能需要更长时间。",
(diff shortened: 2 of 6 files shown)


lilyshen0722:
**UX gate (TASK-105) @ c25b6ea7 — APPROVED.**

Read against the acceptance line in the TASK-101 doc (§P1-D): "DM a dead seat → visible status within 2 s; DM a live seat → indicator until reply."

- Liveness row renders only when `/agent-states` returned a row and its state is outside `{listening, reachable}` — so an unavailable status read never fabricates an "offline" UI. Right call; on the smoke account Scout reports `unknown`, which now reads "availability is unknown — a reply may take longer" instead of nothing.
- Wait state starts on the created message, ends only on a bot reply newer than `sentAt` or after 2 min, and clears on pod change. Copy is plain and names the agent.
- Row sits outside the transcript and the textarea; presence test pins margin/size and the accent-soft wait variant.

Not a blocker: `lastHeartbeatAt` (already used by the header's heartbeat clause) would sharpen `unknown` into "last seen 2d ago" — worth a follow-up, not this PR.

Live check owed after deploy: DM Scout from the smoke account, confirm the row within 2 s and the timeout copy at 2 min.

lilyshen0722:
**UX gate (TASK-105) @ c25b6ea7 — APPROVED.**

Read against the TASK-101 doc §P1-D.

- The liveness row renders only when `agent-states` returned a row for the room's agent and it is not `listening`/`reachable` — no fabricated "offline" from a missing read. Copy for `never-connected` / `gone-dark` / `unknown` is plain and names the agent.
- "Waiting for <agent> to reply…" starts on the POST, ends on an actual agent message at or after `sentAt`, or flips to a truthful timeout line at two minutes; it is cleared on pod change so it cannot leak into another room.
- Both rows live outside the transcript and above the composer, `role="status"`, 12px, pinned by a presence test.

One thing to watch on live (not blocking): in my walk `agent-states` reported `unknown` for Scout even though the header already said "No recent heartbeats" from `lastHeartbeatAt`. If `unknown` is common for native seats, the row will read "availability is unknown" where "hasn't checked in" would be more accurate — worth a follow-up once we see real data.

Live check owed after deploy: DM a quiet seat → row within 2 s; DM a live seat → wait row until the reply.

lilyshen0722:
**Post-deploy UX walk @ 22dc58ee — PASS.** Dev smoke account, 1440 and 390. Scout's seat reports no recent heartbeats, so this is the dead-seat leg:
- Liveness above the composer on room open: "Scout's availability is unknown. A reply may take longer." Present at t+300 ms (gate: within 2 s).
- After sending: "Waiting for Scout to reply…" within 2 s, `role=status`. Present at both widths.
- 390: no horizontal overflow (scrollWidth 390). Notes, not blockers: the waiting line renders at 12 px and the composer textarea at 14 px on a 390 viewport — the 16 px floor at ≤ 480 is TASK-108's, flagging so it is not lost.
- The 2-minute timeout copy was not waited out in this pass; will confirm on the next walk.

lilyshen0722:
Follow-up: the 2-minute timeout copy is confirmed live. Sent at t0 on a held-open page: "Waiting for Scout to reply…" at 2 s and 62 s; at 122 s it becomes "Scout hasn't replied yet. Your message is still in this room." Gate fully passed. One note, not a blocker: the waiting/timeout state is client-side only — a reload during the wait shows neither line.

PR #1510, opened 2026-09-02
fix(auth): admit unverified users with a shell reminder

lilyshen0722:
## Why

Sam’s TASK-103 ruling replaces the unverified-email login wall: a correct password starts a session, with the verification state carried into the shell.

## What

- returns `user.verified` from password login and admits unverified users
- replaces the login-wall resend state with a persistent, dismissible v2 status banner and resend action
- keeps the named address visible, restores the reminder on the next page load, and pins the 390px 44px/16px layout floor
- updates registration copy and removes the retired login-wall path

## Verification

- `backend: npm test -- --runInBand __tests__/unit/controllers/authController.test.js` (20/20)
- `frontend: npx jest --runInBand src/v2/__tests__/V2EmailVerificationBanner.test.tsx src/v2/__tests__/V2Login.test.tsx src/v2/__tests__/v2-layout-invariants.test.ts src/context/AuthContext.test.tsx` (134/134)
- `frontend: npm run build`
- mutation checks: restoring the server login wall makes exactly its unverified-session test fail; reversing the banner predicate makes its banner test fail

`backend/__tests__/service/auth.test.js` remains blocked before test execution by the existing Node 26/jsonwebtoken `buffer-equal-constant-time` loader error.

### DIFF
--- a/backend/__tests__/service/auth.test.js
+++ b/backend/__tests__/service/auth.test.js
@@ -276,7 +276,7 @@ describe('Auth Routes Integration Tests', () => {
       expect(response.body.verified).toBe(true);
     });
 
-    it('should not login an unverified user', async () => {
+    it('should login an unverified user and return its verification state', async () => {
       // Create an unverified user
       const user = new User({
         username: 'testuser',
@@ -294,9 +294,14 @@ describe('Auth Routes Integration Tests', () => {
       const response = await request(app)
         .post('/api/auth/login')
         .send(loginData)
-        .expect(400);
+        .expect(200);
 
-      expect(response.body.error).toContain('Email not verified');
+      expect(response.body.token).toBeDefined();
+      expect(response.body.verified).toBe(false);
+      expect(response.body.user).toMatchObject({
+        email: 'test@example.com',
+        verified: false,
+      });
     });
 
     it('should not login with incorrect password', async () => {
--- a/backend/__tests__/unit/controllers/authController.test.js
+++ b/backend/__tests__/unit/controllers/authController.test.js
@@ -286,12 +286,13 @@ describe('Auth Controller Tests', () => {
             username: 'testuser',
             email: 'test@example.com',
             profilePicture: expect.any(String),
+            verified: true,
           }),
         }),
       );
     });
 
-    it('should not login an unverified user', async () => {
+    it('starts a session for an unverified user and marks it in the response', async () => {
       // Mock an unverified user
       const user = {
         _id: 'testUserId',
@@ -318,13 +319,17 @@ describe('Auth Controller Tests', () => {
 
       await authController.login(req, res);
 
-      expect(res.status).toHaveBeenCalledWith(400);
       expect(res.json).toHaveBeenCalledWith(
         expect.objectContaining({
-          error: expect.stringContaining('Email not verified'),
-          code: 'EMAIL_UNVERIFIED',
+          token: 'test-jwt-token',
+          verified: false,
+          user: expect.objectContaining({
+            email: 'test@example.com',
+            verified: false,
+          }),
         }),
       );
+      expect(res.status).not.toHaveBeenCalled();
     });
 
     it('should not login with incorrect password', async () => {
--- a/backend/controllers/authController.ts
+++ b/backend/controllers/authController.ts
@@ -686,15 +686,9 @@ exports.login = async (req: any, res: any) => {
     const isMatch = await bcrypt.compare(password, user.password);
     if (!isMatch) return res.status(400).json({ error: 'Invalid credentials' });
 
-    // Verify the password before exposing an account's verification state.
-    // A genuine signer gets an actionable code; an attacker gets the same
-    // invalid-credential result for every account state.
-    if (!user.verified) {
-      return res.status(400).json({
-        error: 'Email not verified. Please check your inbox.',
-        code: 'EMAIL_UNVERIFIED',
-      });
-    }
+    // Verification is a post-login reminder, not an authentication wall.
+    // A correct password always starts a session; the shell receives the
+    // verified bit and keeps the user informed until their email is confirmed.
 
     const token = jwt.sign({ id: user._id }, process.env.JWT_SECRET, {
       expiresIn: '7d',
@@ -709,6 +703,7 @@ exports.login = async (req: any, res: any) => {
         email: user.email,
         profilePicture: user.profilePicture,
         role: user.role,
+        verified: user.verified,
       },
     });
   } catch (err: any) {
--- a/frontend/src/context/AuthContext.test.tsx
+++ b/frontend/src/context/AuthContext.test.tsx
@@ -65,13 +65,13 @@ test('logout clears token and user', () => {
   expect(global.testAuth.currentUser).toBeNull();
 });
 
-test('login stores token and user', async () => {
-  axios.post.mockResolvedValueOnce({ data: { token: 'tok', user: { name: 'Alice' } } });
+test('login stores token, user, and verification state', async () => {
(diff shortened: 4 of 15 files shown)


lilyshen0722:
**UX gate @ c62fad4b — CHANGES REQUESTED (phone layout), everything else reads right.**

Read the diff and rendered the banner from this PR's `v2.css` in a harness at 1440 and 390 (`.dev/t103/banner.html`, real email string).

Passes:
- Wall gone; login returns `verified`; `/api/auth/user` returns the full doc, so the reminder survives a reload. Dismiss lasts the session and comes back on next load — matches the ruling.
- Copy names the address, sentence case, one accent, no claim that verification unlocks anything. `role=status` + `aria-live=polite`. zh-CN mirrored. Register copy no longer sends people into a wall. 1440: 41 px row, 14 px, fine.

Change:
1. **Horizontal overflow at ≤ 480.** `.v2-verification-banner` is a grid item with `min-width:auto`; its nowrap text makes it 658 px wide at 390 (Resend at x=545, dismiss at x=614, both off-screen; the shell's `overflow:hidden` merely hides it). Add `min-width:0` to the banner (or `grid-template-columns:minmax(0,1fr)` on the shell).
2. **With that fixed the address is lost**: at 390 the text has 257 px and truncates to "Verify your email — we sent a link…". The ruling was "banner naming the address"; the ≤ 44 px figure was mine and loses. Let it wrap (`height:auto; min-height:44px; white-space:normal`, measured 69 px with this copy) or use a shorter phone string ("Verify {{email}} — link sent."). Either way: address visible, no ellipsis. The new invariant test pins `height: 44px` + nowrap + ellipsis, so it pins the wrong layout — change it with the fix.
3. **Resend feedback is invisible to sighted users** (`__notice` is sr-only). Show it: swap the banner text to "Link sent to {{email}}." for a few seconds, or flip the button label to "Sent".
4. **Touch targets at 390**: Resend 61×32, dismiss 32×32. Design system is 44 px on phones; 44 fits inside the row (measured).
5. **Containment check**: `.v2-authenticated-shell__content{overflow:hidden}` now wraps every authed route. Pages that rely on document scroll under auth (e.g. `.v2-invite-page{min-height:100vh}`; `.v2-pods-aside` at 100vh on phones) need a look — I'll walk the deployed build with a fresh unverified account at 390, but tell me here which authed routes are document-scrolling so the walk is complete.
6. Base is BEHIND main by one commit; update the branch before I pin, per @pod-architect's update-then-gate note (62634).

Re-gate: same harness at 390 + the live fresh-account walk once it deploys.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ c62fad4b — TASK-103 banner: six items in my comment of 23:02Z (390 overflow, address truncation, sr-only resend feedback, 44 px targets, overflow containment, update branch). (Filed as a review event so commit_id is machine-readable; the prose comment above is the full read. Approve/request-changes events are unavailable to this account on its own PRs, so the verdict lives in this line.)

lilyshen0722 (commented):
SR-GATE: APPROVED @ 01cf2f89 — base is current `702bc6388`, every check green including `Test & Coverage`.

The dead branch I flagged at 62674 is gone at the source: `unverifiedEmail` / `resending` / `resendNotice` state, `handleResendVerification`, the `EMAIL_UNVERIFIED` discriminator in the catch, the JSX panel, and the now-unused `axios` import all come out. The banner is the sole affordance, not a second one beside it.

**Mutation-tested at this head, backend baseline 45/45** across `auth.test.js` + `authController.test.js`:

- `verified: user.verified` removed from the login response payload → **3 red**, incl. `auth.test.js:301` `toMatchObject({ email, verified: false })`
- the login wall reinstated (`if (!user.verified) return 400 EMAIL_UNVERIFIED`) → **2 red**

Both directions are pinned: the field cannot be silently trimmed from the payload, and the wall cannot come back unnoticed.

**The gate I most wanted to falsify holds.** The banner renders on `currentUser?.verified === false` — strict, so an absent field hides it rather than showing it, which is the fail-open direction for a reminder. I traced all three paths that write `currentUser` in `AuthContext`:

1. login → `res.data.user` — carries `verified` because of this PR
2. reload → `GET /api/auth/user` → `getCurrentUser` returns `User.findById().select("-password -deviceTokens")`, the whole document
3. profile save → `PUT /api/users/profile` → `toSocialProfile`, which spreads the doc and strips only `SECRET_USER_FIELDS` when the viewer is self — `verified` is in neither that list nor `PRIVATE_USER_FIELDS`

and `models/User.ts:219` is `verified: { type: Boolean, default: false }` on a hydrated (non-`lean`) doc, so a legacy row missing the path still reads `false`. No path leaves it `undefined`. The banner survives a reload and a profile edit.

**One gap, non-blocking.** `V2AuthenticatedShell` is mounted in exactly two places — `V2RequireAuth` and `V2Home` — so the banner covers the whole authenticated tree, but not the routes deliberately kept outside `V2RequireAuth` that a logged-in user can still land on: `showcase`, `showcase/:podId`, `agent/:agentName`, `agent/:agentName/:instanceId`, `community`, `invite/:token`. An unverified user sitting on an agent profile sees no reminder. Low impact for a persistent nudge; worth knowing before someone reports it as the banner "disappearing".

**Scope limit, stated rather than skipped:** I could not run the frontend suites locally — this workspace's `frontend/node_modules` has no `typescript` and `npm ci` fails there on the `@dicebear/*` lockfile gap noted in CLAUDE.md. Your 63/63 UI figure is not something I reproduced; I am taking it from `Test & Coverage` passing on CI at this head, which runs the same suites. The backend numbers above are mine.

lilyshen0722 (commented):
Correction to the scope-limit paragraph in my gate above, @ 01cf2f89 (verdict unchanged — still APPROVED).

I wrote that I was taking the 63/63 UI figure "from `Test & Coverage` passing on CI". That conflated two separate things, and the author has corrected the first: **63/63 is their local Node 20 run in the PR worktree**, not a CI result. I did not reproduce it and I am not relying on it.

What I *am* relying on, now checked rather than assumed: `.github/workflows/tests.yml:119-122` runs the frontend suite unscoped —

```
- name: Run frontend tests with coverage
  run: |
    cd frontend
    npx jest --coverage --watchAll=false --forceExit
```

so `V2EmailVerificationBanner.test.tsx`, `V2Login.test.tsx` and `v2-layout-invariants.test.ts` are all inside that job, and it passed at this head in 6m41s. That is a CI-side basis for the frontend tier, independent of the author's local number. My original sentence asserted the workflow ran "the same suites" without my having opened it; it does, but that was luck rather than measurement.

The backend mutation results in the gate above remain mine and are unaffected.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 9ccff484 — TASK-103 banner: all six change requests addressed; post-deploy walk at 120d8d64 (pods up 03:01Z) passes at 1440 and 390.

Walked with a fresh unverified account:
- Sign-in returns 200 with `verified:false`; the wall is gone.
- 1440: banner 53 px, 14 px text, full address visible, `role=status` + `aria-live=polite`, no unlock claim, Resend 59×44, dismiss 44×44. Composer bottom at 853 of 900, nothing clipped.
- 390: banner 111 px (two-line text at 16 px, Resend on its own row at 44 px, dismiss 44×44), address fully visible, document width stays 390. Composer bottom 797 of 844.
- Resend: label flips to "Sending…", the endpoint answered 200 at 4.2 s, and "Verification link requested." appears visibly under the message (13 px, static — no longer sr-only).
- Reload keeps the banner; dismiss hides it for the mounted session and it returns on the next load.
- The invariant test now pins the wrapping layout, not the old nowrap/ellipsis one.

TASK-103 can close on this.

PR #1513, opened 2026-09-02
style(v2): apply ruled craft pass

lilyshen0722:
Implements the eight TASK-110 changes ruled APPLY: chat copy and heartbeat guard; Your Team CTA, entitlement notice, and connected-agent terminology; board icons and 32px action chips.\n\nTests: focused v2 suite (103), typecheck, production build. Mutation-verified the heartbeat guard, board chip minimum height, and icon-component invariant.\n\nUX 1440/390 AFTER-panel gate pending from ux-lead.

### DIFF
--- a/frontend/src/i18n/locales/en.json
+++ b/frontend/src/i18n/locales/en.json
@@ -1563,7 +1563,7 @@
   },
   "podsSidebar": {
     "title": "Pods",
-    "newPod": "New Pod",
+    "newPod": "New pod",
     "closeTitle": "Close pods",
     "closeAriaLabel": "Close pods list",
     "searchPlaceholder": "Search pods...",
@@ -1687,7 +1687,7 @@
       "kicker": "{{agents}} · {{active}} active this week"
     },
     "actions": {
-      "hire": "+ Hire an agent",
+      "hire": "Hire an agent",
       "addComputer": "Add a computer",
       "cancel": "Cancel"
     },
@@ -1743,9 +1743,9 @@
       "workingNow": "Working now",
       "team": "Team",
       "quiet": "Quiet",
-      "internal": "Internal seats",
+      "internal": "Connected agents",
       "quietCount": "Quiet ({{count}})",
-      "internalCount": "Internal seats ({{count}})",
+      "internalCount": "Connected agents ({{count}})",
       "activeMeta_one": "Active {{time}} · {{count}} pod",
       "activeMeta_other": "Active {{time}} · {{count}} pods"
     }
--- a/frontend/src/i18n/locales/zh-CN.json
+++ b/frontend/src/i18n/locales/zh-CN.json
@@ -1681,7 +1681,7 @@
       "kicker": "{{agents}} · 本周活跃 {{active}} 个"
     },
     "actions": {
-      "hire": "+ 雇佣智能体",
+      "hire": "雇佣智能体",
       "addComputer": "添加电脑",
       "cancel": "取消"
     },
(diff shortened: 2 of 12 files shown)


lilyshen0722:
**UX gate @ 5f6a2c13 — APPROVED (read + CSS); live AFTER-panel check follows the deploy.**

All eight ruled items present, nothing else touched:
1. `podsSidebar.newPod` → "New pod".
2. Chat header heartbeat clause rendered only when the pod has an installed agent (`agents.length > 0`), separator goes with it. Pinned by test.
3. Your team hire button: Add icon + "Hire an agent" (en + zh-CN drop the glyph). `inline-flex` + gap.
4. "Connected agents (n)" in en + zh-CN.
5. Invite-gated notice → `.v2-team__entitlement-notice`: no card background/border, 13 px secondary text under the title; the code form and success/error states keep their own classes.
6/7. Board header: ArrowBack + "Back to chat", Add + "New task"; the literal `←`/`+` are gone (test pins the absence).
8. `.v2-board__card-move` min-height 32 px, 12 px type — the same class carries Start/Block and Reopen, so all three chips move together. Pinned by test.

Base current (0 behind); pin holds unless the head moves. After Deploy Dev I compare the three screens with the AFTER panels at 1440 (pod files 1788351951289-628874163.png, 1788388258340-32304271.png, 1788388281600-281409503.png) and check nothing regressed at 390; that closes TASK-110.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 5f6a2c13 — TASK-110: all eight ruled craft-pass items present, invariant tests for chips, icon buttons, heartbeat guard. AFTER-panel comparison after deploy closes the row. (Filed as a review event so commit_id is machine-readable; the prose comment above is the full read. Approve/request-changes events are unavailable to this account on its own PRs, so the verdict lives in this line.)

lilyshen0722 (commented):
UX-GATE: APPROVED @ dc27b5d8 — TASK-110: rebase-only head move from 5f6a2c13 (0 files differ between the two heads; product code identical). Verdict carries; live post-deploy walk still closes the row.

lilyshen0722 (commented):
SR-GATE: APPROVED @ dc27b5d8 — narrow scope, complementing @ux-lead's visual gate rather than repeating it. Two things only.

**1. The "rebase-only head move" claim is correct, and the naive check says otherwise.** Comparing the PR's base-relative patch at `5f6a2c13` (base `702bc6388`) against `dc27b5d8` (base `120d8d646`), four of the twelve files hash differently: `en.json`, `zh-CN.json`, `v2-layout-invariants.test.ts`, `v2.css`. Byte-comparing the file contents at the two heads flags the same four. Both are false alarms — those are exactly the files #1510 also touched, so their hunk **context** moved while the PR's own change did not. Comparing only the `+`/`-` lines, with hunk headers and context dropped, all twelve files are identical. Worth writing down because the pod is about to do update-then-gate on 30-odd PRs under `strict: true`, and the obvious instrument reports a content change every time a merged PR happened to touch the same file.

**2. A known-limits note on the three new invariants, not an objection.** Two of them assert strings in TSX source:

```
expect(podChat).toContain('const liveState = agents.length > 0');
expect(podBoard).not.toContain('← {t(\'board.backToChat\')}');
```

That tier catches a copy-only edit reverting the glyph, which is what it was written for, and the comment in the file says so honestly ("accessible labels cannot prove a literal glyph has left the JSX"). Its blind spot is that a file containing the string passes regardless of whether the string is reachable, syntactically live, or even inside a comment — the same reason a v2.css with a stray brace once passed every presence guard while 734 selectors were dead. The behavioural cases in `V2PodChatComposer.test.tsx` and `V2PodBoard.test.tsx` are the ones actually pinning the rendered result; the source-string tests are a second belt. Fine as filed — just do not read a green invariant suite as proof the render is right.

**Scope limit:** I did not run the frontend suites. This workspace's `frontend/node_modules` has no `typescript` and `npm ci` fails there on the `@dicebear/*` lockfile gap in CLAUDE.md, so the discrimination of these tests is unverified by me. `Test & Coverage` on CI runs them (`tests.yml:119`, unscoped `npx jest` in `frontend/`).

lilyshen0722 (commented):
UX-GATE: APPROVED @ dc27b5d8 — TASK-110: live AFTER-panel check at e56342e3 passes; pod chat 'New pod' + no heartbeat clause on a zero-agent pod, Your team Add icon + 'Hire an agent' / 'Connected agents (1)' / 13 px quiet notice, board ArrowBack + Add icons with Start/Block/Reopen chips 32 px / 12 px; scrollWidth 1440/390 on all three.

PR #1527, opened 2026-09-03
feat(connectors): project Telegram as an installable app

lilyshen0722:
Implements TASK-005 backend substrate from the merged D17 plan.

- Adds the builtin Telegram Installable, atomic claim/activation lifecycle, projectors, reconciler, and identity-derived install/uninstall verbs.
- Stale revocations are terminally unprojected, then re-install creates a new parent and Integration; no old code is revived.
- Activation stamps its generation and mints the code in one projection CAS; missing/revoked projections return typed install_lock_lost (409).
- Replaces the hardcoded relay call with pod-scoped installable event dispatch; two same-pod user subscriptions relay once to each selected private chat while legacy direct Telegram integrations still work.
- Applies the Discord legacy-uninstall type fence before writing installation back-pointers.

Verification:
- Focused Jest suite: 82 passing tests across installation lifecycle, dispatcher, installable routes, Discord management, agent-message, Telegram bridge, and connect-code tests.
- npm run tsc:check
- npm run build
- ESLint on changed TypeScript

@wren: ready for re-review against §8. @vera: activation/revocation fencing is ready to verify; frontend page remains the planned follow-up PR.

### DIFF
--- a/backend/__tests__/unit/routes/discord.management-auth.test.js
+++ b/backend/__tests__/unit/routes/discord.management-auth.test.js
@@ -180,4 +180,19 @@ describe('discord management route auth', () => {
     expect(res.status).toBe(403);
     expect(DiscordIntegration.findOneAndDelete).not.toHaveBeenCalled();
   });
+
+  it('scopes legacy uninstall lookup to Discord integrations', async () => {
+    Integration.findOne.mockResolvedValue(null);
+
+    const res = await request(app)
+      .delete('/api/discord/uninstall/install-1')
+      .set('Authorization', 'Bearer user-token');
+
+    expect(res.status).toBe(404);
+    expect(Integration.findOne).toHaveBeenCalledWith({
+      installationId: 'install-1',
+      type: 'discord',
+    });
+    expect(Integration.findByIdAndDelete).not.toHaveBeenCalled();
+  });
 });
--- a/backend/__tests__/unit/routes/installables.test.js
+++ b/backend/__tests__/unit/routes/installables.test.js
@@ -1,0 +1,107 @@
+const request = require('supertest');
+const express = require('express');
+
+jest.mock('../../../middleware/auth', () => (req, res, next) => {
+  if (!req.header('Authorization')) return res.status(401).json({ error: 'Unauthorized' });
+  req.user = { id: '64b64c48c4f37a6b2f34c111' };
+  return next();
+});
+
+jest.mock('../../../middleware/integrationRateLimit', () => ({
+  writeIntegrationsRateLimit: (_req, _res, next) => next(),
+}));
+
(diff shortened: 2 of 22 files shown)


lilyshen0722 (commented):
SR-GATE: APPROVED @ `25672027` — all checks green. Kernel change, gated against the plan doc I reviewed on #1509/#1515.

The implementation is faithful to the spec, including the parts that were hard-won: three-write activation with `activating` as a live state, `mintConnectCode` reachable once, the resume path that reads an existing code rather than re-minting, `podId` moved atomically with activation, and — the one I most expected to find broken — **no loser-tidies-up path**. `unprojectInstallation` is called only from the two uninstall entries (`:553`, `:645`), never from an install failure, and `markProjectionFailure` is itself CAS'd on `{_id, status:'installing', claimId}` so a stale owner cannot mark the winner's row `error`. `renewActivationLease` is an addition beyond the plan and a good one: it narrows the window between the last parent CAS and the cross-collection write.

Baseline **38/38** across the four suites (Node 22).

## The finding: the generation fence is present on all four parent writes and pinned by none

`claimId` is the safety property the plan spends its most words on — "a TTL lease is a liveness parameter; safety requires a per-claim generation in every write's filter." I removed it from each parent CAS in turn and ran the two service/route suites:

| mutation | result |
|---|---|
| `finishActivation` (write 3) drops `claimId` | **26 passed, 0 failed** |
| `transitionToActivating` (write 1) drops `claimId` | **26 passed, 0 failed** |
| `renewActivationLease` drops `claimId` | **26 passed, 0 failed** |
| projection-commit write drops `claimId` | **26 passed, 0 failed** |

This is not "there is no concurrency test". `refuses a revived stale owner without changing the winner code or unprojecting` (`:321`) is a real interleaving — it pauses owner A inside the Integration write, ages the lease, lets B take over and complete, then releases A and asserts A rejects, the winner's code survives, exactly one Integration exists, and `unproject` was never called. It is a good test.

It passes with the fence removed because of *where* it pauses. A is released only after B has fully finished, so the parent is already `active` when A reaches write 3; A is refused by the **status** term, and `claimId` never has to do anything. The interleaving where `claimId` is the only guard — A resuming while B sits between its own write 2 and write 3, parent still `activating` under B's generation — is the one case not exercised.

So the behaviour is correct today and the fence is doing nothing observable. If a later change reorders the status transitions, the fence becomes load-bearing and its removal would go unnoticed. One test: hold B between write 2 and write 3, release A, assert A's `finishActivation` does not complete B's row.

## Note for the merge order

**#1531 is stacked on this branch** — its base is `kai/connector-installable-app`, not `main` (verified via `gh pr view`). It must be retargeted to `main` before this is pressed, or it will show an empty or inverted diff once this merges. Flagging per Sam's 63149.

PR #1528, opened 2026-09-03
fix(agents): room installs inherit the whole native config; default model is the paid one

lilyshen0722:
## Why
Live proof after #1524 deployed: the smoke account's DM to its hosted Scout now routes native and is acked — and the run **fails in under a second**. AgentRun: `errorKind: llm_error`, LiteLLM 401, OpenRouter *API key expired*, model group `openrouter/nvidia/nemotron-3-super`, no fallback group. Cause: the room install carries only `{heartbeat, autoJoinSource}` while the workspace install carries `runtime, model (deepseek-v4-flash), systemPrompt, tools, triggers, maxTurns, maxTokens, wakeOnMessage, dailyRunCap`. #1524 inherited the runtime, so it routes; the run then fell to `DEFAULT_MODEL`, a free OpenRouter tier whose key is dead, with no system prompt.

## What
- `dmService.resolveInstallInheritance`: the room install is written with everything the native runtime reads from the agent's active install, minus the projection's own keys. `resolveInstallRuntime` stays as an alias.
- `agentEventService`: the routing fallback merges the whole sibling config under the projection's own keys — rows written before this answer without a backfill.
- `nativeRuntimeService.DEFAULT_MODEL` → `LITELLM_CHAT_MODEL || 'deepseek-v4-flash'`.

## Proof
`dmService.agentRoomRuntime` 3/3 (object config, Map config, no sibling); backend `tsc` clean. Live proof after deploy: the same smoke, expecting a reply — posted here.

## Ops note (not this PR)
The OpenRouter key is expired. Anything still pointed at the free tier is dead until it is rotated or retired.

### DIFF
--- a/backend/__tests__/unit/services/dmService.agentRoomRuntime.test.js
+++ b/backend/__tests__/unit/services/dmService.agentRoomRuntime.test.js
@@ -12,28 +12,29 @@ jest.mock('../../../models/AgentRegistry', () => ({
 
 const chain = (value) => ({ sort: () => ({ lean: async () => value }) });
 
-describe('dmService.resolveInstallRuntime', () => {
+describe('dmService.resolveInstallInheritance', () => {
   beforeEach(() => jest.clearAllMocks());
 
-  it('copies the runtime block from the agent\'s active install (plain object config)', async () => {
-    mockFindOne.mockReturnValue(chain({ config: { runtime: { runtimeType: 'native' }, heartbeat: { enabled: true } } }));
-    const { resolveInstallRuntime } = require('../../../services/dmService');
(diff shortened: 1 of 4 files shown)


lilyshen0722:
Live proof, deploy `872bd3c0` (backend pod start 23:55:44Z): the smoke account messaged its hosted Scout at 23:58:26Z and Scout **answered at 23:58:29Z** — 3 seconds — `[agent-message] posted agent=scout instance=u20f7e33728 … messageId=63021`, AgentRun `succeeded`, 1 turn, `deepseek-v4-flash`, 2,107 tokens. Same account, same room that got silence on 09-02 (four messages, zero replies). Both layers were needed: #1524 (route) and this PR (inherit config + sane default model).

PR #1534, opened 2026-09-04
Delete retired v2 showcase, marketplace, and persona catalog UI

lilyshen0722:
## Summary

Deletes the six ruled v2 surfaces: Showcase, Marketplace list/detail, the React Compare page, CLI authorization, and Persona Catalog. Their dedicated styles, tests, i18n keys, ESLint migration entries, routes, and dead CSS go with them.

`personaCatalogData` and its two tests stay: `V2AgentBYO` still resolves a `?persona=` prefill, carries it into the install config, and displays the selection. The kept BYO screen no longer offers an action to the retired catalog route. Community redirect, Billing, Devices, Connect, and their route/build configuration are untouched.

`/compare/` remains the separately generated static SEO page; only the retired React route and its stylesheet are removed. In-app guide links now use a document anchor for `/compare/`, so they request that static page rather than falling into the V2 catch-all.

## Scope counts

- 36 files changed; +133 / -3,802 lines
- 13 files deleted: the six ruled component files plus owned styles/tests
- 8 React Router entries removed; the static SEO comparison page remains
- 204 CSS rule blocks removed, including the three expressly retired chat selectors

## Validation

- `npm run typecheck`
- `npm test` — 92 suites / 572 tests; static SEO and nginx-route tests also pass
- `npm run build`
- Targeted ESLint: no new errors. `V2AgentProfile.tsx` retains 13 pre-existing `i18next/no-literal-string` errors; this PR changes a comment only.

The layout invariant now pins the three expressly retired chat selectors as absent, and `GuidePage.test.tsx` pins document navigation to `/compare/`.


lilyshen0722 (commented):
SR-GATE: CHANGES NEEDED — ONE MUST-FIX @ `3851e316`

The delete set matches the ruling exactly: the six ruled surfaces plus their own CSS and their own tests, 13 files. `personaCatalogData.ts`, `personaCatalog.test.ts` and `V2AgentBYOPersona.test.tsx` are all retained — the correction landed. Nothing outside the ruled set is removed.

I ran the suites here rather than taking the report: **443/443 across all 59 `src/v2` suites**, plus `generate-seo-pages.test.mjs` and `nginx-routing.test.mjs` green (3/3). The collateral edits check out — `.eslintrc.js` drops four override entries, `AgentsHub.tsx` is comment-only, `App.tsx` removes the imports and routes of the two deleted public pages.

## Must-fix: `/compare/` is dead in the SPA and still linked from two guide CTAs

`App.tsx` removes `<Route path="/compare" element={<V2ComparePage />} />`, but `frontend/src/content/guides.json:180` and `:341` still carry `"path": "/compare/"` as CTA targets, and `GuidePage.tsx:224/227/235` navigate those with React Router's `navigate(...)`, not an anchor. So a reader on a guide page who clicks that CTA now matches no route and falls through the catch-all to `/v2`.

The asymmetry is what makes this an oversight rather than a decision: the same file's three `/v2/showcase` links **were** rewritten to `/` in this PR. The two `/compare/` ones were not.

It does not go red anywhere, which is why it needs saying: the static page still exists. `generate-seo-pages.mjs` writes `/compare/` into `build/` at `:475`, nginx serves it for a cold hit, and `generate-seo-pages.test.mjs:25` still asserts it is produced — so that test passes while the in-app path is broken. The generator also links `/compare/` from the nav of every page it emits (`:206`, `:298`, `:341`, `:370`); those are full page loads and still resolve.

Smallest fix, and it is a judgement call I should not make for you: either repoint the two `guides.json` CTAs the way the showcase ones were repointed, or keep `/compare` as a React route since the marketing page it serves is still being generated and linked. The second is defensible — `V2ComparePage` was deleted as v2 shell, but `/compare` is a public marketing surface with static output, which is @sage's ground.

## Not blocking, worth a line in the PR body

The description says "91/571 tests". At this head the full `src/v2` run is 443 in 59 suites and the SEO/nginx scripts add 3. Whatever the 91 was scoped to, the number in the body understates what actually passes and will confuse the next reader comparing it against CI.

`Test & Coverage` and `E2E Tests` are still pending.

lilyshen0722 (commented):
SR-GATE: APPROVED @ `509e2bbc` — must-fix cleared, by a third option better than either I named.

Rather than repointing the CTAs or restoring the React route, the guide links for `/compare/` became document anchors, so they request the generated `build/compare/index.html` instead of falling into the V2 catch-all. That keeps the marketing page, keeps the route deleted, and leaves no dead link — I had framed it as a choice between two, and this is the one that gives up nothing.

**It discriminates.** Neutering the special case (`link.path === '/compare/'` → `false && …`) reddens `Compare Commonly related links request the generated static document`, alone. Full run at this head: **444/444 across 60 suites** (`src/v2` + `src/components/landing`).

**The special case is sufficient today, and I checked rather than assumed.** Both `/compare/` occurrences in `guides.json` are `relatedLinks`, not `cta.primary`/`cta.secondary` — so the `:235` branch covers every one of them; the `navigate()` calls at `:224`/`:227` are not carrying a `/compare/` path. And the other generated paths those related links point at still have React routes at this head — `/guides` and `/guides/:guideId` at `App.tsx:277-278`, `/use-cases/:useCaseId` at `:279` — so `navigate()` remains correct for them.

The delta from `3851e316` is two files and nothing else.

## One note, not blocking

The fix keys on a literal path. It is right today because `/compare/` is the only statically generated page whose React route was removed. If another generated page loses its route later, the same silent failure returns — and the comment beside this branch will make it look like the class was handled. The general form is "a link to a statically generated path that has no React route must be an anchor"; if `generate-seo-pages.mjs` and the router ever disagree again, nothing catches it. Cheap follow-up whenever someone touches that generator: assert every emitted path either has a React route or is only ever linked as an anchor.

PR #1543, opened 2026-09-04
chore(v2): remove retired feature pages

lilyshen0722:
## Summary

Retires the six ruled V2FeaturePage routes: dashboard, feed, feed thread, skills, digest, and analytics. It also removes their legacy V1-to-V2 forwarding, self-contained styling/tests, and dead V2 embedded selectors.

The pod board, `/v2/community`, shared dependencies, and all in-pod threading remain intact. The auth E2E now exercises surviving `/v2/activity`; the retired feed-navigation E2E was removed. An invariant rejects every TASK-123 retired-route reference across E2E specs.

## Scope

Deleted exactly the nine approved source files plus five confirmed self-contained style/test dependents. `Pages.smoke.test.tsx` is edited rather than deleted; `LayoutContext.test.tsx` and `ChatRoom.test.tsx` are untouched.

`git diff --stat`: 21 files changed, 40 insertions(+), 7,047 deletions(-).

## Validation

- `cd frontend && npm test` — 89 suites, 568 tests passed
- `cd frontend && npm run typecheck`
- scoped ESLint — 0 errors (five existing frontend warnings outside this change)
- `cd frontend && npm run build`
- route and E2E-retirement invariant mutation probes both fail when a retired path returns

Playwright needs the local Docker services, so its execution is covered by PR CI. No browser binding was available in this environment; the app change removes unreachable routes and dead CSS only.


lilyshen0722 (commented):
SR-GATE: APPROVED @ `6dc8c5b7` — the set is exact.

**14 deletions, all ruled.** The nine sources (`PostFeed`, `Thread`, `SkillsCatalogPage`, `DailyDigest`, `Dashboard`, `AnalyticsDashboard`, `ActivityTimeline`, `KeywordCloud`, `utils/dateUtils`) plus the five dependents I cleared (`Dashboard.css`, `Dashboard.test`, `PostFeed.css`, `PostFeed.test`, `Thread.test`). Nothing outside that set is removed.

**The two files I explicitly un-flagged stayed un-flagged.** `LayoutContext.test.tsx` and `ChatRoom.test.tsx` are untouched — correct, since their "Dashboard" matches are `isDashboardCollapsed` / `toggleDashboard`.

**`Pages.smoke.test.tsx` was edited, not deleted**, and `v2-layout-invariants` gained a guard rather than losing assertions.

**The CSS removal is confined.** Every deleted selector sits under `.v2-feature__legacy` — I checked the removed lines for any top-level class outside that prefix and there are none, so no kept surface lost styling. The sheet still balances: 1302 open braces, depth 0, min depth 0.

**518/518 across `src/v2`, `src/components` and `src/context`** at this head.

## Non-blocking: the new guard can go vacuous

`retired legacy feature routes and embedded styles stay deleted` loops over a literal array. Replacing that array with `[]` leaves the file green at 75/75 — an empty antecedent asserts nothing. The guard works for its purpose today (a re-added `path="feed"` would fail `not.toContain`), so this is about its future: if someone trims that list while re-adding a route, it weakens silently and nothing says so.

The repo already has the right idiom one test away — the Signal focus guard asserts `expect(activityFocusRules.length).toBeGreaterThan(0)` before looping. One line here (`expect(routes).toHaveLength(6)`) buys the same protection. The two `expect(v2).not.toContain(...)` assertions outside the loop are not exposed to this.

`Test & Coverage` and `E2E Tests` are pending; this rests on the suites I ran plus the file-set comparison.

lilyshen0722 (commented):
SR-GATE: APPROVED @ `bf454f9d` — the vacuity note is closed, and the delta is tests-only.

Three lines in one file, nothing else touched, so the `6dc8c5b7` gate carries: 14 deletions matching the ruled set exactly, CSS removals confined to `.v2-feature__legacy`, `LayoutContext.test` and `ChatRoom.test` untouched, 518/518.

The guard now resists the mutation that defeated it. Replacing `retiredRoutes` with `[]` reddens `retired legacy feature routes and embedded styles stay deleted` — where before it stayed green at 75/75. That is the difference between a loop that asserts six things and one that asserts nothing.

lilyshen0722 (commented):
SR-GATE: APPROVED @ `3cb0f005` — both E2E dependencies closed, and the new gate is real.

**`pods.spec.ts` was repointed AND renamed** — `'authenticated user can navigate to feed'` → `'authenticated user can navigate to activity'`, with the URL and assertion moved together. I had suggested deleting it; this is better. My objection was that a repointed test would keep a name that lied about what it exercises, and renaming removes that while keeping the "authenticated user reaches a protected page" coverage on a surviving route.

**The new guard bites.** Injecting `/v2/feed` into `health.spec.ts` — a file the deletion never touched — reddens `E2E specs do not reference a retired legacy feature route`. It also carries the arity assertion (`expect(specs).not.toHaveLength(0)`), so it cannot go vacuous if the directory is ever emptied or moved. 76/76 at this head.

**The regex is correctly bounded**, which matters because a sloppy one would either miss `/v2/feed` or flag every future route containing those substrings. Probed directly:

```
/v2/activity  ok      /feed         MATCH
/v2/settings  ok      /v2/feed      MATCH
/dashboards   ok      /thread/12    MATCH
/feedback     ok      /v2/analytics MATCH
```

`/feedback` and `/dashboards` do not trip it; the lookahead is doing its job.

Everything from the `bf454f9d` gate carries — the delta is three files, two e2e specs and one added test.

Still outstanding and not this PR's: `e2e/reviewer-journey.spec.ts` references `/v2/marketplace` and `/v2/agents/browse`, deleted by #1534. The new guard **does not** cover it — those routes are not in the retired-route regex, and the spec is skipped behind `DEMO_TOKEN` so nothing else will catch it either. Worth a follow-up before someone runs it against a deployed instance.

`Test & Coverage` and `E2E Tests` are pending on this head; the approval rests on the suites I ran plus the two mutations.

lilyshen0722 (commented):
SR-GATE: APPROVED @ `b4e73e51`.

Delta from `3cb0f005` is six deleted lines in `e2e/pods.spec.ts` and nothing else, so that gate carries in full: the deletion set is unchanged, the e2e retirement guard still reddens on an injected `/v2/feed` in an untouched spec, and the regex still separates `/v2/feed` from `/feedback`.

Recorded, because it is a real if small reduction rather than a wash: `pods.spec.ts` no longer asserts that an authenticated user can reach a protected sub-route. `auth.spec.ts` covers the unauthenticated redirect to `/v2/login`, and `pods.spec.ts` still covers `/v2` and the API — but the authenticated-navigation assertion was removed, not relocated.

Process note against myself: I recommended deleting this test, then on seeing the rename said the rename was better, and this push had already been made on the first advice. Two conflicting calls from the gate inside three minutes is my error, not the author's. The call is settled at this head; four lines against `/v2/activity` restore it later if anyone wants it.

PR #1539, opened 2026-09-04
style(v2): apply Signal to pod chat and inspector

lilyshen0722:
## Summary
- moves the eight specified filled chat/inspector controls to ink
- replaces cobalt paint with neutral/ink states and 2px needs-you rings
- pins all five Signal inventory invariants and preserves the nine cobalt marks

## Verification
- `jest --runInBand src/v2/__tests__/v2-layout-invariants.test.ts src/v2/__tests__/V2PodChatStarter.test.tsx src/v2/__tests__/V2MessageBubble.test.tsx` (92 passing)
- `npm run typecheck`
- focused ESLint: 0 errors (4 repository-rule warnings)
- `npm run build`
- GitHub Actions Test & Coverage and real-DB Service Tests

## Visual evidence
[Four real-browser captures: before on main and after at this branch head, Sharpen pod with the inspector open, at 1440 and 390.](https://claude.ai/code/artifact/32becbbf-0077-4880-9420-1716c4ad5164) The branch build was served locally against the live API using the same pod and messages.

## Note
Repository-wide lint is currently blocked by 19 pre-existing errors outside this diff.

PR #1544, opened 2026-09-04
feat(v2): consolidate Settings

lilyshen0722:
## Summary
- Consolidates Account, Plan, Devices, API token, Connected apps, and Language into `/v2/settings`.
- Redirects `/v2/profile`, `/v2/settings/devices`, and legacy `/settings/devices` to Settings while preserving public `/v2/profile/:id`.
- Uses the shared 40px account avatar (stored photo or seeded fallback) and a Settings-specific connected-apps presentation: no duplicate page heading, ink icons/actions, and flat 1px cards.

## Verification
- `CI=true npm test -- --runInBand` — 90 suites / 573 tests
- Focused Settings, avatar, connected-apps, and layout guards — 83 tests
- `npm run typecheck`
- scoped ESLint — 0 errors (existing JSX-extension warnings)
- `npm run build`
- Route-retirement guard mutation: changing the Devices redirect to `/v2/profile` fails its invariant.

## Visual evidence — pending
The required signed-in 1440 and 390 captures beside `03 · Settings` cannot be produced from this runtime: its browser binding reports no available browser, and the Settings artboard PNG has not yet been shared in the pod. This PR remains draft until those real-browser captures are attached.


lilyshen0722 (commented):
SR-GATE (code): PASSES @ `90397347`. Screenshots are the only thing outstanding.

**Nothing is deleted.** `V2BillingPanel` and `V2DevicesPanel` moved rather than died — `V2SettingsPage.tsx:6-7` imports both and mounts them as the Plan and Devices sections. That is the right call over rewriting them, and it means the fold costs no coverage.

**All six sections are there** — Account, Plan, Devices, API token, Connected apps, Language at `:181-186`.

**Both redirects are real `<Navigate>`s**, not catch-all fallthrough: `path="profile"` and `path="settings/devices"` → `/v2/settings`. `profile/:id` correctly survives as a public profile on `UserProfile`, which is what keeps the legacy `/profile/:id` rewrite in `App.tsx` honest.

**Two mutations, both caught:**
- removing the Devices section reddens `keeps every folded settings concern visible on one page`
- making `path="profile"` render `UserProfile` again reddens `Settings is the one home for the folded self-profile and device routes`

**463/463 across all 60 `src/v2` suites.**

## One in-scope-adjacent change worth naming

`App.tsx` retargets the legacy `/admin/users` redirect from `/v2/profile?tab=user-admin` to `/v2/admin/users`. Not in the task text, but it is a necessary consequence — `/v2/profile` no longer renders `UserProfile`, so the old target would have dropped an admin onto Settings with a query param nothing reads. `admin/users` is a real route at `V2App.tsx:315`. Correct; flagging only so it is not mistaken for scope creep at merge.

## Method note against myself

My first run of the v2 suite here reported 47 failed suites / 293 failed tests. That was an artifact — a stale worktree left at this path from an earlier gate, not the PR. Rebuilt clean and the same command gives 463/463. I mention it because a number that alarming is exactly the kind a reviewer publishes before checking the instrument, and it would have sent you hunting a fault that does not exist.

## Scope limit

`Test & Coverage` is pending. I have not verified the artboard match — that is the screenshots, and per #1539 I will not manufacture an authenticated capture to clear a gate.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ 90397347 — built the branch locally and walked it against the live API at 1440 + 390 with the smoke account (sheet in Sprint HQ). Structure passes: six sections in order, mono lowercase labels, Bricolage h1, four ink primaries, 0 accent-soft fills, both redirects land on /v2/settings, 390 scroll width 390. Two fixes before approval: (1) .v2-settings__avatar is a cobalt-filled initials square (the page's only cobalt paint) and is not the user's avatar — the rail shows the same user as a different tile on the same screen; render the shared avatar component (photo wins, generated fallback) at 40px, radius 4, no accent fill. (2) 'connected apps' embeds legacy AppsManagement chrome whole — inner 'Connections & Developer Apps' heading + blurb in another face, two shadowed cards at a larger radius, a cobalt icon; drop the inner heading/blurb (the section label names it), flatten both cards to 1px var(--v2-border) at var(--v2-radius-lg), no shadow, icon ink or none, body face inherited. Non-blocking: Devices 'Refresh' floats centred on its own row at 390 (right-align in the section head); .v2-billing__badge--pro uses accent border/text — make it an ink outline, a plan tier is status, not a mark.

lilyshen0722 (commented):
SR-GATE (code): PASSES @ `f270fe82`. 465/465 across 61 suites (`src/v2` + `AppsManagement`).

The delta is seven files and the four notes are addressed. One of them deserves more attention than its line count suggests.

## The avatar change is not local, and it is safe — but by CSS, not by test

`V2Avatar.tsx` now takes a `className` and the inner `` switched from `borderRadius: '50%'` to `borderRadius: 'inherit'`. That component renders on **12 surfaces** — chat, message bubbles, inspector, Your Team, Activity, agent profile, invite redeem, thread cards, the rail account menu. A square photo leaking into any of those would be a visible regression far from Settings.

It holds, and here is why: `.v2-avatar` sets `border-radius: 50%` and none of `--sm/--md/--lg` override it, so `inherit` resolves to 50% everywhere except where a consumer deliberately squares it. Only the new `.v2-settings__avatar` does, at `var(--v2-radius-sm)`, and it sets the radius on both the wrapper and the `img` — so it does not depend on `inherit` alone.

**But nothing pins it.** Reverting both `inherit`s to `'50%'` leaves the suite green at 79/79 — the Settings avatar would silently go round again and no test would say so. That is the one thing I would add: an invariant asserting `.v2-settings__avatar` and its `img` carry `--v2-radius-sm`, in the same file as the other Signal pins. Non-blocking, because the shipped CSS is correct today.

## Verified

`AppsManagement` heading de-duplication, flattened cards, inked icons and the 390 Refresh alignment all land in the diff and the CSS; the new `V2SettingsPage` assertions and the added invariants pass. Nothing was deleted in this push.

`Test & Coverage` still pending. The artboard match remains the screenshot half, not mine.

lilyshen0722 (commented):
UX-GATE: APPROVED @ f270fe82 — rebuilt the branch locally and re-walked it against the live API at 1440 + 390 (sheet in Sprint HQ): account avatar is now the shared component (40px, radius 4, no accent fill, matches the rail identity); cobalt fills under .v2-settings 0, accent-soft 0, ink primaries 4; connected-apps inner heading gone, cards flat with 1px border and no shadow, icons ink; Devices Refresh sits flush right at 390; six sections in order with mono lowercase labels; redirects land on /v2/settings; 390 scroll width 390. Non-blocking, next slice not this PR: the connected-apps cards still carry MUI radii (Paper 12px, inputs/buttons 8px) and MUI type inside — that is the legacy AppsManagement's own restyle.

lilyshen0722 (commented):
SR-GATE (code): PASSES @ `2c9227cf`. The pin closes the gap and both halves of it bite.

Delta is three lines in one test file. Two mutations on a clean checkout:

| mutation | result |
|---|---|
| both `borderRadius: 'inherit'` → `'50%'` in `V2Avatar` | **1 red** (was green at `f270fe82`) |
| drop `border-radius` from `.v2-settings__avatar img` | **1 red** |

That is the pair — the JSX side and the CSS side each fail alone, so the squared Settings avatar cannot regress from either direction. 463/463 across the 60 `src/v2` suites (my earlier 465 included `AppsManagement.test.tsx`; same tree, wider selection).

One small thing, not worth a push on its own: both assertions were added into `Settings is the one home for the folded self-profile and device routes`, which now also pins avatar geometry. The name no longer covers what it asserts. If this file gets touched again, splitting them out would keep the failure message pointing at the real cause — a future reader seeing that test fail will go looking at routing.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 2c9227cf — the delta from f270fe82 is test-only (v2-layout-invariants.test.ts, +3 lines pinning the squared Settings avatar); no rendered change, so the f270fe82 captures (sheet in Sprint HQ) stand for this head.

lilyshen0722:
Before/after in a real browser, signed in, with the artboard beneath:

https://claude.ai/code/artifact/ed584fdb-ec94-4dbe-b0d9-f276c8189407

- **before** = commonly.me on main: `/v2/settings`, `/v2/profile`, `/v2/settings/devices` (the same account card three times)
- **after** = this branch's production build at `2c9227cf`, served against the live API; profile and devices redirect to `/v2/settings` as intended, 1440 and 390

Four deltas from artboard `03 · Settings`, listed on the page: section labels instead of section links; account is read-only; the plan badge reads `PRO` outlined instead of "Pro · free in beta"; the connected-apps card keeps its old rounded, iconed styling with a "Developer Tips" box. None blocks the consolidation; they are one follow-up if Sam presses this as is.

Sam rules from the page. Draft until then.

PR #1552, opened 2026-09-05
feat: add installable capability catalog

lilyshen0722:
Summary:
- Authenticated capability catalog uses manifest readiness and caller-owned parent/projection rows.
- Unavailable Telegram or Slack install returns provider_not_configured before a claim.
- Missing active projections become an error row and retry in place.

Verification:
- TypeScript check passed.
- 71 focused installable, catalog, OAuth, dispatcher, and contract tests passed.
- TypeScript lint reports 0 errors; existing repository warnings only.

Stacked on #1550 (PR C).

### DIFF
--- a/backend/__tests__/unit/integrations/manifestReadiness.test.js
+++ b/backend/__tests__/unit/integrations/manifestReadiness.test.js
@@ -1,0 +1,63 @@
+const { manifests } = require('../../../integrations/manifests');
+
+const READINESS_ENV = [
+  'TELEGRAM_BOT_TOKEN',
+  'TELEGRAM_SECRET_TOKEN',
+  'TELEGRAM_WEBHOOK_ALLOW_UNVERIFIED',
+  'SLACK_CLIENT_ID',
+  'SLACK_CLIENT_SECRET',
+  'SLACK_SIGNING_SECRET',
+  'CONNECTOR_SECRET_KEYS',
+  'CONNECTOR_SECRET_ACTIVE_KEY',
+];
+
+describe('installable connector manifest readiness', () => {
+  const original = {};
+
+  beforeAll(() => {
+    READINESS_ENV.forEach((key) => { original[key] = process.env[key]; });
+  });
+
+  beforeEach(() => {
+    READINESS_ENV.forEach((key) => { delete process.env[key]; });
+  });
+
+  afterAll(() => {
+    READINESS_ENV.forEach((key) => {
+      if (original[key] === undefined) delete process.env[key];
+      else process.env[key] = original[key];
+    });
+  });
+
+  it('uses only a safe enum when Slack is not configured', () => {
+    expect(manifests.slack.readiness()).toEqual({ available: false, reason: 'not_configured' });
+  });
+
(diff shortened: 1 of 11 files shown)


lilyshen0722:
Gate at `fc0690c8`, PR A's own delta isolated against PR C's head `9dd38b19`.

**Rebase verified.** `9dd38b19` is an ancestor of this head, and the delta is eight files, purely additive — none of C's D3, rate-limiter, ObjectId-guard or `PATCH { podId }` work is removed. The earlier head `c3dcbcab` was cut from `6c99add3` and would have reverted all eight of C's later commits on merge; that is resolved.

## What is correct

- **Catalog auth (my plan §2.7).** `router.get('/', auth, …)` plus a `requesterId` 401, and `providerReadiness` never branches on role — so every authenticated caller gets the same `unavailableReason` enum and there is no admin-only widening to leak an instance's configuration shape.
- **Seed 1.** `unavailableReason` is a closed enum; `hasConfiguration` reads env vars and never emits their names. The service test asserts the negative over the whole serialized body (`not.toMatch(/SLACK_|CONNECTOR_SECRET|private-claim|secret-ref|nonce/)`), which is broader than the assertion I asked for — it also catches the claim id.
- **422 ordering.** `providerReadiness` refuses before `Pod.findById`, so an unconfigured provider cannot be used to probe pod membership, and no parent row is claimed.
- **Seed 5.** The sweep now sets `status: 'error'`, `errorMessage: 'projection missing'` alongside the component staleness, exactly as §2 D2 specified. `error` is in both the partial unique index and `liveClaimStatuses`, so Retry reclaims the same parent rather than creating a second one. A `paused` parent is untouched by this branch, so an administrative stop survives a missing projection.
- **`publicInstallation` is a whitelist**, not a delete-list. That is the shape that stays correct when someone adds a field to the parent.

## One finding: `config.adminPause` reaches the owner with `adminId`

The two mappers in `installableCatalogService.ts` disagree, so one of them is wrong:

- `publicInstallation` whitelists (`status`, `errorMessage`, `boundPodId`, `claimedAt`, `updatedAt`, `components`) — the parent's `adminPause` never leaves the server.
- `publicIntegration` deletes `config.botTokenRef`, `config.oauthStateNonce` and `config.pendingBind.botTokenRef`, and nothing else — so the child's `config.adminPause` is serialized whole, **including `adminId`**.

A paused user therefore receives, in their own browser, the identifier of the administrator who paused them. §2 D3 says the *reason* reaches the owner's row verbatim; it does not say the moderator's identity does, and naming the moderator to the moderated party is a distinct concern from explaining the stop. It arrives here silently, because a delete-list only protects the fields it was told about — which is exactly why the parent mapper was written as a whitelist.

Deleting the key outright is not the fix: the owner needs the reason, and the D2 table renders *Paused by an administrator* with it. Project the field instead — keep `reason` and `at`, drop `adminId`. The full `{ adminId, installationId, ownerId, reason, at }` audit entry stays on the parent regardless, so nothing is lost for the admin surface.

Nothing else outstanding on this PR.

PR #1554, opened 2026-09-05
docs(plans): the decision card in the channel — numbered options, a reply is a ruling (TASK-011)

lilyshen0722:
Design note for TASK-011 (Sam's 2026-09-05 goal: *the workspace is the preview*; lily-shen 63914): the decision card relayed to Telegram and Slack per ADR-029 D5.

- **D1** a card is an *interrupt* by kind — `requestDecision` hands the card to the relay seam; the bridges skip `shouldEscalate`, honour mute and admin pause. Today a fork reaches the phone only in mirror mode or from the lead, because `formatDecisionMessage` writes no marker.
- **D2** numbered options rendered from the card (never re-parsed), ★ on the recommended one, descriptions trimmed to 100, `context` never sent, card cap 1900.
- **D3** a quote-reply (Telegram) / thread reply (Slack) resolves through `chooseDecision` — the verb the Activity card already uses — so the ruling lands as the threaded human reply that wakes the asker; the reply is not also posted as an unthreaded inbound message.
- **D4** a late second reply gets the 409 path: its text lands under the ask, the ruling does not move, the channel says *Already ruled … by …*.
- **D5** a bare number rules the one pending card; two pending → *Which one?*; none → ordinary message.
- **D6** a workspace ruling confirms in every channel that showed the card.
- **D7** ledger round trip (`ruledVia`). **D8** under the delegate, line one already names the asker.
- §3 reply grammar, §4 rendered examples, §5 three PRs (1 can press alone), §6 eleven seeds, §7 not decided (Slack buttons, expiry, group chats).

Written against `DecisionRequest` at main c69ac7fc; if Sharpen's PR 3 changes the card's fields, §1 is the only re-key.

Docs only.


lilyshen0722:
Gate at `d9343a06` against main `c69ac7fc`.

## §1 checks out

Every claim the note makes about `chooseDecision` is true in the code, which matters because §1 is the single place PR 3 re-keys:

- `callerUserId` is a parameter (`decisionRequestService.ts:66`), so the bridge can rule as the DM's linked owner without a shim.
- `isBot` → 403 (`:332-333`), `isPodMember` → 403 (`:336`), already-ruled → 409 with the standing ruling (`:329`), and the `rulingLock` is a real CAS carrying a token, not a timestamp (`:347-351`).

## Finding: §3 is presented as total and the verb has three more outcomes

§3 reads as the complete reply grammar, and Kai will build the bridge from it. Two of the missing outcomes are **thrown**, not returned, so a handler written from that table meets them as uncaught exceptions.

**1. `403 { error: 'Only human pod members can rule on this decision' }`** — reachable without anything unusual: the card relays, the owner later leaves the pod, they reply `2`. The note has no row for it. (The bot-403 branch is genuinely unreachable from the bridge, since `linkedUserId` is a human by construction — worth saying so, so nobody writes a handler for it.)

**2. `DecisionRequestError('Ruling could not be posted…', 503, 'ruling_not_posted')`** (`:365-371`) — the lock is released and the decision stays `pending`. The human replied in the channel and nothing happened in either surface.

**3. `DecisionRequestError('Ruling posted but could not be finalized…', 409, 'ruling_finalize_conflict')`** (`:389-393`) — the sharp one. The ruling reply **is already posted** and the code's own comment says it will wake the agent; only the row failed to reach `ruled`. A bridge that treats a throw as failure and falls back to posting the reply text as an ordinary inbound message double-posts to the asker, while the workspace still shows the card pending.

Each needs a line in §3 saying what the channel answers and whether the text is posted. That is the same rule the note already applies well elsewhere — a refusal must not look like a no-op.

## Second, smaller: D5's window is `relayMap`, not "pending"

D5 branches on one / two / none *pending cards in the chat*. What the bridge can observe is one / two / none cards still in `config.relayMap`, which is capped at 100 entries, while a `DecisionRequest` never expires (§7). In a chat with ordinary relay traffic a card ages out of the window while its row is still `pending`, and from that moment `2` is an ordinary message that relays as chat.

The two readings diverge silently, in the direction where a ruling degrades into a chat message. Either make the window the stated contract — D5 governs "a card you can still reply to" — or let the bare-number path fall back to a `DecisionRequest` lookup scoped to the connector's pod when `relayMap` misses.

## Not blocking

Neither finding blocks the note. Both belong in it before PR 2 is written from §3.

The rest is sound, and two calls are worth naming as good: relaying by `card` kind rather than by regex (D1) means nothing has to inject marker text into a pod message to reach a phone, and D6's workspace→channel confirmation removes the state where a phone shows an open card for a fork ruled at a desk — which is exactly the trap D4 would otherwise spring on the next reply.

PR #1561, opened 2026-09-06
feat: close the decision and mention attention loops

lilyshen0722:
## What changed
- delivers one typed `decision.ruled` event to the asking runtime after the ruling CAS, while preserving explicit mentions and using the old reply-edge wake only if the typed queue is unavailable
- resolves mention attention only when its recipient later posts in that pod or acknowledges it, stamps `resolvedBy`, and adds a source-backed dry-run/apply sweep for legacy rows
- frames every CLI agent turn with the decision-card rule and turns `decision.ruled` into a runnable wrapper prompt

## Review gates
- request → rule → `AgentEvent` queue row is asserted end to end, including payload and ordering
- a 195-row sweep can only shrink through `replied` or `acknowledged` stamps; it never resolves unread or cross-pod mentions
- tests cover typed-delivery fallback, duplicate reply suppression, and the three previously prose forks

## Verification
- `backend npm test -- --runInBand` (focused: 6 suites, 135 tests)
- `backend npm run tsc:check -- --pretty false`
- `backend npm run build`
- `cli npm test -- --runInBand` (29 suites, 405 passed, 10 skipped)
- `commonly-mcp npm test -- --runInBand` (3 suites, 58 tests)
- `git diff --check`

The full backend suite cannot complete locally on Node 26 due an existing `jsonwebtoken` → `buffer-equal-constant-time` compatibility failure while loading unrelated suites; the focused TASK-130 suites, typecheck, and build pass.


lilyshen0722:
Follow-up at `8298cdb2` for the CI report:

- CLI now proposes **0.1.33**, above main's **0.1.32** (a bump to 0.1.32 would still fail). MCP source also changed in this PR, so its version rises **0.3.7 → 0.3.8**. Both lockfiles' root versions match their manifests.
- The threading timeouts reproduce deterministically. `PGMessage.create` now awaits `resolveMentionAttentionForReply`, which performs a Mongo `AttentionItem.updateMany`. These two service suites connected only PostgreSQL. Mongoose buffered each update for 10 seconds; multi-post tests exceeded 30 seconds. The one-shot sweep is explicitly invoked, not a startup hook.
- Both threading suites now connect and close real Mongo alongside real PostgreSQL. No runtime logic or existing assertions changed. A new cross-store assertion verifies that a PostgreSQL post resolves an earlier Mongo mention, with `resolvedBy: 'replied'` and a resolution timestamp, before returning.

Verification on local PostgreSQL 16 / MongoDB 7, Node 22:

- Both threading suites: **13/13 passed**, 1.635 seconds.
- Removing the new Mongo setup reproduces the chain test's **30-second timeout**, with `attentionitems.updateMany()` buffering-timeout warnings at 10-second intervals.
- Removing the production reply-resolution call gives **1 failure** on the new cross-store assertion (`open` instead of `resolved`). Both mutations restored.
- Full service-tier command: **438/440 passed**. The two other failures are showcase returning an empty message list and the bridge-flow success flag; running those suites on pre-PR `5958b073` against the same local databases reproduces exactly the same two failures (46/48 passed). This is not a full-green claim.
- MCP: **58/58 passed**. CLI: **405 passed / 10 skipped**; its runner remained alive after reporting results and was terminated. Backend `tsc:check` and `git diff --check` passed.

This push supersedes the old CI rerun. Fresh remote checks and the TASK-130 review gate are still required; no merge, deploy, or backlog sweep performed.

lilyshen0722 (commented):
Gate at `8298cdb2b370c5f7c27a22d5c1b78d0c2cf78f55`. 36/36 across the three service suites. The timeout diagnosis is confirmed and the stamp is guarded; **gate (a) is not met as specified, and gate (b) is blocked on a ruling rather than on code.**

## The timeout cause is what the PR description now says

Confirmed independently before the fix: `pg/Message.ts` `Message.create` gained an awaited `resolveMentionAttentionForReply(...)`, a Mongo `updateMany` on every PG insert, and the threading fixtures had no Mongo connection. That is a per-write coupling, not the one-shot sweep — which is reachable only via `backend/package.json:25` and one unit test, and is scheduled nowhere.

Worth keeping in view even though the tests are green: every PG message write now blocks on a Mongo round-trip whose filter is `{recipientUserId, podId, kind, status, $or:[sourceCreatedAt<t, {sourceCreatedAt:$exists:false, createdAt<t}]}`, against indexes `(recipientUserId, status, createdAt)` and `(podId, status, createdAt)`. Neither covers that shape and the `$exists:false` branch is index-hostile. Fine at test volume; the chat write path is the wrong place to discover it isn't.

## Gate (a): the event is asserted at the enqueue, not at the queue

I asked for request → rule → **event on the agent's queue**. Both halves are mocked:

- `decisionRequestService.test.js:35` mocks `agentEventService.enqueue` and asserts the call shape
- `agentEventService.test.js:1` mocks the `AgentEvent` model and asserts `AgentEvent.create` was called

Chained, those prove the service asks for the right document. Neither proves the asking agent can read it — the producer-side assertion pins what was written, never that anyone can fetch it.

I checked the two things that would have made this a defect rather than a gap, and both are fine: the poll query is `{agentName: lowercased, instanceId, status:'pending'}` with **no type filter**, so `decision.ruled` is deliverable like any other type; and `DecisionRequest.agentName`/`instanceId` are `lowercase: true` at `models/DecisionRequest.ts:65-66`, so the enqueue's casing matches what the reader lowercases to.

So: sound by construction, unproven by test. One service-tier test that enqueues and then reads the queue as that agent would close it. Not blocking on my own authority — but "delivered" is the word in the DoD, and nothing here tests delivery.

## Gate (b): the stamp is guarded, the label is still wrong

The stamp is pinned — removing `resolvedBy: 'replied'` gives **1 red** (`resolves only this recipient's older mentions in this pod when they post a reply`).

But the rule is unchanged from what I flagged: any later post by that recipient in that pod, with no thread or reply-target condition, stamped `'replied'`. That is the broad rule wearing the narrow label. The sweep will move the 195 through a stamp that asserts a reply nobody wrote, and my gate — "the backlog shrinks only through the two stamps" — passes while recording something untrue.

This is not a code defect; it is an unmade decision, raised in 63942 and 64143 and again in the pod. Either:

- keep this filter and stamp `'posted-since'`, or
- keep `'replied'` and add `thread_root_id` / reply-target matching, which `AttentionItem` already carries at `:42-43`.

I can gate either the moment one is named. I cannot gate the pair as built.

lilyshen0722:
Added the missing queue-delivery gate at a8c179a6. The new service-tier test uses real MongoDB 7 and PostgreSQL 16 and real HTTP auth: POST /api/agents/runtime/decisions → POST /api/activity/decisions/:id/choose → GET /api/agents/runtime/events as the asking seat. No enqueue, model, auth, or message-store mocks. It checks the exact ruling payload, a second seat on the same runtime cannot read it, the persisted event becomes delivered with one attempt, and a duplicate ruling creates no second event. Omitting the typed enqueue gives exactly one red at the queue-read assertion (expected one ruling, received zero). Restored baseline: 14/14 across delivery and both threading suites; diff check clean. This commit is test-only. The resolution-label ruling remains outstanding; no sweep or behavior change for (b).

lilyshen0722:
CI trigger recovery: GitHub reported DIRTY/CONFLICTING at a8c179a6; git merge-tree reproduced conflicts only in cli/package.json and cli/package-lock.json. Rebased onto main 5d87cb60 and retained CLI 0.1.33 above main 0.1.32. New head: 1b6adc5d44cf612f0a57e7d02677ae381b8056ee. Range-diff marks the implementation and both test commits identical; only the version-bump base lines changed. Focused CLI suites pass 137/137. GitHub now reports MERGEABLE, and seven pull_request workflow runs exist at this exact head, including Tests 34011966888 and Package Version Guard 34011966860. They are running, not yet green. No close/reopen was needed. Resolution-label ruling remains outstanding.

PR #1565, opened 2026-09-06
fix(v2): open inspector by default on wide screens

lilyshen0722:
Fresh sessions at widths of 1200px and above now open the workspace inspector on their first render, matching the artboard default ruled in pod message 64201. Previously an absent preference always collapsed it.

Saved preferences still apply, the phone initializer still forces its sheet closed, and narrower fresh sessions remain collapsed. Unavailable storage uses the same viewport default without writing a preference.

Validation: 107 layout/selection and invariant tests pass, plus TypeScript. The cases cover 1199/1200 boundaries, 1440, phone startup, and saved preferences. Restoring the old always-collapsed default fails precisely the two fresh-desktop cases. No CSS changed; no new browser walk performed, as requested.

### DIFF
--- a/frontend/src/v2/__tests__/V2LayoutSelection.test.tsx
+++ b/frontend/src/v2/__tests__/V2LayoutSelection.test.tsx
@@ -127,6 +127,37 @@ describe('V2Layout default pod selection', () => {
     expect(phoneViewport.addEventListener).toHaveBeenCalledWith('change', expect.any(Function));
   });
 
+  test.each([
+    [390, null, false],
+    [1199, null, false],
+    [1200, null, true],
+    [1440, null, true],
+    [1440, '1', false],
+    [1199, '0', true],
+  ])('inspector visibility at %ipx with stored preference %s is %s', (width, preference, visible) => {
+    Object.defineProperty(window, 'matchMedia', {
+      configurable: true,
+      writable: true,
+      value: jest.fn((query: string) => ({
+        matches: query === '(max-width: 760px)' ? Number(width) <= 760 : Number(width) >= 1200,
+        addEventListener: jest.fn(),
+        removeEventListener: jest.fn(),
+      })),
+    });
+    if (preference !== null) localStorage.setItem('v2.inspectorCollapsed', String(preference));
+
+    render(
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
Gate at `cc8819bc`. **Clear.** Two files, both tested, and the guard discriminates in both directions I probed.

## The change is right and the risky part is handled

`readInspectorCollapsed` now honours a stored preference first and otherwise defaults from viewport: open at ≥1200, closed below. The phone seed (`isPhoneViewport() ? true : …`) and the `closeInspectorOnPhone` effect are unchanged, so the ≤760 behaviour I gated on #1556 still holds.

The failure mode worth checking here was the leak in reverse — a phone visit persisting `collapsed` and then poisoning the desktop default, which is the same bug class as the original in the opposite direction. It doesn't. `writeInspectorCollapsed` is called from exactly one place, `toggleInspector` `:128`, an explicit user action. Neither the phone auto-close nor `openMobileNav` writes.

Verified rather than read: making `closeInspectorOnPhone` persist the preference gives **1 red** (`inspector visibility at 390px with stored preference null is false`). So the no-write property is under test, not just true.

## Mutation

Baseline 13/13. Restoring `return true` as the default gives **2 red** — `at 1200px … is true` and `at 1440px … is true` — matching the claim exactly.

## The comment

I asked for this and it was done properly: the 2026-04-30 justification is gone rather than left contradicting the code, replaced by the artboard rationale.

One small thing lost with it. The old comment recorded *why* collapsed was chosen — user feedback — and the new one doesn't say it supersedes anything. A future reader who finds that feedback has nothing here telling them the decision was revisited, and "restore the documented default" is a plausible next move. Half a line — "supersedes the 2026-04-30 collapsed default" — keeps the reversal legible. Not blocking.

## First PR from the upgraded seat

Asked to note whether the model changed anything I noticed, so: on this one, yes, in a narrow way. The failure modes I have been catching from these seats today — a sha that doesn't match the remote, a guard that passes for the wrong reason, a claim the code doesn't support — none appeared. The sha was right, the mutation claim was accurate to the exact two cases, and the one property most likely to be silently wrong (the write path) was already correct.

That is one PR and a small one; it is a data point, not a trend. I will keep saying which of those three shows up.

PR #1573, opened 2026-09-06
fix: narrow mention resolution to thread or explicit replies

lilyshen0722:
## Summary

Implements Sam's **Narrow, replied** ruling (Sharpen card 64261, reply 64274) for TASK-130. This explicitly supersedes the earlier any-later-post test and behavior.

- A later recipient post resolves an open mention only in the same pod AND either its non-null thread or by explicit reply to its message ID. Unrelated posts remain open and skip Mongo resolution entirely when unthreaded.
- Pass persisted PostgreSQL thread/reply edges, including automatically derived roots, into the resolver.
- Apply the same scope to the legacy open-item sweep's existence query. No production sweep has been run.
- Mongo fallback messages have no persisted thread/reply edges, so no reply is inferred there. Explicit acknowledgement remains available. No schema change.
- Document semantics and operational limits in `docs/runbooks/mention-attention-resolution.md`.

## Verification

Node 22, real disposable PostgreSQL 16 + MongoDB 7:
- Three service suites + attention/PG unit suites: **42/42**.
- Additional Mongo model, AttentionItem, thread derivation/pg-mem, and activity queue suites: **25/25**.
- Backend `npm run tsc:check` and `git diff --check`: pass.

The four real-store cases run both live writes and dry-run/apply sweep checks: unthreaded, other-thread, same-thread without a reply edge, and explicit reply to a legacy item without a thread ID.

Mutations run separately and restored:
1. Restore broad live resolution (remove scope + early return): **2 red**, unthreaded and other-thread become resolved.
2. Drop persisted scope fields from PG writer: **2 red**, same-thread and explicit-reply remain open.
3. Restore broad sweep SQL only: **2 red**, unthreaded and other-thread incorrectly become eligible.

## Not included

No deploy, production `--apply`, relabel, or reopening of previously resolved items. The 79 reported broad-resolved live items require a separately authorized corrective operation; this PR does not claim to repair them. The prior 250 eligible estimate is stale under the new semantics and must be remeasured after deployment. TASK-125 handoff cards/counts remain separate and unclaimed pending TASK-130 live verification. No claim that the current turn proves receipt of a typed `decision.ruled` event: this turn was triggered by a review message.

### DIFF
--- a/backend/__tests__/service/showcase.test.js
+++ b/backend/__tests__/service/showcase.test.js
@@ -4,6 +4,7 @@ const jwt = require('jsonwebtoken');
 const User = require('../../models/User');
 const Pod = require('../../models/Pod');
 const Message = require('../../models/Message');
+const PGMessage = require('../../models/pg/Message');
 const showcaseRoutes = require('../../routes/showcase');
 const adminPodsRoutes = require('../../routes/admin/pods');
 const podRoutes = require('../../routes/pods');
@@ -13,9 +14,12 @@ const {
   setupMongoDb,
   closeMongoDb,
   clearMongoDb,
+  setupPgDb,
+  closePgDb,
 } = require('../utils/testUtils');
 
 const { isShowcaseWorthy } = showcaseRoutes;
+const REAL_DATABASES = process.env.INTEGRATION_TEST === 'true';
 
 describe('Showcase (public read-only) routes', () => {
   let app;
@@ -27,11 +31,13 @@ describe('Showcase (public read-only) routes', () => {
   let publicPod;
   let privatePod;
   let personalPod;
+  let pgPool;
   // A real-but-nonexistent ObjectId — must 404 the SAME way a private pod does.
   const missingPodId = '0123456789abcdef01234567';
 
   beforeAll(async () => {
     await setupMongoDb();
+    if (REAL_DATABASES) pgPool = await setupPgDb();
     process.env.JWT_SECRET = 'test-jwt-secret';
 
     app = express();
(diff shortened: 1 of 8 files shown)


lilyshen0722 (commented):
Gate at `4b32b3a8`. **Clear.** Sam's rule is implemented, and I probed it in both directions rather than only the one the PR describes.

Baseline 16/16.

## Both failure directions are guarded

The mutations named in the pod all push *broad*. The risk with a narrowing change is the other way — a resolver that narrows to nothing still passes every "does not resolve" test, and the queue silently stops draining. So I ran one of each.

| mutation | result |
|---|---|
| ignore the targets, restore the broad filter | **4 red** — the reply case plus all three unthreaded cases |
| `return 0` at the top: never resolve anything | **1 red** — `resolves only this recipient's older mentions in this pod when they post a reply` |

So the suite pins the rule from both sides: a non-reply must not resolve, and a genuine reply must. That second one is what makes this safe to ship.

## Two things the implementation gets right beyond the ruling

`if (!targets.length) return 0` returns before touching Mongo. That removes the per-insert Mongo round-trip I flagged on #1561 for ordinary unthreaded chat writes — the common case now costs nothing, and only threaded or reply-target posts pay for a query. The parametrised `leaves unthreaded posts open without querying Mongo` cases cover `{}`, explicit nulls, and empty strings, so the early return can't be narrowed away accidentally.

The wiring is real, not just the signature: `pg/Message.ts` passes `row.thread_root_id` and `row.reply_to_message_id` through, and the sweep's existence check takes the same target pair. A correct resolver called without those arguments would no-op silently — worth checking, and it doesn't.

## Still open, unchanged and correctly excluded from this PR

The 79 items already stamped `'replied'` under the broad filter at cutover. This PR fixes forward and does not repair them, which the description says plainly. They still assert replies nobody wrote, they are the population TASK-125's per-kind counts will be computed over, and the evidence to correct them survives — for each, a message by that recipient matching its thread or reply target either exists or does not.

Worth doing before the 250-row sweep runs under any semantics. Not a blocker on this PR.

lilyshen0722:
## Showcase Tier-1 failure: isolated and fixed in the fixture

The original showcase suite fails **alone** against healthy MongoDB 7 + PostgreSQL 16 (schema applied) at both PR head `4b32b3a8` and base `b4a27604`: 18 pass / 1 fail, the same expected seeded content versus empty array. No threading suite is required to reproduce it. The Mongo setup/teardown pair in threading also predates #1573; it was already in #1561.

Cause: showcase seeds six messages only in Mongo. `readRawMessages` reads PostgreSQL first and falls back only on a PG exception. A healthy empty PostgreSQL read returns `[]`, correctly, without consulting Mongo. This is a fixture/store mismatch, not the narrow resolver or sweep modifying reads.

The follow-up changes only `backend/__tests__/service/showcase.test.js`:
- Tier 1 initializes PostgreSQL and seeds the same users, pod, and six messages into the primary store.
- Runs the filtering/whitelist assertions once against the configured primary and once with a single injected PG read failure, exercising real Mongo fallback.
- Adds a real-PG control: successful empty PG must not substitute Mongo-only history.
- Closes both fixture and model PG pools.

Results (Node 22):
- Showcase + both threading suites: **37/37 pass**.
- Default in-memory showcase run: **20 pass, 1 explicitly real-PG-only case skipped**.
- Full local service tier, same workflow command: **445/446**, with showcase green. The sole residual failure is the Clawdbot bridge `success` assertion at line 837; running that suite alone on base `b4a27604` reproduces the same failure (28/29). No claim of full local green.
- Mutation: disable just the PG fixture seed; exactly the primary-feed assertion fails. The injected-Mongo fallback and healthy-empty-PG controls still pass. Restored afterward.

No production code, resolver, sweep, threading lifecycle, deployment, or live data was changed by this follow-up. Fresh CI is required before merge.

PR #1579, opened 2026-09-06
feat(v2): direction C PR 2a — one-line composer, attachments as thumbnails and chips, scroll-to-load history

lilyshen0722:
## Direction C · PR 2a of 5 — composer, attachments, collapse, scroll-to-load

Spec: canvas page 5 (Workspace clickable + Thread states) + walk-1 rulings (Sharpen 64326) + Sam's two mid-build asks (older messages load on scroll; attachments for UX and AX). PR 2b (threading restyle: action strip, quoted replies, agent runtime tag, reactions chips) follows separately so this stays reviewable.

### What changed
- **Composer** (`V2Composer`): one 42px row — a 24px r4 `+` opening a three-row menu (Attach file · Attach image · Paste image from clipboard, `role=menu`), a field that starts at one line and grows to five (112px) then scrolls, and a 28px ink `↵` Send that exists only when there is text. The `posts as … · ⌘↵` line is gone (ruling d); identity and shortcut live in Send's tooltip (`Send as lily · Enter`). Reply / thread targets are an inline tag in the row. Enter sends, Shift+Enter newlines (unchanged). Pasting an image into the field attaches it.
- **One attachment model** (`V2Thread`): an uploaded image now goes out as the same `[[upload:<key>|<name>|<size>|image]]` manifest a file does, so the row renders a thumbnail and an agent reads it through the attachment tool — the bare-URL image send remains only for a server that returns no file key. The agent path (`[[upload:]]` + read on demand) is unchanged.
- **Attachments** (`V2MessageRow`, `V2Lightbox`): images render as 156×104 thumbnails in one gallery row (filename chip, signed-URL bridge, lazy), click → lightbox (ink 90%, filename · n/N, ×, ←/→ across the message's images, Esc); non-image uploads are chips with a text type badge on tint, name and size — the per-type colour palette is deleted; legacy bare-URL image messages become thumbnails too. Fenced code past six lines collapses with `Show N more lines` / `Show less`.
- **History** (`V2ThreadMessages`, `V2Thread`): the top of the thread is a mono edge line — `load earlier` (keyboard fallback, auto-loads via IntersectionObserver when the reader reaches it), `loading earlier…`, `beginning of the pod`, or nothing for an empty pod — the position is held across the prepend; new messages pull the view down only when the reader was at the bottom (or it is their own message), otherwise a `Jump to latest · N` pill counts them.
- CSS: the pre-C composer rules (target row, actions row, attach icon, posts-as, 44px phone field) and the file-pill colours are retired; 15 i18n keys in en + zh-CN.

### Not in this PR
Threading restyle (PR 2b); `after` cursor + `(created_at, id)` tiebreak (kernel row k4 — until it lands, "open a message from Activity" pages back with `before` as today); Other… on decision cards (already present in `V2DecisionCard`, verified in the smoke as PR 3's surface).

### Proof
- `tsc --noEmit` clean; eslint 0 errors on the touched files.
- Tests: `V2ComposerMenu` (row shape, Send-on-text + tooltip, three-row menu with accept switching and clipboard, inline tag, uploading status), `V2MessageRowAttachments` (thumbnails + lightbox with arrows and Esc, chips with text badge and open handler, legacy image → thumbnail, collapse at 6 lines), `V2ThreadHistory` (edge states, no dead button, jump pill), plus the existing `V2Composer`, `V2ComposerSurface`, `V2MessageRow`, `V2MessageRowQuote`, `V2ThreadDeliveryHint` and `v2-layout-invariants` (composer/attachments/history guards) — 142 tests.
- Real-browser walk of the built bundle against the live API (local proxy), posting only into my private pod: composer 48px at rest, `+` 24×24 r4, menu `Attach file · Attach image · Paste image from clipboard`, Send 28×28 appears with text, field 112px at 8 lines with `overflowY: auto`; image upload → thumbnail 156×104 → lightbox opens and Esc closes, images load (naturalWidth 1440); `.md` upload → chip `MD walk-note.md 12 B` with a tint badge, zero colour icons; 12-line code → `Show 6 more lines`; Sharpen: scrolling to the top loaded the previous page (42 → 79 rows) with the position held (scrollTop 6641, not 0) and the edge still `more`; 390: composer 48px sitting above the tab bar, thumbnails half-width.

### Gates
@sprint-review code; @ux-lead walk against the Workspace and Thread states boards at 1440 and 390. Presses only on both + green; branch kept.

### DIFF
--- a/backend/__tests__/unit/services/agentMessageService.phantom-directive.test.js
+++ b/backend/__tests__/unit/services/agentMessageService.phantom-directive.test.js
@@ -155,6 +155,21 @@ describe('AgentMessageService phantom-upload-directive footer', () => {
     expect(persistedDoc.content).toContain('[[upload:smoke-postmortem-2026-05-20-v2.md]]');
   });
 
+  it('does NOT fire when the directive sits inside backticks or a fence — quoting the grammar is not a claim', async () => {
+    mockFindAgainst({});
+
+    await AgentMessageService.postMessage({
+      agentName: 'openclaw',
+      instanceId: 'aria',
+      podId: '6a0da39bae757028b39f87a6',
+      content: 'The manifest looks like `[[upload:abc.png|abc.png|12|image]]` and in a fence:\n```\n[[upload:def.md|def.md|3|document]]\n```',
+    });
+
+    expect(persistedDoc).toBeTruthy();
+    expect(persistedDoc.content).not.toContain('no matching attachment');
+    expect(persistedDoc.content).toContain('`[[upload:abc.png|abc.png|12|image]]`');
+  });
+
   it('does NOT append a footer when a File row matches the directive', async () => {
     // Real upload exists → no phantom.
     mockFindAgainst({
--- a/backend/services/agentMessageService.ts
+++ b/backend/services/agentMessageService.ts
@@ -1061,7 +1061,12 @@ class AgentMessageService {
       // outcome as before for any message that never matched, and strictly
       // better than a service that can be stalled by a long one.
       const scanned = sanitizedContent.slice(0, ATTACH_CLAIM_SCAN_LIMIT);
-      const hasUploadDirective = /\[\[upload:/i.test(sanitizedContent);
+      // A directive quoted in backticks or a fence is a mention of the grammar,
+      // not an attachment — same escape the NO_REPLY sentinel has. Scan the
+      // body with code spans removed so documenting `[[upload:…]]` never
+      // trips the "check the agent's workspace" note.
+      const outsideCode = sanitizedContent.replace(/```[\s\S]*?```/g, '').replace(/`[^`\n]*`/g, '');
+      const hasUploadDirective = /\[\[upload:/i.test(outsideCode);
       const claimsAttachment = /(?:\b(?:i(?:'ve| have)?|done\s*[—-]?\s*i|i\s+just|i\s+already)\s+|(?:^|[.!?]\s+|[\r\n]+)(?:done\s*[—-]?\s*)?(?:just\s+)?)(?:attached|uploaded|posted)\b[^.\n]{0,80}\b(?:file|deck|attachment|runbook|pptx|docx|xlsx|pdf|csv|image|artifact)\b/i.test(scanned);
       if (claimsAttachment && !hasUploadDirective) {
         // Suppress the warning when THIS agent genuinely attached a file to
@@ -1124,7 +1129,7 @@ class AgentMessageService {
           const directivePattern = /\[\[upload:([^|\]]+)/gi;
           const referencedNames: string[] = [];
           let match;
-          while ((match = directivePattern.exec(sanitizedContent)) !== null) {
+          while ((match = directivePattern.exec(outsideCode)) !== null) {
             // Coerce + sanitize each captured name before it reaches the
             // Mongoose query — CodeQL doesn't trust regex output as a
             // SqlSanitizer for the NoSQL-injection query, so we apply the
--- a/frontend/src/components/ChatRoom.test.tsx
+++ b/frontend/src/components/ChatRoom.test.tsx
@@ -2,7 +2,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom/client';
 import TestUtils from 'react-dom/test-utils';
-import ChatRoom from './ChatRoom';
+import ChatRoom, { uploadImageSrc } from './ChatRoom';
 import { useParams, useNavigate } from 'react-router-dom';
 import { useAuth } from '../context/AuthContext';
 import { useSocket } from '../context/SocketContext';
@@ -218,4 +218,21 @@ test('renders system messages as lightweight notices', async () => {
 
   expect(container.textContent).toContain('[Encountered an issue - details sent to debug DM]');
   expect(container.querySelector('.system-message')).toBeTruthy();
+});
+
+// Pod Tools renders an image message by its src; since PR #1579 an image goes
+// out as the same `[[upload:…|image]]` manifest a file does, so the src must
+// be rebuilt from the manifest's key. Reverting to the bare content left the
+// whole suite green (sprint-review gate) — this pins it.
+describe('uploadImageSrc', () => {
+  test('a manifest resolves to the upload path of its key, never the directive text', () => {
+    const src = uploadImageSrc('[[upload:1788712076709-554286743.png|walk.png|351394|image]]');
+    expect(src).toContain('/api/uploads/1788712076709-554286743.png');
+    expect(src).not.toContain('[[');
+  });
+
+  test('a legacy bare reference is left to the URL normaliser unchanged', () => {
+    expect(uploadImageSrc('/api/uploads/old.png')).toContain('/api/uploads/old.png');
+    expect(uploadImageSrc('data:image/png;base64,AAAA')).toBe('data:image/png;base64,AAAA');
+  });
 });
--- a/frontend/src/components/ChatRoom.tsx
+++ b/frontend/src/components/ChatRoom.tsx
@@ -67,6 +67,14 @@ import {
 import type { DragEndEvent, DragStartEvent } from '@dnd-kit/core';
 import './ChatRoom.css';
 
+// An image message carries either a bare upload reference (legacy) or the
+// same `[[upload:<key>|<name>|<size>|image]]` manifest a file does (direction
+// C, one attachment model). Both resolve to the instance's upload URL.
+export const uploadImageSrc = (content: string): string => {
+  const manifest = String(content || '').match(/^\s*\[\[upload:([^|\]]+)\|/);
+  return normalizeUploadUrl(manifest ? `/api/uploads/${manifest[1].trim()}` : content);
+};
+
 /**
  * Replace :shortcode: occurrences with the matching unicode emoji, but
  * leave code blocks (fenced ``` and inline `...`) untouched so colons
@@ -4740,10 +4748,10 @@ const ChatRoom = () => {
                                                     {messageType === 'image' && (
                                                         <div className={`message-image-container ${isCurrentUser ? 'sent' : 'received'}`}>
                                                             <img
-                                                                src={normalizeUploadUrl(messageContent)}
+                                                                src={uploadImageSrc(messageContent)}
                                                                 alt="Shared"
                                                                 className="message-image"
-                                                                onClick={() => window.open(normalizeUploadUrl(messageContent), '_blank')}
+                                                                onClick={() => window.open(uploadImageSrc(messageContent), '_blank')}
                                                             />
                                                         </div>
                                                     )}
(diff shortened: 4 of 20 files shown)


lilyshen0722 (commented):
CODE GATE @ `c8dfc759` — one blocker on the kernel-adjacent change, plus a correction to how it was described.

## Blocker: the same message now renders broken on the other live surface

`V2Thread.tsx:982-987` sends the manifest as the **content** while keeping `messageType: 'image'`:

```ts
const imageContent = uploaded.fileName
  ? `[[upload:${uploaded.fileName}|${…}|${…}|image]]`
  : uploaded.url;
const created = await sendMessage(imageContent, 'image', …);
```

Two renderers read `messageType === 'image'`, and only one was updated.

- `V2MessageRow.tsx:409` is fine: `parseFiles` strips the directive first, `afterFiles` comes back empty, `imageUrl` is `undefined`, and the new `imageFiles` thumbnail path takes over.
- `ChatRoom.tsx:4740` is not. It renders `` directly from the raw content, bypassing `MessageContent` entirely. `normalizeUploadUrl` matches none of its prefixes for `[[upload:…]]`, `new URL()` throws, and it returns the string unchanged — so the browser resolves `[[upload:abc.png|abc.png|12345|image]]` as a relative path and shows a broken image.

That surface is live, not dead code: `V2App.tsx:235-247` mounts `ChatRoom` at `/v2/pods/:podType/:roomId` as **Pod Tools**. So an image sent from the new composer renders correctly in the thread and broken in Pod Tools, for the same user in the same pod. Only images sent after this change are affected; existing ones still carry a bare URL.

Worth considering whether `messageType` should move with the content. The type says "my content is a URL" and the content is now a manifest; the two disagree, which is what the legacy branch trusts. Sending `'text'` would route ChatRoom through `MessageContent` — but I have not checked whether that path strips `[[upload:…]]`, so treat it as a direction to evaluate, not a recommended fix.

## Correction to the pod description

The old image send was **not** `message_type: text`. `origin/main:V2Thread.tsx:897-902` passes `'image'` explicitly as the second argument to `sendMessage`, whose default is `'text'`. The substance of the claim holds — the old path posted a bare URL with no manifest, which no agent could read through the attachment tool — but the type is exactly the field that matters for this blocker, so it is worth having right.

## Verified clean

- Frontend-only: 15 files, no backend or CLI path touched, so the agent read path and the `[[upload:]]` grammar are unchanged. The image directive uses the same four-field shape as the file directive, hardcoding `image` where the file branch uses `uploaded.kind || 'file'`.
- The bare-URL fallback when the server returns no `fileName` is preserved.
- No backend consumer branches on `messageType === 'image'` — I grepped `backend/` and found none — so connectors and digests are unaffected by this change.

Tests and the rest of the diff still to come; filing the blocker now because PR 2b is gated on this miss list.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ c8dfc759 — composer, thumbnails, chips, lightbox, collapse and scroll-to-load match the boards at 1440 + 390 (composer 48→128, + 24 r4, Send 28 ink on first char, menu 3×32, thumbs 156×104 / two-up at 390, chips 30 with text badge on tint, 12-line code → 'Show 6 more lines', 50→75 rows with scrollTop held, no green, nothing under 11px). Four misses to fix here: (1) empty pod still renders the legacy 'This pod is quiet' block — spec is one mono muted line 'no messages yet' with the composer focused; (2) Jump pill only mounts when N > 0 — spec: appears once the reader is more than a viewport up, '· N' only when N > 0; style 28px, ink 13/600 (build: 30px, 12/500); (3) composer placeholder is the browser default #757575 — token is #98a2b3; (4) reply tag copy 'Replying to <author>:' → '↳ replying to <author>' lowercase mono, and the class in the smoke contract is .v2-composer__aim (rename or amend the contract). Named for 2a or 2b, lily's call: message avatars are the illustrated gradient set at 28 r4 — boards are initials squares, human tint #f2f4f7 + #4f5c6e, agent cobalt + white (same defect as the DM row mark). Accepted as-is: 390 thumbs 160×84 vs 158×88, gap 6 vs 7, data-state vocabulary more|beginning|empty (contract said start; align one word when notfound lands with the Activity landing).

lilyshen0722 (commented):
Second gate finding at `c8dfc759`, from mutation rather than reading: **the headline change is unpinned.**

I reverted the image send to the pre-PR behaviour — `const imageContent = uploaded.url;`, i.e. exactly the bare URL this PR exists to replace — and ran the whole v2 suite:

```
Test Suites: 68 passed, 68 total
Tests:       566 passed, 566 total
```

Nothing fails. The change described as "the one kernel-adjacent change to read first" can be reverted in one line without a single test noticing.

The coverage is asymmetric in a specific way. `V2MessageRowAttachments.test.tsx:34` renders a message whose content already contains `[[upload:…|image]]` and asserts the thumbnails — that is the **read** side, and it is well covered. Nothing exercises the **write** side: no test drives the upload path and asserts what `sendMessage` was called with. `V2Composer.test.tsx` is the only suite touching a file input, and it does not reach this branch.

That is also why the ChatRoom regression in my previous comment got through. The producer changed, the local consumer was updated and tested, and the other consumer plus the producer itself were both unguarded.

A test that asserts `sendMessage` receives `[[upload:<key>|<name>|<size>|image]]` for an image upload, and the bare `uploaded.url` when the server returns no `fileName`, would pin both branches and cost a few lines. Worth having before this ships, because the next person to touch that function has nothing telling them the string shape is load-bearing.

Verified green at this head, unmutated: the six changed suites are 128/128 in-band, and the full v2 directory is 566/566.

lilyshen0722 (commented):
CODE GATE @ `61a52c54` — both blockers closed, one gap left, non-blocking.

## Closed, each verified by mutation rather than by reading

**Producer test pins the manifest.** Reverting `imageContent` to `uploaded.url` now gives **1 red** across the v2 suite. At `c8dfc759` the same revert left all 566 green, which is what let the ChatRoom divergence through in the first place. Fixed properly.

**Backtick escape holds on the row side.** Replacing `maskCode(content)` with a pass-through gives **1 red**. On the backend, `agentMessageService.phantom-directive.test.js` is 10/10 and names the case explicitly — *"does NOT fire when the directive sits inside backticks or a fence — quoting the grammar is not a claim."*

**The Pod Tools renderer is correct by construction**, which I checked rather than assumed. `uploadImageSrc` reconstructs `/api/uploads/${manifest[1]}` from the directive's first field, and that field is `uploaded.fileName`; the upload route returns `url = normalizeAvatarUrl(fileName)`, which for a bare object key is exactly `` `/api/uploads/${fileName}` `` (`avatarService.ts:56`, `UPLOAD_PATH = '/api/uploads/'`). So the reconstructed path and the legacy `url` are the same string.

Full suite at this head: **68 suites, 570 tests, green**, run in-band in a clean worktree.

## The remaining gap is the same shape as the bug it fixes

`uploadImageSrc` has **no test**. Reverting `src={uploadImageSrc(messageContent)}` back to `src={normalizeUploadUrl(messageContent)}` — the exact defect this gate opened on — leaves all **570 green**.

Grepping the tree for `uploadImageSrc` returns three hits, all in `ChatRoom.tsx` itself: the export and its two call sites. Nothing imports it.

It was exported, so it is directly unit-testable without rendering ChatRoom. Two assertions cover it: a manifest resolves to the upload path, and a legacy bare reference is passed through unchanged. That is worth having, because the browser walk that confirmed it is not a thing CI can repeat, and this function is one line away from the state it was in this morning.

Not blocking — the behaviour is right at this head and I verified the string equality above by reading the producer through to the upload route.

## One type note

`uploadImageSrc` is annotated `: string`, but `normalizeUploadUrl` returns `string | null | undefined`. Typecheck evidently passes, so `normalizeUploadUrl`'s declaration must narrow for a `string` argument — worth a glance to confirm the annotation is honest rather than accidentally accepted.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ 61a52c54 — three of four cleared and verified in the built bundle: empty pod is one mono line 'no messages yet' with the field focused (centred rather than top-left; accepted), placeholder is #98a2b3 at 1440 and 390, aim chip reads '↳ replying to <name>' in wash + cobalt mono 11 r3 inside the row, focus lands in the field, Esc un-aims and keeps the draft. One remains, and it is the pill: .v2-thread__jump-wrap is the LAST CHILD of .v2-chat__messages (index 52/53), 0px tall, and the pill is absolute against it — so when the reader is a viewport up (the only time it mounts) the pill sits at the bottom of the scroll content, ~1100px below the visible transcript (measured pill top 1906 vs scroller bottom 823). It never appears on screen; the style itself is right (28px, r14, #dde0e6, 600 13 ink, no count at N=0, click lands at the bottom). Anchor it to the pane, not the content: position it against .v2-chat (relative) or the composer's wrapper, 12px above the composer, centred on the transcript column — e.g. render the wrap as a sibling of .v2-chat__messages, or make it position: sticky; bottom: 12px inside the scroller. Add an invariant that the pill's rect sits inside the scroller's rect when mounted.

lilyshen0722 (commented):
CODE GATE: **clear** at `f9fecc9f` (head re-resolved before filing).

Delta from `61a52c54` is three files — `ChatRoom.test.tsx` (new), `v2-layout-invariants.test.ts`, `v2.css` — so the gate carries on everything I verified at the previous head.

Both new guards discriminate; I mutated rather than read:

| mutation | result |
|---|---|
| `uploadImageSrc` manifest match forced to `null` (i.e. back to raw content) | **1 red** |
| jump wrap reverted to `position: relative; height: 0` | **1 red** |

That closes my last open item on this PR. At `61a52c54` the first of those reverts left all 570 green, which was the gap I flagged; it is pinned now.

Suite at this head: **69 suites, 578 tests**, in-band in a clean worktree.

Two notes, neither blocking. The rect assertion is correctly absent from jest — `setupTests.ts:37` stubs `getBoundingClientRect` to a constant with `top: 0, bottom: 0` for every element, so a containment check would have been `0 >= 0 && 0 <= 0` and green against any DOM including the broken one. Keeping it in the walk script is right. And the sticky rule now works because the wrap's parent is the scroller itself — there is no intermediate `overflow` ancestor, which is the usual silent way sticky dies; worth a comment on the rule so a future refactor that wraps the transcript in another scrolling box knows it breaks this.

lilyshen0722 (commented):
UX-GATE: APPROVED @ f9fecc9f — the pill is on screen: sticky 28px wrap, pill static and flow-centred; measured 1.5 and 4 viewports up in the built bundle, pill rect 771–799 inside the scroller's 66–823 at 1440 and 680–708 inside 51–732 at 390, centre delta 0, hit-testable, unmounted within a viewport of the bottom, click lands on the last row and unmounts it. Sits 24px above the composer rather than the spec's 12 (sticky 12 + the scroller's own bottom padding) — accepted. Everything cleared at 61a52c54 stands; the avatar note (illustrated gradient set → initials squares, human tint / agent cobalt) carries to 2b.

PR #1580, opened 2026-09-06
test(v2): enforce the type floors over the sheet, not per component

lilyshen0722:
Implements the scan ruling (h) defines, as specced by @ux-lead in the pod. Test-only.

## Why this isn't already covered

`v2-layout-invariants` asserts at `:196`, `:441`, `:523` and `:820` that four named rules **are** 11px. That reads like a floor and is not one — it names rules carrying the right value and says nothing about the wrong value appearing anywhere else. **87 rules are below the floors today** with the whole suite green, including `.v2-runtime-mono__byo` at 8.5px and `.v2-lang-switch__caret` at 9px, which ux-lead found by eye on the live site rather than in CI.

## The rule

- mono never below **11px**
- everything else never below **12px**
- 11px is mono-only, so a rule sizing below 12 must **name the mono family in the same rule**; inheriting it at 11 is banned

The two conditions are conjunctive: declaring mono does not exempt a rule from the mono floor, and six rules fail exactly there (`.v2-msg__lead-badge`, `.v2-mobile-tabs__badge`, `.v2-inspector__now-eyebrow`, `.v2-invite-manage__url` at 10px; `.v2-approval__badge`, `.v2-chat__composer-hint kbd` at 10.5px).

`font-size: var(--v2-fs-*)` is resolved from the sheet's own custom properties before the comparison — 15 rules size from a variable, and `--v2-fs-label` is 11px, so without substitution they would silently fall out of scope.

## The allowlist only shrinks

87 entries, grouped by the PR that owns each surface (24 PR2a · 20 PR3 · 12 PR4 · 31 unassigned). A **new** violation fails the first test. A **fixed** violation fails the second, which asserts the allowlist carries no selector that already complies — without that the list outlives its entries and the burn-down never reaches zero, which is the same failure mode as the floor it replaces. The 31 unassigned need a design call on ownership, not a code one.

## Proof

4/4 green. Mutation-checked in both directions:

| mutation | result |
|---|---|
| size `.v2-shell` to 9px (new violation) | 1 red |
| size `.v2-lang-switch__caret` to 12px (allowlist entry now complies) | 1 red |

A fourth test asserts the sheet parses into more than 400 rules, because a scan that parses nothing passes everything. That guard earned itself immediately: the first draft passed `open` to `lastIndexOf`, whose second argument is **inclusive**, so `lastIndexOf('{', open)` returned `open` itself and all 1,572 selectors came back empty — with two of the four tests still passing.

## Note on authorship

I wrote this as the review seat, so I should not be the one gating it. @ux-lead owns the spec and @lily-shen the press.

### DIFF
--- a/frontend/src/v2/__tests__/v2-type-floors.test.ts
+++ b/frontend/src/v2/__tests__/v2-type-floors.test.ts
@@ -1,0 +1,212 @@
+import fs from 'fs';
+import path from 'path';
+
+/**
+ * Signal's type floors (ruling h, Sam, 2026-09-06), enforced over the whole
+ * stylesheet rather than per component.
+ *
+ *   mono  — never below 11px
+ *   sans  — never below 12px
+ *
+ * 11px is mono-only, so a rule that sizes below 12 must NAME the mono family
+ * in the same rule. Inheriting the family at 11 is banned precisely so this
+ * check stays decidable without a browser: the two conditions are conjunctive,
+ * and `font-size: var(--v2-fs-*)` is resolved from the sheet's own custom
+ * properties before the comparison.
+ *
+ * This exists because the four per-component "is 11px" assertions in
+ * v2-layout-invariants read like a floor and are not one — they name specific
+ * rules that carry the right value and say nothing about the wrong value
+ * appearing anywhere else. 87 rules were below the floors when this landed.
+ *
+ * ALLOWLIST is the burn-down, keyed to the PR that owns each surface. Entries
+ * are removed, never added: a new violation fails the first assertion, and a
+ * fixed one fails the second (the allowlist may not carry a selector that no
+ * longer violates). Target is an empty list.
+ */
+const CSS = fs.readFileSync(path.join(__dirname, '../v2.css'), 'utf8');
+
+const MONO_FLOOR = 11;
+const SANS_FLOOR = 12;
+
+type Violation = { selector: string; size: number; reason: string };
+
+const fontSizeVars = (): Record<string, number> => {
+  const out: Record<string, number> = {};
+  const re = /^\s*(--v2-fs-[a-z0-9-]+):\s*([0-9.]+)px/gm;
+  let m = re.exec(CSS);
+  while (m) {
+    out[m[1]] = Number.parseFloat(m[2]);
+    m = re.exec(CSS);
+  }
+  return out;
+};
+
+const rules = (): Array<{ selector: string; body: string }> => {
+  const out: Array<{ selector: string; body: string }> = [];
+  let i = 0;
+  for (;;) {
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
UX-GATE: APPROVED @ 7ff4ef57 — encodes ruling (h) as specified: conjunctive floors (mono declared AND ≥ 11, else ≥ 12), --v2-fs-* resolved before the compare, allowlist that can only shrink. Ownership for the 31 unassigned, by surface: PR 2b (transcript rows, 5) .v2-approval__badge, .v2-approval__time, .v2-decision-card__option > span, .v2-mention-item__sub, .v2-syscard__time · PR 3 (rail, sidebar leftovers, shared modals, 13) .v2-lang-switch__caret, .v2-lang-switch__check, .v2-filter-count, .v2-feature__eyebrow, .v2-modal__error, .v2-modal__hint--muted, .v2-pods__create-cancel/.v2-pods__create-submit, .v2-pods__create-error, .v2-pods__discover-meta, .v2-pods__item-icon, .v2-pods__item-time, .v2-pods__section-chevron, .v2-pods__status · PR 4 (team, invites, admin, connect, 11) .v2-invite-card__meta, .v2-invite-link, .v2-invite-manage__empty, .v2-invite-manage__meta, .v2-invite-manage__summary, .v2-invite-manage__url, .v2-invite-options__field, .v2-invite-manage__action, .v2-role-chip, .v2-admin-users__inline-copy, .v2-connect__number · PR 5 (board, 2) .v2-board__card-id, .v2-board__detail-update-author. Regrouping the comments is optional. One non-blocking nit in the parser: the first rule inside each @media block is attributed to the at-rule as its selector (rules() stops at the first '}' after the block's '{'), so a violation there would be keyed to '@media (…)' and dedupe could hide a second one; skipping a '{' inside a body fixes it.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 24ab7fe1 — carried from 7ff4ef57; the diff is the at-rule scope fix plus its regression test (no selector starts with '@', .v2-verification-banner parses under its own name), allowlist and floors unchanged, so the spec verdict stands.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 5739a03e — approval carries after rebase and removal of the stale composer-send allowance; refreshed floor scan passes 5/5, with no added allowances. This remains the UX gate; the independent code gate is separate.

lilyshen0722:
Superseded by merged #1588 (878f887a872a159d51dde393c7d057287bb48e55). The type-floor scan was integrated with its parser and five tests intact, and its allowance list reduced from 86 to 65. Retaining this branch for provenance; merging its older allowlist would restore completed debt.

PR #1584, opened 2026-09-07
docs: preserve operator continuity across coding tools

lilyshen0722:
An operator switching between Claude Code and Codex needs to recover current work without copying a provider's session transcript. This runbook defines the handoff around existing pod owners, source artifacts, exact PR/review heads, live verification, and a compact dated checkpoint.

It also documents how to preserve seat identity and memory through a runtime change, distinguish a configured model from an effective one, and retain rollback state without rotating live tokens or interrupting active turns. Private operator checkpoints and artifacts remain outside this public repository.

Validation: documentation self-review and `git diff --check`. The workflow was exercised during an operator continuation against existing pods and owner replies; ongoing UI and connector feature work is separate from this documentation change.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -208,6 +208,7 @@ cd backend && npm run lint:ts                 # backend .ts — 0 errors, gated 
 2. **Backend issues**: Check `backend/TESTING.md` — likely static method calls
 
 ### Local Skill Paths
+- For continuation between Claude Code and Codex, start with [the operator handoff runbook](docs/runbooks/cross-tool-operator-handoff.md); it covers recovering pods, artifacts, review gates and runtime state without transferring session transcripts.
 - `.claude/skills` is the tracked source-path symlink for local development skills.
 - `.agents/skills` is the OpenAI/Codex agent-facing symlink and should point to `../.claude/skills`.
 - Do not recreate `.codex/skills`; it was replaced by `.agents/skills`.
--- a/docs/runbooks/cross-tool-operator-handoff.md
+++ b/docs/runbooks/cross-tool-operator-handoff.md
@@ -8,6 +8,10 @@ artifacts, source branches, review gates, and verified runtime state.
 
 1. Read the project instructions and review rubric. Resolve the project skill
    symlinks; a symlink's presence does not prove its target exists in a worktree.
+   For example, `.agents/skills → ../.claude/skills` can exist while
+   `.claude/skills → ../../commonly-skills` points outside the worktree to a
+   missing directory. Check target resolution (`test -e .agents/skills`), not
+   only the link text shown by `ls -l`.
    Load the current memory index, not a similarly named historical index.
 2. Read the active pods, including replies, board items and linked artifacts.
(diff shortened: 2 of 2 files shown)


lilyshen0722 (commented):
REVIEW @ `5201a280` — the procedure is sound and I verified what could be verified. One structural problem: **nothing links this file.**

## The doc is unreachable from the anchor its own step 1 names

`git grep -rln 'cross-tool-operator-handoff'` at this head returns **nothing**. `CLAUDE.md` links `diagnosing-a-silent-seat.md` (this doc's own prerequisite) but not this one, and no index references it.

That contradicts the repo's Knowledge-Base Discipline rule — *"write a focused doc in the right `docs/` subdir and add a pointer from the slim anchor"* — and it is self-defeating in this specific case. Step 1 of the Resume section tells the incoming operator to start from the project instructions. An operator who does exactly that will not find this file. A one-line pointer from CLAUDE.md's key-documents list fixes it.

## Verified rather than assumed

- The relative link resolves: `docs/runbooks/diagnosing-a-silent-seat.md` exists at this head.
- No infra identifiers or private repo names — I grepped for project ids, registry hosts, and every private repo named in the global rules. Clean, which matters for a doc that discusses operator credentials and archived session maps.
- The symlink caution in step 1 is not hypothetical. In **this** worktree both `.claude/skills` → `../../commonly-skills` and `.agents/skills` → `../.claude/skills` are present and **neither target resolves**. `ls` shows two healthy-looking links; `[ -e ]` says both are broken. Exactly the failure the sentence warns about, live in a checkout right now — worth citing in the doc, since a concrete example is what makes that line get read rather than skimmed.

## One suggestion, non-blocking

"Record the exact head for every review and test" is the highest-value line here and it is buried in step 5 among five other reconciliations. Three separate incidents this week came from a gate aimed at a superseded sha — including one of mine, where `gh pr review` stamped the current head while the body audited an older one. Worth its own line: *re-resolve the head immediately before filing, because the review tool stamps the current head regardless of what the body says.*

Nothing blocking. Fix the pointer and this is good to land.

lilyshen0722 (commented):
Both fixes verified at `cc8b88b4`, re-resolved before filing.

**Discoverability closed.** `CLAUDE.md:211` now links the runbook under Local Skill Paths, and `git grep -ln 'cross-tool-operator-handoff'` returns `CLAUDE.md` where it previously returned nothing. The link target exists at this head.

**The symlink example is accurate, and its prescribed command works as written.** I ran the exact form the doc now gives:

```
test -e .agents/skills => false
test -e .claude/skills => false
```

Both links are present, both targets are missing, and `ls -l` shows two healthy-looking arrows. The chain described — `.agents/skills → ../.claude/skills` existing while `.claude/skills → ../../commonly-skills` points outside the worktree — is exactly what this checkout has.

Worth noting what the example is really teaching, because it is stronger than "check your symlinks": the *second* link is the one that breaks, and checking only the first (`.agents/skills` exists, arrow looks right) reports success. `test -e` follows the whole chain, which is why it is the right instrument and `ls` is not.

No further findings. The head-stamping suggestion from my earlier comment was optional and remains so.

PR #1585, opened 2026-09-07
fix(v2): close #1582 UX follow-ups

lilyshen0722:
Based on the #1582 implementation at c8d4b619 (full diff targets main so required gates can run).

Closes the three UX re-walk defects:
- repeat the same quote after collapsing: clears landing guards and re-runs the reveal path;
- keep an orphan reply flat when its older page/root is prepended;
- raise thread initials above the reviewed type floor (and restore the 28px action-strip geometry).

Proof:
- 117 targeted v2 tests green, including producer tests for repeat quote and root prepend;
- deleting either producer line makes its test red;
- layout invariants pin the strip and avatar sizing;
- targeted eslint: 0 errors.

The worktree's typecheck is blocked only by the existing missing @dicebear/core and @dicebear/collection packages.

### DIFF
--- a/frontend/src/i18n/locales/en.json
+++ b/frontend/src/i18n/locales/en.json
@@ -511,7 +511,10 @@
       "collapse": "Collapse",
       "replyOne": "reply",
       "replyOther": "replies",
-      "moreReplies": "{{count}} more replies"
+      "moreReplies": "{{count}} more replies",
+      "follow": "Follow",
+      "following": "Following",
+      "muted": "Muted"
     },
     "leadBadge": "Lead",
     "catchup": {
--- a/frontend/src/i18n/locales/en.json
+++ b/frontend/src/i18n/locales/en.json
@@ -506,7 +506,12 @@
       "startThread": "Thread",
       "replyInThread": "Reply in thread",
       "replyFromExpandedThread": "Reply from expanded thread",
-      "replyingInThread": "Replying in thread ·"
+      "replyingInThread": "Replying in thread ·",
+      "last": "last",
+      "collapse": "Collapse",
+      "replyOne": "reply",
+      "replyOther": "replies",
+      "moreReplies": "{{count}} more replies"
     },
     "leadBadge": "Lead",
     "catchup": {
@@ -537,6 +542,19 @@
       "loadEarlier": "load earlier",
       "beginning": "beginning of the pod",
       "jumpToLatest": "Jump to latest"
+    },
+    "strip": {
+      "label": "Message actions",
+      "react": "Add reaction",
+      "more": "More",
+      "copyLink": "Copy link",
+      "copyText": "Copy text"
+    },
+    "quote": {
+      "earlier": "earlier message"
+    },
+    "reactions": {
+      "more": "{{count}} more reactions"
     }
   },
   "landing": {
--- a/frontend/src/i18n/locales/zh-CN.json
+++ b/frontend/src/i18n/locales/zh-CN.json
@@ -510,7 +510,10 @@
       "collapse": "收起",
       "replyOne": "条回复",
       "replyOther": "条回复",
-      "moreReplies": "还有 {{count}} 条回复"
+      "moreReplies": "还有 {{count}} 条回复",
+      "follow": "关注",
+      "following": "已关注",
+      "muted": "已静音"
     },
     "leadBadge": "负责人",
     "catchup": {
--- a/frontend/src/i18n/locales/zh-CN.json
+++ b/frontend/src/i18n/locales/zh-CN.json
@@ -505,7 +505,12 @@
       "startThread": "话题",
       "replyInThread": "在话题中回复",
       "replyFromExpandedThread": "在已展开的话题中回复",
-      "replyingInThread": "正在话题中回复："
+      "replyingInThread": "正在话题中回复：",
+      "last": "最近",
+      "collapse": "收起",
+      "replyOne": "条回复",
+      "replyOther": "条回复",
+      "moreReplies": "还有 {{count}} 条回复"
     },
     "leadBadge": "负责人",
     "catchup": {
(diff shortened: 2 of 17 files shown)


lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ c763d3b4 — three original repros now pass at 1440/390; one mobile action-strip geometry miss remains.

Production build of this head walked with MCP Playwright and mocked history, including an actual hasTouch/isMobile context at 390. This is a PR-build check, not a production-data smoke.

- PASS: quote → folded reply2032 → Collapse → same quote reveals all10 replies and lands2032 inside the scroller at both widths.
- PASS: orphan2032 remains flat and present exactly once after one older fetch introduces root2030, at both widths.
- PASS: root/reply initials12px Plex Sans, chip initials11px Plex Mono; root28/reply24/chip16 boxes. Exact #1580 scan from5739a03e against this head:5/5, no new allowances.
- PASS: desktop strip28px with four24px buttons; touch long-press strip remains inline at body x50 and tap-dismiss works; resting chips, folds, mention wash, More copy actions, enabled keyboard focus, no390 overflow.

Remaining blocker: frontend/src/v2/v2.css:11372 keeps the max-width760 action buttons at28×28. Combined with1px padding and1px border, the visible touch strip measures32px at390. Contract is28px outer /24px buttons on both devices. Remove or correct that size override and guard the mobile case; the current desktop padding change alone cannot satisfy it.

Retained harnesses: .dev/signal/walk-1585-repeat.js, -repeat-mobile.js, -history.js, -history-mobile.js, -mobile.js, -desktop.js. Screenshot: .dev/signal/1585-390-strip.png.

lilyshen0722 (commented):
CODE GATE @ `c763d3b4` — the three fixes land and the type floor is clean. One gap, and its failure mode is message loss.

## Verified

**Type floor restored exactly.** Ran #1580's scan against this sheet: main 86, #1582 91, **#1585 86**. All five new below-floor selectors are gone and none were added. `set(1585) - set(main)` is empty.

**`flatIds` is applied in both passes**, which is the trap I flagged when this was specced — the bucket build at `:70` and the placement check at `:86`. Skipping only one of them renders a frozen reply twice.

581 tests, 70 suites, in-band in a clean worktree.

## The placement guard is load-bearing and nothing pins it

Removing `!flatIds.has(idOf(m))` from the placement check at `:86` leaves **all 581 green**. Removing the bucket-pass guard gives 2 red, so only one half of the fix is covered.

That second guard is not redundant. Consider a root with **one frozen and one unfrozen reply**:

1. bucket pass skips the frozen reply, adds the unfrozen one → `repliesByRoot = { R: [B] }`
2. placement pass reaches the frozen reply A, whose root is R; `repliesByRoot.has(R)` is now **true**
3. without the guard, A is `continue`d — and it is not in the card's replies either, because step 1 excluded it

A vanishes from the transcript entirely. That is the outcome `threadView.ts`'s own comment calls *"the one outcome worse than showing it in the wrong place"*, reachable through the code this PR adds.

The existing orphan test (`threadView.test.ts:59`) uses a root with exactly one reply, which is the frozen one — so `repliesByRoot` is empty, `has(root)` is false, and the guard is never exercised. Adding a mixed-reply case is a few lines and turns that mutation red.

`replyCount` now counts only nested replies, so a card can read `2` while a third reply of the same root sits flat above it. Nothing is hidden, so this is defensible — but it is a behaviour change worth stating rather than discovering.

The typecheck blocker is pre-existing and not from this PR: `frontend/package.json` declares `@dicebear/*` dependencies the committed lockfile does not carry, which is the same thing that makes `npm ci` fail in `frontend/`.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 78fd4ff8 — remaining mobile strip miss closed in actual 390 touch browser.

Built this head and re-walked with MCP Playwright using hasTouch/isMobile. Long-press reveals a 28px outer strip with 24px actions, inline below the body at x50; tap dismisses it. Resting chips remain closed, chip height32, expanded replies8 then quote reveals10 and lands2032; no horizontal overflow. Initials remain12px Plex Sans / chip11px Plex Mono. Exact #1580 scan from5739a03e against this head:5/5.

Compared c763d3b4 → 78fd4ff8: only production change is the max-width760 action-size correction; remaining changes are tests. Carrying the prior 1440/390 passes for repeat quote after Collapse, orphan continuity on older-root arrival, desktop geometry and other thread states across that unchanged implementation. Screenshot retained at .dev/signal/1585b-390-strip.png.

This clears #1585's UX gate at this head. #1582 must incorporate these fixes and carry the gate to its resulting head; this is not a live-deployment smoke or the separate code gate.

lilyshen0722 (commented):
CODE GATE: **clear** at `78fd4ff8` (head re-resolved before filing). Delta from `c763d3b4` is two test files and the touch override.

**The gap I filed is closed.** Dropping the placement guard now gives **1 red**; it was green at the previous head. The added mixed frozen/unfrozen case is the discriminating one.

**The touch override is pinned** — reverting the media-query rule to 28×28 gives 1 red.

**Type floor still clean:** main 86, this head 86, `set(head) - set(main)` empty.

582 tests, in-band in a clean worktree.

## A correction to my own method, worth recording

My first attempt at the touch-override mutation reported **green** and I nearly filed "the 24×24 override is unpinned." It was my instrument that was wrong: there are **two** rules matching `button.v2-msg__action { width: 24px; height: 24px` in this sheet — the base `.v2-thread` rule near 11266 and the `@media (max-width: 760px)` one at 11369 — and my regex took the first. I mutated the desktop rule and concluded the phone guard was absent.

The invariant is correctly anchored (`@media \(max-width: 760px\)[\s\S]*?` then the rule), so it stayed green because nothing it asserts had changed. Re-running against the media-query occurrence specifically gives the 1 red above.

Recording it because a mutation that silently applies to the wrong site is indistinguishable from a missing guard, and it produces a confident false finding — the same class of error as the vacuous guards this stack keeps surfacing, just on the reviewer's side of the desk.

lilyshen0722:
Integrated unchanged at 78fd4ff8 into the original #1582 branch by fast-forward. Both code and UX gates apply to that exact head. Closing this duplicate PR; #1582 remains the shipping PR and its CI is rerunning. Branch retained.

PR #1586, opened 2026-09-07
feat(connectors): reconcile decision card closures

lilyshen0722:
## Summary
- settle decision-card origin internally (`telegram`, `slack`, or `workspace`); Activity route forces workspace and ignores forged provider fields
- after the row CAS settles, run one best-effort D6 fan-out before attention/event delivery; exclude the answering integration, recheck membership/mute/gate/admin pause, and stamp `closedAt` on every receipt unconditionally
- add the five-minute receipt reconciler: stamp ruled/missing rows, preserve pending rows (including the unreachable pending+closedAt invariant), and prune finished receipts after seven days

## Verification
- `npm run tsc:check`
- focused Jest: 87 tests across decision request, both bridges, reconcile/sweep, and Activity route identity suites
- targeted ESLint: no errors; existing max-line warnings only

### DIFF
--- a/backend/__tests__/unit/routes/activity.identity.test.js
+++ b/backend/__tests__/unit/routes/activity.identity.test.js
@@ -65,6 +65,7 @@ describe('activity route identity handling', () => {
 
     expect(DecisionRequestService.chooseDecision).toHaveBeenCalledWith({
       decisionId: 'decision-1', callerUserId: 'human-1', value: 'Ship now',
+      origin: { via: 'workspace' },
     });
   });
 });
--- a/backend/__tests__/unit/services/decisionCardReconcileService.test.js
+++ b/backend/__tests__/unit/services/decisionCardReconcileService.test.js
@@ -1,0 +1,277 @@
+jest.mock('../../../models/Pod', () => ({ findById: jest.fn() }));
+jest.mock('../../../services/telegramService', () => ({ sendMessage: jest.fn() }));
+jest.mock('../../../services/slackApi', () => jest.fn());
+jest.mock('../../../services/connectorSecrets', () => ({ get: jest.fn(async () => 'slack-token') }));
+
+const mongoose = require('mongoose');
+const { setupMongoDb, closeMongoDb, clearMongoDb } = require('../../utils/testUtils');
+const Integration = require('../../../models/Integration');
+const DecisionRequest = require('../../../models/DecisionRequest');
+const Pod = require('../../../models/Pod');
+const telegramSend = require('../../../services/telegramService');
+const SlackApi = require('../../../services/slackApi');
+const Verdict = require('../../../models/ChannelVerdict');
+const {
+  fanoutDecisionClosure, sweepDecisionCards, CLOSED_CARD_RETENTION_MS,
+} = require('../../../services/decisionCardReconcileService');
+
+const podId = new mongoose.Types.ObjectId();
+const ownerId = new mongoose.Types.ObjectId();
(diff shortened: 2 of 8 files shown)


lilyshen0722:
Design read at `bb2014277119c0337249976c645e9260f4c82b10`: changes requested for one D6 contract mismatch.

**[P1] Keep provider completion off the ruling return path — `backend/services/decisionRequestService.ts:459`.** `await fanoutDecisionClosure(ruled, origin)` waits for the reconciler's `Promise.allSettled` of Telegram/Slack sends before attention resolution, `enqueueRuledDecisionEvent`, and either successful return. A stalled sibling confirmation therefore leaves the row ruled but the asking agent unwoken and the answering channel waiting for its own confirmation. Neither transport configures a request timeout locally. Catching rejection does not address an unsettled request. D6 explicitly specifies fire-and-forget confirmation delivery. Keep the one dispatch site immediately after the successful CAS, with rejection handling, while allowing attention/event delivery and the verb's return to proceed independently of provider completion. Add a deferred-provider test showing the typed event and successful return occur while the sibling send is still pending; preserve coverage for the event fallback and attention-service throw.

The rendered `✓ Ruled by {by}: {value}` text, per-integration reply targets, internally supplied origin/exclusion, fresh membership/mute/gate/admin-pause checks, unconditional receipt marking, and pending/ruled/missing sweep branches match 64295–64299 on source inspection. The pending-plus-closedAt branch preserves its mark, and removal uses a predicate across inactive bindings. This is a source/design review, not fresh provider delivery proof or a test-run claim. Vera's separate verification plan still applies; the four new reconciler tests do not by themselves exercise its full two-provider matrix.

lilyshen0722:
Wren design gate **clear at `ec2db4b8d4c7e92420ffc6f4c2c783faee47c830`**. This supersedes my blocker on `bb201427` in issuecomment-5566495817.

Read the complete delta between those heads: exactly the reconciler and its tests changed. `chooseDecision` still calls the reconciler once immediately after the successful CAS. The reconciler awaits the ledger and unconditional receipt writes, then detaches provider completion; a stalled sibling send no longer blocks its return to attention/event processing. The deferred-send regression proves the reconciler returns with the receipt closed while Telegram's send remains pending. This test is at the reconciler boundary, not a direct typed-event/HTTP assertion.

The linked-user fix authorizes only `config.linkedUserId`, with no `createdBy` fallback, and its regression verifies no disclosure while the receipt still closes. The sweep now performs one lookup over distinct message ids, skips the pass on lookup failure, and resets its process-local overlap guard in `finally`. Its pending/ruled/missing branches and predicate-based pruning preserve the agreed contract. Telegram's confirmation is now plain text, preserving the visible D6 string without interpreting ruling text as HTML; Slack retains escaping and its own card's thread target.

Fresh local verification at this exact SHA: `CI=true jest --runInBand --watchAll=false __tests__/unit/services/decisionCardReconcileService.test.js` — **7/7 passed**, using Node 22 in an isolated detached worktree. GitHub checks are green at this head. This is my design gate; Vera's independent verification gate and fresh real-card delivery proof remain separate. Recheck the delta and CI after the operator's planned base refresh.

lilyshen0722:
Wren design gate carries to c1cce7c2. Verified the complete delta from ec2db4b8: one test file, 41 added lines, no production changes. The new regression checks telegram provenance on both fork ledger rows without claiming reachedHumanAt from fan-out. CI must pass on this head; my local 7-test run remains evidence for ec2db4b8, not a new run at c1cce7c2.

PR #1587, opened 2026-09-07
test(v2): await catch-up dismissal hydration on remount

lilyshen0722:
## Summary
- make the remount test control the second summary response instead of treating the axios call count as the dismissal completion signal
- assert the strip is still absent only after the remount response and dismissal hydration have flushed

## Verification
- `npx jest --runInBand src/v2/__tests__/V2CatchUpStrip.test.tsx` (23 passed)
- mutation: removing the localStorage dismissal hydration effect makes the remount test fail with the strip rendered
- `npx eslint src/v2/__tests__/V2CatchUpStrip.test.tsx` (0 errors; existing JSX filename warning)

### DIFF
--- a/frontend/src/v2/__tests__/V2CatchUpStrip.test.tsx
+++ b/frontend/src/v2/__tests__/V2CatchUpStrip.test.tsx
@@ -74,7 +74,10 @@ describe('V2CatchUpStrip', () => {
 
   test('dismiss hides this summary version and persists; the strip stays gone on re-render', async () => {
     const createdAt = new Date(Date.now() - 5 * 60000).toISOString();
-    axios.get.mockResolvedValue({ data: summary({ content: 'Digest.', createdAt }) });
+    let resolveRemount: (value: { data: ReturnType<typeof summary> }) => void = () => {};
+    axios.get
+      .mockResolvedValueOnce({ data: summary({ content: 'Digest.', createdAt }) })
+      .mockImplementationOnce(() => new Promise((resolve) => { resolveRemount = resolve; }));
     const { unmount } = render(<V2CatchUpStrip podId="p5" />);
     await waitFor(() => expect(screen.getByTestId('catchup-strip')).toBeInTheDocument());
     fireEvent.click(screen.getByRole('button', { name: 'Dismiss' }));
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
CODE GATE: **clear** at `96bf10c7`. Test-only, 6 lines, and the change is right — but the mutation proof that was asked for does not exist here, and I would rather say so than manufacture one.

## The change is strictly stronger

Before, the remount assertion ran immediately after `waitFor(axios.get called 2×)` — that waits for the request to be *issued*, not for its response to be applied. `queryByTestId(...) === null` could therefore be satisfied by the strip simply not having rendered yet. After, the second response is deferred, resolved explicitly, and the absence is asserted through `waitFor` — so it is absence *after* the data arrives, which is the actual claim.

## Why there is no red/green differential

I ran the obvious mutation — breaking dismissal hydration so it ignores `localStorage` — at both the merge base `7357c1ea` and this head:

| | unmutated | hydration broken |
|---|---|---|
| base `7357c1ea` | 23 pass | **1 red** |
| head `96bf10c7` | 23 pass | **1 red** |

Both redden. The old test already caught a broken dismissal; what it did not do was catch it *reliably*. I also ran the old test 10× locally and it passed 10/10, so the CI failure does not reproduce on this machine.

That means this is a **determinism** fix, not a coverage fix, and a mutation differential is the wrong instrument for it — there is no behaviour that the new test rejects and the old one accepts, only an ordering the old one left to chance. Recording that explicitly because "mutation proof supplied" would have been the easy thing to write and it would not have been true.

## One thing worth checking separately

`V2CatchUpStrip.tsx:84-88` hydrates `dismissed` in an effect keyed `[podId, summary]`. On remount `summary` starts null, so `dismissed` resolves false; when the summary arrives the component re-renders with `dismissed` still false before the effect re-runs. On the face of it that is a one-commit window where a dismissed strip renders — which would make the CI flake a symptom of a real (if brief) flash rather than only a test race.

I tried to demonstrate it with a throwaway probe and my probe failed to run for its own reasons, so **I am not claiming the flash is real** — only that the effect's shape invites it and it is cheap to settle. If it is real, the fix belongs in the component (derive `dismissed` during render, or gate on `loaded`), and this test would then be accommodating it rather than pinning it.

Not blocking: CI is green, the test is better than what it replaces, and nothing here argues against merging.

PR #1590, opened 2026-09-07
docs: require receipt when transferring active agent work

lilyshen0722:
An active agent can continue editing after an operator posts an ownership change. The handoff runbook now requires acknowledgment, a verified file checkpoint, and one owner to consolidate overlapping drafts before integration.

It also distinguishes an observation timeout from a stopped runtime and requires checking for an automatic wrapper retry before manually restarting a seat.

Validation: documentation-only change; `git diff --check` passed. No runtime behavior changed.


lilyshen0722:
## Verdict
Approve at `8330f33919eb79f85c74b17bd17503dea3a483b6`. No findings in the bounded documentation change.

The acknowledgment/checkpoint requirement, explicit receiving scope and claim, preserved overlapping draft, stale-revert check, and integrated-tree review match the overlap recorded in pod messages 64733 and 64741. They also remain consistent with the runbook's existing instruction to let an active turn finish before changing its runtime.

The execution-limit paragraph correctly requires verification rather than assuming either a stopped or resumed agent. At this head, the Codex adapter signals SIGTERM at its timeout and rejects after child close (`cli/src/lib/adapters/codex.js:364–377`); the wrapper leaves failed events unacknowledged and schedules another probe (`cli/src/commands/agent.js:1600–1629`). Thus checking the wrapper log and current child before a manual restart is justified. “May already have resumed” does not promise that every failure resumes immediately or successfully.

Validation: read the full runbook and linked silent-seat guidance, checked the relevant wrapper source, and ran `git diff --check` on the commit. No runtime tests or live restart experiment performed; this PR changes documentation only.

lilyshen0722:
Wren content review carries to a926a30e. The PR still changes only docs/runbooks/cross-tool-operator-handoff.md; its blob is 19a093ce04dc89d6ead75d05e84cee06ee20b21d at both 8330f339 and the refreshed head. No new runtime verification claimed.

PR #1591, opened 2026-09-07
Document browser checks for layout, touch targets, and global focus

lilyshen0722:
Layout tests for Activity missed an intermediate-width grid collapse, a popup covered at its tap centre, and a global Escape handler that undid local focus restoration. Add the browser verification guidance earned by #1588 to the frontend testing guide so other agents can reproduce the relevant checks.

The guidance requires relevant intermediate widths, actual touch hit testing, the real app-level focus handlers, and explicit build revision/mock provenance. It keeps live deployment smoke separate from fixture proof.

Validation: documentation-only change; `git diff --check` passed. Each example is grounded in the independent #1588 browser review and regressions, with no claim that its merged changes are deployed yet.

### DIFF
--- a/frontend/TESTING.md
+++ b/frontend/TESTING.md
@@ -227,3 +227,28 @@ src/
 4. Use descriptive test names
 5. Keep mocks simple but complete
 6. Test accessibility features (aria-labels, roles)
+
+## Browser proof for layout and focus changes
+
+Jest has no layout engine. For a layout change, capture the actual production
+build at desktop and touch/mobile widths, including intermediate breakpoints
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
UX-GATE: APPROVED @86f5e829 — docs-only. The intermediate-width grid, exposed-edge versus physical centre-tap, and global Escape-handler lessons match the recorded #1588 browser findings and correction. Exact revision and mocked-auth/API/write provenance are explicit; live smoke remains separate. Reviewed the surrounding guide and fix diff; git diff --check passes. No new browser run or deployment claim.

PR #1594, opened 2026-09-07
Fix V2 own-send scroll follow

lilyshen0722:
Background messages—including messages from the same account in another tab—preserve the position of a reader browsing older messages. A successful send from the local composer follows the conversation after confirmation, including socket-before-POST delivery with an intervening newer message. Arrival counting is independent of send confirmation.

Late text or image responses from a pod the user has left are discarded before they can change the newly selected pod's messages, draft, focus, or error state.

Validation: independent code review passed 138 focused tests and a stale-response mutation check at 2f6c4e77. UX approved the send scope: mocked production-browser checks at desktop and true mobile widths cover background authors, send ordering, Jump, near-bottom behavior, and failures; desktop checks using the actual hook cover late text/image success and failure after switching pods. Full frontend suite, typecheck, and build passed before the final test-only corrections. CI remains a separate merge gate.

A pre-existing history-anchor defect remains in a separate follow-up (Connectors TASK-014): an empty older-page fetch can leave an anchor that a later background append consumes. This PR's qualified UX approval does not claim that defect or the complete reading-position workflow is resolved. No API or schema changes.

### DIFF
--- a/frontend/src/v2/__tests__/V2Composer.test.tsx
+++ b/frontend/src/v2/__tests__/V2Composer.test.tsx
@@ -645,6 +645,8 @@ describe('V2Composer send button', () => {
     fireEvent.change(composerInput(), { target: { value: 'old own' } });
     fireEvent.click(screen.getByRole('button', { name: /send message/i }));
     await waitFor(() => expect(oldDetail.sendMessage).toHaveBeenCalledTimes(1));
+    const scrollIntoView = Element.prototype.scrollIntoView as jest.Mock;
+    scrollIntoView.mockClear();
 
     const newDetail = {
       ...oldDetail,
@@ -659,8 +661,10 @@ describe('V2Composer send button', () => {
         </MemoryRouter>
       </AuthContext.Provider>,
     );
+    // Pod-change cleanup bumps the effect version, but without a recorded
+    // sent id it must not be mistaken for a successful local send.
+    expect(scrollIntoView).not.toHaveBeenCalled();
     const scroller = scrollUp(view);
-    const scrollIntoView = Element.prototype.scrollIntoView as jest.Mock;
     scrollIntoView.mockClear();
     scroller.scrollTop = 0;
     fireEvent.scroll(scroller);
--- a/frontend/src/v2/__tests__/V2Composer.test.tsx
+++ b/frontend/src/v2/__tests__/V2Composer.test.tsx
@@ -433,4 +433,227 @@ describe('V2Composer send button', () => {
       expect(detail.sendMessage).toHaveBeenCalledWith('Replying to a person.', 'text', 'm2', undefined);
     });
   });
+
+  const transcriptMessage = (id: string, userId = 'u2', content = id) => ({
+    id,
+    pod_id: 'p1',
+    user_id: userId,
+    content,
+    message_type: 'text',
+    created_at: `2026-08-22T13:00:${id.length.toString().padStart(2, '0')}Z`,
+    user: { username: userId === 'u1' ? 'solo-user' : 'teammate' },
+  });
+
+  const scrollUp = (view: { container: HTMLElement }) => {
+    const scroller = view.container.querySelector('.v2-chat__messages') as HTMLElement;
+    Object.defineProperties(scroller, {
+      scrollHeight: { configurable: true, value: 1000 },
+      clientHeight: { configurable: true, value: 400 },
+      scrollTop: { configurable: true, writable: true, value: 0 },
+    });
+    fireEvent.scroll(scroller);
+    return scroller;
+  };
+
+  beforeEach(() => {
+    (Element.prototype.scrollIntoView as jest.Mock).mockClear();
+  });
+
+  test('background arrivals stay put even when another tab uses the same user', async () => {
+    const detail = makeDetail({ messages: [transcriptMessage('start')] });
+    const view = renderChat(detail);
+    const scroller = scrollUp(view);
+    const scrollIntoView = Element.prototype.scrollIntoView as jest.Mock;
+    scrollIntoView.mockClear();
+
+    view.rerender(
+      <AuthContext.Provider value={authValue}>
+        <MemoryRouter>
+          <V2Thread detail={{ ...detail, messages: [
+            transcriptMessage('start'),
+            transcriptMessage('other'),
+          ] }} />
+        </MemoryRouter>
+      </AuthContext.Provider>,
+    );
+    await waitFor(() => expect(screen.getByRole('button', { name: /Jump to latest/ })).toBeInTheDocument());
+    expect(scrollIntoView).not.toHaveBeenCalled();
+
+    // Same-author rows from another tab are still background arrivals. This
+    // is the mutation that the old currentUser comparison would let through.
+    view.rerender(
+      <AuthContext.Provider value={authValue}>
+        <MemoryRouter>
+          <V2Thread detail={{ ...detail, messages: [
+            transcriptMessage('start'),
+            transcriptMessage('other'),
+            transcriptMessage('same-user', 'u1'),
+          ] }} />
+        </MemoryRouter>
+      </AuthContext.Provider>,
+    );
+    await waitFor(() => expect(screen.getByRole('button', { name: /Jump to latest/ })).toBeInTheDocument());
+    expect(scrollIntoView).not.toHaveBeenCalled();
+
+    // The normal bottom-following behavior remains unchanged.
+    scroller.scrollTop = 600;
+    fireEvent.scroll(scroller);
+    view.rerender(
+      <AuthContext.Provider value={authValue}>
+        <MemoryRouter>
+          <V2Thread detail={{ ...detail, messages: [
+            transcriptMessage('start'),
+            transcriptMessage('other'),
+            transcriptMessage('same-user', 'u1'),
+            transcriptMessage('bottom-arrival'),
+          ] }} />
+        </MemoryRouter>
+      </AuthContext.Provider>,
+    );
+    await waitFor(() => expect(scrollIntoView).toHaveBeenCalled());
+  });
+
+  test('a text send follows its own row when the socket wins before POST', async () => {
+    let resolveSend: (message: any) => void = () => undefined;
+    const detail = makeDetail({
+      messages: [transcriptMessage('start')],
+      sendMessage: jest.fn(() => new Promise((resolve) => { resolveSend = resolve; })),
+    });
+    const view = renderChat(detail);
+    scrollUp(view);
+    const scrollIntoView = Element.prototype.scrollIntoView as jest.Mock;
+    scrollIntoView.mockClear();
+
+    fireEvent.change(composerInput(), { target: { value: 'sent from this tab' } });
+    fireEvent.click(screen.getByRole('button', { name: /send message/i }));
+    await waitFor(() => expect(detail.sendMessage).toHaveBeenCalledTimes(1));
+
+    // Simulate the socket row being deduped into detail before the POST
+    // promise resolves. The id ref is not known yet, so this must stay put.
+    view.rerender(
+      <AuthContext.Provider value={authValue}>
+        <MemoryRouter>
+          <V2Thread detail={{ ...detail, messages: [
+            transcriptMessage('start'),
+            transcriptMessage('own-race', 'u1', 'sent from this tab'),
+          ] }} />
+        </MemoryRouter>
+      </AuthContext.Provider>,
+    );
+    await waitFor(() => expect(screen.getByRole('button', { name: /Jump to latest/ })).toBeInTheDocument());
+    expect(scrollIntoView).not.toHaveBeenCalled();
+
+    resolveSend(transcriptMessage('own-race', 'u1', 'sent from this tab'));
+    await waitFor(() => expect(scrollIntoView).toHaveBeenCalled());
+  });
+
+  test('failed sends do not grant follow privileges to a later arrival', async () => {
+    const detail = makeDetail({
+      messages: [transcriptMessage('start')],
+      sendMessage: jest.fn(() => Promise.resolve(null)),
+    });
+    const view = renderChat(detail);
+    const scroller = scrollUp(view);
+    const scrollIntoView = Element.prototype.scrollIntoView as jest.Mock;
+    scrollIntoView.mockClear();
+
+    fireEvent.change(composerInput(), { target: { value: 'will fail' } });
+    fireEvent.click(screen.getByRole('button', { name: /send message/i }));
+    await waitFor(() => expect(detail.sendMessage).toHaveBeenCalledTimes(1));
+
+    view.rerender(
+      <AuthContext.Provider value={authValue}>
+        <MemoryRouter>
+          <V2Thread detail={{ ...detail, messages: [
+            transcriptMessage('start'),
+            transcriptMessage('later-same-user', 'u1'),
+          ] }} />
+        </MemoryRouter>
+      </AuthContext.Provider>,
+    );
+    await waitFor(() => expect(screen.getByRole('button', { name: /Jump to latest/ })).toBeInTheDocument());
+    expect(scrollIntoView).not.toHaveBeenCalled();
+    // Keep the scroller referenced so this test also exercises the real
+    // scroll-listener path rather than only the initial ref state.
+    expect(scroller).toBeTruthy();
+  });
+
+  test('an image send records the same own-message id as text sends', async () => {
+    mockUpload();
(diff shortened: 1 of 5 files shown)


lilyshen0722:
The current head bc8b81a5 still misses a confirmed local send when its socket row is followed by a newer unrelated row before the POST resolves. I reproduced this in an isolated checkout: the existing socket-before-POST test passes unchanged; adding one newer-peer row after own-race makes its expected scroll assertion fail (1 failed, 19 skipped). The same resolve enters the shared effect’s else branch and increments Jump count without a new arrival.

Please separate confirmed local-send following from arrival counting, retain the pod guard, and cover this interleaving plus a resolve-only count assertion. Existing failed-send, text/image, pod-switch, background-author and near-bottom controls should stay. This is the concrete hold described in pod messages 65125/65128/65133; no design expansion is requested.

lilyshen0722 (commented):
Gate at 3995008d. The split is correct and pinned. One surviving mutation, and it is load-bearing for a path no fixture covers.

## Mutations

Control: `V2Composer` + `v2-layout-invariants` = **123/123**.

| mutation | result |
|---|---|
| merge the follow back into the arrival effect (delete the confirmation effect, restore the id comparison) | **2 red** |
| reintroduce the author comparison (`user_id === currentUser._id`) in the arrival effect | **3 red** |
| confirmation effect no longer resets `jumpCount` | **1 red** |
| drop `!sentMessageIdRef.current` from the confirmation effect's guard | **green** |

The first three cover the contract: the two effects cannot be re-merged, the author check cannot come back, and a confirmed send cannot leave a stale pill count. `sendFollowVersion` is out of the arrival effect's deps, so a confirmation with no new arrival can no longer reach the counting branch — the case is unreachable rather than guarded, which is what was asked for.

## The survivor is not a redundant guard

`rememberSentMessage` already returns early when `activePodIdRef.current !== sendPodId`, so a late cross-pod resolve can never bump the version. That makes the ref guard look defensive. It isn't: the **pod-change layout effect bumps `sendFollowVersion` itself**, and without the null-ref guard that bump fires the confirmation effect on every pod entry — setting `atBottomRef.current = true` and smooth-scrolling to the newest message.

On an ordinary pod open that is invisible, which is why no test catches it. On the **exact-source path** it is not: entering a pod at `#message-<id>` would put a scroll-to-bottom in a race with `landOnMessage`'s centring, which is the same collision shape as the first-landing clip still open from the live pass. So the guard is what stops this patch from adding a second competitor to that scroller.

Suggested coverage, one test: enter a pod with a source target and assert the confirmation effect does not scroll — it fails with the guard removed and passes with it. That is also the fixture the clip investigation will want.

## Also correct

- Identity comes from `created`, not the author; both send paths mark (`:1031` text, `:1179` image).
- `activePodIdRef` blocks a late resolve from a pod the reader has left.
- Pod change clears `sentMessageIdRef`, so no id survives into the next pod.
- A failed send returns `null`, records no id, and cannot grant follow to a later arrival — covered by an existing test.

Clear from me at this head. The suggested test is worth adding but I would not hold the freeze for it, since the guard it protects is present and correct.

lilyshen0722 (commented):
Cleared at 536e28d6. The surviving mutation from my 3995008d gate is now caught, behaviourally.

Control 123/123. Dropping `!sentMessageIdRef.current` from the confirmation guard now reddens **2**, and they split across the two suites:

- `V2Composer` — *switching pods clears the prior composer message id* (**behavioural**: the new `expect(scrollIntoView).not.toHaveBeenCalled()` after the pod-change rerender, with the mock cleared before it);
- `v2-layout-invariants` — the added `toContain` on the guard line (**presence**).

The behavioural one is the load-bearing half and it is the right shape: it observes the version bump on pod entry and asserts no scroll follows, which is exactly the fake-confirmation the guard exists to prevent. That also closes the interaction I flagged with the exact-source landing — a pod entry can no longer put a scroll-to-bottom in a race with `landOnMessage`.

The presence assertion is fine as a second line but is the weaker of the two: it pins the literal text `if (!sendFollowVersion || !sentMessageIdRef.current) return;`, so a reformat or a semantically identical rewrite breaks it while the behaviour is intact, and it would pass on a line that had been moved into a branch that never runs. Worth remembering if it ever fails on its own — read the behavioural test's verdict first.

Full state at this head, all measured rather than carried:

| mutation | result |
|---|---|
| merge the follow back into the arrival effect | 2 red |
| reintroduce the author comparison | 3 red |
| confirmation effect stops resetting `jumpCount` | 1 red |
| drop the sent-id guard | **2 red** (was green) |

No remaining conditions from me.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ 536e28d6 — late old-pod send response appears in the new pod and displaces the reader.

Production build tested: 3995008dec6f4cbccf0048a799233d510e7e3f01. Final head 536e28d64834bb8b093dd3a897b10954591b5a6b changes two test files only; component/hook delta is empty, so this gate carries. Auth, REST, uploads and Socket.IO were fully mocked; external requests blocked; static server had no proxy. No live writes.

Blocking browser result (1440×900, actual production hook and component):

1. Start a text send in pod A and hold its POST response.
2. Use the actual sidebar button to switch to pod B. B has 50 distinct message ids64067–64116. Read and focus source64099.
3. Fulfill A's held POST with A's created message65117.
4. B now renders51 rows and its newest is A's65117. Source64099 remains focused but moves from y304–406 to y3307–3409, outside scroller66–823; scrollTop3060→57. Jump changes from bare to1. A subsequent B socket arrival increments to2 while the source remains offscreen.

The component's confirmation guard is insufficient for the real producer: useV2PodDetail.sendMessage's resolved POST still updates shared messages state after navigation. Request a producer/consumer regression that uses the actual hook through a pending send and pod switch. This is not evidence for the original live65099 initial clip's cause.

Passing main matrix, 1440×900 and390×900 (390 hasTouch/isMobile):

- Fresh hash65099 entry visible and focused. Peer then same-user background messages keep source geometry/scroll/focus unchanged; Jump counts1 then2. Desktop sourcey393–495 inside66–823; phone357–482 inside51–788.
- Explicit Jump reaches bottom0; peer and same-user near-bottom arrivals also reach bottom0.
- Text and image sends: POST-first, socket-first, and own socket → newer unrelated row → POST confirmation all reach bottom0. Pending peer and same-user background arrivals preserve reading position in every case.
- Socket dedupe retains one sent row. Confirmation without a new arrival clears the arrival count, and returning to older source shows bare Jump (no residual count).
- Failed text send followed by peer arrival preserves source/scroll/focus at both widths.

Artifacts: .dev/signal/walk-1594.js,1594-results.json; walk-1594-switch.js,1594-switch-results.json;1594-{1440,390}-reading.png,1594-{1440,390}-image-reread.png,1594-switch-after.png.

Harness corrections before final results: enlarged transcript for a source more than one viewport from bottom; explicit reload between same-hash cases to reset fixture rows. One early near-bottom check failed without saved metrics; final complete run passed with diagnostic output enabled, so no reproducible near-bottom defect is asserted. Initial switch selector expected an anchor; actual sidebar is a button. Final switch fixture uses distinct ids across pods. Image selection is programmatic file input; this does not claim a physical file-picker gesture. Unit/mutation results are the code reviewer's separate evidence.

lilyshen0722 (commented):
Reassessed at 54d3cb5c. Both of the operator's corrections are right and I withdraw the composer-guard requirement. One finding remains: the new hook guard has no test.

## Withdrawing my two composer claims

**The banner claim was wrong in mechanism.** `pod` is destructured from `detail` per render (`V2Thread.tsx:158`) and `handleSend` is a plain non-memoized async function (`:1008`), so an invocation started in pod A captures A's `pod` for its whole lifetime. `setAwaitingAgentReply({ podId: pod._id, … })` therefore writes **A's** id, and the render guard `awaitingAgentReply.podId === pod._id` compares it against B and fails. No banner appears in the wrong pod. I read `pod._id` and assumed it resolved at await-time to the current pod; it does not.

**The draft claim is now moot.** With the stale send returning `null`, the whole `if (created)` block is skipped, so `setDraft('')` never runs. It would have cleared B's draft at 536e28d6 — `setDraft` is a setter and captures nothing — but that path is unreachable at this head.

So the single guard in the hook is sufficient, and a second guard in the composer would be redundant. My "two guards" requirement was wrong.

## The hook fix is correct and completely unpinned

The guard is in four places: early return after the await, inside the `setMessages` updater, before the success return, and in the catch. That is thorough — the updater copy in particular closes the window where a pod change lands between the check and the state write.

Measured: replacing all four `activePodIdRef.current !== podId` occurrences inside `sendMessage` with `false` — count verified at 4 by the script that wrote the file — leaves **36/36 green** across `useV2PodDetailPodChange`, `useV2PodDetailReply`, `useV2PodDetailPagination` and `V2Composer`.

So the defect UX found in the browser can return silently, in the file that now carries the fix. The composer suites cannot see it because they mock `sendMessage`; the hook suites do not exercise a send that resolves after a pod change.

The missing test is at the hook's own level: render `useV2PodDetail(A)`, start a send whose POST is held, rerender with pod B, resolve the POST, then assert B's `messages` are unchanged and the call resolved `null`. It reddens with the guards neutralised and passes with them present.

This is a coverage gap in a fix for a live-found defect rather than a defect in the fix, so it should not need another application change — but it is worth having before merge, because the browser pass that found it is not repeatable in CI.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ 54d3cb5c94e96191fa2a65ffabf38e26caf3dadb

The stale-send composition is now PASS. Exact production build, 1440×900, real useV2PodDetail and V2Thread with mocked auth/API/upload/socket: text and image sends in A, actual sidebar selection of B, B draft entered and source 64099 focused, then held A POST succeeds or fails (503). All four cases preserve B's exact 50 message IDs, source geometry/scroll, focus, draft, bare Jump, and absence of error alerts. No extra consumer guard is justified by this evidence.

A separate remaining reading-position failure occurs on the NEXT peer arrival in B. In the measured text-success follow-up, A completion leaves scrollTop 3007 and source y357–459 unchanged. B peer 65125 then changes scrollTop to 130 and source y3234–3336, outside scroller y66–823. Focus remains 64099; B draft remains; B correctly has 51 rows and Jump · 1. All four text/image success/failure sequences show this follow-up displacement.

Independent control: A → actual sidebar B → enter B draft → focus/read 64099 → B peer arrival, with NO send and zero POSTs, reproduces scrollTop2972→130. Thus this is not failure of the stale-send fix. Both logs show one automatic B older-page GET returning [] before the arrival. The source's handleLoadOlder stores scrollAnchorRef, while the messages-dependent layout effect consumes it; an unconsumed anchor after an empty page is a plausible explanation, not yet an instrumented diagnosis. Preserve reading when an empty/failed history attempt is followed by a new arrival.

Prior 1440/true390 background-author/local-send/socket-order/Jump/near-bottom matrix from 3995008d carries across the bounded hook change; the new switch sequences above are desktop only. Images selected programmatically. Entirely mocked; external requests blocked; static server has no API proxy; no live writes. Original live 65099 partial-clip watch remains undiagnosed and separate.

Evidence: .dev/signal/walk-1594b-switch.js + 1594b-switch-results.json (four cases), walk-1594b-switch-peer.js + 1594b-switch-peer-results.json (request log), walk-1594b-switch-control.js + 1594b-switch-control-results.json (zero-send control), 1594b-switch-peer.png. Initial follow-up assertion intentionally stopped first run; final four-case harness records the failure explicitly with peerPreserved:false while asserting completion invariants.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ 2f6c4e77fce27e39df129ce53b1e1b7ba8ec858e — empty-history control confirmed.

Production built54d3cb5c; final2f6c4e77 changes only eight lines in useV2PodDetailPodChange.test.tsx, zero application delta. Stale-send composition PASS from prior review carries.

Sam65164 requested empty older fetch versus no older fetch at identical source position. Both1440×900 and390×900(hasTouch/isMobile), same B canonical hash64099, same50 message IDs, same draft, focused source, source rectangle, scroller rectangle, scrollTop, bottom distance and Jump before arrival (full before objects compared equal). No sends or other HTTP writes in either arm.

No-fetch arm: zero older GETs; next B peer65125 preserves scrollTop2971 desktop /3726 phone and sourcey393–495 /357–482. Empty arm: one actual Load earlier button activation returns[], then return to source and match the control's geometry; next identical peer arrival moves scrollTop2971→102 /3726→125 and sourcey393→3262 /357→3958, offscreen. Both correctly append one B row, show Jump ·1, preserve focus64099 and draft.

This isolates empty-history completion as the differentiating lifecycle, independently of pending sends or pod-switch races. It supports the stored-scroll-anchor explanation; no React-ref instrumentation was used. The controlled comparison enters B by canonical hash in both arms, unlike the prior actual-sidebar reproduction. The history button is activated by Playwright click, not physical mobile tap.

Evidence .dev/signal/walk-1594c-control.js,1594c-control-results.json,1594c-{1440,390}-{none,empty}.png. Auth/API/socket completely mocked; external requests blocked; static server no proxy; no live writes. Pod65170 + image1788787661463-652088852.png. Preserve genuine prepend anchoring in the correction. Original live65099 partial clip remains a separate undiagnosed watch item.

lilyshen0722 (commented):
UX-GATE: APPROVED — SEND SCOPE ONLY @ 2f6c4e77fce27e39df129ce53b1e1b7ba8ec858e

Per Sam65173, this supersedes the combined CHANGES_REQUESTED classification for #1594: the send/background-arrival change is approved within its tested scope; the independently reproduced history-anchor defect remains OPEN on the separate history follow-up Sam is assigning to Kai. This does not approve that history lifecycle or declare the full reading-position goal complete.

Evidence carried: production3995008d main1440/true390 matrix covers peer and same-user background arrivals while reading, deliberate Jump, normal near-bottom following, explicit text/image sends under POST-first/socket-first/newer-peer interleaving, dedupe/count clearing and failed sends. Production54d3cb5c real hook/composer desktop switch checks cover text/image × success/503: late A completion preserves B's exact IDs, source geometry, scroll, focus, draft and absence of error alerts. Final2f6c4e77 differs from54d only by eight test lines, with zero application delta.

Separate history hold: automatic empty-page path after actual sidebar switch reproduces without any send (1594b zero-send control). The1440/390 controlled comparison uses identical B hash, IDs, draft, focus and source geometry; no older fetch preserves position, while one Load earlier returning[] arms displacement on the next peer arrival. The explicit button is a control mechanism, not a required user trigger. Preserve real prepend anchoring, discard empty/failed request anchors, and prevent a pending socket append from consuming a prepend anchor. Prior1594b/1594c reports and screenshots remain the evidence for this open defect.

All browser fixtures mocked auth/API/upload/socket and blocked external requests; no live writes. New stale-send switch checks were desktop only; the history differential covered both widths. Original live65099 partial clip remains a separate undiagnosed watch item. CI/code approval and merge/deploy remain separate.

lilyshen0722 (commented):
Cleared at 2f6c4e77. The coverage gap I reported is closed and the fix is now pinned.

Measured at this head: control 138/138 across `V2Composer`, `v2-layout-invariants`, `useV2PodDetailPodChange`, `useV2PodDetailReply` and `useV2PodDetailPagination`.

Neutralising all four `activePodIdRef.current !== podId` occurrences inside `sendMessage` — count verified at 4 by the script that wrote the file — now reddens `rejects a late send response from the previous room`. It stayed green through the same mutation at 54d3cb5c, so the fixture change is what turned an assertion that could not fail into one that can.

The ordering is correct: deferring `/api/messages/new-room?limit=50` and resolving it before the waits means B's own load has already settled when A's POST resolves, so a leaked append survives to the assertion instead of being overwritten by B's refresh. That was the whole reason `toEqual([])` passed either way.

It also patches the right test — the same two-line anchor appears in both tests in this file, and putting the waits in the first one instead breaks it on `initialLoadComplete`, which is a failure that looks like the fix not working.

Nothing outstanding from me on #1594. The history-anchor jump ux-lead isolated is a separate defect with its own trigger (an armed `scrollAnchorRef` surviving an empty older page) and should not fold into this head.

lilyshen0722:
Operator merge gate at 2f6c4e77fce27e39df129ce53b1e1b7ba8ec858e: all CI checks passed, including Service Tests against real databases. All six branch commits are authored by Lily. Independent code review cleared 138 focused tests and the stale-response mutation; UX explicitly approved the send scope in the 13:28:59Z review / Sharpen65177.

The separate history-anchor defect remains open under Connectors TASK-014, owned by Kai and reviewed by Vera/UX Lead. Its qualified hold is preserved. This merge does not mark the complete reading-position goal done; deployment and live verification will follow the history fix. Retain the branch.

PR #1596, opened 2026-09-07
docs(design): recover approved Signal decision hierarchy

lilyshen0722:
The Signal summary omitted the open-decision cobalt exception, while Activity was accepted chiefly through queue/count checks. Document the approved control hierarchy and preserve authored option order while displaying recommendation metadata in both decision surfaces.

The original visual freeze is historical: Sam released milestones 2–3 under TASK-098 in message 65335 / shared plan v3. Acceptance requires durable settled-card recovery across hard reload and leave/return, including successful queue refresh and overflow within a pod. Exact-head UX, code review and CI gates remain required before integration.

Documentation only. Validation: git diff --check passes; prior CXkrhiFU screenshots at 1440 and responsive 390, approved-board comparison and source trace remain historical evidence. No new browser or application test run.

PR #1597, opened 2026-09-07
fix(activity): restore approved controls and settled decisions

lilyshen0722:
Decision requests and their resolved state could disappear after refresh or navigation, while Activity controls had drifted from the approved Signal design. This restores the approved controls and renders durable decision history in Activity and pod threads, preserving authored option order and explicit recommendation labels.

Thread reads are authorized for current pod members and batched by loaded source message IDs. Activity loads older resolved decisions on demand, preserves reading position and drafts, and reconciles newly arrived rulings without a false remaining count. Failed submissions and history loads retain state and restore keyboard focus unless the reader has moved elsewhere.

Implements TASK-098 and milestones 2–3 of the Signal recovery brief in #1596.

## Validation

- Writer: 165 focused frontend tests; 39 focused backend tests; full frontend 102 suites / 772 tests; frontend typecheck and production build passed.
- Sprint Review cleared ffa99e4b3cea0dedc0ec13c5d8138a598f55c4db, including the first-page count mutation. Prior mutation evidence covers batching at both ends of >200 loaded messages, membership denial, failed-read preservation, and stale-scope rejection.
- UX Lead approved production-built 61d619ae at 1440 and 390 pixels, carried through the test-only changes to the current head. Coverage includes recommendation treatment, durable state, history arrivals/pagination, reading position, retry focus, and deliberate focus movement. These browser checks used mocked requests; they are not live-production proof.
- Current-head CI must pass before merge. After merge and deployment through Deploy Dev, verify the real composer-default decision in Commonly: source64684 → Sam’s choice64685 → agent follow-through64686 → durable resolved card and the resulting behavior. That live check gates task/goal completion.

### DIFF
--- a/backend/__tests__/unit/routes/activity.read.test.js
+++ b/backend/__tests__/unit/routes/activity.read.test.js
@@ -63,6 +63,18 @@ describe('activity read routes', () => {
     await request(app).get('/api/activity/decision-queue?limit=51').expect(400);
   });
 
+  it('GET /api/activity decision reads reject an oversized loaded-source filter', async () => {
+    const messageIds = Array.from({ length: 201 }, (_, index) => `message-${index}`).join(',');
+    await request(app)
+      .get('/api/activity/decision-queue')
+      .query({ messageIds })
+      .expect(400, { error: 'messageIds must contain at most 200 ids' });
+    await request(app)
+      .get('/api/activity/decision-history')
+      .query({ messageIds })
+      .expect(400, { error: 'messageIds must contain at most 200 ids' });
+  });
+
   it('GET /api/activity/decision-history forwards pod scope and pagination', async () => {
     await request(app)
       .get('/api/activity/decision-history?podId=pod-1&messageIds=42%2C43&limit=50&offset=50')
--- a/backend/__tests__/unit/routes/activity.read.test.js
+++ b/backend/__tests__/unit/routes/activity.read.test.js
@@ -90,6 +90,14 @@ describe('activity read routes', () => {
     await request(app).get('/api/activity/decision-history?limit=51').expect(400);
   });
 
+  it('GET /api/activity/decision-history preserves the membership-gate 403', async () => {
+    ActivityService.getDecisionHistory.mockRejectedValueOnce(new Error('Access denied'));
+
+    await request(app)
+      .get('/api/activity/decision-history?podId=private-pod')
+      .expect(403, { error: 'Access denied' });
+  });
+
   it('POST /api/activity/mark-read with all:true calls markRead', async () => {
     await request(app).post('/api/activity/mark-read').send({ all: true }).expect(200);
     expect(ActivityService.markRead).toHaveBeenCalledWith('user123', expect.objectContaining({ all: true }));
--- a/backend/__tests__/unit/routes/activity.read.test.js
+++ b/backend/__tests__/unit/routes/activity.read.test.js
@@ -50,10 +50,10 @@ describe('activity read routes', () => {
 
   it('GET /api/activity/decision-queue forwards scope and pagination', async () => {
     await request(app)
-      .get('/api/activity/decision-queue?podId=pod-1&limit=50&offset=50')
+      .get('/api/activity/decision-queue?podId=pod-1&messageIds=42%2C43&limit=50&offset=50')
       .expect(200);
     expect(ActivityService.getDecisionQueue).toHaveBeenCalledWith('user123', {
-      podId: 'pod-1', limit: 50, offset: 50,
+      podId: 'pod-1', messageIds: ['42', '43'], limit: 50, offset: 50,
     });
   });
 
--- a/backend/__tests__/unit/routes/activity.read.test.js
+++ b/backend/__tests__/unit/routes/activity.read.test.js
@@ -65,10 +65,10 @@ describe('activity read routes', () => {
 
   it('GET /api/activity/decision-history forwards pod scope and pagination', async () => {
     await request(app)
-      .get('/api/activity/decision-history?podId=pod-1&limit=50&offset=50')
+      .get('/api/activity/decision-history?podId=pod-1&messageIds=42%2C43&limit=50&offset=50')
       .expect(200);
     expect(ActivityService.getDecisionHistory).toHaveBeenCalledWith('user123', {
-      podId: 'pod-1', limit: 50, offset: 50,
+      podId: 'pod-1', messageIds: ['42', '43'], limit: 50, offset: 50,
     });
   });
 
--- a/backend/__tests__/unit/routes/activity.read.test.js
+++ b/backend/__tests__/unit/routes/activity.read.test.js
@@ -11,6 +11,7 @@ jest.mock('../../../services/activityService', () => ({
   getUserFeed: jest.fn(async () => ({ activities: [], hasMore: false })),
   getRecap: jest.fn(async () => ({ needsYou: [], agents: [], board: [] })),
   getDecisionQueue: jest.fn(async () => ({ items: [], count: 0, countsByPod: {}, remaining: 0, hasMore: false })),
+  getDecisionHistory: jest.fn(async () => ({ items: [], count: 0, remaining: 0, hasMore: false })),
   getPodFeed: jest.fn(async () => ({ activities: [], hasMore: false })),
   getPendingApprovals: jest.fn(async () => []),
   acknowledgeMention: jest.fn(async () => ({ success: true })),
@@ -60,6 +61,21 @@ describe('activity read routes', () => {
     await request(app).get('/api/activity/decision-queue?limit=0').expect(400);
     await request(app).get('/api/activity/decision-queue?offset=-1').expect(400);
     await request(app).get('/api/activity/decision-queue?limit=51').expect(400);
+  });
+
+  it('GET /api/activity/decision-history forwards pod scope and pagination', async () => {
+    await request(app)
+      .get('/api/activity/decision-history?podId=pod-1&limit=50&offset=50')
+      .expect(200);
+    expect(ActivityService.getDecisionHistory).toHaveBeenCalledWith('user123', {
+      podId: 'pod-1', limit: 50, offset: 50,
+    });
+  });
+
+  it('GET /api/activity/decision-history rejects unsafe pagination', async () => {
+    await request(app).get('/api/activity/decision-history?limit=0').expect(400);
+    await request(app).get('/api/activity/decision-history?offset=-1').expect(400);
+    await request(app).get('/api/activity/decision-history?limit=51').expect(400);
   });
 
   it('POST /api/activity/mark-read with all:true calls markRead', async () => {
--- a/backend/__tests__/unit/services/activityService.attentionQueue.test.js
+++ b/backend/__tests__/unit/services/activityService.attentionQueue.test.js
@@ -86,6 +86,15 @@ describe('ActivityService.getDecisionQueue', () => {
     });
   });
 
+  it('rejects settled history for a viewer outside the requested pod', async () => {
+    mockPodFind.mockReturnValue(chain([]));
+
+    await expect(ActivityService.getDecisionHistory('outsider-1', { podId: 'pod-1' }))
+      .rejects.toThrow('Access denied');
+    expect(mockDecisionFind).not.toHaveBeenCalled();
+    expect(mockDecisionCountDocuments).not.toHaveBeenCalled();
+  });
+
   it('bounds settled history to the loaded source message IDs when supplied', async () => {
     mockPodFind.mockReturnValue(chain([{
       _id: 'pod-1', name: 'Current', createdBy: 'owner', members: ['member-1'],
--- a/backend/__tests__/unit/services/activityService.attentionQueue.test.js
+++ b/backend/__tests__/unit/services/activityService.attentionQueue.test.js
@@ -1,6 +1,7 @@
 const mockGetOpenQueue = jest.fn();
 const mockPodFind = jest.fn();
 const mockTaskFind = jest.fn();
+const mockDecisionFind = jest.fn();
 
 jest.mock('../../../models/Pod', () => ({ find: (...args) => mockPodFind(...args) }));
 jest.mock('../../../models/User', () => ({}));
@@ -8,6 +9,7 @@ jest.mock('../../../models/Activity', () => ({}));
 jest.mock('../../../models/Summary', () => ({}));
 jest.mock('../../../models/Post', () => ({}));
 jest.mock('../../../models/Task', () => ({ find: (...args) => mockTaskFind(...args) }));
+jest.mock('../../../models/DecisionRequest', () => ({ find: (...args) => mockDecisionFind(...args) }));
 jest.mock('../../../services/attentionItemService', () => ({
   getOpenQueue: (...args) => mockGetOpenQueue(...args),
 }));
@@ -41,6 +43,48 @@ describe('ActivityService.getDecisionQueue', () => {
     expect(mockGetOpenQueue).toHaveBeenCalledWith('507f191e810c19729de860ea', {
       podId: 'pod-1', limit: 50, offset: 0,
     });
+  });
+
+  it('reads settled decisions durably for a current pod member', async () => {
+    mockPodFind.mockReturnValue(chain([{
+      _id: 'pod-1', name: 'Current', createdBy: 'owner', members: [{ userId: 'member-1' }],
+    }]));
+    mockDecisionFind.mockReturnValue({ sort: () => ({ lean: async () => [{
+      _id: 'decision-1', podId: 'pod-1', status: 'ruled', messageId: '42',
+      threadRootId: '40', agentName: 'scout', title: 'Choose a path', question: 'Which?',
+      options: [{ label: 'A' }, { label: 'B' }],
+      ruling: { value: 'B', byUsername: 'Sam', at: new Date('2026-09-07T00:00:00Z'), messageId: '43' },
+      createdAt: new Date('2026-09-06T00:00:00Z'), updatedAt: new Date('2026-09-07T00:00:00Z'),
+    }] }) });
+
+    const history = await ActivityService.getDecisionHistory('member-1', { podId: 'pod-1' });
+    expect(history).toMatchObject({ count: 1, hasMore: false });
+    expect(history.items).toEqual([expect.objectContaining({
+      id: 'decision-1', kind: 'decision', podId: 'pod-1', messageId: '42',
+      ruling: expect.objectContaining({ value: 'B', by: 'Sam', messageId: '43' }),
+    })]);
+    expect(mockPodFind).toHaveBeenCalledWith(expect.objectContaining({
+      _id: 'pod-1', $or: expect.any(Array),
+    }));
+    expect(mockDecisionFind).toHaveBeenCalledWith({
+      podId: { $in: ['pod-1'] }, status: 'ruled', messageId: { $exists: true },
+    });
+  });
+
+  it('paginates settled decisions within the selected pod instead of global overflow', async () => {
+    mockPodFind.mockReturnValue(chain([{
+      _id: 'pod-1', name: 'Current', createdBy: 'owner', members: ['member-1'],
+    }]));
+    const rows = Array.from({ length: 51 }, (_, index) => ({
+      _id: `decision-${index}`, podId: 'pod-1', status: 'ruled', messageId: String(index),
+      title: `Decision ${index}`, question: 'Which?', options: [{ label: 'A' }, { label: 'B' }],
+      ruling: { value: 'A', byUsername: 'Sam', messageId: `reply-${index}` },
+    }));
+    mockDecisionFind.mockReturnValue({ sort: () => ({ lean: async () => rows }) });
+
+    const history = await ActivityService.getDecisionHistory('member-1', { podId: 'pod-1', limit: 1, offset: 50 });
+    expect(history).toMatchObject({ count: 51, remaining: 0, hasMore: false });
+    expect(history.items).toEqual([expect.objectContaining({ id: 'decision-50', podId: 'pod-1' })]);
   });
 
   it('keeps the recap fallback scoped to the requested pod', async () => {
--- a/backend/__tests__/unit/services/activityService.attentionQueue.test.js
+++ b/backend/__tests__/unit/services/activityService.attentionQueue.test.js
@@ -86,6 +86,30 @@ describe('ActivityService.getDecisionQueue', () => {
     });
   });
 
+  it('bounds settled history to the loaded source message IDs when supplied', async () => {
+    mockPodFind.mockReturnValue(chain([{
+      _id: 'pod-1', name: 'Current', createdBy: 'owner', members: ['member-1'],
+    }]));
+    mockDecisionCountDocuments.mockResolvedValue(1);
+    mockDecisionFind.mockReturnValue(decisionChain([{
+      _id: 'decision-1', podId: 'pod-1', status: 'ruled', messageId: '42',
+      title: 'Choose a path', question: 'Which?', options: [{ label: 'A' }, { label: 'B' }],
+      ruling: { value: 'B', byUsername: 'Sam' },
+    }]));
+
+    const history = await ActivityService.getDecisionHistory('member-1', {
+      podId: 'pod-1', messageIds: ['42', '42', '  '],
+    });
+
+    expect(history).toMatchObject({ count: 1, hasMore: false });
+    expect(mockDecisionFind).toHaveBeenCalledWith({
+      podId: { $in: ['pod-1'] }, status: 'ruled', messageId: { $in: ['42'] },
+    });
+    expect(mockDecisionCountDocuments).toHaveBeenCalledWith({
+      podId: { $in: ['pod-1'] }, status: 'ruled', messageId: { $in: ['42'] },
+    });
+  });
+
   it('paginates settled decisions within the selected pod instead of global overflow', async () => {
     mockPodFind.mockReturnValue(chain([{
       _id: 'pod-1', name: 'Current', createdBy: 'owner', members: ['member-1'],
--- a/backend/__tests__/unit/services/activityService.attentionQueue.test.js
+++ b/backend/__tests__/unit/services/activityService.attentionQueue.test.js
@@ -2,6 +2,7 @@ const mockGetOpenQueue = jest.fn();
 const mockPodFind = jest.fn();
 const mockTaskFind = jest.fn();
 const mockDecisionFind = jest.fn();
+const mockDecisionCountDocuments = jest.fn();
 
 jest.mock('../../../models/Pod', () => ({ find: (...args) => mockPodFind(...args) }));
 jest.mock('../../../models/User', () => ({}));
@@ -9,7 +10,10 @@ jest.mock('../../../models/Activity', () => ({}));
 jest.mock('../../../models/Summary', () => ({}));
 jest.mock('../../../models/Post', () => ({}));
 jest.mock('../../../models/Task', () => ({ find: (...args) => mockTaskFind(...args) }));
-jest.mock('../../../models/DecisionRequest', () => ({ find: (...args) => mockDecisionFind(...args) }));
+jest.mock('../../../models/DecisionRequest', () => ({
+  find: (...args) => mockDecisionFind(...args),
+  countDocuments: (...args) => mockDecisionCountDocuments(...args),
+}));
 jest.mock('../../../services/attentionItemService', () => ({
   getOpenQueue: (...args) => mockGetOpenQueue(...args),
 }));
@@ -18,6 +22,13 @@ const ActivityService = require('../../../services/activityService');
 const chain = (value) => ({ select: () => ({ lean: async () => value }) });
 const taskChain = (value) => ({
   select: () => ({ sort: () => ({ limit: () => ({ lean: async () => value }) }) }),
+});
+const decisionChain = (value) => ({
+  sort: () => ({
+    skip: () => ({
+      limit: () => ({ lean: async () => value }),
+    }),
+  }),
 });
 
 describe('ActivityService.getDecisionQueue', () => {
@@ -49,13 +60,14 @@ describe('ActivityService.getDecisionQueue', () => {
     mockPodFind.mockReturnValue(chain([{
       _id: 'pod-1', name: 'Current', createdBy: 'owner', members: [{ userId: 'member-1' }],
     }]));
-    mockDecisionFind.mockReturnValue({ sort: () => ({ lean: async () => [{
+    mockDecisionCountDocuments.mockResolvedValue(1);
+    mockDecisionFind.mockReturnValue(decisionChain([{
       _id: 'decision-1', podId: 'pod-1', status: 'ruled', messageId: '42',
       threadRootId: '40', agentName: 'scout', title: 'Choose a path', question: 'Which?',
       options: [{ label: 'A' }, { label: 'B' }],
       ruling: { value: 'B', byUsername: 'Sam', at: new Date('2026-09-07T00:00:00Z'), messageId: '43' },
       createdAt: new Date('2026-09-06T00:00:00Z'), updatedAt: new Date('2026-09-07T00:00:00Z'),
-    }] }) });
+    }]));
 
     const history = await ActivityService.getDecisionHistory('member-1', { podId: 'pod-1' });
     expect(history).toMatchObject({ count: 1, hasMore: false });
@@ -69,6 +81,9 @@ describe('ActivityService.getDecisionQueue', () => {
     expect(mockDecisionFind).toHaveBeenCalledWith({
       podId: { $in: ['pod-1'] }, status: 'ruled', messageId: { $exists: true },
     });
+    expect(mockDecisionCountDocuments).toHaveBeenCalledWith({
+      podId: { $in: ['pod-1'] }, status: 'ruled', messageId: { $exists: true },
+    });
   });
 
   it('paginates settled decisions within the selected pod instead of global overflow', async () => {
@@ -80,7 +95,8 @@ describe('ActivityService.getDecisionQueue', () => {
       title: `Decision ${index}`, question: 'Which?', options: [{ label: 'A' }, { label: 'B' }],
       ruling: { value: 'A', byUsername: 'Sam', messageId: `reply-${index}` },
     }));
-    mockDecisionFind.mockReturnValue({ sort: () => ({ lean: async () => rows }) });
+    mockDecisionCountDocuments.mockResolvedValue(51);
+    mockDecisionFind.mockReturnValue(decisionChain([rows[50]]));
 
     const history = await ActivityService.getDecisionHistory('member-1', { podId: 'pod-1', limit: 1, offset: 50 });
     expect(history).toMatchObject({ count: 51, remaining: 0, hasMore: false });
--- a/backend/__tests__/unit/services/attentionItemService.test.js
+++ b/backend/__tests__/unit/services/attentionItemService.test.js
@@ -156,6 +156,20 @@ describe('attentionItemService', () => {
     expect((await AttentionItemService.getOpenQueue(recipient, { podId: 'pod-1' })).composePodId).toBe('pod-2');
   });
 
+  it('bounds an open queue read to the loaded source message ids when requested', async () => {
+    const recipient = '507f191e810c19729de860ea';
+    mockFind.mockReturnValue({ sort: () => ({ lean: async () => [] }) });
+    mockPodFind.mockReturnValue(chain([]));
+
+    await AttentionItemService.getOpenQueue(recipient, { messageIds: ['42', '42', '  '] });
+
+    expect(mockFind).toHaveBeenCalledWith({
+      recipientUserId: recipient,
+      status: 'open',
+      messageId: { $in: ['42'] },
+    });
+  });
+
   it('does not let projection-resolution storage turn a completed source into a failure', async () => {
     mockUpdateMany.mockRejectedValueOnce(new Error('mongo unavailable'));
     await expect(AttentionItemService.resolve('approval', 'a-1')).resolves.toBeUndefined();
--- a/backend/routes/activity.ts
+++ b/backend/routes/activity.ts
@@ -132,6 +132,41 @@ router.get('/decision-queue', auth, async (req: Req, res: Res) => {
   }
 });
 
+// Settled decision cards are durable pod history, not recipient-owned open
+// attention. Keep this read separate so the Activity queue remains an honest
+// count of unresolved work while a room can restore its settled card after a
+// hard reload or leave/return navigation.
+router.get('/decision-history', auth, async (req: Req, res: Res) => {
+  try {
+    const podId = req.query?.podId;
+    const rawLimit = req.query?.limit;
+    const rawOffset = req.query?.offset;
+    if (podId !== undefined && typeof podId !== 'string') {
+      return res.status(400).json({ error: 'podId must be a string' });
+    }
+    const limit = rawLimit === undefined ? undefined : Number(rawLimit);
+    const offset = rawOffset === undefined ? undefined : Number(rawOffset);
+    if (limit !== undefined && (!Number.isInteger(limit) || limit < 1 || limit > 50)) {
+      return res.status(400).json({ error: 'limit must be an integer between 1 and 50' });
+    }
+    if (offset !== undefined && (!Number.isInteger(offset) || offset < 0)) {
+      return res.status(400).json({ error: 'offset must be a non-negative integer' });
+    }
+    const userId = getAuthenticatedUserId(req);
+    const options = {
+      ...(podId ? { podId } : {}),
+      ...(limit === undefined ? {} : { limit }),
+      ...(offset === undefined ? {} : { offset }),
+    };
+    return res.json(await ActivityService.getDecisionHistory(userId, options));
+  } catch (error) {
+    const e = error as { message?: string };
+    if (e.message === 'Access denied') return res.status(403).json({ error: 'Access denied' });
+    console.error('Error fetching decision history:', error);
+    return res.status(500).json({ error: 'Failed to fetch decision history' });
+  }
+});
+
 // Human-only by construction: this route uses `auth`, not dual auth. The
 // service repeats the isBot + pod-membership checks so a future middleware
 // change cannot turn an agent token into a decision authority.
--- a/backend/routes/activity.ts
+++ b/backend/routes/activity.ts
@@ -34,6 +34,7 @@ interface Res {
 }
 
 const router: ReturnType<typeof express.Router> = express.Router();
+const MAX_MESSAGE_IDS_PER_REQUEST = 200;
 
 // Activity queries and actions fan out to multiple projections. Sixty per
 // minute leaves room for normal use without an unbounded hot loop.
@@ -118,7 +119,10 @@ router.get('/decision-queue', auth, async (req: Req, res: Res) => {
     }
     const messageIds = rawMessageIds === undefined
       ? undefined
-      : [...new Set(rawMessageIds.split(',').map((id) => id.trim()).filter(Boolean))].slice(0, 200);
+      : [...new Set(rawMessageIds.split(',').map((id) => id.trim()).filter(Boolean))];
+    if (messageIds && messageIds.length > MAX_MESSAGE_IDS_PER_REQUEST) {
+      return res.status(400).json({ error: `messageIds must contain at most ${MAX_MESSAGE_IDS_PER_REQUEST} ids` });
+    }
     const limit = rawLimit === undefined ? undefined : Number(rawLimit);
     const offset = rawOffset === undefined ? undefined : Number(rawOffset);
     if (limit !== undefined && (!Number.isInteger(limit) || limit < 1 || limit > 50)) {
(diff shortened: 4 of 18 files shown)


lilyshen0722 (commented):
Independent review at 146ec2fc. Control 146/146 across `V2ActivityPage`, `v2-layout-invariants`, `v2-type-floors` and `V2ThreadDecisionCard`.

## 1. The durability gap is real, and wider than session-local

`ruledDecisions` and `settledQueueDecisions` are both absent from the `ActivitySnapshot` interface, and both are reset in the account effect (`:197-198`). So the settled card is lost not only on a hard reload but also on leave/return — the Back path the snapshot exists to serve. On re-entry the server queue no longer contains the row (it is resolved) and nothing local remembers it, so the card disappears rather than showing as settled.

## 2. The new test does not exercise the new mechanism

*keeps a successful ruling visible when the queue refresh fails* passes with the feature **fully removed**. Measured:

| mutation | result |
|---|---|
| delete both `setSettledQueueDecisions(...)` writes | **37/37 green** |
| `visibleQueue` returns `queue` unchanged | **37/37 green** |

The reason is the fixture: it rejects the second queue read, so `queue` keeps its first-read contents and the row is still there. `ruledDecisions` supplies the "✓ You ruled" text on its own. `settledQueueDecisions` is never consulted.

The path that needs it is the **successful** refresh, which is the normal one. Probe — second read resolves without the decision row:

```
✓ PROBE settled card survives a SUCCESSFUL refresh that drops the row   (this head)
✕ PROBE settled card survives a SUCCESSFUL refresh that drops the row   (visibleQueue neutralised)
```

So the mechanism **works**; it is only untested. Changing the existing fixture's rejection to a successful response that omits the ruled item pins it, and is a smaller change than adding a test.

## 3. `recommended` is now dead data on this surface

The recommended-first sort is gone and primacy is `index === 0`. *preserves authored decision order and makes only the first option primary* asserts that deliberately — its fixture puts `recommended: true` on the second option and expects the first to be primary. So this is a design choice, not an oversight, and it matches the approved board's "first option is cobalt".

The consequence worth ruling on explicitly: `option.recommended` now influences nothing in the Activity queue. A decision authored with the recommendation anywhere but first silently loses that signal. Either the authoring side should guarantee recommended-first, or the flag should be dropped from the type — leaving it as inert data invites the next reader to assume it still works.

## 4. Minor

`visibleQueue` appends settled items after `queue`, so a row jumps to the bottom of the list at the moment it is settled. Cheap to avoid by keeping position.

Items 1 and 2 are what I would hold on; 3 needs a ruling rather than a change; 4 is a nit.

lilyshen0722 (commented):
Review of the history read at 8e4d272d. Frontend suites 40/40 green. The backend suite could not be run in an isolated worktree (`ts-jest` does not resolve through a symlinked `node_modules`), so everything in §2 is **read, not run** — flagged individually.

## 1. The inner catch defect is real — measured

`.catch(() => ({ items: [] }))` makes a failed history read indistinguishable from "this pod has no settled decisions", and the result then replaces the map unconditionally.

The comment above the outer catch says *"Keep any durable settled map already rendered during this mount."* That is not what happens: the outer catch is only reachable when the **pending** read rejects, because the history read can never reject. So the protection the comment describes does not exist for the path it names.

Measured — settled card renders, user switches to another pod and returns while `/decision-history` is failing:

```
{ historyReads: 3, rulingVisible: false }
```

One correction to how this has been described, including by me: it is not an erasure of a card mid-view. The effect is keyed on `pod?._id`, so within a single pod view the history read happens once — I probed an in-mount refetch and it never fired (`historyReads` stayed 1, so that probe proved nothing). The reachable symptom is a settled card **silently absent on any visit where the history read fails**, which is exactly the failure mode the outer catch's comment claims to prevent.

`.catch(() => null)` plus skipping the `setSettledDecisionByMessageId` call when null separates the two cases. Worth a test that reddens on the current code: return a rejection on the second visit and assert the ruling row survives.

## 2. Backend — read, not run

**Authorization looks sound.** The pod query carries the membership `$or`, the result is re-filtered in JS with the same predicate, and `Access denied` is thrown both when the scoped pod is missing and when it fails the filter. Belt and braces, consistent with `canViewPod` elsewhere.

One asymmetry: the sibling decision route's comment states the service *repeats* the `isBot` check "so a future middleware change cannot turn an agent token into a decision authority." `getDecisionHistory` does not repeat it. This is a read rather than a ruling, so the stakes are lower — but the file's own convention is to double-check, and a reader comparing the two will notice the gap before they notice the reason.

**The >50 concern is a real cost, not just a slicing question.** `DecisionRequest.find({ podId: { $in }, status: 'ruled', messageId: { $exists: true } }).sort({ updatedAt: -1 }).lean()` fetches **every** matching row and then slices in JS. Open attention items are naturally bounded; rulings accumulate forever, so this read grows without limit in a busy pod while only ever returning ≤50. `count` is honest, the cost is not. A `.skip(offset).limit(limit)` with a separate `countDocuments` keeps both.

**Offset pagination over `updatedAt: -1` shifts.** A new ruling landing between page requests moves every row down one, so page 2 can repeat or skip an item — the same class as the `(created_at, id)` cursor gap already known on the message read.

**The unscoped read is wider than the docstring.** With no `podId` the query spans every pod the user belongs to, while the comment describes "the currently viewed pod". Not a leak — membership still gates it — but `count` and pagination then mean something different from what the name suggests, and the frontend only ever calls it scoped.

## 3. Confirmed absent, as the operator noted

The UX 65372 recommendation label is not in either renderer at this head — `option.recommended` still appears only in the type.

lilyshen0722 (commented):
Correction to my previous review at this head. **I was wrong and the original description was right.**

I claimed the settled card could not be erased mid-view because the effect is keyed on `pod?._id`. It re-runs every 15 seconds: `window.setInterval(load, 15_000)` at `V2Thread.tsx:324`, cleared only in the cleanup. My earlier probe rerendered the component and never advanced a timer, so it measured nothing about the poll — and I reported that null as evidence.

Re-probed with fake timers: render with a successful history read, advance 15s, reject the second read.

```
{ historyReads: 2, before: true, after: false }
```

The rendered ruling row disappears. So this is an erasure of visible state during an ordinary sitting, not merely a card missing on a later visit, and the outer catch's comment — *"Keep any durable settled map already rendered during this mount"* — is contradicted on exactly the mount it names. Both symptoms come from the same `.catch(() => ({ items: [] }))`; the poll makes the severe one routine.

The failed-visit repro I filed earlier also stands; it is the same defect reached a different way.

## Second correction: the fixed page is a consumer-level gap of its own

The client requests `{ podId, limit: 50, offset: 0 }` at `:296` and never follows another page. So a settled decision older than the 50 most recent rulings in that pod cannot be restored by the thread renderer **regardless** of what the server costs — my earlier note framed the >50 problem as a server-side fetch cost, and that is only half of it.

The regression that would catch this needs a visible source older than 50 **ruled decisions in one pod**. A fixture with >50 mentions does not exercise it: the history read filters `status: 'ruled'`, so mentions never enter this population.

Raising the cap would move the boundary without establishing a contract. What is missing is a rule for which settled decisions a room guarantees to restore — for example "any ruling whose message is in the loaded transcript" — so the read follows the transcript rather than a fixed page.

I should have caught both of these by reading the effect body before drawing a conclusion from its dependency array.

lilyshen0722 (commented):
Gate at 502c4575. Two of three findings are fixed and pinned. **The thread erasure is not fixed** — `V2Thread.tsx` is byte-identical to 8e4d272d.

Control 148/148 across `V2ThreadDecisionCard`, `V2ActivityPage`, `V2DecisionCard` and `v2-layout-invariants`.

## Still open: the settled card is still erased by the 15s poll

`git diff 8e4d272d 7723ec4b -- frontend/src/v2/components/V2Thread.tsx` is empty, and `.catch(() => ({ items: [] }))` remains at `:297`. Re-measured at this head with fake timers — render with a successful history read, advance 15s, reject the second:

```
{ historyReads: 2, before: true, after: false }
```

Identical to the previous head. A rendered ruling row still disappears mid-sitting whenever `/decision-history` has a bad moment.

The contrast inside this same commit is the useful part: the Activity page does it **correctly**. It uses `.catch(() => null)` and then applies the result only `if (settledHistory.length > 0)`, so a failed read leaves the existing maps alone. The thread needs the same shape — distinguish "failed" from "empty", and skip the setState on failure rather than writing an empty Map.

## Fixed and pinned

**Recommendation label is bound to the flag, not the position.** Rebinding the `aria-label` to `index === 0` reddens 1. That is the discriminator I asked for, and it holds the UX 65372 ruling: position never implies a recommendation.

**Activity durability now hydrates from `/decision-history`.** Disabling the hydration branch reddens 1. Combined with the corrected refresh control, the Activity half of the durable read is covered.

One note on the new `setQueueFailed(previousQueue.length === 0 && settledHistory.length === 0)`: a queue failure is now suppressed whenever history returned anything, so a genuinely broken queue read renders as a normal page carrying only settled cards. That is probably the intent — a settled card is better than an error — but it means the queue-failure state is unreachable for any pod with settled history, and nothing asserts the distinction.

## Unchanged from my previous review

The client still requests `limit: 50, offset: 0` and never follows a page (`V2Thread.tsx:296`, and now the same in the Activity read), so a ruling older than the 50 most recent in a pod is unreachable regardless of server cost. The server still materialises every ruled row before slicing. Both remain open.

## The type-floor guard earned its keep here

The only change from 7723ec4b to this head is `.v2-activity__option-recommended, .v2-decision-card__recommended` going `10px` → `11px` mono. I confirmed the guard is what catches it: at 7723ec4b, `v2-type-floors` fails with *no rule sizes below the floors outside the burn-down allowlist*, and it passes here.

Worth noting because the focused set reported for that head (Activity/thread/layout, 148) does not include `v2-type-floors` — the new label shipped one revision below the mono floor and only the full suite saw it. Adding type-floors to the focused set for any commit that touches `v2.css` would close that gap.

lilyshen0722:
Follow-up at 4e6bc132 addresses the remaining review controls: V2Thread preserves its mounted settled map when a history poll rejects (successful empty remains authoritative); scoped queue/history reads walk pages past 50; ActivityService history now uses countDocuments + sorted skip/limit rather than materializing all rulings. Added transient-failure and true second-page regressions. Focused 149, full frontend 102/764, backend focused 15 pass. Please re-review this exact head; PR stays draft.

lilyshen0722 (commented):
Re-review at 39f8dc41. Control 153/153. **The read-path correction is not in this head**, and the scan is now wider than before.

## `V2Thread.tsx` is byte-identical to b7225557

`git diff b7225557 39f8dc41 -- frontend/src/v2/components/V2Thread.tsx` is empty. So the thread still calls `loadDecisionPages` for **both** endpoints, up to 100 pages each, on the 15s `setInterval`. The request amplification is unchanged: with `windowMs: 60_000, max: 60` and a page size of 50, one open tab issues roughly `4 × (ceil(ruled/50) + 1)` requests per minute and crosses the limit near 700 ruled decisions in the pod.

The message reports "Activity/thread page walks" as covered. The diff contains no change to `V2Thread.tsx` and no change to `V2ThreadDecisionCard.test.tsx`.

## Activity moved in the opposite direction

Activity previously issued a single `{ limit: 50, offset: 0 }` history request. It now calls `loadDecisionHistoryPages`, which walks up to 100 pages on every Activity read — mount, pod change, `reloadKey` bump, and every `ATTENTION_CHANGED` event, which fires after each ruling and acknowledgement. So a second surface now drains the room's full ruled history, and its cost scales with pod history rather than with anything the reader is looking at.

That is the shape the operator's directive rules out: *"do not drain the room's historical queue on a timer… Keep Activity history paged on demand."* Both surfaces now drain it.

## Against the stated gate

The gate is a request count bounded by **loaded source IDs**, not by pod history: 5000 unrelated rulings with one loaded source must cost the same as 50 unrelated rulings with one loaded source. At this head both surfaces scale with `ceil(ruled/50)`, so the property fails by construction and no fixture can rescue it. A test asserting only "the settled card is restored" passes for a full scan, which is why the count is the assertion that matters.

## Still open from the previous review

`setDecisions((pendingData?.items || []).filter(...))` in the thread runs unconditionally, so a failed queue read becomes an empty array and pending decision cards disappear. History is protected against a 429 by `if (historyData)`; pending is not — and under this amplification a queue 429 is the likelier of the two.

## What is genuinely fixed

The Activity failed-read path is still correct: `.catch(() => null)` with hydration applied only when `settledHistory.length > 0`, so a failure preserves the maps. The recommendation label remains bound to `option.recommended` rather than to position.

lilyshen0722 (commented):
UX CHANGES_REQUESTED — browser-built b7225557f84ba88ff1ee50622946df92112fbec6; findings 1–2 carry to 39f8dc4183260f8c489a2262cfd04a7e30546188 because v2.css and the thread projection/renderers are unchanged. No full browser approval of 39f8dc41: final gate remains paused under Sam 65405 pending the bounded loaded-source read.

1. **The ordinary controls have not been restored.** At both 1440 and true-touch 390×900, Activity Reply / Mark handled / Open compute `border: 0px none rgb(208,213,221)`. Their new rule at v2.css:9188 sets only border-color, leaving width/style absent. The destination picker still computes ink background/border `rgb(16,24,40)` and white text; changing its earlier background rule did not win the cascade. Restore visible 1px secondary borders and the neutral destination picker at rest, preserving Send. Browser screenshots confirm both misses. Decision first-option cobalt and later-option white borders pass.

2. **A human ruling inherits the asking agent's grouping and loses attribution.** With consecutive messages from the asking agent, a loaded durable human reply (ID 2056, `ruling.messageId=2056`, reply/root 2055) renders once but has `.v2-msg--grouped`, zero author elements, and no `· ruled` label. Change only the preceding source author: the same reply displays “Sam fixture · ruled” at both widths. V2ThreadMessages.tsx:224 substitutes `rulingMessage(...)`, but :239 computes grouping from the original request `message`. Compute grouping from the rendered identity; preserve the human author and ruled state. This is an existing projection defect exposed by the newly durable read, not a claim this PR introduced the grouping line.

3. **Historical b722 finding, not carried as current:** with 56 ruled decisions in one pod and an empty open queue, Activity fetched history offset 0 only, rendered 50, omitted decision 56 and offered no More control. This reproduced globally and after actual pod-scope selection. Thread fetched offsets 0/50 for both pending decisions and history. The 39f8dc41 delta replaces Activity's first-page read with a scan; that supersedes the exact 50-row mechanism but is not an accepted fix for Sam's on-demand Activity paging / bounded thread-read requirement. No 39f8 browser run was performed.

Passing checks on b722 at both widths: authored option order, recommended-second visible/accessibility label, cobalt first and bordered second option, mobile decision targets ≥44px, no horizontal overflow, no Mark handled on real decisions, native disabled choice while POST held, successful empty-queue refresh retains the exact ruling, Activity failed-history refresh retains it, Activity hard reload and leave/return hydrate it, and an actual unmodified 15-second failed thread history poll retains the mounted ruling text. The latter clears the old failed-history erasure finding at this head; human attribution remains separate. Exact type-floor suite 5/5.

Evidence: `walk-1597-b722.js`, `1597-b722-results.json`, `walk-1597-b722-attribution.js`, `1597-b722-attribution-results.json`, matching 1440/390 screenshots. Fully mocked auth/API, external requests blocked, static server without proxy; two intentional intercepted choice POSTs in the completed main run, zero live writes; attribution/scope controls zero writes. Phone uses hasTouch/isMobile; the choice activation was keyboard Enter, not a physical tap claim. Initial harness waited on the request ID after transformation to a ruling ID; corrected before the completed runs. Full Other/409/stale-response/reading regression matrix and genuine live decision loop remain pending the corrected candidate.

lilyshen0722 (commented):
Gate cleared at **c389bf45f1609d1274ac97c537278179d5abde31**.

Method, per the operator's instruction: detached checkouts only, source reading and every probe taken from one path at a time, each restored to clean before use.

The mutation battery ran at **777a1615086d1d8063b5fcf5af8b92cdbd608e06** in `/tmp/gate-1597-777a1615`. The head then moved to c389bf45, so I re-checked out at `/tmp/gate-1597-c389bf45` and confirmed the five files those findings rest on — `V2Thread.tsx`, `V2ActivityPage.tsx`, `activityService.ts`, `attentionItemService.ts`, `routes/activity.ts` — are byte-identical between the two shas. The delta is `v2.css` plus two test files. Control re-run at c389bf45: **159/159** across the five suites.

Control at the mutation sha: 153/153 across `V2ThreadDecisionCard`, `V2ActivityPage`, `v2-layout-invariants` and `v2-type-floors`.

## All four gate items discriminate

| mutation | result |
|---|---|
| strip `messageIds` from both thread reads | shipped suite **3 red**; my probe goes 1+1 → **100+100** requests |
| remove `if (pendingData)` | **2 red** |
| remove `if (historyData)` | **2 red** |
| remove the post-await `if (!active) return` | **1 red** |

**Request count follows loaded source IDs, not pod history.** My probe simulates a server that pages only when it is *not* given a filter: with the filter the client issues 1 queue + 1 history request, and without it 100 of each. Both carry `messageIds`.

One correction to my own work along the way: my first version of that probe returned `items: []`, and `loadDecisionPages` breaks on an empty page regardless of `hasMore`, so it reported 1 request whether or not the filter was sent — a fixture that could not fail. I only found it because the mutation didn't move the number. The result above is from the repaired fixture, which does move.

## Server side, read from the same checkout

Both endpoints now honour the filter — `getDecisionHistory` and `getOpenQueue` each build `messageId: { $in: messageIds }` when the array is present. `getDecisionHistory` also replaced the materialise-then-slice with `countDocuments` plus `.skip(offset).limit(limit)` in a `Promise.all`, so my earlier unbounded-fetch finding is resolved.

Authorization is unchanged and still sound: membership `$or` in the pod query, re-filtered in JS, `Access denied` on both empty results.

The backend suite could not be executed from a worktree (`ts-jest` does not resolve through a symlinked `node_modules`), so the backend paragraph is read, not run.

## Previously open, now closed

- pending cards survive a failed queue read (`if (pendingData)`, test at `V2ThreadDecisionCard.test.tsx:187`);
- settled cards survive a failed history read (`if (historyData)`, test at `:144`);
- the 15s poll no longer walks pod history on either surface.

No remaining conditions from me. UX and CI gates are separate.

lilyshen0722 (commented):
Additional verification at **777a1615086d1d8063b5fcf5af8b92cdbd608e06**, one detached checkout at `/tmp/gate-1597-777a1615`, restored clean and removed.

## Initial empty transcript — safe, but two wasted requests

Measured. At mount with no messages the client issues both reads with `messageIds: ""`. Server side that is not an unfiltered read: `rawMessageIds` is `''`, so `messageIds` becomes `[]`, and because an empty array is truthy the first branch passes `{ messageIds: [] }`; the service then builds `messageId: { $in: [] }`, which matches nothing. So the empty-transcript case costs two requests that can only ever return nothing — cheap, not a scan.

The refetch is prompt rather than poll-delayed: the effect depends on `detailInitialLoadComplete`, so when the transcript arrives it re-reads with the real ids and the card renders. Probe result across the sequence:

```
atMount:   ["", ""]
afterLoad: ["", "", "9", "9"]
cardVisible: true
```

Skipping the read when there are no loaded ids would remove two round trips per pod open. Minor.

## The real bound is `.slice(0, 200)`, and it drops the wrong end

`routes/activity.ts` caps the filter at 200 ids on both endpoints. `messages` is chronological ascending, and the ref maps it in order, so the 200 sent are the **oldest** loaded messages. A reader who has paged back to 300 messages gets decision cards for the oldest 200 and **not** for the 100 nearest the bottom — the ones actually on screen.

This is silent: no error, no indicator, the card simply never appears. It is also the new answer to "an older loaded source", which is now reachable *if* it falls inside the first 200 and unreachable otherwise, regardless of how recently it was loaded.

Two options, both cheap: send the last 200 rather than the first, or chunk the ids across requests so the count still tracks loaded sources. The first keeps one request and matches what the reader is looking at.

## Stale pod responses

Covered. Removing the post-await `if (!active) return` reddens 1.

## Failed reads

Both covered, as reported at this SHA: removing `if (pendingData)` reddens 2, removing `if (historyData)` reddens 2.

## Request bound

Confirmed with a probe that can fail: against a server that pages only when unfiltered, the client issues 1 queue + 1 history request with the filter and 100 of each without it.

lilyshen0722 (commented):
Backend suites now executed — thank you for the dependency pointer. At **c389bf45f1609d1274ac97c537278179d5abde31**, one detached checkout, `backend/node_modules` symlinked to the existing dependency directory, no install and no tracked change. Restored and removed.

Control: **16/16** across `activity.read` and `activityService.attentionQueue`.

## Mutations

| mutation | result |
|---|---|
| history ignores `messageIds` (always `messageId: { $exists: true }`) | **1 red** |
| drop the membership `$or` from the pod query | **1 red** |
| drop the `$or` **and** the JS re-filter | **1 red** |
| drop the first `Access denied` throw | green |
| drop **both** `Access denied` throws | **green** |

The loaded-source filter is pinned, and so is the authorization *gate* — removing the membership condition from the query reddens, and removing the JS re-filter as well still reddens. Those are the parts that would leak, and they are covered.

## What is not covered: the 403 contract

Both `Access denied` throws can be deleted with the suite green. I checked that this is genuinely uncovered rather than merely unobservable — removing the first alone is masked by the second, so I removed both, and it stays green.

The consequence today is mild: with the throws gone a non-member gets `{ items: [], count: 0 }` instead of a 403, because the query is still membership-gated. No data leaks. But the route's 403 path is unreachable by any test, so a later refactor that moves membership out of the query — exactly the mutation that reddens above — would leave the throws as the only guard, and nothing would notice if they were dropped in the same change.

One route-level test asserting 403 for a pod the caller is not a member of closes it, and it belongs with this PR because this PR introduces the endpoint.

## Correcting my earlier note

I previously reported the backend suite as unrunnable and attributed it to `ts-jest` not resolving through a symlinked `node_modules` from a worktree. That was wrong: `ts-jest` is simply absent from this checkout's dependency directory and present in the one the operator named. The symlink mechanism works. I diagnosed from the error text instead of checking whether the package was there.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ c389bf45f1609d1274ac97c537278179d5abde31

Activity still cannot reach resolved decision 56 on demand, and a failed thread custom answer loses keyboard retry focus. The secondary borders, destination picker, and human ruling attribution now pass.

Candidate provenance: functional browser runs used detached checkout `/Users/xcjsam/agents/ux-lead/commonly/.dev/wt1597-777` at `777a1615086d1d8063b5fcf5af8b92cdbd608e06`. Following Sam65428, picker browser runs used a separate detached checkout `/Users/xcjsam/agents/ux-lead/commonly/.dev/wt1597-c389` at `c389bf45f1609d1274ac97c537278179d5abde31`. Both were production-built locally. The complete 777→c389 delta is picker CSS plus two test assertions; application TypeScript/TSX is byte-identical. Functional results below carry explicitly across that verified delta; they were not rerun on c389. Both tracked frontend source trees stayed clean.

Remaining findings:

1. **Resolved history is inaccessible after the first 50.** With 56 resolved decisions in one pod and no pending decisions, Activity renders 50; decision 56 is absent and no More control exists. Reproduced at 1440 and true 390×900, both global and actual selected-pod scope. History requests stop at offset 0. `V2ActivityPage.tsx:325` fetches one history page; the existing More handler only fetches the pending queue. Minimum plan control: **Show more resolved decisions · N remaining**, beneath the loaded results, using the existing text-button treatment and a 44px phone target. Fetch the next 50 only on activation, append/dedupe without replacing loaded rows, and preserve reading position. Failure retains rows and offers Retry for the same offset; reject stale-scope responses. Its remaining count is independent of Needs you. For keyboard use, preserve a useful focus target when the final page removes the control. No automatic history scan or new design/task is needed.

2. **Failed thread Other submission drops keyboard focus.** Open Other, type a custom ruling, focus Send ruling and press Enter; hold the POST, then return 503. The draft and inline alert remain and retry is enabled, but activeElement becomes BODY. In a separate desktop rerun, the next Tab goes to the composer Add button, skipping the failed answer's retry control. Activity retains BUTTON focus in the equivalent case. Restore focus to the failed thread answer's input or retry button after enabling it. This is a newly exercised recovery issue, not a claim that this PR introduced it. Source: `V2DecisionCard.tsx:45–85`; no focus recovery accompanies saving/error state.

Cleared and retained controls:

- At 777, Reply / Mark handled / Open compute white with `1px solid rgb(208,213,221)` borders at both widths; phone targets are 44px. The picker was still ink at 777. At c389 it computes white / ink text / the same neutral border, hover `rgb(242,244,247)` with stronger border, and keyboard focus has a cobalt border plus visible 3px focus ring. Both widths were visually inspected; picker checks made zero HTTP writes.
- Human reply 2056 for source 2055 renders exactly once with Sam fixture and “ruled”, even after consecutive asking-agent messages. Changing only the preceding author also passes. Both widths; screenshots inspected. This clears the prior hidden-attribution finding.
- Authored choice order, cobalt first choice, Recommended on the second choice visually and in its accessible name, and no horizontal overflow pass. A decision cannot be marked handled.
- Activity and thread Other/409: 8 cases across both widths pass exact custom/standing answer, named human, native disabled state while held, no reopened choices, and durable reload. Other 503 preserves the custom draft and supports successful retry; focus limitation is finding 2. Desktop activation used keyboard Enter; phone activation used actual Playwright touchscreen taps.
- Successful queue omission, failed Activity history refresh, hard reload, leave/return hydration, and an actual unmodified 15-second failed thread history poll preserve the ruling. Thread fixtures exercise 56 loaded sources and second-page results; this is not a >200-source or real-backend bound proof.
- All 24 send/reading cases pass: peer and same-human background arrivals preserve source position/focus, Jump and near-bottom follow, explicit local text/image sends follow across POST/socket ordering and newer interleaves, dedupe holds, and failed sends do not arm later background following. Four additional real-browser history cases pass plain prepend and user scrolling during a held prepend at both widths, retaining IDs/order/draft/focus and restoring native anchoring. No history-case HTTP writes.

Evidence: `.dev/signal/walk-1597-777{,-attribution,-actions,-focus,-reading-send,-reading-bounded}.js`, corresponding `1597-777-*-results.json`, and `.dev/signal/walk-1597-c389-picker.js` / `1597-c389-picker-results.json`, with 1440/390 screenshots. All auth/API/socket traffic was mocked, external requests blocked, and the static server had no API proxy. Completed runs made 36 intentional intercepted mock writes; an earlier discarded action harness attempt made one additional intercepted POST before cleanup. Zero live writes. The action harness's pending button locator was corrected because its accessible label becomes Working; the picker harness's initial hover expectation was corrected from #f9fafb to the measured #f2f4f7 token before the completed run.

Sprint Review owns the independent request-bound, pending/ruled failure, stale-pod, and >200-source chunking gates. Sam65435's chunking requirement remains open; this UX review does not clear it. No application approval, merge, deploy, or genuine live-decision acceptance is claimed.

lilyshen0722 (commented):
Review at **b3dfe71a75c759e2195d29ad1e4ff13356e5e192**, one detached checkout at `/tmp/gate-1597-b3dfe71a`, frontend and backend dependencies linked, restored clean.

Controls: frontend **161/161** across five suites, backend **18/18**.

## Batching and the oversized-filter rejection are pinned

| mutation | result |
|---|---|
| truncate to the **first** 200 ids instead of batching | **1 red** |
| truncate to the **last** 200 ids | **1 red** |
| remove the oversized-batch 400 from both routes | **1 red** |

Both truncation directions redden, which is the property the operator asked for — neither end may be dropped, and a fixture that only asserted "a card is restored" would have passed both. The silent `.slice(0, 200)` is gone from the routes and replaced by an explicit 400 with the limit named in the message, and the client batches under that limit rather than relying on it.

The empty-transcript case is handled deliberately: `if (batches.length === 0) batches.push([])`, so the read stays authoritative rather than being skipped.

## The new 403 test does not pin the authorization decision

`GET /api/activity/decision-history preserves the membership-gate 403` mocks the service and asserts the route's mapping:

```js
ActivityService.getDecisionHistory.mockRejectedValueOnce(new Error('Access denied'));
await request(app).get('/api/activity/decision-history?podId=private-pod').expect(403, ...);
```

That pins the route's error translation, which is worth having. It cannot see the service's decision to throw. Measured: deleting **both** `Access denied` throws from `getDecisionHistory` leaves the backend suite at **18/18 green**.

So the mutation that matters is still invisible. Today the consequence remains mild — a non-member receives an empty list rather than a 403, because the pod query is membership-gated and that gate *is* pinned (removing the `$or` reddens, removing the JS re-filter as well still reddens). The gap is that the throws could be dropped in a refactor with nothing noticing, and they are the guard that would matter if the query gate ever moved.

What closes it is a service-level test rather than a route-level one: call `getDecisionHistory` with a `podId` the user is not a member of and assert it rejects. The existing `activityService.attentionQueue` suite already has the Pod mock needed.

## Carried explicitly

The three preservation controls and the request-count bound were measured at 777a1615 and re-confirmed at c389bf45; the files they rest on are unchanged in this delta apart from the batching helper, which the two truncation mutations above cover directly.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ 440b335728426acd427e0431ceb29f0e3ff8e76b

On-demand history now reaches resolved decision 56, and the failed thread Other focus correction passes. Two history interaction issues remain: focus is lost after failure/final-page completion, and refreshing after the last page incorrectly restores an already-consumed remaining count.

Fixed production builds/checkouts: history at `b3dfe71a75c759e2195d29ad1e4ff13356e5e192`, `/Users/xcjsam/agents/ux-lead/commonly/.dev/wt1597-b3`; thread focus at `440b335728426acd427e0431ceb29f0e3ff8e76b`, separate `/Users/xcjsam/agents/ux-lead/commonly/.dev/wt1597-440`. Verified complete b3→440 delta: V2DecisionCard focus code and its test only. History source is byte-identical, so the b3 history findings explicitly carry to 440; no history rerun at 440 is claimed.

1. **History loading loses keyboard position.** In All pods and actual selected-pod scope, activate Show more settled with Enter, return 503, then Retry and finish the final page. Failure leaves activeElement at BODY. Final-page completion also leaves BODY because the More button disappears; the next Tab goes to the rail Pods button on desktop or Today on phone, rather than the new results. `V2ActivityPage.tsx:578` only focuses historyMoreButtonRef, with no target after its removal, and the catch provides no focus recovery. Restore Retry focus after failure when the user has not deliberately moved; on the final page focus the first newly added row with preventScroll, preserving the reader's position. The existing pending-queue handler already has a first-added-row fallback, but history lacks it.

2. **Focus refresh restores a false remaining count after loading everything.** With 56 resolved decisions plus one pending decision, successful pagination yields 57 unique visible rows and no More button. A subsequent focus refresh retains all 57 rows but shows Show more settled · 6 remaining again. The ordinary focus/Tab sequence exposed this at both widths; a separate desktop control dispatched a window focus event and reproduced it directly. The history refresh at `V2ActivityPage.tsx:363–369` resets historyOffsetRef/remaining from the first page while retaining previously loaded history rows. Preserve the loaded history extent and accurate remaining count across refresh; do not invite refetching a page whose six rows are already shown. The initial first-page refresh is not an automatic historical scan.

Passing history controls, freshly run at b3 at 1440 and true 390×900, All pods and selected-pod scope:

- Initial 50 resolved rows, explicit Show more settled · 6 remaining; no automatic offset>0 request. One pending decision keeps Needs you at 1 regardless of resolved history length.
- Held native-disabled control; failed page retains IDs/order/draft and offers Retry. Failure and retry each request offset 50 once. The keyboard focus-out/reentry diagnostic additionally triggers an offset-0 refresh, recorded separately rather than counted as an extra historical page.
- Retry appends the final six without duplicates. All original rows stay in order; the last loaded row shifts 0px and scroller stays exactly 4649px desktop / 10160px phone. Draft survives. Decision 56 is actually scrolled into view and its exact human ruling read, not merely counted in the DOM.
- A held Alpha history page returned after switching to Beta is discarded at both widths; Beta's rows and focus remain unchanged. All history probes make zero HTTP writes.

Thread correction, freshly run at 440: four cases across both widths. Other → held POST → 503 restores focus to the custom-answer INPUT, retains draft/alert/retry, and leaves scroller at 2272px desktop / 2395px phone. If the user deliberately focuses the composer while the POST is held, failure preserves that TEXTAREA focus instead. Successful retry settles the exact custom value with Sam fixture, removes choices, and survives reload. Desktop used Enter; phone used actual Playwright touchscreen taps. Eight intentional intercepted mock POSTs, zero live writes.

Evidence: `.dev/signal/walk-1597-b3-history.js`, `1597-b3-history-results.json`, `1597-b3-history-final-results.json`, `walk-1597-b3-refresh.js` / `1597-b3-refresh-results.json`, and `walk-1597-440-focus.js` / `1597-440-focus-results.json`. Phone final-history/decision56 screenshots inspected. The augmented history harness initially assumed no offset-0 refresh during keyboard focus-out/reentry; two diagnostic runs stopped at that assertion, then the completed run distinguished first-page refresh from historical offset-50 requests. No application code was modified for testing.

Both builds pass. All auth/API/socket traffic was mocked, external requests blocked, and static servers had no API proxy. This is a targeted follow-up to review5135214147, not a rerun or blanket carry of its full reading matrix over the intervening V2Thread batching changes. Independent batching/failure/backend authorization gates remain Sprint Review's responsibility; their latest service-level 403 test finding is separate. No application approval, merge, deploy, or live durable-rendering claim. Real request64684 → Sam64685 → agent64686 remains the operator-proposed post-deployment case.

lilyshen0722 (commented):
Final code verdict at **6462a9d1e8db8539406cb3aaaff5388701ef3193** — one detached checkout at `/tmp/gate-1597-6462a9d1`, both dependency trees linked, restored clean and removed.

**Not clear.** The one outstanding item is not in this head.

## The service-level 403 test is absent

The delta from b3dfe71a touches five frontend files only — `V2ActivityPage.tsx`, `V2DecisionCard.tsx` and three test files. No backend test changed.

Re-measured here rather than carried: backend control **18/18**, and deleting **both** `Access denied` throws from `getDecisionHistory` still leaves **18/18 green**. So the branch remains uncovered, exactly as reported at the previous head.

The existing route test mocks the service and asserts the 403 mapping, which cannot reach the service's decision to throw. What closes it is a direct call: `getDecisionHistory(userId, { podId })` for a pod the user is not a member of, asserting rejection. The `activityService.attentionQueue` suite already mocks `Pod.find`, so the fixture is a few lines.

## Everything else passes at this SHA

Frontend control **165/165** across six suites.

`V2Thread.tsx`, `routes/activity.ts`, `activityService.ts` and `attentionItemService.ts` are byte-identical to b3dfe71a, so these results carry explicitly rather than by assumption:

- batching pinned in both directions — truncating to the first 200 reddens 1, to the last 200 reddens 1;
- oversized-batch 400 pinned — removing it from both routes reddens 1;
- loaded-source filter pinned — making history ignore `messageIds` reddens 1;
- the authorization *gate* pinned — removing the membership `$or` reddens, removing the JS re-filter as well still reddens;
- pending and ruled preservation, and stale-pod discard, each redden.

## Verdict

One item, one small test. With it, the code side is clear from me; without it the endpoint ships with its 403 contract unexercised. Nothing else outstanding.

lilyshen0722 (commented):
**Code side clear at b14ffe7fa08f7ac7773215889049603af42058c4.** One detached checkout at `/tmp/gate-1597-b14ffe7f`, restored clean and removed.

The head moved while my previous verdict was being filed; that verdict applies to 6462a9d1 and is superseded here.

## The 403 branch is now covered

Backend control **19/19**. Deleting **both** `Access denied` throws from `getDecisionHistory` now reddens 1 — at 6462a9d1 the same mutation left the suite green, so the new `rejects settled history for a viewer outside the requested pod` is what closes it, and it closes it at the service rather than through the route's mocked mapping.

Removing only the *second* throw is invisible (7/7). That is expected rather than a gap: the pod query already excludes non-member pods, so `pods.length === 0` fires first and the later check is defence in depth behind it. The pair is pinned; the redundant half is unreachable while the query keeps its membership clause — which is itself pinned, since removing the `$or` reddens.

## Carried explicitly

`frontend/` is byte-identical between 6462a9d1 and this head, so these stand without re-running:

- frontend control 165/165 across six suites;
- batching pinned in both directions (first-200 reddens 1, last-200 reddens 1);
- oversized-batch 400 pinned;
- loaded-source filter pinned;
- pending preservation, ruled preservation, and stale-pod discard each redden;
- request count bounded by loaded source IDs — 1 request per endpoint with the filter against 100 without, measured with a fixture I first had to repair because an empty-items response made it unable to fail.

## Verdict

No remaining code conditions from me. The UX history focus and remaining-count items are ux-lead's and still open; CI is separate.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ b14ffe7fa08f7ac7773215889049603af42058c4

History and thread focus recovery now pass. Refreshing an unchanged history no longer revives More. One remaining-count edge still invites a duplicate fetch: a newly resolved item already loaded by the refreshed first page is counted as unseen.

Browser candidate: `6462a9d1e8db8539406cb3aaaff5388701ef3193`, production-built from the fixed detached checkout `/Users/xcjsam/agents/ux-lead/commonly/.dev/wt1597-6462`; tracked frontend source remained clean. Verified complete 6462→b14 delta: one backend service-test file, nine added lines. All application/frontend files are byte-identical. These browser results explicitly carry to b14; no b14 rebuild/rerun is claimed.

Remaining finding — **new first-page results leave a false More count and stale offset**:

1. Load all 56 resolved decisions, alongside one pending decision: 57 unique rows; no More control.
2. Resolve one additional, newer decision. On focus refresh, the first page returns this new ruling and the client merges it into the retained history. All 57 resolved decisions are now loaded (58 unique rows including the pending decision).
3. More nevertheless says **1 remaining**. Activating it requests offset 56, which returns the already-loaded old final row; visible unique rows stay at 58 and More disappears. Reproduced at 1440 and true 390×900; both global and selected-pod refresh show the false count, and separate desktop/phone controls activate it and verify zero added rows.

`V2ActivityPage.tsx` retains the previous raw extent but subtracts it from the increased server count without accounting for newly merged first-page IDs. Keep the history cursor/remaining count aligned with the unique rows now loaded. Do not simply suppress count growth: the paired control adds one genuinely unseen older ruling, which correctly shows More1, fetches offset56, and adds exactly one new row. Both widths pass that control. This is the remaining edge of the existing accurate-count requirement, not a new design direction.

Freshly cleared at 6462:

- History 503 focuses Retry. Final-page success focuses the first newly added article, decision-50; existing tail shifts 0px and scroller stays fixed. If the user focuses the composer while the page is held, both failure and success preserve that TEXTAREA focus, with no scroll change.
- An unchanged same-scope focus refresh after all history is loaded retains all rows and keeps More absent. Core on-demand access to decision56, native disabled state, failed-page draft/order preservation, retry offset50, append/dedupe, Needs you1 independent of resolved count, and stale Alpha response rejection after switching to Beta all pass.
- Four thread Other cases rerun because this candidate also changes V2DecisionCard: restore custom INPUT after 503, respect deliberate composer focus, preserve scroll at 2272px desktop / 2395px phone, retain draft/alert/retry, then settle the exact custom answer with Sam fixture and survive reload. Desktop Enter and actual phone touchscreen taps; eight intercepted mock POSTs, zero live writes.

Evidence: `.dev/signal/walk-1597-6462-{history,moved,newest,newest-phone,focus}.js` and corresponding `1597-6462-*-results.json`; 1440/390 screenshots. The first augmented history attempt stopped in the separate stale-scope setup because its newly grown first page no longer contained the old tail prerequisite; the fixture was reset before that independent case and the complete run passed. No diagnostic application edits. All history probes made zero HTTP writes; all auth/API/socket traffic was mocked, external requests blocked, and the static server had no API proxy.

This targeted follow-up clears the focus and unchanged-count findings from review5135265280 and retains only the growth edge above. Earlier visual/reading results are not represented as a new full-matrix run. Sprint Review's reported b14 code/authorization verdict and CI remain separate. No final application approval, merge, deployment, or live durable-rendering proof.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 8d3d1e0d589b5fbcc57668a9158d3956413c0a54 — reported history/focus/count findings cleared.

Production browser ran at fixed 61d619ae05a89b554db7002e4381b2056c8c358d, 1440×900 and true mobile 390×900. Full 61d619ae→8d3d1e0d diff adds only one frontend test (41 lines); application source is byte-identical. Results explicitly carry; no 8d browser/build rerun.

- Global and actual selected-pod scope: 56 resolved + one pending initially show 50 resolved, More6, Needs you1. History loads only on activation. Held button is disabled; 503 retains rows/order/draft and focuses Retry. Retrying offset50 appends six unique rows, preserves the existing tail position (0px shift), and focuses first-added decision50 without scrolling. Decision56 is actually scrolled into view and its Sam ruling read.
- Unchanged-count refresh keeps all loaded rows and More absent. Newest ruling already merged by first-page refresh gives 57/57 resolved loaded and no More, globally and scoped at both widths.
- Unseen older ruling still gives More1 and actual activation at offset56 adds it. Combined newest plus unseen older arrivals also show exactly More1, then append the unseen row without duplicates (59 total including the pending card).
- Deliberately moved composer focus survives both failed and successful history loads. Stale Alpha history returned after switching Beta leaves Beta empty and its scope focus unchanged at both widths.
- Desktop uses keyboard Enter; phone uses actual touchscreen taps. All three completed scripts pass first run. All browser auth/API/socket data mocked, external requests blocked, static server has no API proxy. Zero HTTP writes, including zero live writes.

The 6462 thread Other retry-focus evidence carries over unchanged V2DecisionCard; the earlier computed picker/border and human-attribution component findings remain cleared. The old 28-case reading/send matrix was not rerun or claimed wholesale over later thread batching changes. Code/service/batching mutation review, CI, and post-deployment real 64684→64685→64686 durable-render proof remain separate.

Evidence: .dev/signal/walk-1597-61-{history,moved,combined}.js and matching 1597-61-*-results.json; phone combined-arrivals screenshot inspected. Build passed; tracked frontend clean. No application edits, merge or deployment.

lilyshen0722 (commented):
Count-delta review at **8d3d1e0d589b5fbcc57668a9158d3956413c0a54** — detached checkouts at `/tmp/gate-1597-61d619ae` and `/tmp/gate-1597-8d3d1e0d`, both restored clean and removed. Everything outside this delta is carried from my cleared b14ffe7f as instructed.

Control: **145/145** at this head (`V2ActivityPage` 41, `v2-layout-invariants` 104); 144/144 at 61d619ae where the mutation battery ran.

The head moved once during this review. The new test at 8d3d1e0d — *keeps an unseen older ruling available after a newest ruling arrives* — is a genuine addition, and I re-ran the condition mutation against it rather than carrying the earlier result.

## Two thirds of the change is pinned

| mutation | result |
|---|---|
| count only the refreshed page, ignoring already-rendered cards | **1 red** |
| drop the refreshed page from the union | **1 red** |
| revert the guard to `previousHistoryExtent > historyItems.length` | **green at 61d619ae, still green at 8d3d1e0d** |

The union itself is well covered from both sides. The **condition** that selects it is not.

## The condition change is untested

The delta changes `sameScope && previousHistoryExtent > historyItems.length` to `sameScope && previousHistoryExtent > 0`. Restoring the old condition leaves the suite green, and the fixture shows why: the third history read returns 50 items while `previousHistoryExtent` is 51 (50 + the older page), so `51 > 50` holds and both conditions take the same branch. The test exercises the union, never the choice to use it.

The two conditions can only diverge when `previousHistoryExtent > 0` **and** `previousHistoryExtent <= historyItems.length` — that is, when the reader has loaded only the first page and never pressed "show more". In that state a new ruling arrives, the refreshed page still returns a full page, and the old condition falls through to the server's `remaining`, which counts the new row as unseen older history; the new condition takes the union and does not. That is the reported symptom.

So the missing control is the *simpler* of the two paths: first page only, no "show more", then a new ruling. Both fixtures — the original and the one added at 8d3d1e0d — press "Show more settled" first, which puts `previousHistoryExtent` at 51 against a 50-item page, so `51 > 50` holds and the two conditions agree. The new test broadens what happens *after* the union is chosen; it does not exercise the choice.

I am flagging the divergence analytically — from the fixture's numbers and the two expressions — rather than from a run. What I measured is that the shipped test cannot separate the conditions.

## Carried explicitly from b14ffe7f

`V2Thread.tsx`, `routes/activity.ts`, `activityService.ts` and `attentionItemService.ts` are untouched in this delta, so the batching, oversized-batch 400, loaded-source filter, preservation, stale-pod and 403 results all stand unchanged.

Not a blocker on its own — the behaviour at this head is the intended one. But the guard that produces it is currently free to regress silently, and the fixture that would catch it is smaller than the one already written.

lilyshen0722 (commented):
Cleared at **ffa99e4b3cea0dedc0ec13c5d8138a598f55c4db** — one detached checkout at `/tmp/gate-1597-ffa99e4b`, restored clean and removed.

Control **146/146** (`V2ActivityPage` 42, `v2-layout-invariants` 104).

The gap is closed. Reverting the guard to `previousHistoryExtent > historyItems.length` now reddens *does not revive More after a newest ruling arrives before older history is requested* — the same mutation was green at 61d619ae and at 8d3d1e0d. The new fixture reaches the state the earlier two could not: first page loaded, no show-more, extent equal to the page length, so the two conditions finally disagree.

The union assertions strengthened alongside it: emptying `existingIds` now reddens **3** rather than 1.

## Full picture at this SHA

Everything I have gated on this PR, with the basis for each stated:

- condition selecting the union — **pinned here** (1 red);
- union over already-rendered cards and over the refreshed page — pinned (3 red, 1 red);
- batching, both truncation directions — pinned, carried from b3dfe71a where `V2Thread.tsx` was last touched;
- oversized-batch 400 and loaded-source filter — pinned, carried from b3dfe71a;
- pending preservation, ruled preservation, stale-pod discard — pinned, carried;
- service-level 403 — pinned at b14ffe7f (deleting both throws reddens 1);
- request count bounded by loaded source IDs — 1 request per endpoint with the filter against 100 without.

No remaining code conditions from me.

PR #1599, opened 2026-09-07
Preserve resolved decision cards and source links in pod threads

lilyshen0722:
Expanded decision requests inside pod threads previously remained raw request prose after Activity showed the recorded ruling. Render the same durable human answer in both root and reply positions, suppress its duplicate raw row, and retain both source-request and durable-answer links on the visible ruling.

When decision history arrives after Activity has already landed on the source, preserve the landing through the replacement while respecting deliberate focus movement. Keep that recovery separate from folded-thread reveal and older-history lookup. The AX audit records why root-only fixtures missed the reply path.

Validation at final head 7c549726: independent code approval and full frontend run (102 suites / 776 tests); UX approval after a fresh production build and 10 reveal/hydration/focus cases at desktop1440 and mobile390, with zero HTTP writes. UX also ran the full frontend suite and three Node tests. Source-alias, duplicate-suppression, durable-identity and hydration probes remain covered. Exact-head CI gates merge; verification of the genuine request64684 → recorded choice64685 → agent action64686, plus durable rendering after reload/navigation, gates task completion after deployment.

Follow-up to #1597, deployed as 713e1f6c. The Activity composer dropdown correction is a separate follow-up.

### DIFF
--- a/docs/development/agent-experience-audit.md
+++ b/docs/development/agent-experience-audit.md
@@ -3720,3 +3720,21 @@ looks like the absence of a problem.
 - Related: entry 36 (the fleet's checkout tracks no revision) — the same
   hand-synced worktree is what made "the fix is merged" and "the fix is
   running" two different questions here.
+
+## 55. A shared state map does not mean every renderer consumes it (2026-09-07, sprint-impl)
+
+*Origin observation: @ux-lead, msg 65479; verification and path comparison: @sprint-review, msgs 65482 and 65484; implementation: @sprint-impl, PR #1599.*
+
+The settled-decision map was hydrated and the flat transcript renderer already
+projected it into a human ruling row. The test fixtures therefore looked
+covered while always putting the decision request at the root. In production,
+the real request (64684) was a reply under 64679. The expanded-replies branch
+passed the raw `V2MessageRow`, so the live thread showed ordinary request/reply
+prose with no ruled marker even though Activity correctly said “Sam ruled”.
+
+**Lesson:** shared state is not shared behavior. When a stateful surface has
+multiple renderers, a fixture must exercise each structural path that can
+consume the state. A green root-only test can prove the map is populated while
+missing the user-visible path entirely. The safe test shape is the real
+source → reply → durable ruling relationship, and the implementation must
+assert both the transformed row and suppression of the duplicate durable row.
--- a/frontend/src/v2/__tests__/V2ThreadDecisionCard.test.tsx
+++ b/frontend/src/v2/__tests__/V2ThreadDecisionCard.test.tsx
@@ -242,6 +242,67 @@ describe('V2Thread decision cards', () => {
     }
   });
 
+  test('preserves deliberate composer focus through settled hydration', async () => {
+    const settled = {
+      id: 'decision-42', kind: 'decision', podId: 'pod-1', messageId: '42',
+      title: 'Choose the workspace cutover', detail: 'Which implementation should ship?',
+      options: [{ label: 'Ship the rebuilt workspace' }], status: 'ruled',
+      ruling: {
+        value: 'Ship the rebuilt workspace', by: 'Lily', messageId: '43', at: '2026-09-05T12:01:00.000Z',
+      },
+    };
+    const pending = { ...settled, status: 'pending', ruling: undefined };
+    let historyReads = 0;
+    mockGet.mockImplementation((url) => {
+      if (url === '/api/activity/decision-queue') return Promise.resolve({ data: { items: [pending] } });
+      if (url === '/api/activity/decision-history') {
+        historyReads += 1;
+        return historyReads === 1
+          ? Promise.resolve({ data: { items: [] } })
+          : Promise.resolve({ data: { items: [settled] } });
+      }
+      return Promise.resolve({ data: {} });
+    });
+    const settledDetail = {
+      ...detail,
+      messages: [
+        ...detail.messages,
+        {
+          id: '43', pod_id: 'pod-1', user_id: 'human-1',
+          user: { username: 'lily', isBot: false }, content: 'Ship the rebuilt workspace',
+          message_type: 'text', created_at: '2026-09-05T12:01:00.000Z',
+          thread_root_id: '42',
+        },
+      ],
+    };
+
+    jest.useFakeTimers();
+    try {
+      const { container } = render(
+        <AuthContext.Provider value={auth}>
+          <MemoryRouter initialEntries={['/v2/pods/pod-1#message-42']}>
+            <V2Thread detail={settledDetail} />
+          </MemoryRouter>
+        </AuthContext.Provider>,
+      );
+      expect(await screen.findByTestId('decision-card')).toBeInTheDocument();
+      await waitFor(() => expect(container.querySelector('#message-42')).toHaveClass('v2-msg--landed'));
+      const composer = container.querySelector('textarea');
+      expect(composer).toBeTruthy();
+      composer.focus();
+
+      await act(async () => {
+        jest.advanceTimersByTime(15_000);
+        await Promise.resolve();
+      });
+      await waitFor(() => expect(historyReads).toBe(2));
+      expect(composer).toHaveFocus();
+      expect(container.querySelector('#message-43')).not.toHaveClass('v2-msg--landed');
+    } finally {
+      jest.useRealTimers();
+    }
+  });
+
   test('returns focus to a failed custom answer instead of skipping to the composer', async () => {
     mockPost.mockRejectedValueOnce(new Error('temporary failure'));
     render(
--- a/frontend/src/v2/__tests__/V2ThreadDecisionCard.test.tsx
+++ b/frontend/src/v2/__tests__/V2ThreadDecisionCard.test.tsx
@@ -184,6 +184,64 @@ describe('V2Thread decision cards', () => {
     }
   });
 
+  test('preserves Activity landing when hydration replaces the focused source row', async () => {
+    const settled = {
+      id: 'decision-42', kind: 'decision', podId: 'pod-1', messageId: '42',
+      title: 'Choose the workspace cutover', detail: 'Which implementation should ship?',
+      options: [{ label: 'Ship the rebuilt workspace' }], status: 'ruled',
+      ruling: {
+        value: 'Ship the rebuilt workspace', by: 'Lily', messageId: '43', at: '2026-09-05T12:01:00.000Z',
+      },
+    };
+    const pending = { ...settled, status: 'pending', ruling: undefined };
+    let historyReads = 0;
+    mockGet.mockImplementation((url) => {
+      if (url === '/api/activity/decision-queue') return Promise.resolve({ data: { items: [pending] } });
+      if (url === '/api/activity/decision-history') {
+        historyReads += 1;
+        return historyReads === 1
+          ? Promise.resolve({ data: { items: [] } })
+          : Promise.resolve({ data: { items: [settled] } });
+      }
+      return Promise.resolve({ data: {} });
+    });
+    const settledDetail = {
+      ...detail,
+      messages: [
+        ...detail.messages,
+        {
+          id: '43', pod_id: 'pod-1', user_id: 'human-1',
+          user: { username: 'lily', isBot: false }, content: 'Ship the rebuilt workspace',
+          message_type: 'text', created_at: '2026-09-05T12:01:00.000Z',
+          thread_root_id: '42',
+        },
+      ],
+    };
+
+    jest.useFakeTimers();
+    try {
+      const { container } = render(
+        <AuthContext.Provider value={auth}>
+          <MemoryRouter initialEntries={['/v2/pods/pod-1#message-42']}>
+            <V2Thread detail={settledDetail} />
+          </MemoryRouter>
+        </AuthContext.Provider>,
+      );
+      expect(await screen.findByTestId('decision-card')).toBeInTheDocument();
+      await waitFor(() => expect(container.querySelector('#message-42')).toHaveClass('v2-msg--landed'));
+      await waitFor(() => expect(historyReads).toBe(1));
+
+      await act(async () => {
+        jest.advanceTimersByTime(15_000);
+        await Promise.resolve();
+      });
+      await waitFor(() => expect(container.querySelector('#message-43')).toHaveClass('v2-msg--landed'));
+      expect(document.activeElement).toBe(container.querySelector('#message-43'));
+    } finally {
+      jest.useRealTimers();
+    }
+  });
+
   test('returns focus to a failed custom answer instead of skipping to the composer', async () => {
     mockPost.mockRejectedValueOnce(new Error('temporary failure'));
     render(
--- a/frontend/src/v2/__tests__/V2ThreadRestyle.test.tsx
+++ b/frontend/src/v2/__tests__/V2ThreadRestyle.test.tsx
@@ -198,6 +198,50 @@ describe('direction C threading restyle (PR 2b)', () => {
     expect(container.querySelector('.v2-thread-block--open')).toBeNull();
   });
 
+  test('expanded settled source replies render one ruled answer instead of a duplicate durable reply', () => {
+    const root = msg(20);
+    const source = msg(21, { thread_root_id: '20', content: 'Choose a path' });
+    const durableRuling = msg(22, { thread_root_id: '20', content: 'Ship it', user: { username: 'Sam' } });
+    const replies = [source, durableRuling];
+    const { container } = renderThread({
+      messages: [root, ...replies],
+      threadView: [
+        { kind: 'message', message: root },
+        { kind: 'card', rootId: '20', replyCount: replies.length, participants: [], lastActivityAt: null, replies },
+      ],
+      settledDecisionByMessageId: new Map([['21', { value: 'Ship it', by: 'Sam', messageId: '22' }]]),
+    });
+
+    fireEvent.click(screen.getByRole('button', { name: 'Expand thread, 2 replies' }));
+    const block = container.querySelector('.v2-thread-block--open');
+    expect(block.querySelectorAll('.v2-thread-replies .v2-msg')).toHaveLength(1);
+    expect(block.querySelector('[data-testid="decision-ruling-row"]')).toHaveTextContent('Ship it');
+    expect(block.querySelectorAll('#message-22')).toHaveLength(1);
+    Element.prototype.scrollIntoView = jest.fn();
+    expect(landOnMessage('21')).toBe(true);
+    const sourceTarget = block.querySelector('#message-22');
+    expect(sourceTarget).toHaveClass('v2-msg--landed');
+    expect(landOnMessage('22')).toBe(true);
+    expect(block.querySelector('#message-22')).toBe(sourceTarget);
+  });
+
+  test('settled root keeps source and durable links on the same visible ruling', () => {
+    const root = msg(30, { content: 'Choose a path' });
+    const durableRuling = msg(31, { content: 'Ship it', user: { username: 'Sam' } });
+    const { container } = renderThread({
+      messages: [root, durableRuling],
+      threadView: [{ kind: 'message', message: root }],
+      settledDecisionByMessageId: new Map([['30', { value: 'Ship it', by: 'Sam', messageId: '31' }]]),
+    });
+    const ruling = container.querySelector('[data-testid="decision-ruling-row"]');
+    expect(ruling).toBe(container.querySelector('#message-31'));
+    Element.prototype.scrollIntoView = jest.fn();
+    expect(landOnMessage('30')).toBe(true);
+    expect(landOnMessage('31')).toBe(true);
+    expect(container.querySelectorAll('[data-testid="decision-ruling-row"]')).toHaveLength(1);
+    expect(container.querySelector('#message-31')).toBe(ruling);
+  });
+
   test('landing on a reply folded behind the chip reveals the whole thread instead of fetching history', () => {
     const onRevealed = jest.fn();
     const { container, rerender } = renderThread({ revealMessageId: '22', onRevealed });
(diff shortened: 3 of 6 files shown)


lilyshen0722 (commented):
Reply-path review at **e09fa2e84ad36f3a1f8579e2527f11ea72ac0c66** — mutation battery run in one detached checkout at `/tmp/gate-1599-46edb182`, restored clean and removed.

The head moved to e09fa2e8 while I was measuring; that delta is the AX-audit note only (`docs/development/agent-experience-audit.md`, +18 lines), with no source or test change, so every result below carries unchanged.

Control **26/26** across `V2ThreadRestyle`, `V2ThreadDecisionCard` and `V2ThreadHistory`.

## Against the four gates

| gate | mutation | result |
|---|---|---|
| reply renders the ruling row | reply branch ignores the settled lookup | **1 red** |
| duplicate durable reply suppressed | drop the `settledRulingReplyIds` filter | **1 red** |
| pending card suppressed on a settled reply | `decision={decisionByMessageId.get(...)}` unconditionally | green |
| attribution / grouping match the root path | — | see below |

Two of the three mutable gates discriminate. The new regression does place the decision under a root and asserts one row, the ruled marker, and no duplicate `#message-22`, which is the shape that was missing.

## The surviving mutation is redundancy, not a gap

I checked rather than reporting it as uncovered. `rulingMessage` rewrites the id to `decision-ruling-${source.id}`, and the `decision=` prop is evaluated against the **transformed** message, so `decisionByMessageId.get('decision-ruling-21')` can never match a pending decision. The `ruling ? undefined :` guard is unreachable while the transform rewrites the id.

That makes it harmless defence, but it reads as load-bearing. Worth either a comment saying why it cannot fire, or dropping it — an unreachable guard beside a reachable one is how the next reader learns the wrong model.

## Two smaller notes, neither blocking

**Attribution is implied, not asserted.** The test checks the ruling row's text is "Ship it"; the author comes from `ruling.by`, which the fixture sets to `Sam`, but nothing asserts the rendered author. Swapping `ruling.by` for the source's agent name would not redden.

**Reload is not exercised on this path.** `V2ThreadRestyle` renders `V2ThreadMessages` directly with a supplied map, so there is no hydrate dimension here; the root path's reload coverage lives in `V2ThreadDecisionCard`. The rendering code is the same on hydrate, so I would not hold for it — but the operator asked for reload in the brief, and it is currently satisfied only by inference.

## Verdict

Clear from me. The defect is fixed and the fix is pinned by the mutation that matters; the three items above are quality notes rather than conditions.

lilyshen0722 (commented):
Navigation check at **e9332dfe4250eb1b15b9740fb26ec92669d9021d**, one detached checkout, restored clean and removed. Control 26/26.

The head moved twice while I measured; I re-ran the probe here rather than carrying it. The e9332dfe delta routes the pending lookup through `sourceId` instead of the transformed id, which aligns the reply path with the root path — but it does not add a DOM anchor for the source, so the finding below is unchanged at this head.

## The risk is real, and measured

Probing the new regression's own fixture — source reply `21`, durable ruling `22`, settled:

```
{ sourceAnchor: 0, durableAnchor: 1 }
```

`#message-21` does not exist in the rendered thread. Activity's Open appends `#message-<sourceId>`, and for a settled history row that id is the **decision's source message**, so `landOnMessage` finds nothing and the landing silently fails. Links to the durable answer still work.

## It is not new to the reply branch

`rulingMessage` returns the durable row itself when `ruling.messageId` is loaded, falling back to `decision-ruling-${source.id}` only for the response-before-socket race. `V2MessageRow` anchors on `message.id`. The **root** path has done exactly this since before this PR — `message={renderedMessage}` at `:255` — so the source anchor is lost there too.

So scoping a fix to the new reply code would repair half of it. Whatever alias is chosen — keeping `id={message-${sourceId}}` on the projected row, or emitting a hidden anchor beside it — belongs in the shared projection so both paths get it, and the assertion should be that **both** `#message-<sourceId>` and `#message-<durableId>` resolve to the one visible ruling.

## Correcting something I said earlier

I reported the surviving `ruling ? undefined : decisionByMessageId.get(...)` mutation as *unreachable*, on the grounds that the id is always rewritten to `decision-ruling-…`. That was wrong: I read only the tail of `rulingMessage` and missed the `if (durable) return durable;` above it. On the loaded path the id is a real message id, so the lookup is not structurally impossible — it merely requires a pending decision keyed to the ruling reply's own id, which is unlikely rather than unreachable.

The mutation still survives, so the conclusion "not pinned by any test" stands. My explanation of *why* did not.

lilyshen0722 (commented):
# PR1599 reply presentation browser gate

Reply presentation PASS on production-built e09fa2e84ad36f3a1f8579e2527f11ea72ac0c66. Final UX clearance OPEN: Sam's source/durable-link landing check (65492) is held by Sprint Review. The latest head at this read is e9332dfe4250eb1b15b9740fb26ec92669d9021d, whose source-ID decision lookup delta was inspected but was not built or browser-tested here. No blanket carry to that head.

At 1440×900 and true mobile 390×900, a mocked root → agent decision reply → durable human ruling → agent follow-through passed:

- Pending reply renders interactive choices before the normal 15-second refresh hydrates the settled history.
- Hydrated nested reply becomes exactly one durable human row, with Sam fixture's visible attribution and ruled marker; pending card and duplicate answer disappear.
- Reload then expand, leave to Your team and return, and a different previous author all retain the same result.
- A real, unmodified 15-second timer read returning mocked 503 preserves the mounted ruling and attribution.
- Agent follow-through remains visible, and there is no document horizontal overflow.
- Desktop expansion used Enter; phone expansion used touch taps. Phone screenshot inspected.

All authentication, APIs and sockets mocked; external requests blocked; static server has no API proxy. Completed run: zero HTTP writes, including zero live writes. Both owned browser contexts closed in finally. This is a topology-faithful fixture, not a live recheck of messages 64684–64686; that remains a post-deploy gate. Passed live Activity evidence on 713e1f6c from 65479/65480 is retained; no Activity rerun here.

Evidence: walk-1599-e09.js, 1599-e09-results.json, 1599-e09-{1440,390}-nested-ruling.png in .dev/signal. Build passed; tracked frontend clean. One setup attempt ran build from the worktree root rather than frontend and failed to locate index.html; corrected build passed. The first fixture incorrectly held history while waiting for pending decisions, although the app joins both reads; corrected to normal pending hydration then timer refresh before the completed passing run.

No application edits, merge, deploy, choice submission or task mutation. Reviewer unit/mutation coverage is separate.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ 673ce99f358087b274ce7d3ed763a1eb51f98353

Root Activity Open/source reload loses its landing when decision history hydrates. At both 1440×900 and true mobile 390×900, the original source row receives focus and the landed marker; replacing it with the durable ruling leaves BODY focused and removes the marker. The alias correctly resolves links after hydration, but does not preserve the initial landing across replacement.

A separate held-history browser control confirms the transition at both widths: before fulfillment, active element is message-2055 with v2-msg--landed; after fulfillment, the single ruling is message-2056 with data-source-message-id=2055, while active element is BODY and the ruling is unmarked. This is an end-to-end hydration gap, not a failure of calling landOnMessage on an already-settled DOM. V2Thread.tsx's landing effect returns early for landedTargetRef.current === target and does not depend on the settled projection. Preserve focus/landing through replacement when the reader has not deliberately moved focus; retain durable action identity and avoid unnecessary scrolling. No claim that this behavior was introduced by this PR.

Passed at both widths on this exact production build:

- Expanded reply: actual Activity Open, source-hash reload, durable-answer link after leaving, and the human ruling's source quote all scroll to and focus the single marked ruling. Both IDs reach message-2056; source alias 2055 remains on that same row.
- Root: durable-answer link and source quote after hydration both land correctly on that one row.
- All 16 navigation cases retain one visible Sam fixture author, one ruled marker, one durable answer, and no pending decision card. Four root initial/reload cases fail focus/landing; 12 cases pass fully.
- The 25 later messages make scrolling observable. The phone ruling sits wholly inside its scroller after successful navigation. Phone screenshots inspected.

Desktop Open/quote activation used Enter; phone used actual touch taps. Root/reply topology follows the real case, with synthetic identities/content and APIs. Both completed scripts made zero HTTP writes. Auth/API/socket mocked, external traffic blocked, static server has no proxy. Owned contexts closed in finally. Production build passed; no application edits, merge, deploy, live choice or task mutation.

Evidence: .dev/signal/walk-1599-673-navigation.js and 1599-673-navigation-results.json; walk-1599-673-hydration-focus.js and 1599-673-hydration-focus-results.json; 1599-673-390-reply-Activity-Open.png and 1599-673-390-root-after-held-hydration.png. Both scripts completed on first run.

Prior e09 reply presentation/poll-failure evidence remains separately recorded; it is not a blanket browser carry to changed files. Unrelated live Activity approval on deployed 713e1f6c is retained. The genuine 64684→64685→64686 source-thread verification remains for Sam after integration/deployment. Code/mutation tests and CI are separate gates.

lilyshen0722 (commented):
**APPROVED — alias fix gated at `673ce99f358087b274ce7d3ed763a1eb51f98353`.**

Method: one detached checkout, `/tmp/gate-1599-acd6cd1e`, `git status --porcelain` empty before and after every mutation. Source, tests and probes all from that tree. Note the announced sha was `acd6cd1e`; `origin/pr/1599` resolves to `673ce99f…` — the worktree is named after the announcement, the SHA above is what was measured.

Control at that SHA (per-suite, one file per invocation): V2ThreadRestyle 13, V2MessageRow 4, V2MessageRowQuote 5, V2ThreadReveal 20, V2ThreadHistory 6, V2ActivityPage 42 — 90 passed, 0 failed.

My stated gate was that landing must be pinned on **both** projections, not just the expanded reply, and that removing the alias from either path alone must redden. Measured, each mutation applied to a verified single anchor and reverted:

| mutation | result |
|---|---|
| root caller `V2ThreadMessages:258` `sourceMessageId` → `undefined` | 1 red — *settled root keeps source and durable links on the same visible ruling* |
| reply caller `V2ThreadMessages:334` `sourceMessageId` → `undefined` | 1 red — *expanded settled source replies render one ruled answer…* |
| the alias attribute itself, `V2MessageRow:557` | 2 red — both of the above |
| `landOnMessage`'s alias-resolution branch disabled | 2 red — both of the above |

So the two projections are independently pinned and the shared resolution is pinned by both. The tests assert landing, not presence: `landOnMessage('21')`/`('30')` and `('22')`/`('31')` each return `true`, the reply case additionally asserts the landed element is the durable row (`v2-msg--landed` on `#message-22`), and both assert exactly one ruling row survives so the alias cannot become a second visible row. That satisfies the strengthening — actual source landing plus durable-answer landing.

Two non-blocking observations, both measured, neither a merge condition:

1. The alias is also emitted on the open decision-request card (`V2MessageRow:416`, the `v2-message-row--decision` branch). Removing it there changes nothing in any of the six suites — no ruling renders through that branch, so it is an inert addition rather than covered behaviour. Fine to keep for symmetry; just don't read the green suite as evidence that site works.
2. The root test pins that `landOnMessage('30')` succeeds and that one ruling row exists, but — unlike the reply test — it does not assert *which* element the source id landed on. Today only one element in that DOM carries an alias, so the two are equivalent; if a second aliased row ever renders in the root projection the assertion would stop discriminating. Copying the reply test's `toHaveClass('v2-msg--landed')` identity check into the root case would close that.

lilyshen0722 (commented):
**CHANGES REQUESTED at `0fc78caf800b46a9757b32dd7b23c0aae39d2b8d`** — this supersedes my approval of `673ce99f`, which did not cover the land-then-hydrate ordering ux-lead found. My tests there called `landOnMessage` after the settled projection was already rendered; the reported bug is the replacement happening *under* an already-landed row, and no assertion I wrote could see it.

Method: one detached checkout, `/tmp/gate-1599c`, at the SHA above; tree clean before and after every mutation; one suite per jest invocation.

### The new regression does pin the reported bug
Reverting **only** `V2Thread.tsx` to `673ce99f` while keeping the new test file: *preserves Activity landing when hydration replaces the focused source row* fails (1 failed / 8 passed). It exercises the real transition — focused pending card, 15s poll delivers settled history, durable `#message-43` replaces it — and asserts both the landed marker and `document.activeElement`. That is the right test.

### Blocking: the fix regresses `V2ThreadReveal`, 20/20 → 16/20

| tree | V2ThreadReveal |
|---|---|
| `0fc78caf` as pushed | **4 failed, 16 passed** |
| same tree, `V2Thread.tsx` reverted to `673ce99f` | 20 passed |

Failures, all on the reveal-vs-fetch invariant, all `expect(detail.loadOlder).not.toHaveBeenCalled()` receiving 1 call:
- a target folded inside a collapsed thread is revealed — the thread opens and NO history is fetched
- legacy `?message=` landing uses the same source reveal path
- canonical hash wins when a legacy query target is also present
- quoting the same folded reply again after collapse reopens it without fetching history

**Localized to the dependency array.** Reverting only `decisionByMessageId, settledDecisionByMessageId` from the deps at `:955`, keeping the entire new body: Reveal returns to 20/20 and the new hydration test goes red. So the two are in direct conflict rather than this being a typo — the effect *must* re-enter on decision-map identity to notice the replacement, and re-entering is what breaks reveal. Both maps are stable identities (`useMemo` on `decisions`; state), so the extra entry is the decision fetch settling, not render churn.

Reading the body for the mechanism (this part is read, not isolated): on re-entry the target is still absent from the DOM, `landOnMessage` fails, `revealTriedRef.current === target` from the first pass so the reveal branch is skipped, and control reaches `searchOlderForMessage`. Splitting the re-land into its own effect keyed on the decision maps — leaving the reveal/search effect's deps untouched — would satisfy both.

### Also missing: the deliberate-focus condition is untested
`focusedElsewhere` is claimed in the description ("does not steal focus after deliberate movement") and named as a gate condition, but no test covers it. I could not measure this cleanly because the suite is already red; it should land with the fix above. `landingShapeChanged` is likewise unexercised.

### Unchanged from the previous head
`V2ThreadDecisionCard` 9/9, `V2ThreadRestyle` 13/13, `V2ThreadHistory` 6/6 — the three suites cited in the description. `V2ThreadReveal` is the suite that owns the effect this diff rewrites, and it was not in that set.

lilyshen0722 (commented):
UX-GATE: focused hydration/focus PASS @ 0fc78caf800b46a9757b32dd7b23c0aae39d2b8d. Integration remains blocked by the separate reveal regression reported by Sprint Review (65517–65518).

Production-built this exact head; final GitHub head unchanged. At 1440×900 and true mobile 390×900, actual Activity Open lands on root source 2055 while settled history is held. Releasing history replaces it with durable ruling 2056, preserves source alias 2055, and transfers focus and landed highlighting to the one visible human ruling. If the reader first clicks/taps the composer and enters an unsent draft, hydration preserves textarea focus and the exact draft instead. Four cases passed; measured scroller position remained 0→0 in all four. Each retained one Sam fixture author, one ruled marker and one durable answer, with no pending card or duplicate.

All auth/API/socket data mocked, external requests blocked, static server without proxy; zero HTTP writes. Phone screenshot inspected. No live decision was created or answered. Evidence: `.dev/signal/walk-1599-0fc-focus.js`, `1599-0fc-focus-results.json`, and `1599-0fc-{1440,390}-{stay,moved}-after-held-hydration.png`.

Per operator scope, retain the prior 673ce99f passing reply/source/durable-link browser evidence and unrelated Activity evidence. This run only rechecks root hydration and deliberate focus movement; it is not a broad navigation or history revalidation. The 673→0fc delta changes V2Thread and its hydration regression test, with the shared message projections and styling unchanged.

The new reveal regression is a separate code gate and requires a corrected candidate. The composer Post in open-menu defect also remains a separate bounded follow-up: large real pod list, light menu, bounded internal scrolling, keyboard and mobile. Real source 64684→64685→64686 remains for the operator’s post-deploy verification.

lilyshen0722 (commented):
**APPROVED at `ba579d56e6547ba4cab486fd10e330a5a45b3260`** — the `V2ThreadReveal` regression I filed against `0fc78caf` is resolved, and it is resolved at the cause rather than around it.

Method: one detached checkout, `/tmp/gate-1599d`, at the SHA above; `git status --porcelain` empty before and after every mutation; one suite per jest invocation.

### The block is cleared
The reveal-vs-fetch effect's dependency array is now byte-identical to `673ce99f`:
```
}, [landingTarget, initialLoadComplete, messages, threadView, revealTick, loadingOlder, loading, historySearch.targetId, historySearch.status, searchOlderForMessage]);
```
and the recovery moved into its own effect keyed on `[landingTarget, initialLoadComplete, messages, decisionByMessageId, settledDecisionByMessageId]`. `V2ThreadReveal` is back to 20/20.

### The hydration fix still pins the bug
Reverting **only** `V2Thread.tsx` to `673ce99f` in this tree: `V2ThreadDecisionCard` 1 failed / 8 passed (*preserves Activity landing when hydration replaces the focused source row*), while `V2ThreadReveal` stays 20/20. So the two behaviours are now independent — which is the whole point of the split.

Control: DecisionCard 9, Restyle 13, History 6, Reveal 20, MessageRow 4, ActivityPage 42 — **94 passed, 0 failed**.

### Guard-level mutations

| mutation | result |
|---|---|
| `landingShapeChanged := false` | 1 red (hydration test) — load-bearing |
| `focusedElsewhere := false` | **0 red across all four thread suites** |

### Non-blocking: `focusedElsewhere` cannot be shown load-bearing
"Does not steal focus after deliberate movement" is a stated requirement, so I checked it directly rather than reporting the surviving mutation as missing coverage. I wrote a throwaway probe — land on the source, focus an unrelated element, advance the poll so hydration replaces the row, assert `document.activeElement` is unchanged.

- At HEAD the probe **passes**: the requirement genuinely holds.
- Under `focusedElsewhere := false` the probe **also passes**, with identical instrumentation (`active=elsewhere, 43landed=false, 42present=false`).

So the surviving mutation is *not* missing coverage — it is no observable difference. The protection comes from the `!landingNodeReplaced && !landingShapeChanged` early return, not from `focusedElsewhere`; and `landingNodeReplaced` is itself gated on `activeElement === document.body`, which cannot hold at the same time as "the reader focused something else". I could not construct a state where `focusedElsewhere` is the deciding term.

Not a merge condition — the user-facing behaviour is correct and I verified it. But the guard currently reads as protection that is doing nothing, which is the kind of line a later reader trusts. Worth either deleting it or landing the case that needs it.

lilyshen0722 (commented):
**APPROVED at `7c549726647196cd283671d6dc3f9207f0d93554`** — final receipt. Carries forward the `ba579d56` approval; the two commits since are the deliberate-focus test and the guard reconciliation that removes `focusedElsewhere`.

### Full frontend suite, from a clean detached checkout at this SHA
```
Test Suites: 102 passed, 102 total
Tests:       776 passed, 776 total
```
Run `--runInBand`, whole suite rather than a focused subset — focused subsets are what let the Reveal regression through two heads ago.

### The `focusedElsewhere` removal is confirmed inert
I recommended deleting it on the evidence that mutating it to `false` changed nothing in any suite, including the test written specifically for it. `7c549726` deletes it, and the full suite is green at 776 — the same 776 as at `f0effc41` with the guard still present. The prediction and the outcome agree, which is the only reason I'll restate it: the protection was always the `!landingNodeReplaced && !landingShapeChanged` early return, and the new comment now says so accurately.

`landingShapeChanged` remains load-bearing — mutating it to `false` reddens the hydration regression.

### State of the gate across the series
- alias landing on both root and reply projections — pinned, each caller independently (`673ce99f`)
- one visible ruling, durable row rather than the synthetic stub — pinned (`673ce99f`)
- landing survives hydration replacing the focused source row — pinned (`0fc78caf` onward)
- reveal-vs-fetch not regressed by the above — restored and pinned (`ba579d56`)
- deliberate focus movement respected — verified by direct probe; characterized by test, though that test would pass without the guard it was written for

No outstanding blockers from my side.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 7c549726647196cd283671d6dc3f9207f0d93554.

Production-built and browser-tested this exact head. Final GitHub head is unchanged. The outstanding reveal-versus-hydration conflict passes at 1440×900 and true mobile 390×900.

- Root hydration, four cases: actual Activity Open lands on source 2055 while settled history is held. Releasing history transfers focus and landed treatment to durable ruling 2056 with source alias 2055. Deliberate composer click/tap and an unsent draft before hydration preserve textarea focus and the exact draft. Each case retains one human author, one ruled marker and one durable answer, with no pending card or duplicate. Measured scroller position stays 0→0.
- Loaded reply reveal, six cases: canonical hash, legacy query and canonical-hash precedence each open the collapsed thread and reveal an early reply outside the newest-eight window. Exact target 2002 is focused, marked and inside the transcript viewport. A further two-second observation preserves focus and position. Zero older-history requests in every case despite a full loaded page and more history available.
- Full frontend suite on this exact head: 102/102 suites, 776/776 Jest tests, plus 3/3 Node SEO/routing tests. Production build passed. Tracked checkout clean.

Browser data is synthetic, topology-faithful and fully mocked: auth/API/socket interception, external requests blocked, static server without proxy, zero HTTP writes. Desktop Enter and actual mobile taps used for Activity Open and deliberate focus movement. Phone reveal and composer-focus screenshots visually inspected.

Per operator direction, retain prior passing alias/identity/durable-action, reply/source/durable-link and unrelated Activity evidence. These are explicit scoped carries; no broad history/send sweep was repeated. The composer Post in open-menu defect remains a separate follow-up, and the genuine 64684→64685→64686 post-deploy check remains with the operator.

Evidence: `.dev/signal/walk-1599-7c-{focus,reveal}.js`, `1599-7c-{focus,reveal}-results.json`, `1599-7c-full-frontend.log`, and matching phone screenshots. Predecessor ba579d56 also passed hydration and its full suite. Its first reveal run recorded one offscreen phone geometry sample; this did not recur in three phone confirmations, the complete six-case ba rerun, or the complete exact-7c run. That initial observation remains undiagnosed, rather than being attributed to this change or erased. One diagnostic wait incorrectly required the temporary landed class after its timeout; the subsequent position observation uses the stable message ID.

PR #1600, opened 2026-09-07
fix(activity): restore bounded post destination picker

lilyshen0722:
Activity’s Post in menu inherited dark button styling and rendered every pod in an unbounded list. With 108 pods, the live menu was 3,322px tall and extended offscreen. Use the existing light Signal treatment and a 240px internal scroller, while preserving the selected destination and unsent draft.

Support Arrow/Home/End selection, Escape with focus retained or restored to the trigger, and outside mouse/touch dismissal. Keep the existing top-level scope menu and Send button styling intact.

Validation: freshly built integrated head334cdb74 passed the actual open-menu checks with120 pods at1440×900 and390×844, including last-option reach, both Escape paths, outside dismissal, selection and draft preservation. Phone menu bottom528 remained inside the844px viewport. Independent code review confirmed the style cascade and dismissal behavior. The final a5470294 delta adds only a trigger-Escape regression; UI code is identical. Relevant tests, full frontend102 suites/779 tests plus3 Node tests passed at334; exact final-head CI gates merge. Live verification follows deployment.

This branch includes source-thread fix #1599 from main(eab498a3).

### DIFF
--- a/frontend/src/v2/__tests__/V2ActivityPage.test.tsx
+++ b/frontend/src/v2/__tests__/V2ActivityPage.test.tsx
@@ -1094,6 +1094,20 @@ describe('V2ActivityPage', () => {
     expect(picker).toHaveFocus();
   });
 
+  test('keeps the trigger focused when Escape closes an open destination menu', async () => {
+    const activityRecap = { ...recap, pods: [...recap.pods, { id: 'pod-2', name: 'GTM Programs' }] };
+    mockGet.mockImplementation((url: string) => Promise.resolve({ data: url === '/api/activity/decision-queue' ? decisionQueue : activityRecap }));
+    renderPage();
+    await screen.findByRole('heading', { name: 'Tell your agents' });
+    const picker = document.querySelector<HTMLButtonElement>('.v2-activity__compose-picker-button');
+    expect(picker).not.toBeNull();
+    picker.focus();
+    fireEvent.click(picker);
+    fireEvent.keyDown(picker, { key: 'Escape' });
+    expect(screen.queryByRole('listbox')).not.toBeInTheDocument();
+    expect(picker).toHaveFocus();
+  });
+
   test('closes the destination menu on an outside mouse or touch press', async () => {
     const activityRecap = { ...recap, pods: [...recap.pods, { id: 'pod-2', name: 'GTM Programs' }] };
     mockGet.mockImplementation((url: string) => Promise.resolve({ data: url === '/api/activity/decision-queue' ? decisionQueue : activityRecap }));
--- a/frontend/src/v2/__tests__/V2ActivityPage.test.tsx
+++ b/frontend/src/v2/__tests__/V2ActivityPage.test.tsx
@@ -1079,6 +1079,38 @@ describe('V2ActivityPage', () => {
     expect(compose).toHaveValue('Keep this draft');
   });
 
+  test('closes the destination menu on Escape without blurring its trigger', async () => {
+    const activityRecap = { ...recap, pods: [...recap.pods, { id: 'pod-2', name: 'GTM Programs' }] };
+    mockGet.mockImplementation((url: string) => Promise.resolve({ data: url === '/api/activity/decision-queue' ? decisionQueue : activityRecap }));
+    renderPage();
+    await screen.findByRole('heading', { name: 'Tell your agents' });
+    const picker = document.querySelector<HTMLButtonElement>('.v2-activity__compose-picker-button');
+    expect(picker).not.toBeNull();
+    fireEvent.click(picker);
+    const option = within(screen.getByRole('listbox')).getByRole('option', { name: 'GTM Programs' });
+    option.focus();
+    fireEvent.keyDown(option, { key: 'Escape' });
+    expect(screen.queryByRole('listbox')).not.toBeInTheDocument();
+    expect(picker).toHaveFocus();
+  });
+
+  test('closes the destination menu on an outside mouse or touch press', async () => {
+    const activityRecap = { ...recap, pods: [...recap.pods, { id: 'pod-2', name: 'GTM Programs' }] };
+    mockGet.mockImplementation((url: string) => Promise.resolve({ data: url === '/api/activity/decision-queue' ? decisionQueue : activityRecap }));
+    renderPage();
+    await screen.findByRole('heading', { name: 'Tell your agents' });
+    const picker = document.querySelector<HTMLButtonElement>('.v2-activity__compose-picker-button');
+    expect(picker).not.toBeNull();
+    fireEvent.click(picker);
+    expect(screen.getByRole('listbox')).toBeInTheDocument();
+    fireEvent.mouseDown(document.body);
+    expect(screen.queryByRole('listbox')).not.toBeInTheDocument();
+    fireEvent.click(picker);
+    expect(screen.getByRole('listbox')).toBeInTheDocument();
+    fireEvent.touchStart(document.body);
+    expect(screen.queryByRole('listbox')).not.toBeInTheDocument();
+  });
+
   test('keeps an overflow pod selection visible and returns focus to its trigger', async () => {
     const extraPods = [2, 3, 4].map((id) => ({ id: `pod-${id}`, name: `Pod ${id}` }));
     mockGet.mockImplementation((url: string) => Promise.resolve({ data: url === '/api/activity/decision-queue' ? decisionQueue : { ...recap, pods: [...recap.pods, ...extraPods] } }));
--- a/frontend/src/v2/__tests__/V2ActivityPage.test.tsx
+++ b/frontend/src/v2/__tests__/V2ActivityPage.test.tsx
@@ -1060,6 +1060,25 @@ describe('V2ActivityPage', () => {
     await waitFor(() => expect(picker).toHaveFocus());
   });
 
+  test('supports keyboard destination selection without losing the draft', async () => {
+    const activityRecap = { ...recap, pods: [...recap.pods, { id: 'pod-2', name: 'GTM Programs' }] };
+    mockGet.mockImplementation((url: string) => Promise.resolve({ data: url === '/api/activity/decision-queue' ? decisionQueue : activityRecap }));
+    renderPage();
+    const compose = await screen.findByRole('textbox', { name: i18n.t('activity.compose.placeholder') });
+    fireEvent.change(compose, { target: { value: 'Keep this draft' } });
+    const picker = document.querySelector<HTMLButtonElement>('.v2-activity__compose-picker-button');
+    expect(picker).not.toBeNull();
+    fireEvent.keyDown(picker, { key: 'ArrowDown' });
+    const options = within(screen.getByRole('listbox')).getAllByRole('option');
+    await waitFor(() => expect(options[0]).toHaveFocus());
+    fireEvent.keyDown(options[0], { key: 'ArrowDown' });
+    await waitFor(() => expect(options[1]).toHaveFocus());
+    fireEvent.keyDown(options[1], { key: 'Enter' });
+    await waitFor(() => expect(picker).toHaveFocus());
+    expect(picker).toHaveTextContent('GTM Programs');
+    expect(compose).toHaveValue('Keep this draft');
+  });
+
   test('keeps an overflow pod selection visible and returns focus to its trigger', async () => {
     const extraPods = [2, 3, 4].map((id) => ({ id: `pod-${id}`, name: `Pod ${id}` }));
     mockGet.mockImplementation((url: string) => Promise.resolve({ data: url === '/api/activity/decision-queue' ? decisionQueue : { ...recap, pods: [...recap.pods, ...extraPods] } }));
--- a/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
+++ b/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
@@ -404,6 +404,12 @@ describe('v2 layout invariants (CSS rule presence)', () => {
     const borderedAction = ruleBody(v2, '.v2-root .v2-activity__queue-actions button.v2-activity__queue-action--bordered');
     expect(borderedAction).toContain('border: 1px solid var(--v2-border)');
     expect(ruleBody(v2, '.v2-root .v2-activity__compose .v2-activity__compose-picker-button')).toContain('border: 1px solid var(--v2-border)');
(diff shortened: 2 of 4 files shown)


lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ 02edefebcb61e59b2c0dc3347e2cc1c131018ad6

The production-built integrated candidate loses keyboard focus when Escape closes the composer Post in menu. At both 1440×900 and true mobile 390×900, ArrowDown/End/Enter selects Destination 120; reopening and pressing Escape closes the menu but leaves document.activeElement as BODY. V2ActivityPage.tsx:1003–1006 restores the trigger synchronously without stopping propagation; the adjacent scope menu explicitly stops Escape propagation to protect its focus from the global handler. Fix both trigger/option Escape paths and cover the actual shell interaction.

Outside click/tap on the Activity heading also leaves the listbox open at both widths. This is observed dismissal behavior, not a claim that this PR introduced it. Preserve the chosen destination and draft on dismissal.

Cleared with 120 synthetic pods: white menu, transparent/tint options with ink text, height 240px, overflow-y auto, scrollHeight 3608px/clientHeight 238px. Menu bounds desktop x940/y134/right1120/bottom374; phone x183/y288/right363/bottom528. End reaches focused Destination 120 with scrollTop3370 and the option entirely inside the menu. Arrow/Home/End navigation, keyboard and click/tap selection, exact unsent draft and destination preservation pass. Phone screenshot visually inspected. No HTTP writes; auth/API/socket fully mocked, external requests blocked, local static server has no proxy. This is actual production UI with synthetic data, not live Sam data.

Initial harness samples raced requestAnimationFrame focus; added settle intervals. A phone outside-textarea tap hit the overlapping menu instead; final outside dismissal uses the unobscured Activity heading. Final completed two-width run reports the Escape and outside-dismissal findings. No application edits. Build passed; tracked checkout clean. Existing source-card/alias/identity/action evidence is unchanged by this four-file dropdown delta; no broad review sweep.

Evidence: walk-1600-02-picker.js, 1600-02-picker-results.json, 1600-02-{1440,390}-{open,last}.png.

lilyshen0722 (commented):
**APPROVED on the code side at `02edefebcb61e59b2c0dc3347e2cc1c131018ad6`.** Browser/UX acceptance is ux-lead's; one conditional item for their pass is at the end.

The head moved from `ea060d04` to `02edefeb` while I was reviewing — the new commit is the merge of `main` (`eab498a3`, #1599). That is the same merge I had already performed and tested locally, and the two are **byte-identical**: my merge commit and `02edefeb` share tree `da76339e9c4f03df8664378a51b3985e1efb1668`, `git diff --name-only` between them is 0 files. Neither `v2.css` nor `V2ActivityPage.tsx` is touched by the merge, so the measurements below transfer exactly rather than by inference.

Method: one detached checkout, `/tmp/gate-1600`; clean before and after every mutation.

### The cascade fix is proven in a real engine, not only by the presence test
The added `v2-layout-invariants` assertions are string presence — they cannot see the cascade, and would pass even if the ink rule still won. I served the actual `v2.css` over http and read computed styles from Chromium, with the pre-fix sheet as a positive control:

| | option `background` | `color` | menu `max-height` / `overflow-y` |
|---|---|---|---|
| `713e1f6c` (pre-fix) | `rgb(16, 24, 40)` | `rgb(255,255,255)` | `none` / `visible` |
| this head | `rgba(0,0,0,0)` | `rgb(16,24,40)` | `240px` / `auto` |

The control reproduces the reported defect exactly, so the instrument discriminates. With 108 options the menu is `scrollHeight 3248` clamped into a scrolling 240px box, matching the reported 3322px wall.

Mechanism is as the new comment states: `.v2-root .v2-activity__compose button` is (0,2,1) and beat the old (0,2,0) option rule; the new (0,3,0) selector wins. **Send is still `rgb(16,24,40)`** — the override did not over-reach, which is the failure mode a stronger selector invites. Sheet integrity: 1600 parsed `cssRules`, last rule a well-formed `@media` — no truncation from the insert.

### Mutations (single verified anchor each, reverted)

| mutation | result |
|---|---|
| trigger `ArrowDown`/`ArrowUp` handler removed | 1 red |
| option `Enter`/`Space` select removed | 1 red |
| new CSS option rule renamed out of the cascade | 1 red (layout invariants) |

### Interaction with main
Full frontend suite on this exact tree: **102 suites, 777 tests, all passing** — 776 on main plus the one new keyboard test.

### For the live pass — a question, not a finding
`max-height: 240px` is a fixed bound, not viewport-relative, and the menu is absolutely positioned at `top: calc(100% + 4px)` with no flip-up. In my harness, with the composer docked near the bottom, `menu.getBoundingClientRect().bottom` landed **187px past the viewport**. That arrangement is my own construction, not necessarily the real Activity layout, so it may not reproduce — but "bounded to 240px" and "bounded to the viewport" are different guarantees and the acceptance criterion asks for the second. Worth measuring `bottom` against `innerHeight` at 390×844 with the composer where it actually sits.

Two non-blocking notes: `composePickerOptionRefs` is never pruned, so entries for removed pods accumulate for the page's lifetime; and the first `ArrowDown` from a closed menu opens it and focuses inside a single `requestAnimationFrame`, which holds in jsdom — worth confirming one frame is enough after the state commit in a real browser.

lilyshen0722:
Follow-up on integrated head: 334cdb74 adds Escape propagation guards and closes the Post destination picker on outside mouse/touch press, preserving trigger focus. Focused Activity/layout tests: 149/149; typecheck and targeted ESLint pass. Base-freshness merge from main is retained; CI is rerunning.

lilyshen0722 (commented):
**Correcting my own approval at `02edefebcb61e59b2c0dc3347e2cc1c131018ad6`.** Dismissal was on the acceptance list and I did not test it. ux-lead's findings are real, and I have now reproduced both **at the unit tier**, which means they can be pinned by jest rather than only by a live pass.

Measured in jsdom on this head (probes written into the existing `V2ActivityPage` suite, then removed; tree clean):

| path | result |
|---|---|
| open picker, focus trigger, press `Escape` | menu closes, `document.activeElement` is **BODY** |
| arrow into an option, press `Escape` | menu closes, `document.activeElement` is **BODY** |
| open picker, click outside | **menu stays open** |

### A caution on the proposed mechanism
The report attributes the focus loss to the handler lacking `event.stopPropagation()`, by analogy with the scope menu, whose Escape handler carries `stopPropagation()` and the comment *"This menu restores focus; the global Escape handler blurs it."*

That reading is not supported by what I measured:

- It reproduces in **jsdom**, where no such global listener is registered.
- There is no `blur()` call anywhere in `src/v2` outside test files.
- I could not locate a global `keydown`/`Escape` handler in the v2 tree at all. The five files registering a global `keydown` are `V2PodsSidebar`, `V2Lightbox`, `V2FirstRunHero`, `V2Composer`, `V2LangSwitch`; none is the described blur-on-Escape handler.

And on the trigger path specifically, instrumented immediately before the key event: focus **is** on the trigger, the button is **not** disabled, and after the close its DOM node is **reused and still in the document** — so this is not the node-replacement mechanism from #1599 either. Something blurs a focused, mounted, enabled element, and the cause is not established.

Separately, the trigger's Escape handler at `:963` closes the menu without any `focus()` call, unlike the option handler at `:1003` and unlike the scope menu — so an explicit refocus is likely needed on that path regardless of what the blur turns out to be.

**Ask:** whatever lands, verify it against the three rows above rather than against the propagation theory. All three are cheap jest assertions and should ship as regressions with the fix — a live-only check is what let this through my review.

The cascade, keyboard-navigation, and main-integration findings in my previous comment are unaffected and still hold.

lilyshen0722:
Final dismissal/test pin: a5470294 (on 02edefeb, preserving eab498a3 merge) adds explicit trigger-Escape focus coverage. Trigger Escape, option Escape, and outside mouse/touch dismissal now pass; focused Activity/layout is 150/150. Typecheck/build/targeted lint pass. Awaiting CI and final 1440/390 viewport review.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 334cdb745635577bf6643e3d76aef75cb9837a30

Both dismissal blockers from02edefeb are cleared on this freshly production-built integrated head. In the actual Activity page at1440×900 and true mobile390×844, Escape from a focused option restores the Post in trigger; Escape from the trigger keeps its focus. Outside mouse click / phone tap on the unobscured Activity heading closes the listbox. Exact unsent draft and chosen destination survive all dismissal paths.

The actual OPEN menu contains120 synthetic pods: white surface, transparent/tint options with ink text; height240px, scrollHeight3608/clientHeight238, overflow-y auto. Desktop menu x940/y134/right1120/bottom374; phone x183/y288/right363/bottom528 inside844px viewport. No artificial low-composer positioning. Last option120 is visible entirely inside the menu after End and after wheel scrolling (scrollTop3370), and can be selected by Enter or actual click/phone tap. ArrowUp/Down and Home/End navigation pass. Selection closes the menu and preserves the draft. Top Pod scope more-menu remains white,240px/auto; its Escape still dismisses. Phone last-option screenshot visually inspected.

Full frontend suite directly on334cdb74:102 suites/779 Jest tests plus3 Node SEO/routing tests PASS. Production build PASS, tracked checkout clean. Final GitHub head verified unchanged. Browser run completed first try after the earlier02 harness corrections; both widths return pass:true and zero HTTP writes. Auth/API/socket mocked, external traffic blocked, static server has no API proxy. Actual application UI with synthetic120-pod data; not a live deployment walk. Phone selection/dismissal used touch, menu scroll probe used wheel.

02→334 changes only V2ActivityPage21 implementation lines and32 test lines; the eab498a3 main merge remains. Existing source-card/alias/identity/action and live1599 evidence carries unchanged. No unrelated review sweep or application edits.

Evidence:walk-1600-334-picker.js,1600-334-picker-results.json,1600-334-{1440,390}-{open,last}.png,1600-334-full-frontend.log.

lilyshen0722 (commented):
UX-GATE: APPROVED @ a547029468db3e050c27d62a39390d2132bfeb54 — explicit browser carry from334cdb745635577bf6643e3d76aef75cb9837a30.

The head advanced while my prior review was filed: its body identifies334cdb74 accurately, but GitHub attached it to a5470294. I fetched and inspected the entire334→a547 delta: only14 added lines in V2ActivityPage.test.tsx, adding the focused-trigger Escape regression. Zero application or stylesheet changes. Therefore the freshly production-built334 browser evidence at1440×900 and390×844 carries unchanged to a547:120-option OPEN menu styling/bounds/internal scroll/last-option reach, keyboard and click/tap selection, both Escape paths, outside dismissal, exact draft/destination preservation, and unchanged top scope menu all PASS.

The direct full frontend run remains334:102 suites/779 Jest tests plus3 Node SEO/routing tests. I did not build or rerun tests at a547; this is a stated zero-runtime-delta browser carry, not a new exact-head test claim. Own test contexts/server/worktree were cleaned up. The global Escape blur handler is in frontend/src/utils/focusUtils.ts:7–11, initialized by frontend/src/App.tsx:248; it exists outside src/v2.

lilyshen0722 (commented):
**Retraction, and the dismissal gate cleared at `a547029468db3e050c27d62a39390d2132bfeb54`.**

### I was wrong about the mechanism — ux-lead was right
My earlier comment argued the `stopPropagation()` reading was unsupported, because I could find no global `keydown` listener in the v2 tree and no `blur()` outside tests. That negative came from an instrument that could not see what it was looking for: I grepped for **native** `addEventListener('keydown')` and for `blur()`, and neither can detect a **React synthetic** handler on an ancestor.

Isolated by mutation — removing only the `stopPropagation()` calls, leaving the outside-press effect in place:

| row | with | without |
|---|---|---|
| Escape from trigger | picker button focused | **BODY** |
| Escape from option | picker button focused | **BODY** |
| outside press | closes | closes |

`stopPropagation()` is load-bearing on both Escape paths, and the outside-press effect is independent of it. A propagating handler demonstrably exists; I never located it by name, and the mutation settles the question without that.

### Both fixes are pinned by shipped tests, not just working

| mutation (single verified anchor set, reverted) | result |
|---|---|
| drop `stopPropagation()` from all three Escape sites (incl. the scope menu's pre-existing one) | 2 red |
| drop the outside-press `mousedown`/`touchstart` registration | 1 red |

All three rows I filed now pass at this head, and they ship as suite assertions rather than as a live-only check — which was the ask, since a live-only check is what let this past my first review.

Full frontend suite at this head: **102 suites, 780 tests, all passing.**

### A process note against myself
My previous comment stamped `334cdb74` because my pre-filing head check read a **stale local ref** instead of re-fetching, so it attached to a commit I had not measured. Its claims happened to survive there, but that was luck rather than the guard working — and the guard exists precisely so a verdict cannot land on an unmeasured head. Both this comment and the head check above re-fetch first.

The cascade, keyboard-navigation and main-integration findings from my first comment are unaffected. No outstanding code-side blockers from me at this head. The viewport-vs-240px question in that comment remains open for the live pass.

lilyshen0722 (commented):
**Correcting my own correction.** In my previous comment I explained my false negative as native-vs-synthetic: that I grepped for native `addEventListener('keydown')` and `blur()`, which "cannot detect a React synthetic handler on an ancestor."

That diagnosis is wrong. sprint-impl named the handler and I read it — `frontend/src/utils/focusUtils.ts:8`:

```ts
export const setupFocusManagement = (): void => {
  document.addEventListener('keydown', (event: KeyboardEvent) => {
    if (event.key === 'Escape') {
      blurActiveElement();
    }
  });
```

It is a plain **native** document listener calling **`blurActiveElement()`** — precisely the two constructs I grepped for. My search construct was correct; my **scope** was wrong. I searched `src/v2/` because that was the subtree under review, and the handler lives one directory up in `src/utils/`. Both greps would have found it immediately at repo scope.

That is the lesson worth recording, and it is a sharper one than what I filed: when reasoning about what reaches a component, the search scope has to be the whole app, because propagation does not respect the directory you happen to be reviewing. The earlier framing would have taught the next reader to distrust grep rather than to widen it.

### A related gap, non-blocking and outside this PR's scope
The same file's `mousedown` handler dismisses only `[role="menu"]` and `[role="dialog"]`. A `role="listbox"` popup was never covered by it — which is exactly why this picker needed its own outside-press effect rather than inheriting one. Two other `role="listbox"` popups exist in the app: `V2Composer`'s mention dropdown (`:178`) and `V2LangSwitch` (`:56`). Both plausibly carry the identical Escape-blur and outside-press gaps. I have not measured either, and neither belongs in this PR — flagging it so the pattern is visible rather than rediscovered a third time.

Nothing here changes the clearance: the three dismissal rows pass at this head and both fixes are pinned by mutation.

PR #1601, opened 2026-09-07
Keep Activity destination options readable and usable on phones

lilyshen0722:
Activity destination options inherited the composer's dark hover fill, leaving dark labels unreadable. Their 30px desktop row rule also overrode the existing 44px phone control minimum.

Scope the option hover treatment so it keeps the existing light Signal surface, and restore 44px option targets at widths up to 640px while preserving 30px desktop rows. Send styling, selection, draft retention and the bounded scrolling menu are unchanged.

Validation: 102 frontend suites / 780 tests plus three SEO checks reported green by the implementation owner. Independent Chromium measurements on c08be6af confirm 44px phone targets, 30px desktop targets and 16.11:1 hover/selected-hover contrast. UX independently verified the production-built UI with 120 destinations at 390px and 1440px: all phone options measure 44px, the last option remains reachable within the 240px menu, five interaction states remain readable, and dismissal preserves the draft. CI passed, including E2E and service tests against real databases. Live verification follows deployment.

### DIFF
--- a/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
+++ b/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
@@ -408,12 +408,14 @@ describe('v2 layout invariants (CSS rule presence)', () => {
     expect(composeMenu).toContain('max-height: 240px');
     expect(composeMenu).toContain('overflow-y: auto');
     const composeOption = ruleBody(v2, '.v2-root .v2-activity__compose .v2-activity__compose-picker-option');
+    expect(composeOption).toContain('min-height: 30px');
     expect(composeOption).toContain('background: transparent');
     expect(composeOption).toContain('color: var(--v2-text-primary)');
     const composeOptionHover = ruleBody(v2, '.v2-root .v2-activity__compose .v2-activity__compose-picker-menu button.v2-activity__compose-picker-option:hover:not(:disabled)');
     expect(composeOptionHover).toContain('background: var(--v2-surface-hover)');
     expect(composeOptionHover).toContain('color: var(--v2-text-primary)');
     expect(composeOptionHover).toContain('border-color: transparent');
+    expect(v2).toMatch(/@media \(max-width: 640px\) \{[\s\S]*?\.v2-root \.v2-activity__compose \.v2-activity__compose-picker-menu \.v2-activity__compose-picker-option \{ min-height: 44px; \}/);
     const otherOption = ruleBody(v2, '.v2-root .v2-activity__queue-actions button.v2-activity__option.v2-activity__queue-action--secondary');
     expect(otherOption).toContain('color: var(--v2-accent-text)');
     expect(v2).toContain('.v2-activity__option-description');
--- a/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
+++ b/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
@@ -410,6 +410,10 @@ describe('v2 layout invariants (CSS rule presence)', () => {
     const composeOption = ruleBody(v2, '.v2-root .v2-activity__compose .v2-activity__compose-picker-option');
     expect(composeOption).toContain('background: transparent');
     expect(composeOption).toContain('color: var(--v2-text-primary)');
+    const composeOptionHover = ruleBody(v2, '.v2-root .v2-activity__compose .v2-activity__compose-picker-menu button.v2-activity__compose-picker-option:hover:not(:disabled)');
+    expect(composeOptionHover).toContain('background: var(--v2-surface-hover)');
+    expect(composeOptionHover).toContain('color: var(--v2-text-primary)');
+    expect(composeOptionHover).toContain('border-color: transparent');
     const otherOption = ruleBody(v2, '.v2-root .v2-activity__queue-actions button.v2-activity__option.v2-activity__queue-action--secondary');
     expect(otherOption).toContain('color: var(--v2-accent-text)');
     expect(v2).toContain('.v2-activity__option-description');
--- a/frontend/src/v2/v2.css
+++ b/frontend/src/v2/v2.css
@@ -8940,8 +8940,17 @@ body.modern-ui.v2-canvas {
   font-weight: 500;
 }
 
-.v2-root .v2-activity__compose .v2-activity__compose-picker-option:hover,
 .v2-root .v2-activity__compose .v2-activity__compose-picker-option.is-active {
+  border-color: transparent;
+  background: var(--v2-surface-hover);
+  color: var(--v2-text-primary);
+}
+
+/* The generic composer hover rule includes :not(:disabled) and a button
+   element, so it outranks the picker option's class-only hover selector.
+   Match that specificity here to keep the light Signal option treatment in
+   the real hovered and selected+hovered states without changing Send. */
+.v2-root .v2-activity__compose .v2-activity__compose-picker-menu button.v2-activity__compose-picker-option:hover:not(:disabled) {
   border-color: transparent;
   background: var(--v2-surface-hover);
   color: var(--v2-text-primary);
--- a/frontend/src/v2/v2.css
+++ b/frontend/src/v2/v2.css
@@ -9541,6 +9541,10 @@ body.modern-ui.v2-canvas {
   .v2-activity__compose-picker { flex: 1 1 auto; min-width: 0; width: 100%; }
   .v2-root .v2-activity__compose-picker-button { width: 100%; max-width: none; text-align: left; }
   .v2-root .v2-activity__compose button { min-height: 44px; padding: 0 18px; }
+  /* The picker option rule is more specific than the generic composer
+     button floor above; repeat the menu scope so every option keeps a 44px
+     touch target on phones. */
+  .v2-root .v2-activity__compose .v2-activity__compose-picker-menu .v2-activity__compose-picker-option { min-height: 44px; }
   .v2-activity__section-heading { flex-wrap: wrap; }
   .v2-activity__section-heading > p { flex-basis: 100%; }
   .v2-activity__agent-grid { grid-template-columns: minmax(0, 1fr); }


lilyshen0722 (commented):
**APPROVED at `04d4853187e44632079190b869d027e7ddc70afa`** — the requested state matrix, bound to this head.

Measured with real Chromium against the branch's `v2.css`, each state genuinely active (`hover()` / `focus()`, not class simulation), at 1440×900 and again at 390×844. Contrast ratios are WCAG relative-luminance, computed against the effective backdrop.

| state | background | text | contrast |
|---|---|---|---|
| default | transparent → white menu | `rgb(16,24,40)` | 17.75 |
| **hover** | `rgb(242,244,247)` | `rgb(16,24,40)` | **16.11** (was **1.21** on `c83d9f69`) |
| keyboard focus | transparent + focus ring | `rgb(16,24,40)` | 17.75 |
| selected | `rgb(242,244,247)` | `rgb(16,24,40)` | 16.11 |
| **selected+hover** | `rgb(242,244,247)` | `rgb(16,24,40)` | **16.11** (was **1.21**) |

Phone results are identical to desktop for both previously-broken states.

**Controls held.** Send is still `rgb(16,24,40)` on white and the picker trigger is still white — the stronger selector does not reach either, which was the stated constraint.

**Edge checked.** The rule carries `:not(:disabled)`, so I measured a disabled option: it computes transparent, not ink, because the generic composer hover carries `:not(:disabled)` as well. No fall-through.

The selector is `(0,6,1)` against the generic rule's `(0,4,1)`, so it wins on specificity rather than source order — confirmed by the computed values above rather than by arithmetic alone.

### Two notes on the guard, neither blocking

1. **The added invariant cannot see this defect class.** It is a string-presence assertion, and the broken sheet and the fixed sheet both contain the selector text. It will catch deletion of the rule; it cannot catch the rule losing the cascade, which is exactly how this bug arrived — the previous invariant passed while hover computed dark-on-dark. Only computed styles with the state active discriminate here.

2. **Do not assert a contrast ratio on the default state.** Its background is `transparent`, so a naive ratio computes against nothing and reads ~1.2 while the real backdrop is the white menu and the true ratio is 17.75. That number is a measurement artifact; treating it as a finding would send someone chasing a non-defect.

### On my own instrument
My first run of this matrix returned a uniform `rgb(239,239,239)` on black for every state — internally consistent and easy to write up as "all states agree." It was browser-default styling: the stylesheet had failed into a 0-byte file while the server still answered `200`. What exposed it was the Send control, included precisely because it must *not* match the options. I discarded that run and re-measured only after asserting the sheet parsed (1601 rules). Worth stating because the same silent failure would make any future contrast check look clean.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 04d4853187e44632079190b869d027e7ddc70afa — actual production-built UI, 120 synthetic destinations, Chrome 1440×900 and 390×844.

All five states were activated before reading computed styles: default, actual pointer hover, keyboard focus via ArrowDown, selected, and selected+hover. Both widths: foreground rgb(16,24,40); default/focus transparent over the measured white menu (17.75:1), selected/hover/selected+hover rgb(242,244,247) (16.11:1). Actual :hover, aria-selected and :focus-visible flags were checked. Keyboard focus retains its visible 3px rgba(29,63,209,.18) ring. Screenshots visually inspected. One phone sample occurred before deferred focus settled; the subsequent settled DOM sample and screenshot confirm focus on the intended option.

The 120-option menu remains 240px tall with internal auto scrolling (scrollHeight 3608/clientHeight 238), inside both viewports: desktop bottom374/right1120; phone bottom528/right363. Phone End selects Destination120; reopening and actual wheel scrolling reaches it inside the menu, and selected+hover stays light. Outside dismissal preserves the exact unsent draft.

Negative controls at both widths: enabled Send remains rgb(16,24,40) with white text, actual hover rgb(29,41,57) with white text; scope option remains white with rgb(71,84,103) in default and actual hover. The c83→04 delta changes only the picker hover selector/rule and four invariant-test lines. Prior c83 keyboard/Escape/outside/structure controls carry unchanged; this is a focused contrast gate, not a broad rerun.

Validation: own production build passes. Local mock HTTP server serves unmodified candidate assets and 120 synthetic pods; empty decision queues, no fake decisions. Actual local fixture login through UI; server logs contain only that mock login POST and no application writes. No proxy to live API, no message sending. Phone uses an actual 390×844 viewport with pointer/keyboard/wheel input, not hardware touch or touch-event emulation. This clears the candidate; deployed/live contrast verification remains pending. Writer's full frontend suite is relayed, not rerun by UX.

Evidence: .dev/signal/hover-04-results.json, hover-04-{1440,390}-{hover,focus,selected-hover}.png, hover-04-390-last-selected-hover.png, and hover-04-server.log.

lilyshen0722 (commented):
UX qualification @ 04d4853187e44632079190b869d027e7ddc70afa: five-state contrast remains APPROVED; mobile target acceptance is OPEN.

Sam65594's source-level question is confirmed by the existing actual production-browser measurements from this exact candidate at390×844 (hover-04-results.json), not a new browser rerun: Destination001 is x188/y293/170×30, ending323; Destination002 starts323 and ends353, also170×30. Destination120 is170×30 at y493–523 after wheel scrolling. Adjacent buttons have zero vertical gap; these are the actual option button bounds, not merely label heights. V2ActivityPage.tsx binds selection directly to each button's onClick.

Adjudication: restore the existing phone composer target floor of44px for these options at max-width640. The current .v2-root .v2-activity__compose .v2-activity__compose-picker-option min-height30px rule outranks the existing mobile .v2-root .v2-activity__compose button min-height44px rule. V3 requires usable phone targets; the intended existing phone rule is the appropriate pattern here. A desktop30px row does not justify shrinking a phone target. This is a bounded restoration, not a new design or a blanket accessibility-compliance claim.

The30px rule is already present in c83; this qualification does not attribute it to PR1601's hover correction. Prior approval was a focused contrast gate and missed adjudicating target height. Retain all five contrast states, Send/scope styles,240px internal menu bound, keyboard/Escape/outside/draft behavior. After the bounded option-height correction, remeasure actual phone button bounds and last-option reachability with108+ options; confirm desktop30px and the five active contrast states remain intact.

lilyshen0722 (commented):
**APPROVED at `c08be6af63ab7e00af34a5d25b9f17a790c84525`** — the responsive rule wins the cascade, measured rather than inferred.

Computed geometry and color from real Chromium against this head's `v2.css`, states genuinely active:

| | option min-height | option measured | trigger | Send | option gap |
|---|---|---|---|---|---|
| 1440×900 | 30px | 30px | 30px | 30px | 0px |
| 390×844 | **44px** | **44px** | 44px | 44px | 0px |

All four options take the floor, not just the first. Hover contrast carries at both widths — 16.11 for hover and for selected+hover — so the `04d48531` proof is not disturbed by the new rule.

### Why this needed checking rather than reading
The 44px floor already existed in this same media block as `.v2-root .v2-activity__compose button` **(0,2,1)**, and it already matched the option — it simply lost to `.v2-root .v2-activity__compose .v2-activity__compose-picker-option` **(0,4,0)**. A media query contributes no specificity, so restoring the floor by reusing the existing pattern verbatim does nothing at all. I measured that before this push:

| candidate at 390 | option height |
|---|---|
| `04d48531` | 30px |
| naive reuse of the existing `≤640` pattern | **still 30px** |
| selector above (0,4,0) inside the block | 44px, desktop preserved at 30px |

The shipped rule takes the third form, which is why it works.

This is the same defect class as the hover bug one commit earlier: a rule that exists, reads correctly, targets the right element, and is outranked. A diff review and a presence test both pass on the naive version. Only computed geometry separates them — worth remembering when the next `@media` floor is added.

### One factual note, not a finding
The gap between adjacent options remains 0px at both widths. The 44px requirement is about target size and is met; the touching-rows observation in the original report is unchanged. Flagging it so it is a UX decision rather than an oversight.

Instrument control: sheet parsed to 1601 rules, and the Send button retained its ink treatment throughout — the control that previously caught a silently empty stylesheet.

lilyshen0722 (commented):
UX-GATE: APPROVED @ c08be6af63ab7e00af34a5d25b9f17a790c84525 — phone target hold cleared; live verification remains pending.

Production-built this exact head and checked the actual UI in Chrome at 390×844 and 1440×900 using 120 synthetic destinations. All 120 phone option buttons measure 170×44px, computed min-height44px. Desktop options remain170×30px. Zero inter-row gap is acceptable for this approved44px pattern; the earlier gap observation established there was no spacing compensating for undersized30px buttons, not a separate request for gutters.

Phone menu remains240px tall at x183/y288, right363/bottom528. Internal overflow-y:auto, scrollHeight5288/clientHeight238. Keyboard End focuses the complete last option at y479–523; Enter selects it. Reopen plus actual wheel reaches that same full44px last option, scrollTop5050. Selected+hover remains readable. Outside pointer dismissal preserves Destination120 and the exact Unsent contrast verification draft.

Both widths freshly pass default, actual hover, keyboard focus, selected and selected+hover. Foreground rgb(16,24,40); default/focus transparent over measured white gives17.75:1, selected/hover tint rgb(242,244,247) gives16.11:1. Focus-visible state and3px ring verified; phone focus and final-option captures visually inspected. Previous Send/scope/other control clearance carries: full04→c08 delta is only four CSS lines for phone height and two invariant assertions.

Method: unmodified production bundle served by an isolated local Python HTTP fixture with empty decision queues and120 destinations, no API proxy. Existing browser fixture session used through normal UI; zero HTTP writes in this run. No Send or decision choice. True390 viewport with keyboard, pointer and wheel; no hardware-touch or touch-event claim. Draft cleared and verified; viewport reset and own tab closed.

Build PASS. Writer reports102 Jest suites/780 tests plus3 SEO checks PASS; that suite was not rerun by UX. Evidence: picker-c08-results.json, picker-c08-390-hover.png, picker-c08-390-focus.png, picker-c08-390-last.png, picker-c08-1440-selected-hover.png and picker-c08-server.log under .dev/signal. Live gate remains open until this correction is deployed and checked.

PR #1603, opened 2026-09-08
fix(activity): restore readable decision option layout

lilyshen0722:
Long decision options were squeezed beside the question into 150px columns, with recommendation text competing with labels. Activity alternatives also inherited a dark hover fill that made their dark text unreadable.

Stack the options below the question at the available card width, put Recommended on a separate line, and render full descriptions at 14/20 with accessible associations. Add a 12px inline inset to Activity choices and retain light hover surfaces for alternatives. Apply the same option grouping to thread cards while preserving authored order, cobalt primary treatment, choice handling and durable resolution.

Validation covers the real GTM cap question plus long three-option copy with the recommendation second at desktop and 390px phone widths. Independent review clears final 9a32b5d1 geometry, padding and settled hover contrast; the UX lead independently passed 48 production-build state checks across Activity/thread and desktop/phone. Final CI passed, including E2E and service tests against real databases. Tests guard the actual CSS rule bodies and description associations. Live verification follows deployment before Sam’s authorized choice is submitted.

### DIFF
--- a/frontend/src/v2/__tests__/V2ActivityPage.test.tsx
+++ b/frontend/src/v2/__tests__/V2ActivityPage.test.tsx
@@ -141,7 +141,8 @@ describe('V2ActivityPage', () => {
     const second = screen.getByRole('button', { name: 'Rule: Ship now (Recommended)' });
     expect(first).toHaveClass('v2-activity__option--primary');
     expect(second).not.toHaveClass('v2-activity__option--primary');
-    expect(second).toHaveTextContent('Recommended');
+    expect(second).not.toHaveTextContent('Recommended');
+    expect(within(second.parentElement as HTMLElement).getByText('Recommended')).toBeInTheDocument();
     expect(first).not.toHaveTextContent('Recommended');
     expect(first.compareDocumentPosition(second) & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
   });
@@ -152,17 +153,22 @@ describe('V2ActivityPage', () => {
       items: [{
         ...decisionQueue.items[1],
         title: 'Scope of the five-post daily X cap',
-        detail: 'Choose how the daily publication cap should apply across the GTM workspace.',
+        detail: '@Sam should we keep the five-post daily cap across every account post, or change it to count only squad-drafted sends?',
         options: [
           {
-            label: 'Keep the five-post cap for every workspace day',
-            description: 'Preserve the current limit so the publishing rhythm stays predictable for the team.',
+            label: 'Keep account-wide cap',
+            description: 'Recommended default: retain five total X posts per day, including casual replies. Simple to count and limits total volume; casual use reduces squad capacity.',
             recommended: false,
           },
           {
-            label: 'Raise the cap for approved launch windows',
-            description: 'Allow a bounded exception when a launch needs additional posts, with the existing audit trail.',
+            label: 'Change to squad-drafted only',
+            description: 'Change the current rule to five squad-drafted sends per day, excluding Sam’s casual replies. Preserves squad capacity but permits more than five total posts and requires identifying draft origin.',
             recommended: true,
+          },
+          {
+            label: 'Use a rolling seven-day cap for all sends',
+            description: 'Spread the same total volume across a longer window while retaining one auditable limit.',
+            recommended: false,
           },
         ],
       }],
@@ -171,13 +177,19 @@ describe('V2ActivityPage', () => {
     mockGet.mockImplementation((url: string) => Promise.resolve({ data: url === '/api/activity/decision-queue' ? longCopyQueue : recap }));
     renderPage();
 
-    const first = await screen.findByRole('button', { name: 'Rule: Keep the five-post cap for every workspace day' });
-    const second = screen.getByRole('button', { name: 'Rule: Raise the cap for approved launch windows (Recommended)' });
+    const first = await screen.findByRole('button', { name: 'Rule: Keep account-wide cap' });
+    const second = screen.getByRole('button', { name: 'Rule: Change to squad-drafted only (Recommended)' });
+    const third = screen.getByRole('button', { name: 'Rule: Use a rolling seven-day cap for all sends' });
     expect(first).toHaveClass('v2-activity__option--primary');
     expect(second).not.toHaveClass('v2-activity__option--primary');
+    expect(third).not.toHaveClass('v2-activity__option--primary');
     expect(first.compareDocumentPosition(second) & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
-    expect(screen.getByText('Preserve the current limit so the publishing rhythm stays predictable for the team.')).toBeInTheDocument();
-    expect(screen.getByText('Allow a bounded exception when a launch needs additional posts, with the existing audit trail.')).toBeInTheDocument();
+    expect(second.compareDocumentPosition(third) & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
+    expect(screen.getByText('Recommended default: retain five total X posts per day, including casual replies. Simple to count and limits total volume; casual use reduces squad capacity.')).toBeInTheDocument();
+    expect(screen.getByText('Change the current rule to five squad-drafted sends per day, excluding Sam’s casual replies. Preserves squad capacity but permits more than five total posts and requires identifying draft origin.')).toBeInTheDocument();
+    expect(second).not.toHaveTextContent('Recommended');
+    expect(within(second.parentElement as HTMLElement).getByText('Recommended')).toBeInTheDocument();
+    expect(second).toHaveAttribute('aria-describedby', expect.stringContaining('description'));
   });
 
   test('changes the read window and opens the source pod from a factual queue row', async () => {
--- a/frontend/src/v2/__tests__/V2ActivityPage.test.tsx
+++ b/frontend/src/v2/__tests__/V2ActivityPage.test.tsx
@@ -146,6 +146,40 @@ describe('V2ActivityPage', () => {
     expect(first.compareDocumentPosition(second) & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
   });
 
+  test('keeps long decision labels and descriptions readable when the recommendation is second', async () => {
+    const longCopyQueue = {
+      ...decisionQueue,
+      items: [{
+        ...decisionQueue.items[1],
+        title: 'Scope of the five-post daily X cap',
+        detail: 'Choose how the daily publication cap should apply across the GTM workspace.',
+        options: [
+          {
+            label: 'Keep the five-post cap for every workspace day',
+            description: 'Preserve the current limit so the publishing rhythm stays predictable for the team.',
+            recommended: false,
+          },
+          {
+            label: 'Raise the cap for approved launch windows',
+            description: 'Allow a bounded exception when a launch needs additional posts, with the existing audit trail.',
+            recommended: true,
+          },
+        ],
+      }],
+      count: 1,
+    };
+    mockGet.mockImplementation((url: string) => Promise.resolve({ data: url === '/api/activity/decision-queue' ? longCopyQueue : recap }));
+    renderPage();
+
+    const first = await screen.findByRole('button', { name: 'Rule: Keep the five-post cap for every workspace day' });
+    const second = screen.getByRole('button', { name: 'Rule: Raise the cap for approved launch windows (Recommended)' });
+    expect(first).toHaveClass('v2-activity__option--primary');
+    expect(second).not.toHaveClass('v2-activity__option--primary');
+    expect(first.compareDocumentPosition(second) & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
+    expect(screen.getByText('Preserve the current limit so the publishing rhythm stays predictable for the team.')).toBeInTheDocument();
+    expect(screen.getByText('Allow a bounded exception when a launch needs additional posts, with the existing audit trail.')).toBeInTheDocument();
+  });
+
   test('changes the read window and opens the source pod from a factual queue row', async () => {
     renderPage();
     await screen.findByText('Review requested');
--- a/frontend/src/v2/__tests__/V2DecisionCard.test.tsx
+++ b/frontend/src/v2/__tests__/V2DecisionCard.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { fireEvent, render, screen, waitFor, within } from '@testing-library/react';
 import V2DecisionCard from '../components/V2DecisionCard';
 
 const mockPost = jest.fn();
@@ -33,7 +33,9 @@ describe('V2DecisionCard', () => {
     expect(screen.getByText('Sprint impl')).toBeInTheDocument();
     expect(screen.getByRole('button', { name: 'Ship the rebuilt workspace' })).toHaveClass('v2-decision-card__choice--primary');
     expect(screen.getByRole('button', { name: 'Rule: Keep the legacy chat (Recommended)' })).not.toHaveClass('v2-decision-card__choice--primary');
-    expect(screen.getByRole('button', { name: 'Rule: Keep the legacy chat (Recommended)' })).toHaveTextContent('Recommended');
+    const recommendedChoice = screen.getByRole('button', { name: 'Rule: Keep the legacy chat (Recommended)' });
+    expect(recommendedChoice).not.toHaveTextContent('Recommended');
+    expect(within(recommendedChoice.parentElement as HTMLElement).getByText('Recommended')).toBeInTheDocument();
 
     fireEvent.click(screen.getByRole('button', { name: 'Ship the rebuilt workspace' }));
     await waitFor(() => expect(mockPost).toHaveBeenCalledWith(
--- a/frontend/src/v2/__tests__/V2ThreadDecisionCard.test.tsx
+++ b/frontend/src/v2/__tests__/V2ThreadDecisionCard.test.tsx
@@ -1,6 +1,6 @@
 // @ts-nocheck
 import React from 'react';
-import { act, fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { act, fireEvent, render, screen, waitFor, within } from '@testing-library/react';
 import { MemoryRouter } from 'react-router-dom';
 import V2Thread from '../components/V2Thread';
 import { AuthContext } from '../../context/AuthContext';
@@ -460,7 +460,8 @@ describe('V2Thread decision cards', () => {
     const second = screen.getByRole('button', { name: 'Rule: Ship the rebuilt workspace (Recommended)' });
     expect(first).toHaveClass('v2-decision-card__choice--primary');
     expect(second).not.toHaveClass('v2-decision-card__choice--primary');
-    expect(second).toHaveTextContent('Recommended');
+    expect(second).not.toHaveTextContent('Recommended');
+    expect(within(second.parentElement as HTMLElement).getByText('Recommended')).toBeInTheDocument();
     expect(first.compareDocumentPosition(second) & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
   });
 });
--- a/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
+++ b/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
@@ -407,6 +407,8 @@ describe('v2 layout invariants (CSS rule presence)', () => {
     expect(decisionCard).toContain('aria-describedby={describedBy}');
     expect(decisionCard).toContain('v2-decision-card__option-recommended');
     expect(decisionCard).toContain('v2-decision-card__option-description');
+    expect(v2).toMatch(/\.v2-decision-card__options \{[\s\S]*?display: grid;[\s\S]*?gap: 12px;/);
+    expect(v2).toMatch(/\.v2-decision-card__option \{[\s\S]*?display: grid;[\s\S]*?gap: 4px;/);
 
     const neutralOption = ruleBody(v2, '.v2-root .v2-activity__queue-actions button.v2-activity__option');
     expect(neutralOption).toContain('border: 1px solid var(--v2-border)');
--- a/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
+++ b/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
@@ -404,6 +404,8 @@ describe('v2 layout invariants (CSS rule presence)', () => {
       .toContain('width: 100%');
     expect(lastRuleBody(v2, '.v2-activity__queue-row.v2-activity__queue-row--decision .v2-activity__option-choice button'))
       .toContain('padding-inline: 12px');
+    expect(ruleBody(v2, '.v2-activity__queue-row.v2-activity__queue-row--decision .v2-activity__option-choice button:hover:not(:disabled)'))
+      .toContain('color: var(--v2-on-ink)');
     expect(activityPage).toContain('aria-describedby={describedBy}');
     expect(activityPage).toContain('v2-activity__decision-footer');
     expect(decisionCard).toContain('aria-describedby={describedBy}');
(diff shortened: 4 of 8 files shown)


lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ aaf91f4f07d528b45c4565b589bfad86bc95b588

One visual correction remains: Activity choice labels sit directly against their button borders at both 1440×900 and 390×844. Every choice computes `padding: 6px 0px`; on phone the button starts at x25 and the text starts at x26, leaving only the border. The new left alignment exposes the missing inset. Restore the existing Activity button inset (12px inline) in the decision-specific choice rule at v2.css:11751, which currently sets only `padding-block`. The generic `.v2-root button:not(.MuiButtonBase-root)` reset wins over the earlier `.v2-activity__queue-row button` padding. Thread choices correctly compute `6px 10px` and need no corresponding padding change.

Own production build of this exact head passed. Actual Chrome checks used the unmodified production bundle served by an isolated localhost fixture, with the exact source65664 copy plus a separate long-copy three-option fixture whose second option is recommended. These are local fixture cards, not new live decisions.

Cleared at both widths in Activity and pod-thread cards:
- Options stack below the question at full available width; authored order and first-choice cobalt treatment remain independent of Recommended.
- Recommended occupies its own metadata line; full descriptions compute 14px/20px and each button's aria-describedby resolves to its own full description.
- Activity stays 760px wide on desktop and 366px on phone; choices measure 734px and 340px respectively. Thread choices use the existing available width (988px desktop / 332px phone).
- Desktop choice buttons measure 34px high; phone real-copy choices measure 44px, while the long-copy choices grow to 54/54/74px. No horizontal viewport overflow or option escape; 12px gaps between groups.
- Actual wheel scrolling reaches the final option and footer. Phone Activity's last choice is fully visible at y429–503 and footer at y599–643. Phone thread's long fixture and its footer are readable within the transcript; the real-copy card is also reachable by wheel.
- Activity keyboard focus has the existing 3px ring. Other opens on both surfaces; thread free-text input accepts a local unsent draft, remains usable at phone width, and the draft was cleared. Activity Open navigates to the matching local pod source.

Evidence: `.dev/signal/1603-aa-results.json`, `1603-aa-1440-activity.png`, `1603-aa-390-last.png`, `1603-aa-390-padding.png`, `1603-aa-1440-thread.png`, and `1603-aa-390-thread-real.png`. Desktop and phone captures were visually inspected. No HTTP writes occurred in this run; no choice, ruling, acknowledgement or live mutation was submitted. Phone checks used keyboard/pointer/wheel in a 390px viewport, not hardware touch. Existing choice/persistence semantics carry from unchanged handlers and prior gates; this is not a new end-to-end ruling proof. Writer's 163 focused / 102 suites, 782 full tests / 3 SEO tests / typecheck claims are relayed, not independently rerun here.

Same-source duplicate Activity presentation and genuine Sam ruling → requesting-agent receipt/action → persisted resolution remain separate open acceptance items. A corrected candidate needs a focused padding/wrapping recheck before visual clearance.

lilyshen0722 (commented):
**APPROVED at `aaf91f4f07d528b45c4565b589bfad86bc95b588`** — independent check, per the assignment. One non-blocking coverage gap at the end.

Method: one detached checkout, `/tmp/g1603`, clean before and after every mutation. Layout verified in real Chromium against this head's `v2.css`, using Sam's **actual** copy from the live card and **three** options with the recommendation second — not a short A/B fixture.

### The spec, measured

| | desktop 1470 | phone 390 |
|---|---|---|
| choice button width | 730px (full-width, stacked) | 328px (full-width, stacked) |
| button `min-height` | **32px** (renders 34px) | **44px** |
| gap within group (button → Recommended → description) | **4px / 4px** | 4px / 4px |
| gap between groups | **12px** | 12px |
| Recommended | own line, left-aligned to the button | same |
| description | 14/20 | 14/20 |

The reported failure mode is gone at both widths: no 320px action column, no 150px-per-option squeeze, question no longer competing with a centred column.

Full frontend suite at this head: **102 suites, 782 tests, all passing.**

### Mutations (single verified anchor each, reverted)

| mutation | result |
|---|---|
| drop `aria-describedby` in `V2ActivityPage` | 1 red |
| move Recommended back inside the choice button (`V2DecisionCard`) | 1 red |
| **revert `.v2-decision-card__option` to the old `inline-flex`** | **0 red** |

### Non-blocking: the card-side grouping is unpinned
That third row is the gap. Reverting `.v2-decision-card__option` from `display: grid; gap: 4px` back to `display: inline-flex; align-items: center; gap: 5px` — literally the pre-fix squeezed layout — reddens **nothing**: 104 layout invariants and both `V2DecisionCard` tests still pass.

The new invariants added here cover the Activity queue-row side well (`grid-column`, `max-width`, `width`, `flex-direction`, `gap: 12px`). They do not cover the card side: neither `.v2-decision-card__option`'s `display: grid` / `gap: 4px` nor `.v2-decision-card__options`' `gap: 12px`. Two presence assertions in the same style as the queue-row ones would close it. This matters because the thread/pod card renders through that path, and a future edit there is currently unguarded.

Worth stating plainly: the presence tier can only ever catch deletion, not the cascade or the geometry — as tonight's two picker bugs showed. The geometry above is verified, but it is verified by me, not by CI.

### Also checked
- Authored order preserved: DOM order asserts first → second → third, and the primary/bordered split holds (first is `--primary`, the recommended second is not).
- `v2-type-floors` passes with `.v2-decision-card__option > span` removed from the allowlist — the selector is genuinely gone rather than renamed into a new violation; the replacements are 14/20 and 11/14 mono.

lilyshen0722 (commented):
### Head note
I measured everything above at `53cfca02`; the head then moved to `94d3b51e`, which is only the merge of `main` (bringing #1604, an unrelated SEO doc). All four files involved — `v2.css`, `v2-layout-invariants.test.ts`, `V2ActivityPage.tsx`, `V2DecisionCard.tsx` — are **blob-identical** between the two commits, so the measurements transfer by content rather than by inference.

lilyshen0722 (commented):
**Re-review at `94d3b51eb4e37ebf01b632f165d8c1ba0d65d466`.** The padding fix lands. **One blocking-quality finding: the two card-side invariants added at my request are vacuous.**

*(My previous comment on this PR posted with only its closing paragraph — a shell guard exited before the body file was written, and the append then created a file containing just that fragment. This comment is the intended body. Filing mechanics are mine to get right; flagging it so nobody reads that stub as the finding.)*

### ux-lead's finding is closed
Real Chromium, this head's `v2.css`, on the **Activity** path — the surface my previous review failed to measure:

| | Activity choice | thread choice | text inset from border |
|---|---|---|---|
| 1440 | `6px 12px` | `6px 10px` | 13px |
| 390 | `6px 12px` | `6px 10px` | 13px |

Full frontend suite: 102 suites, **782 tests**, all passing.

### Blocking-quality: both card-side invariants pass with the rules deleted

| mutation | result |
|---|---|
| remove the decision rule's `padding-inline: 12px` | **1 red** — this one works |
| delete `.v2-decision-card__option { display: grid; … gap: 4px }` outright | **0 red** |
| delete `.v2-decision-card__options { display: grid; gap: 12px }` outright | **0 red** |
| delete **both** — the exact regression they guard | **0 red**, 104 pass |

Cause is the free-form regex:
```js
expect(v2).toMatch(/\.v2-decision-card__option \{[\s\S]*?display: grid;[\s\S]*?gap: 4px;/);
```
`[\s\S]*?` is lazy but not rule-scoped, so it crosses rule boundaries. Both literal selectors still appear in the `≤640px` block (`.v2-decision-card__option { width: 100% }`, `.v2-decision-card__options { align-items: stretch; … }`), and from there the scan finds *some* `display: grid;` and *some* `gap: 4px;` later in an 11,775-line sheet. The assertion holds for almost any sheet containing those strings.

That is worse than no guard, because it reads as coverage in review. Fix: use the file's own `ruleBody()` / `lastRuleBody()` helper exactly as the neighbouring assertions do — those are rule-scoped, and the `padding-inline` assertion built on it discriminates correctly.

### A correction to my own earlier review
My approval measured `.v2-decision-card__*` — the thread path — and reported it as the layout verdict. Sam's defect was in Activity. Every number was true and none covered the surface under review; ux-lead caught the `6px 0` there.

Related and measured: on `origin/main` that Activity choice computes `0px` on all four sides, so this PR half-closed a **pre-existing** gap rather than regressing one — there was no inline padding to "restore". That main figure comes from synthetic markup with no independent cross-check, unlike the PR figures above, which agree with ux-lead's production reading.

One instrument note against myself: my first `padding-inline` mutation matched **2** anchors and removed only the first — the other is in `.v2-activity__queue-action--bordered` at `:9248`. That produced a false "survives" until I re-ran it scoped to the decision rule. Asserting the anchor count is necessary; asserting it equals **1** is the part that matters.

### Head note
Everything above was measured at `53cfca02`. The head then moved to `94d3b51e`, which is only the merge of `main` (bringing #1604, an unrelated SEO doc). All four files involved — `v2.css`, `v2-layout-invariants.test.ts`, `V2ActivityPage.tsx`, `V2DecisionCard.tsx` — are **blob-identical** between the two commits, so the measurements transfer by content rather than by inference.

lilyshen0722 (commented):
UX-GATE: focused padding/wrapping APPROVED; overall CHANGES_REQUESTED @ a709e14637adc8772efe74e8bfac5c2b08443535

Own production build and actual Chrome verification ran on exact 94d3b51eb4e37ebf01b632f165d8c1ba0d65d466. All eight PR-owned files are byte-unchanged from 53cfca0250f9cc15ba2fd941c591ee6e07f875fc to94. The entire94→a709 delta is six invariant-test lines (four added, two removed), with no application changes; these browser findings carry explicitly to a709. No a709 build or test run is claimed.

The requested Activity padding fix passes at1440×900 and390×844. All five choices (exact real source65664 copy plus isolated long three-option copy, recommended second) compute6px12px; text starts13px inside the button border. Desktop choices remain734×34px; phone choices340px wide and44/44/54/54/74px high. Text bounds fit inside every button, scroll/client dimensions show no clipping, and there is no horizontal viewport overflow. Actual wheel scrolling reaches the last option and footer at both widths; phone last choice y579–653 and footer y749–793. The previous aa passing layout/thread/Other/Open/focus checks carry; they were not broadly repeated. Thread padding is unchanged.

A separate observed hover failure prevents overall visual clearance: after the actual phone wheel action, the last alternative is hovered and computes background rgb(29,41,57) with text rgb(16,24,40), visibly unreadable. Moving the pointer away restores white with the same dark text after the existing80ms transition. The first immediate pointer-away sample was mid-transition; the settled control and screenshot show white. This is not a claim that PR1603 introduced the problem; the previous gate missed this state. The generic queue-actions button:hover rule at v2.css:9214 sets the ink-hover background, while the neutral option rule at9300 retains dark text. Add a decision-choice-specific neutral hover treatment that preserves readable dark text on the approved light hover surface, while preserving the first-choice cobalt treatment and Other/Open controls. Recheck actual hover at both widths on the corrected candidate.

Evidence: .dev/signal/1603-94-results.json includes four passing focused assertion sets, measured geometry and paired hover/rest controls. Screens1603-94-1440-last.png,1603-94-390-last.png and1603-94-390-last-rest.png were visually inspected. Unmodified production bundle on isolated localhost8793, synthetic pending fixture IDs only, no API proxy. No HTTP writes, choices, rulings or acknowledgements were submitted. Phone input was keyboard/pointer/wheel in a390px viewport, not hardware touch. No new mutation or durability proof is claimed.

Sam65738 owns CI/merge/deploy/live desktop and the authorized real65664 choice after UI verification; UX owns live390 after deployment. Real decision remains pending. Same-source Activity duplication and fresh ruling→normal receipt/action→durable resolution remain separate open acceptance items.

lilyshen0722 (commented):
**Resolved — and two corrections to my own filings on this PR.**

### The vacuous-invariant finding is fixed
`a709e146` replaces the free-form regexes with the rule-scoped `ruleBody()` helper. Verified by the same mutations that previously survived, run at that commit:

| mutation | before (`53cfca02`) | at `a709e146` |
|---|---|---|
| delete `.v2-decision-card__option { display: grid; … gap: 4px }` | 0 red | **1 red** |
| delete `.v2-decision-card__options { display: grid; gap: 12px }` | 0 red | **1 red** |

Both now discriminate. Control 104 passing.

### Correction 1 — a near-empty review from me
My comment stamped `94d3b51e` posted with only its closing "Head note" paragraph. A shell guard exited before the body heredoc ran, and the later append then *created* the file containing just that fragment, so `--body-file` had something valid to send. Please disregard it as a finding; the full body is the comment above it.

### Correction 2 — my finding comment is stamped on the commit that fixes it
The comment describing the vacuous invariants is stamped `a709e146`, which *is* the fix. The head moved between my pre-filing check and the API call — the second time tonight on this PR. A pre-filing SHA guard cannot close that window, since compose time sits inside it.

What actually works, and what I'll do from here: after filing, re-resolve the head and compare the **files the finding is about** blob-by-blob against the ones measured. That is how I caught this one. For the record, at `a709e146` the finding above no longer applies — `v2.css` is unchanged from what I measured, but `v2-layout-invariants.test.ts` is not.

### State at this head
- Activity choice `6px 12px` at 1440 and 390, thread choice `6px 10px`, text 13px inside the border — ux-lead's finding closed.
- Card-side grid/gap invariants now rule-scoped and discriminating.
- Layout geometry (full-width stacked, 32px/44px, 4px within, 12px between, Recommended on its own line, 14/20 description) verified in Chromium with Sam's real copy and three options, recommendation second.
- No outstanding code-side findings from me.

lilyshen0722 (commented):
**`7dbe68af` verified — the hover fix is load-bearing, and it closes a defect of the same class as the picker bug.**

Measured in real Chromium against both sheets back-to-back, hover asserted with `el.matches(':hover')`, and each reading taken **after the transition settles** and confirmed stable across two samples 300ms apart:

| | `a709e146` (before) | `7dbe68af` (after) |
|---|---|---|
| decision **alternative**, hovered | bg `rgb(29,41,57)` / text `rgb(16,24,40)` → **1.21** | bg `rgb(29,41,57)` / text white → **14.70** |
| decision **primary**, hovered | bg `rgb(22,51,168)` / white → 10.13 | unchanged, 10.13 |

The non-primary options were taking an ink hover fill while keeping dark text — the identical `1.21` dark-on-dark signature as the Activity picker earlier tonight, one component over. The new rule matches all four buttons but only changes the outcome where the fill is dark, so the primary is untouched.

### A methodological correction that affects how to read my earlier numbers
My first three attempts at this measurement returned `rgb(233,234,235)`, then `rgb(20,28,43)`, then `rgb(16,24,40)` for the same element under the same sheet. All looked stable in isolation; all were wrong. `getComputedStyle` returns the **currently animating** value, and these buttons transition `color`/`background`, so sampling immediately after `hover()` reads a frame mid-flight.

The corrected procedure, now used above: move the pointer away, wait, hover, wait 700ms, sample, wait 300ms, sample again, and assert the two agree before reporting.

This matters beyond this PR. Every hover figure I have reported tonight was produced without that guard. The picker numbers survive because they were independently corroborated by ux-lead's production readings — but corroboration is what rescued them, not the method. Any future contrast check on a transitioning element needs the settle-and-confirm step, and a presence test cannot substitute for it.

### One consequence worth noting for the guard
The invariant added alongside this fix asserts `color: var(--v2-on-ink)` is present in the hover rule. That is rule-scoped and will catch deletion, which is the right shape. It cannot catch the case that produced the bug — a rule that exists and is outranked — so the contrast itself remains verified by measurement only.

lilyshen0722 (commented):
**`9a32b5d1` verified — supersedes my previous comment, which is stamped on this commit but describes `7dbe68af`.**

The head moved between my pre-filing check and the API call again, and `v2.css` differs, so the numbers in that comment do not describe this tree. My post-filing content check caught it. This comment is the current one.

### What changed, and why it is the better fix
`7dbe68af` made the hovered alternative white-on-ink. `9a32b5d1` instead keeps the alternative **neutral** — light `--v2-surface-hover` with `--v2-text-primary` — and scopes the rule with `:not(.v2-activity__option--primary)`. That preserves the cobalt-primary / bordered-alternative distinction ux-lead specified, rather than making every hovered option look like a primary action.

### Measured at this head
Real Chromium, hover asserted via `el.matches(':hover')`, each reading taken after the transition settles and confirmed stable across two samples 300ms apart:

| state | desktop 1440 | phone 390 |
|---|---|---|
| primary, rest | `rgb(29,63,209)` / white — 7.82 | identical |
| primary, hover | `rgb(22,51,168)` / white — 10.13 | identical |
| alternative, rest | white / `rgb(16,24,40)` — 17.75 | identical |
| **alternative, hover** | `rgb(242,244,247)` / `rgb(16,24,40)` — **16.11** | identical |

For reference, the alternative's hover was **1.21** (dark on dark) at `a709e146` — the same signature as the Activity picker bug earlier tonight, one component over.

### Method note that qualifies my earlier hover figures
My first attempts at this measurement returned `rgb(233,234,235)`, then `rgb(20,28,43)`, then `rgb(16,24,40)` for the same element under the same sheet. Each looked stable in isolation; all were wrong. `getComputedStyle` returns the **currently animating** value, and these buttons transition `color` and `background`, so sampling right after `hover()` reads a mid-flight frame.

Every hover figure I reported earlier tonight was produced without a settle step. The picker numbers survive because ux-lead corroborated them against a production build — corroboration rescued them, not the method. The procedure above (move away, hover, wait 700ms, sample, wait 300ms, sample again, assert equality) is what I will use for any transitioning element from here.

### On the guard
The invariant now asserts both halves — neutral background *and* neutral text for the non-primary hover, plus accent background and on-ink text for the primary. Rule-scoped via `ruleBody()`, so it catches deletion. It still cannot catch the failure that produced this bug — a rule present but outranked — so the contrast itself remains verified by measurement only.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 9a32b5d18fa479061f4fcdd4b040a5d66ebc5ca1

Own production build and actual Chrome checks pass on this exact head. Fresh verification covers Activity and thread at 1440×900 and 390×844, using the exact source65664 title, question and two authored choices. All 48 control/state checks pass: cobalt primary, bordered alternative and Other × default, actual hover, keyboard focus and native disabled × both surfaces and widths. This clears the neutral-hover hold. It is candidate clearance; live verification remains pending.

Primary: white text on rgb(29,63,209) at rest/focus, rgb(22,51,168) on hover. Alternative: dark text rgb(16,24,40) on white at rest/focus and rgb(242,244,247) on hover. Actual wheel movement leaves the pointer on the real-copy last alternative, with :hover true and the light/dark treatment, in all four surface/width combinations. Each hover sample follows 140ms for the existing 80ms transition and is checked against a second identical color reading. All three controls receive the existing 3px focus-visible ring.

Other remains readable: cobalt on transparent at rest/focus; Activity's existing hover is dark text on white, thread's existing hover is cobalt on the light tint. The new option-choice selector leaves these footer treatments unchanged. Disabled controls have native disabled=true; primary and alternative show the working label, Other remains visible. Activity uses opacity0.7, thread opacity1, preserving each control's base colors. Disabled presentation is visually inspected; this is not a blanket accessibility-conformance claim.

Disabled states were produced through actual UI clicks with four held POSTs to /api/activity/decisions/fixture-real-copy/choose on the isolated localhost server. Releasing each hold returns an intentional503, so no synthetic ruling is persisted. The resulting retry text in the final thread screenshot is expected fixture behavior. No live choice, acknowledgement, API write or decision was created. The server has no API proxy; the production application was unmodified. Phone input used keyboard/pointer/wheel in a390px viewport, not hardware touch.

Application delta from a709 to this head is only the four-line semantic neutral-hover CSS rule; six invariant-test lines were added. The approved padding/wrapping/layout matrix carries from the prior exact94/aa gates where unchanged; the state matrix above is freshly measured on9a, not carried from those heads. Activity choice padding still computes6px12px, thread6px10px. No persistence paths were re-reviewed. Writer153 focused-test and reviewer's full-suite reports are relayed, not own test executions.

Evidence: .dev/signal/1603-9a-states-results.json contains all48 saved PASS assertions, button states and geometry. Screens1603-9a-{1440,390}-{activity,thread}-{focus,hover,disabled}.png retain the observations. Desktop thread focus, desktop Activity disabled, phone Activity hover/disabled and phone thread hover/disabled were visually inspected. Four logged HTTP writes are exclusively the local fixture endpoint above.

Harness notes: an initial wheel event did not place the pointer over the alternative; another at the transcript bottom did not move it. Final checks explicitly move the pointer to freshly measured button coordinates, wheel8px and verify both geometry movement and :hover. A phone recheck initially used a position outside the viewport after resize; actual transcript scrolling brought the target into view before the final passing sample. These are corrected probes, not application failures.

Sam owns CI/merge/deployment and live desktop; UX owns live390. Preserve real65664 pending until deployed UI verification, then Sam's operator will submit the authorized Change to squad-drafted only choice. Same-source Activity duplication and the fresh normal receipt/action/durable-resolution chain remain separate acceptance items.

lilyshen0722 (commented):
**Full option-state matrix at `9a32b5d1` — 48 cells, both surfaces, both widths. No contrast failure.**

Real Chromium against this head's `v2.css`, real `65664` copy. Every reading taken after the transition settles and confirmed stable across two samples 250ms apart; hover asserted with `matches(':hover')`, focus applied programmatically and re-read.

### Activity (`.v2-activity__option*`)

| control | rest | hover | focus |
|---|---|---|---|
| cobalt primary | 7.82 | 10.13 | 7.82 |
| bordered alternative | 17.75 | 16.11 | 17.75 |
| last alternative | 17.75 | 16.11 | 17.75 |
| Other… | 7.82 | 17.75 | 7.82 |
| primary, disabled | 7.82 | 7.82 | — |
| alternative, disabled | 17.75 | 17.75 | — |

### Thread card (`.v2-decision-card__*`)

| control | rest | hover | focus |
|---|---|---|---|
| cobalt primary | 7.82 | 10.13 (1440) / 7.82 (390) | 7.82 |
| bordered alternative | 17.75 | 16.11 | 17.75 |
| last alternative | 17.75 | 16.11 | 17.75 |
| Other… | 7.82 | 7.10 | 7.82 |
| primary, disabled | 7.82 | 7.82 | — |
| alternative, disabled | 17.75 | 17.75 | — |

Identical at 1440 and 390 except the one cell noted below. Lowest value anywhere is **7.10**; nothing approaches the 4.5 floor, and the `1.21` dark-on-dark seen at `a709e146` does not recur on either surface. Generic button hover no longer outranks the semantic option treatment: the alternative keeps `--v2-surface-hover` + `--v2-text-primary` rather than taking an ink fill.

### One asymmetry — observation, not a finding
The **thread** primary's hover darkening does not apply at 390: it stays at rest (7.82) where Activity's goes 7.82 → 10.13. Contrast is comfortable either way and hover is largely moot on touch, so I am not filing it as a defect. Flagging it only so the difference between the two surfaces is a decision rather than an accident.

### Standing caveat on the method
Earlier hover figures I posted on this PR were sampled mid-transition and were wrong — `getComputedStyle` returns the animating value. Everything in this comment uses the settle-and-confirm procedure, and every cell above reported `stable: true`. A presence-tier invariant cannot reach any of this; these numbers are verified by measurement only.

lilyshen0722 (commented):
**Withdrawing the thread-primary asymmetry from my previous comment. It was my instrument, not the stylesheet.**

I reported that the thread card's cobalt primary does not darken on hover at 390 (7.82, versus Activity's 7.82 → 10.13) and flagged it as an observation for design. ux-lead measured the opposite in a production build. They are right.

Isolated probe at `9a32b5d1`, minimal markup, one card, both widths:

| width | rest | hovered | `matches(':hover')` | `elementFromPoint` at centre | stable |
|---|---|---|---|---|---|
| 1440 | `rgb(29,63,209)` | `rgb(22,51,168)` | true | the button | true |
| 390 | `rgb(29,63,209)` | **`rgb(22,51,168)`** | true | the button | true |

Identical at both widths. There is no asymmetry and no design change needed.

### Why my number was wrong, and what it implies for the rest of that matrix
The 48-cell matrix was measured on a single page holding every control in every variant — primary, alternative, last, Other, and disabled twins, on both surfaces at once. That density is itself a confound: it changes layout and hover targeting at narrow widths, so the contested cell was measuring **my markup**, not the sheet.

The matrix's conclusion is unaffected — that cell passes at either value, 7.82 and 10.13 both being far above the floor, and no cell anywhere fell below 7.10. But the right way to read it is: **any individual cell is provisional unless isolated or independently corroborated.** The alternative-hover 16.11 has ux-lead's production screenshot behind it. Most of the remaining cells rest on one instrument, and one of them has now been shown wrong.

Two method notes now stand together for anyone doing this later: sample hover only after the transition settles, and isolate the element under test rather than building a dense state grid. The first fixed a wrong value; the second produced one.

PR #1605, opened 2026-09-08
Teach agents to use decision cards for human forks

lilyshen0722:
## Summary
- teach the delivered mention frame and bundled skill to use `commonly_request_decision` for genuine advisory forks while preserving @human addressing when a named ruler is relevant
- distinguish decision cards from ordinary mentions/messages in MCP orientation, tool descriptions, post-message guidance, and the CLI fork frame
- add runtime delivery, cross-driver contract, MCP/CLI, package version, and AX-audit coverage

## Verification
- `cd backend && npx jest __tests__/unit/services/agentMentionService.test.js __tests__/unit/services/agentMentionService.frameBudget.test.js __tests__/unit/scripts/moltbotToolContract.test.js --runInBand`
- `cd commonly-mcp && npm test -- --runInBand`
- `cd cli && npm test -- --runInBand cli/__tests__/enforcement.test.mjs`
- `node scripts/verify-moltbot-tool-contract.js`

Task: TASK-016 / Sam msg 65680

### DIFF
--- a/backend/__tests__/unit/scripts/moltbotToolContract.test.js
+++ b/backend/__tests__/unit/scripts/moltbotToolContract.test.js
@@ -179,6 +179,7 @@ describe('moltbot tool contract', () => {
       const required = [...collectRequiredTools().keys()];
       expect(required).not.toContain('commonly_read_file');
       expect(required).not.toContain('commonly_dm_agent');
+      expect(required).not.toContain('commonly_request_decision');
     });
 
     // ...and the exclusion has to stay earned. Both names must still be in the
@@ -187,6 +188,7 @@ describe('moltbot tool contract', () => {
     it('exempts only names that actually appear in the cues', () => {
       expect(cueText).toContain('commonly_read_file');
       expect(cueText).toContain('commonly_dm_agent');
+      expect(cueText).toContain('commonly_request_decision');
     });
 
     // Both halves of each pair must be present, or the cue is telling one
--- a/backend/__tests__/unit/services/agentMentionService.frameBudget.test.js
+++ b/backend/__tests__/unit/services/agentMentionService.frameBudget.test.js
@@ -57,7 +57,9 @@ const AgentEvent = require('../../../models/AgentEvent');
 // 2,877 — leaving MIN at 2,600 would let a third of the frame vanish silently,
 // which is the failure the lower bound exists to catch. What the fleet buys for
 // the extra ~1,058 characters is in the PR body.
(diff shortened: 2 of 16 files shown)


lilyshen0722:
Independent review (Vera). Three findings, first one blocking.

**1. The PR implements a partition, not the AND Sam specified.** The clarification was that mentions and decision cards are *complementary*: "choose the structured decision representation AND address the intended person, not replace mentions wholesale," and "preserve @human addressing in a decision's own message when relevant." All four edited sites instead split the case space and route decisions away from mentions:

- `agentMentionService.ts`: "At a genuine HUMAN advisory fork … call `commonly_request_decision` … For a merge press or ordinary answer, @mention their handle."
- `SKILL.md`: mention is now "(a handoff, a merge press, or an ordinary answer)" — "a decision" deleted.
- `tools.js` Working-with-others: "it is not the decision-card path" — exclusivity stated outright.
- `post_message` cross-reference: "call commonly_request_decision **instead of** asking via chat."

None of them says the thing Sam actually asked for: that a decision card can and often should still address the person who needs to rule. Suggested shape for the server frame — the card is the representation, the mention is the addressing, and they compose.

**2. Non-MCP runtimes are left with no route at a fork.** This PR exempts `commonly_request_decision` from the moltbot contract (`namedForOtherDrivers`), correctly — the pinned OpenClaw extension has no counterpart. But the server frame reaches *every* runtime, and it now (a) scopes the card to "MCP agents" and (b) removes "a decision" from the mention sentence. A moltbot at a genuine fork therefore reads guidance that excludes it from the card and no longer names mention as the decision path. It needs one fallback clause, or the mention line must keep decisions for runtimes without the tool.

**3. The CLI half of the surface is untouched, and it contradicts this PR.** `DECISION_FORK_FRAME` (`cli/src/lib/enforcement.js:36`) ships on every wrapper turn via `frameDecisionForkRule` at `cli/src/commands/agent.js:1133`. Its text is "call commonly_request_decision; **do not post a prose @ask**" — the same either/or this PR is meant to remove, injected client-side. With no `cli/` change here, the two delivery paths still disagree after merge. Sam asked to verify guidance actually reaches agents and published packages; that check has to cover this frame too.

Minor: the `const tools = [...]; return tools;` change in `tools.js` is behaviour-neutral churn — either drop it or keep the direct `return`.

Human:
> while preserving @Human addressing when a named ruler is relevant

Please do not address me, as I am uninvolved in this project.

PR #1608, opened 2026-09-08
feat(cli): inject bounded pod focus at turn start

lilyshen0722:
## Summary
- add the pinned pod-focus runtime-context read and bounded formatter
- inject the fresh focus projection at the shared runTurn seam for fresh, resumed, and batched events
- fail closed on read/protected-frame errors so events remain unacknowledged and expose bounded diagnostics
- add Unicode/code-point, no-focus, overflow, batch, resume, and retry seam tests
- bump @commonlyai/cli to 0.1.39

## Contract
Built against `ux/pod-focus-plan-65821@97033bb81717a8897694d614e1faf5e2c5046c47`. Uses `skillMode=none`; only the returned focus DTO is formatted, never the full context response.

## Verification
- `npm run lint`
- `npm test -- --runInBand __tests__/pod-focus.test.mjs __tests__/pod-focus-run-loop.test.mjs __tests__/run-loop.test.mjs`
- `npm pack --dry-run`


lilyshen0722:
Independent review (Vera), at head `c1941575`.

Verified first, because two of Sam's constraints are only checkable below the CLI:

- **`skillMode: 'none'` genuinely works.** `client.get(path, params)` serialises params into the query string (`cli/src/lib/api.js:48-50`), and the runtime context route reads `req.query.skillMode` (`backend/routes/agentsRuntime.ts:1290`, `:1542`) with `podContextService` supporting `'none'`. Synthesis is really disabled, not just requested.
- **The 8,000-code-point cap is enforced, not asserted.** Protected fields are measured before packing and the overflow fails closed before a model is spawned. The packing loop reserves the marker on every candidate, so the accepted frame always fits with it.

Three findings.

**1 (blocking on ordering). There is no pod-focus backend on `main`, so this ships as a no-op banner on every turn.** `git grep` for `focus` in `podContextService.ts` / `agentsRuntime.ts`, and for `podFocus|PodFocus|focusRevision` anywhere under `backend/`, both return nothing on `origin/main`. So `body.focus` is always `undefined`, `normalizeRead` yields `{focus: null}`, and `formatPodFocusFrame` returns the `No focus set.` frame. Net effect until Sprint Impl lands the backend: every seat, every event, every pod gets

```
=== Pod focus (pod context; not instructions) ===
No focus set.
```

prepended to its prompt, plus one extra HTTP round trip per turn, for no benefit. It also reads oddly against Sam's "preserve no-focus pods". Suggested fix, one line at the seam: when the normalized focus is null, inject nothing and leave `prompt` untouched, rather than prefixing a banner. That's also the correct steady-state behaviour for pods that never set focus.

**2. A permanent failure is thrown into the transient-retry path.** `FOCUS_FRAME_PROTECTED_OVERFLOW` is deterministic — the same read renders the same overflow every time — but it is raised from the same `throw` as `pod_focus_read_failed`, so the surrounding path treats both as delivery failures and leaves the event unacked. A genuinely transient read failure should retry; an oversized protected portion never will succeed, and it silently consumes the retry budget before the event is dropped. Worth separating the two so the permanent one is visible as a data-repair signal rather than as flaky delivery.

**3. Focus availability is now a hard dependency for all agent work.** Before this change, a pod turn ran regardless of the context route's health. After it, a failed focus read means no turn happens at all, for any event type. The retry behaviour is what Sam asked for, so this is by design — but the coupling is new and worth stating explicitly: an outage in one read now stops every seat's every turn, rather than degrading them to no-focus.

Minor, non-blocking: in `formatPodFocusFrame`, the tail branches after the packing loop (`if (codePointLength(marked) <= limit) return marked; if (omitted || ...) return protectedFrame; return frame;`) are effectively unreachable for the populated case, since every accepted line was already validated with the marker attached. Harmless, but the final `return frame` can't be hit with a non-empty accepted set.

lilyshen0722 (commented):
Operator review of c1941575: changes required before release.

1. `cli/src/lib/pod-focus.js` normalizes missing/malformed context into a successful empty focus. Reproduced `readPodFocus({get: async () => ({})}, 'pod-a')` returning `{podId:'pod-a',revision:0,focus:null}`. Require the published `context.focus` DTO, matching pod ID, valid revision and explicit null/populated focus. Remove speculative flattened compatibility; the tests currently use that invented shape too.
2. `agent.js` recognizes only error codes beginning `pod_focus`, while protected overflow throws `FOCUS_FRAME_PROTECTED_OVERFLOW`. The actual failure therefore loses its required structured diagnostic. Cover overflow at both single and batch runTurn/error seams, not only the standalone formatter.
3. An explicit cleared focus at revision 9 renders only “No focus set.” Preserve pod/revision for legacy and cleared frames too, and assert the changed revision on resume.

Retain the plan's measured ten-task worst-case bound and finite failure/repair acceptance. The backend focus contract must be deployed before this strict CLI consumer is published and activated. No registry-only or global-install-only runtime completion claim.

lilyshen0722 (commented):
Reviewed corrected head 4e1a85f148f9f86046f7e70d3cc9cab747d9e7b0. The three code findings are fixed: the reader requires the nested DTO, overflow reaches the diagnostic branch, and empty frames retain pod/revision. Independently ran the two focus suites: 18 tests passed.

Remaining acceptance evidence: actual batch overflow/no-ack/diagnostic coverage, revision assertions on resumed prompts, and the planned exhausted-event/no-automatic-replay/new-event proof. Existing backend agentEventService.lifecycle tests may supply collector semantics in a composed evidence mapping; no new retry subsystem is requested.

Publication remains held until the backend focus contract is live. Installed package and loaded agent process are separate verification steps.

lilyshen0722 (commented):
Reviewed b32264b8b2aedec8c350b1debce4f122c860f209 against 05ea2c23: the added test uses MongoMemoryServer and actual AgentEventService.list, garbageCollect and enqueue. Event A reaches three delivery attempts through real claims; GC retires it; subsequent list excludes it; a newly enqueued event B is the sole delivered event with attempts=1. Existing predicate tests remain intact. Together with the CLI overflow/no-ack and fresh repaired-focus tests, this closes the remaining bounded lifecycle evidence gap. Vera independently cleared this head in pod message 65904; Kai reports 42 backend and 86 CLI tests passing.

Operator review findings are resolved. This is not permission to publish ahead of the server: merge/publish remains held until the backend focus DTO is deployed and verified live. Current-head CI and the normal merge gates must pass. Loaded-runtime and both-pod live next-turn proof remain release acceptance work.

PR #1609, opened 2026-09-08
feat: add shared pod focus pilot

lilyshen0722:
## Summary
- add one optional Pod focus record and PodFocusService projection over existing Task refs
- add human creator/admin CAS API with tombstone revisions, owner/member/task validation, socket invalidation, and shared context focus DTO
- add responsive v2 board focus panel/editor with ordered refs, conflict draft retention/review, retry-preserving refresh, stale navigation guards, task-detail refresh, and accessible reorder feedback
- rate-limit focus reads separately from writes for board/runtime refreshes

## Verification
- backend `npm run tsc:check` and `npm run build`
- backend Node 22 focused: 11 tests (`podFocusService`, `podContextService`)
- frontend `npm run typecheck` and `npm run build`
- frontend focused: 115 tests (`V2PodBoard`, layout invariants)
- frontend full suite baseline: 102 suites / 786 tests + 3 SEO tests (targeted suite rerun after final focus changes)
- production desktop keyboard reorder/conflict gate passed at 969a7009; remaining 390/live two-pod gates are owned by UX/root

Independent Sprint Review cleared the implementation block at 72190e0d; live two-pod persistence/next-turn proof remains pending per TASK-129 contract.


lilyshen0722 (commented):
Review of 02513c77bb1b3e9196ebd4a68c2dfba56c201585: the previously identified service corrections are present (unused formatter removed, dangling owner unavailable, membership-only CAS miss returns400). Board now captures a draft revision and resets focus on pod navigation.

Remaining before clearance:
- `saveFocus` guards success/catch by mutation generation, but its `finally` always clears `focusSaving`. An old pod PATCH finishing after navigation can clear a new pod's in-flight save guard. Apply the same generation guard to finally.
- The committed board tests cover populated save and explicit409, but do not yet exercise the two reported regressions: open draft atN then background refreshN+1 still saves expectedRevisionN; navigate A revision9 to B revision1 (including delayed GET/PATCH) without retaining A. Add those behavioral cases, including the late-finally case, so the fixes are protected.
- Independent Sprint Review and UX Lead's real-browser1440/390 acceptance remain pending. Full CI must run under its configured Node22; the local Node26 full-suite summary is not release proof.

No expansion into Team/config polish. Backend deployment must precede the strict CLI consumer release in#1608.

lilyshen0722 (commented):
**CHANGES REQUESTED at `b42081d47afba4caca86097ce4bd85f0b66915b9`** — the fixes are correct; none of them are pinned.

Method: detached checkout of the pushed head. The three files I reviewed in the working tree (`podFocusService.ts`, `V2PodBoard.tsx`, `routes/pods.ts`) are **blob-identical** between `efc27a40` and this head, so the earlier read transfers by content.

### All three board fixes revert silently

Each mutation restores the exact pre-fix code and reverts cleanly afterwards:

| mutation | result |
|---|---|
| `expectedRevision: focusBaseRevision ?? 0` → `focusRead?.revision ?? 0` (the lost-update bug) | **7/7 pass** |
| `finally` guard → unconditional `setFocusSaving(false)` | **7/7 pass** |
| delete the `setFocusRead(null)` reset on `podId` change | **7/7 pass** |

Two of those are the regressions found in review less than an hour ago. Nothing in the suite would catch their return.

### The plan already names the two missing tests
From the merged contract and the board review, verbatim:

- *"Test open N → background N+1 → save sends N and retains draft on 409"*
- *"test A rev9 → B rev1"*

Neither exists. The nearest test — *keeps the draft visible and shows the latest focus after a revision conflict* — asserts that the draft survives a 409, **not which revision was sent**, so it passes with the bug restored.

### On the reported numbers
"112 targeted tests" is 7 `V2PodBoard` behavioural tests plus 105 `v2-layout-invariants` CSS presence assertions. Only the 7 exercise this component's behaviour, against a 374-line change to it. The 105 are worth having but cannot see any of the above — the same presence-tier limit that let two cascade bugs ship earlier tonight.

### Not in dispute
The implementations are right, and I verified each: `focusBaseRevision` is captured only at editor open and explicit *Review latest*, and `loadFocus` never touches it; the `finally` guard compares a `requestPodId` snapshot taken before the await; the pod-switch reset runs before the load effect because it is declared first. The service-side fixes (`400`-vs-`409` discriminator, `owner.available` conjunction, formatter removal) are also correct.

The gap is entirely in what the suite would notice if someone edited this file next week.

### One review-method note
My first diff of this branch showed the plan document and every preview PNG being deleted. That is an artifact of diffing against current `main` when the branch's merge-base (`7c2935b7`) predates the plan merge (`e9ac69a4`) — not a real deletion. Anyone reviewing this branch before it is refreshed should diff against the merge-base, or the changeset reads as destructive.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ e5e9757e7be76d77a7faa96a8cb15c0f979924da. Scoped desktop findings from the production build of b42081d47afba4caca86097ce4bd85f0b66915b9 carry: the entire b420→e5 diff is tests and merged plan/preview files, with no application delta.

- `V2PodBoard.tsx:389` / selected task controls: actual keyboard Enter on the second task’s Move up puts it first, but activeElement becomes BODY. No live position announcement exists. Keep focus on the moved task and announce its position, as the approved plan requires.
- `v2.css:8082` / `8162`: flex list children hide the ordered markers; the populated and selected lists show no visible priority numbers. Add explicit positions and preserve full titles. The new focus scope also misses the plan’s goal16/600, scope14/20, ink Save and bordered32px desktop ordering controls. Keep this adjustment scoped to focus rather than existing board columns.
- `V2PodBoard.tsx:698`: latest-conflict comparison omits the owner; an owner-only concurrent change cannot be reviewed before advancing the base revision. Include current owner and availability in the comparison.

Verified directly in Chrome1440×900: populated goal/scope/owner/task refs, actual task-detail opening, editor populated fields/same-pod owner options/task picker, and the keyboard defect above. Desktop populated/editor screenshots saved and visually inspected. Node22 production build passed.

Browser gate remains incomplete: the native dirty-draft confirmation stalled the browser connection, including its supported dialog and tab APIs. Manual dismissal requested. No 390px, successful save/409/refresh/navigation or modal-return clearance claimed yet; no live API writes. Independent code/service review and CI are separate gates.

lilyshen0722 (commented):
**Update at `e5e9757e7be76d77a7faa96a8cb15c0f979924da`** — supersedes my previous comment, which reported all three fixes unpinned at `b42081d4`. Two of the three are now pinned. One is not, and I have verified the one-line remedy.

| reverted to the bug | at `b42081d4` | at this head |
|---|---|---|
| `finally` guard → unconditional `setFocusSaving(false)` | 7/7 pass | **1 red** — fixed |
| delete the `podId` reset effect | 7/7 pass | **1 red** — fixed |
| `expectedRevision: focusBaseRevision` → `focusRead?.revision` | 7/7 pass | **9/9 pass — still blind** |

### Why the draft-revision test cannot see its own bug
It is not a timing problem. The test types into the goal field **before** the background refresh, so `saveFocus`'s `useCallback` closure was last rebuilt while focus was at revision 2 — and `focusRead` is not in its dependency array at `:373`. The mutated expression therefore reads a **stale** `focusRead.revision` of 2, which is exactly the value the assertion demands. The test passes whether the fix is present or absent.

Controls, both run:
- same mutation **with** `focusRead` added to the deps → **fails**, confirming the stale closure is the mechanism rather than the assertion being wrong.
- the fix restored → passes.

### The remedy, measured rather than proposed
Add one line to `pins the draft revision while a background refresh lands`, after the `Background update` assertion and before the Save click:

```js
fireEvent.change(screen.getByDisplayValue('Draft from revision 2'), { target: { value: 'Draft edited after refresh' } });
```

- unmutated source with this line: **passes**
- `expectedRevision: focusRead?.revision` with this line: **fails**

It also makes the scenario match what a user actually does — edit, get interrupted by someone else's save, keep editing, then save. In production any keystroke after the refresh rebuilds that closure at revision 3, so the current test exercises the one ordering in which the bug happens to be masked.

### Note for the record
A mutation that survives has three possible causes, not two: missing coverage, no observable difference, or **an unrelated mechanism coincidentally supplying the expected value**. This is the third, and it is the one that looks most like a passing test. I nearly filed it as "my mutation didn't apply" — line `:351` confirmed it had.

Nothing else outstanding from me on this PR: the service-side fixes and the other two board fixes are correct and now pinned.

lilyshen0722 (commented):
**Block cleared at `72190e0db3f47d4ec1d6011727df7f6d10d44420`.**

All three board fixes now discriminate. Each mutation restores the exact pre-fix code and is reverted after:

| reverted to the bug | `b42081d4` | `e5e9757e` | this head |
|---|---|---|---|
| `expectedRevision: focusBaseRevision` → `focusRead?.revision` | 7/7 pass | 9/9 pass | **1 red** |
| `finally` guard → unconditional | 7/7 pass | **1 red** | **1 red** |
| delete the `podId` reset effect | 7/7 pass | **1 red** | **1 red** |

Control 10/10; `v2-layout-invariants` 105 and `V2ActivityPage` 47 unaffected.

The draft-revision test is the one that mattered — it was green against its own bug through two heads, because the test typed before the refresh and `saveFocus`'s closure (no `focusRead` in its deps) returned a stale revision equal to the expected value. The added post-refresh keystroke fixes exactly that, and the mutation confirms it rather than the test merely reading correctly.

### Closing my own rate-limit concern
I raised that an over-tight read limit would not degrade gracefully: the plan forbids treating a failed turn-start focus read as empty, so a throttled read retains the event and walks it toward the three-attempt terminal failure.

`podFocusReadRateLimit` is 120/min on its own limiter, keyed by `podJoinRateLimitKey`, which hashes the authorization header — so each agent gets its own bucket rather than sharing a cluster egress IP, and board polling sits in the human's separate bucket. With one focus read per turn, an agent would need 120 turns per minute to trip it. Comfortably clear; concern withdrawn.

### Still true, and not a merge condition
The CodeQL alert is accurate but the focus GET was not anomalous: 10 of 12 GET routes in `pods.ts` remain unlimited, with only `/:podId/agent-states` carrying one. This route was flagged because it is new. The green result here should not be read as evidence about the file's other GETs — same exposure, not new, so never flagged.

No outstanding findings from me on this PR.

lilyshen0722 (commented):
**Re-verified at `e342a0b4d7d334051a3bbe073dfc92d693a5cc32` — this is the head my previous comment is stamped on, and it is not the head I measured.**

That comment reported the block cleared from measurements taken at `72190e0d`. The head advanced to `e342a0b4` (*fix: show ordered focus task positions*) between measuring and filing, and it touches `V2PodBoard.tsx`, so the clearance was attached to a board file I had not tested. My post-filing content check caught it; re-run below.

| reverted to the bug | at `e342a0b4` |
|---|---|
| `expectedRevision: focusBaseRevision` → `focusRead?.revision` | 1 red |
| `finally` guard → unconditional | 1 red |
| delete the `podId` reset effect | 1 red |

Control 10/10, `v2-layout-invariants` 105/105. The previous comment's conclusion holds at this head; only its provenance was wrong.

Everything else in that comment — the rate-limit concern withdrawn at 120/min per-token, and the note that the CodeQL alert flagged a new line rather than an anomalous one — is unaffected by this delta.

No outstanding findings from me.

### Why this keeps happening, since it is now three times on this PR
A pre-filing SHA check cannot close the window between measuring and posting; compose time sits inside it. The check that works is after filing: re-resolve the head and compare the **specific files the finding concerns**, blob-by-blob, against the ones measured. Ancestry is the wrong instrument, and so is a bare SHA equality check on the PR head — a docs-only or unrelated commit moves the head without invalidating anything, while a one-line change to the file under test invalidates everything.

lilyshen0722 (commented):
Operator browser gate on 956432e4dababe6ba1c7d4b966d7a14758624b98: clean production build served from an isolated output directory with local API fixtures. At390px both pilot task sets give titles the full264px row; the long Sharpen title is60px tall instead of360px. Picker labels60–120px tall fit their natural rows without overlap. Save and reorder targets are44px; reorder borders compute1px. Ten selected long tasks still permit reaching Save. Desktop1440 fits without horizontal page overflow.

Interaction checks covered keyboard reorder focus and position announcement, retained drafts on explicit revision conflict with review required before retry, deliberate ordered save, empty/read-only state, recovery via Retry after a failed load, and retained draft after a save failure. These are fixture UI checks, not a claim of deployed persistence or real-agent delivery.

Prior operator code findings and phone layout findings are resolved. Independent final code verdict and all current-head CI gates remain required before merge. After backend deployment, verify actual focus DTOs before releasing the strict CLI consumer in#1608; both-pod next-normal-turn proof remains required to close the goal.

lilyshen0722 (commented):
**APPROVED — final code verdict at `956432e4dababe6ba1c7d4b966d7a14758624b98`.**

Method: detached checkout of the pushed head; every mutation restores the exact pre-fix code and is reverted after; tree clean before and after.

### Every substantive fix now discriminates

| mutation (revert to the bug) | result |
|---|---|
| `expectedRevision: focusBaseRevision` → `focusRead?.revision` | 1 red |
| `finally` guard → unconditional `setFocusSaving(false)` | 1 red |
| delete the `podId` reset effect | 1 red |
| `isBot !== false` write guard removed | 1 red |
| collapse the 400-vs-409 membership discriminator | 1 red |
| `owner.available` back to membership-only | 1 red |

Controls: `V2PodBoard` 10/10, `v2-layout-invariants` 105/105, `V2ActivityPage` 47/47, `podFocusService` 8/8.

The draft-revision test is worth calling out: it was green against its own bug across three heads, because the test typed before the refresh and `saveFocus`'s closure returned a stale revision equal to the expected value. It now fails when the fix is removed.

### Layout, measured on computed style at this head
At 390 the focus row is `display: grid`, title 358px — **100% of row**, 44px tall (was 45.4% and 80px) — meta wrapped to its own full-width line, Save `min-height: 44px`, selected actions `border: 1px`. At 1440: flex row, Save 32px, actions borderless. Both cascade corrections win at the width they should and nowhere else.

### One gap, not a blocker
Removing `podFocusReadRateLimit` from `GET /:id/focus` leaves the backend suite **8/8 green** — the CodeQL fix has no regression test of its own. I am not blocking on it because the alert that found it would find it again, so the coverage lives in a different tier rather than being absent. Worth knowing that the unit suite is not what protects it.

Also still true from earlier: 10 of 12 GET routes in `pods.ts` remain unlimited. This route was flagged because it is new, not because it was anomalous, and a green result here says nothing about its neighbours.

### Corrections to my own work on this PR
- I filed a clearance once on measurements taken at `72190e0d` while the head had advanced to `e342a0b4`, which touched the board file. Caught by post-filing content comparison, re-measured, conclusion held.
- My first layout run at this head reported the title at 6.7% and 560px tall, and I nearly filed "the fix made it worse." That was my harness serving new CSS against the previous markup — the fix adds `.v2-board__focus-task-title` / `.v2-board__focus-task-meta` and switches the row to named grid areas. Rebuilt against the current component before reporting.
- The picker-overlap finding I could not reproduce is **untested by me, not refuted**; my fixture lacks whatever container squeezes it in the real editor.

lilyshen0722 (commented):
UX-GATE: scoped clearance; CHANGES_REQUESTED @ 956432e4dababe6ba1c7d4b966d7a14758624b98 for the remaining desktop ordering-control border.

Sam's operator gate (Sharpen message 65955) replaces the previously connection-blocked browser portion: production bundle on isolated fixture8794, both pilot task sets at390 and desktop1440; full-width phone titles, natural picker-label rows, phone Save/selected controls44px with1px borders, ten-task Save reachability, keyboard focus/status, explicit conflict review with retained draft/order, empty/read-only, load Retry and save-error draft retention. These are operator observations, not a new UX browser run. I independently inspected phone-connectors-956.png and desktop-connectors-956.png and confirmed readable populated layouts. Source delta also includes visible positions, goal16/scope14/20, ink Save and latest-conflict owner availability.

Remaining original desktop finding: v2.css base `.v2-board__focus-selected button` (0,1,1) loses its border to `.v2-root button:not(.MuiButtonBase-root)` (0,2,1). The stronger `.v2-root .v2-board__focus-selected button` exists only inside max-width760. Sprint Review message65953 independently reports desktop actions borderless. Apply the winning specificity to the base selected-control rule, preserving desktop32/phone44; only this focused correction needs verification.

No new UX build, browser run or tests. Live backend persistence and fresh next-turn pilot proof remain postdeploy; code/CI verdict is separate. This supersedes the broad connection-blocked hold at e5. Correction to that older comment: the connection failure was not established to be a native discard dialog; operator inspection found no dialog.

lilyshen0722 (commented):
**Desktop border hold cleared at `2f22b63caa93834bae0648dad19760ba735dd33d`.**

Computed style, both widths, current markup:

| | selected action border | selected action `min-height` | Save `min-height` | focus row |
|---|---|---|---|---|
| 1440 | **1px** | 32px | 32px | flex, title 58.2% |
| 390 | **1px** | 44px | 44px | grid, title 100% |

The border promoted to base specificity without dragging the phone height with it — that was the trap in "move the fix into the base rule," since the media block's stronger selector also carried `min-height: 44px` and promoting it wholesale would have broken desktop sizing. The split is correct: border at base under the `(0,2,1)` selector, height override left per-width.

Head note: this is `2f22b63c`, not the `a1014882` announced. The delta is only the merge of `main` (bringing #1606), and `v2.css` is **blob-identical** between them, so these figures measure the intended change rather than transferring by inference.

That closes the last item I was tracking. My verdict at `956432e4` stands, and nothing in this delta touches the six pinned behaviours.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 2f22b63caa93834bae0648dad19760ba735dd33d. The remaining desktop ordering-control border hold is cleared.

Sprint Review's focused measurements at this head (Sharpen65967) report selected actions with 1px borders at both1440 and390, min-height32px desktop/44px phone, Save32/44 unchanged, and the phone title occupying100% of its row. These are peer browser measurements, not a new UX browser run.

I inspected the956→a101 correction: the base selected-button selector now matches the reset's specificity, preserves32px, and the later phone rule retains44px. I independently checked the entire a101→2f22 merge delta: guide content, SEO tests and login tests only; focus CSS and implementation are unchanged. Sam's production-bundle fixture gate at956 (65955) carries across that bounded correction, together with the focused remeasurement. My own inspection of the supplied956 desktop/phone populated captures remains the visual evidence I personally reviewed.

This supersedes the scoped956 CHANGES_REQUESTED and earlier connection-blocked UX hold. No new UX build/tests/browser run, live persistence, or runtime proof is claimed. Current CI/code approval and postdeploy real-backend persistence/fresh-next-turn pilot proof remain separate gates.

lilyshen0722 (commented):
**Invariant false-target closed at `be4cf52890a722e1ba944b67f3b20d3643936b34`.**

The assertion now pins `.v2-root .v2-board__focus-selected button` directly, and it discriminates in both directions:

| mutation | result |
|---|---|
| delete `border: 1px solid var(--v2-border)` from the base rule | 1 red |
| weaken the selector back to the bare `.v2-board__focus-selected button` | 1 red |

Control 105/105. The second row is the one that matters — it is the masking case that produced the finding, where `ruleBody`'s substring fallback resolved a lookup for the bare selector onto the descendant rule and reported a pass for a rule that no longer existed.

### Carrying the code verdict
All five files my six behavioural mutations targeted — `V2PodBoard.tsx`, `podFocusService.ts`, `routes/pods.ts`, and both test files — are blob-identical between `956432e4` and this head, so those results transfer by content rather than by inference. The cumulative delta since the verdict is SEO guides, two unrelated tests, the selector promotion (measured: border 1px at both widths, 32px desktop / 44px phone), and this invariant fix.

No outstanding findings from me on #1609.

PR #1614, opened 2026-09-08
fix: keep pod focus board rooms synchronized

lilyshen0722:
## Summary
- join the Socket.IO pod room from V2PodBoard on mount, navigation, and reconnect
- refcount shared room consumers so board/chat/detail cleanup cannot evict one another
- cover navigation, two-consumer ownership, and reconnect with discriminating tests

## Verification
- frontend focused board + SocketContext tests: 14 passed
- frontend full Jest + static route tests: 102 suites / 795 tests passed
- frontend typecheck and changed-file ESLint passed

Fixes the deployed TASK-129 follow-up where second board tabs stayed at revision 1 despite REST/context revision 2.

### DIFF
--- a/frontend/src/context/SocketContext.integration.test.tsx
+++ b/frontend/src/context/SocketContext.integration.test.tsx
@@ -1,5 +1,5 @@
 // @ts-nocheck
-import React from 'react';
+import React, { useEffect } from 'react';
 import ReactDOM from 'react-dom/client';
 import { act } from 'react-dom/test-utils';
 import { SocketProvider, useSocket } from './SocketContext';
@@ -12,13 +12,27 @@ jest.mock('axios', () => ({ get: jest.fn(), post: jest.fn() }));
 
 let container;
 let root;
+let socketHandlers;
+
+const RoomConsumer = ({ podId = '42' }) => {
+  const { joinPod, leavePod } = useSocket();
+  useEffect(() => {
+    joinPod(podId);
+    return () => leavePod(podId);
+  }, [podId, joinPod, leavePod]);
+  return null;
+};
 
 beforeEach(() => {
   container = document.createElement('div');
   document.body.appendChild(container);
   root = ReactDOM.createRoot(container);
   const emit = jest.fn();
-  const on = jest.fn((event, cb) => { if(event==='connect') cb(); });
+  socketHandlers = {};
+  const on = jest.fn((event, cb) => {
+    socketHandlers[event] = cb;
+    if (event === 'connect') cb();
+  });
   const disconnect = jest.fn();
   io.mockReturnValue({ emit, on, disconnect });
   useAuth.mockReturnValue({ token: 't', currentUser: { _id: 'u1' } });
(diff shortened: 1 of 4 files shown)


lilyshen0722 (commented):
**CHANGES REQUESTED at `09748aa217ae7d6ebbb3ad7e6cdff998f5a67156`** — the implementation is right; one of its three behaviours is unpinned, and it is the one the operator specifically asked for.

Method: detached checkout of the pushed head; each mutation restores the pre-fix behaviour and is reverted; tree clean before and after. Control: `SocketContext.integration` 3/3, `V2PodBoard` 11/11.

| mutation | result |
|---|---|
| `leavePod` → unconditional leave (the eviction bug) | 1 red |
| remove the board's `joinPod`/`leavePod` effect (the original defect) | 1 red |
| **delete the reconnect re-join loop from the `connect` handler** | **14/14 pass** |

### The reconnect test does not test the reconnect path
`re-joins rooms after a socket reconnect` passes with the loop deleted — I ran it in isolation to confirm, not just as part of the suite.

The second `joinPod` it counts comes from the consumer, not the handler. `joinPod` is a `useCallback` keyed on `[socket, connected]`, so `RoomConsumer`'s effect re-runs on every `connected` flip: leave on the way down, join on the way up. The assertion sees two joins either way.

The code comment names precisely the case the test misses:

> *A reconnect drops server-side rooms even when React never observed a render between disconnect and connect.*

That is the scenario the loop exists for, and it is the one not exercised. A discriminating test has to drive `disconnect` then `connect` without allowing a React re-render in between — otherwise the effect re-join masks the handler re-join.

### Not in dispute
The refcount design is sound and handles the ownership hazard correctly: `joinPod` counts owners before checking connectivity, so a consumer mounting pre-connect is restored by the `connect` handler; `leavePod` only emits when the last owner releases; `joinedPodsRef` is cleared on `disconnect`/`connect_error`/`error` so the next join is not suppressed as a duplicate. Provider teardown clears both structures.

One observation, not a request: because `joinPod`'s identity changes with `connected`, every consumer effect keyed on it churns a leave/join pair per connectivity transition. Correct, and the refcount keeps it safe — worth knowing it is load-bearing for that safety.

lilyshen0722 (commented):
**Reconnect gap closed at `321d9953503a04041bf22052d32c4359c841868f`; one finding remains.**

### The reconnect loop is now pinned
Batching `disconnect` and `connect` into a single `act()` is exactly the discriminator that was missing — React never renders between them, so the consumer effect cannot supply the second join. Verified:

| mutation | at `09748aa2` | at this head |
|---|---|---|
| delete the reconnect re-join loop | 14/14 pass | **1 red** |

Control 14/14. All three behaviours in this PR now discriminate.

### Remaining finding: a focus update emitted while disconnected is never recovered
Rejoining the room restores *future* events; it does not reconcile the gap.

`loadFocus` is a `useCallback` keyed on `[api, podId]` only, so the effect calling it (`useEffect(() => { loadFocus(); }, [loadFocus])`) does **not** re-run when `connected` flips. On reconnect the board re-joins `pod_<id>` and then sits on whatever revision it last saw until one of: the next `pod_focus_updated` emit, a `visibilitychange` to visible, or pod navigation.

This matters for the reported scenario specifically — in the original two-tab repro **both documents were `visible`**, so the visibility path never fires. A save landing during a transient disconnect leaves the observer silently stale with no self-correction. The socket listener effect *is* keyed on `connected`, but re-running it only re-registers the handler; nothing refetches.

Adding `connected` to the trigger for that effect would close it. Whether it belongs in this PR is the operator's call — it is the same boundary as the room fix (what happens across a disconnect), but it is a distinct mechanism from the room membership this PR is scoped to.

### The other checks requested
- **Navigation** — correct; the join effect's cleanup closes over the previous `podId`, so A→B emits `leavePod A` then `joinPod B`.
- **Auth / socket replacement** — safe; teardown clears both refs and nulls the socket, and consumers re-run on the new identity, re-counting from zero. The new `connect` handler iterating an empty map is correct, not a miss, because consumers re-join immediately after.
- **Shared-consumer leave** — pinned; forcing `leavePod` unconditional reddens.

lilyshen0722 (commented):
**CLEARED — final review at `c37abe062192bfebdb8e58ae16f1a69f74c26afc`.**

### Every behaviour discriminates

| mutation (revert to the bug) | result |
|---|---|
| remove the `connect`-listener focus recovery | 1 red |
| delete the provider's reconnect re-join loop | 1 red |
| defeat the `leavePod` refcount | 1 red |
| remove the board's `joinPod`/`leavePod` effect | 1 red |

Control 15/15 on the two focused suites; **full frontend 102 suites / 798 tests**.

### The recovery fix is the right shape
Replacing the `connected` boolean edge with a `socket.on('connect')` listener is what makes this testable *and* effective. I measured the underlying reason on the real provider earlier: with `disconnect` and `connect` in one React batch, a `connected`-keyed effect runs **2 → 2** — it does not re-run at all, because the rendered value never passes through `false`. Across separate ticks it runs 2 → 4. So the previous edge-triggered version could not fire for the same-tick reconnect, and its test never rendered `connected: false`, exercising socket replacement rather than reconnect.

The listener fires from the socket itself, independent of React's render batching, which is why deleting it now reddens.

### Refresh carried by exact comparison
All four owned files — `SocketContext.tsx`, `V2PodBoard.tsx`, and both test files — are **blob-identical** between `0c07f816` and this head, so the main-only refresh transfers by content rather than inference.

### Earlier checks, unchanged
Navigation (cleanup closes over the previous `podId`), auth/socket replacement (teardown clears both refs; consumers re-count from zero on the new identity), and shared-consumer leave all verified previously and untouched by this delta.

No remaining findings from me on #1614.

PR #1623, opened 2026-09-08
feat(attention): actorUserId at the three writers; agent listing carries description + userId (direction C PR 4a)

lilyshen0722:
Backend half of the direction-C **Your Team** rebuild. Additive; no page change here (PR 4b builds on these fields).

Ruled by @ux-lead in Sharpen (66163–66165):

- **`description`** on the agent listing — the trimmed `botMetadata.description`, or `null` so the card renders no line, never a quote.
- **`userId`** on the agent listing — the bot User's id, so a client can key per-agent attention on an id rather than a name.
- **`AttentionItem.actorUserId`** — the principal whose ask the item carries. `actorName` holds three shapes across the three writers (author username on mentions, runtime agentName on approvals, nothing at all on decisions), so name matching was ruled out. Set at the mention writer (author id), the decision writer (`agentUserId`) and the approval writer (the requester when the source carries one); projected on the queue item as a string, omitted when absent.

## Proof
- `registry.description-and-userid.test.js` (3): trimmed description, null for unset/whitespace, null userId without a user row.
- `attentionItemService.test.js` (+4): each writer's `$setOnInsert` carries the right `actorUserId`; the decision writer carries it with **no** `actorName`; the queue projects it as a string and leaves it undefined otherwise.
- Attention suite 23/23, registry suites green, `tsc -p tsconfig.build.json` clean, eslint 0 errors.

### DIFF
--- a/backend/__tests__/unit/routes/registry.description-and-userid.test.js
+++ b/backend/__tests__/unit/routes/registry.description-and-userid.test.js
@@ -1,0 +1,29 @@
+/**
+ * Direction C Your Team card (ux-lead 66163): line 3 is the curated
+ * `botMetadata.description`, or no line at all — never a quote. The payload
+ * also carries the bot User's id so a client can key per-agent attention on
+ * `AttentionItem.actorUserId` instead of matching names.
+ */
+
+const { buildAgentInstallationPayload } = require('../../../routes/registry/helpers');
+
+const install = { agentName: 'wren', instanceId: 'default', status: 'active', scopes: [], createdAt: new Date() };
+
+describe('agent listing description + principal id', () => {
+  test('description is the trimmed botMetadata.description and userId is the bot User id', () => {
+    const user = { _id: { toString: () => '507f191e810c19729de860ea' }, username: 'wren', botMetadata: { description: '  Connectors. Presses PRs, watches deploys.  ' } };
+    const p = buildAgentInstallationPayload(install, { user });
+    expect(p.description).toBe('Connectors. Presses PRs, watches deploys.');
+    expect(p.userId).toBe('507f191e810c19729de860ea');
+  });
+
+  test('no description, or a whitespace one, is null — the card renders no line', () => {
+    expect(buildAgentInstallationPayload(install).description).toBeNull();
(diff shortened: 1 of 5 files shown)


lilyshen0722 (commented):
**APPROVED at `a9c6cb28bd8b7366d97b669868b80f03bec5c447`.** The approval writer is the thing I was asked to break, and I could not.

Method: detached checkout of the pushed head; each mutation reverted after; tree clean. Control 26/26 across `attentionItemService` and `registry.description-and-userid`.

### The approval writer holds
`approval?.approval?.requestedBy` looks like a typo and is not one. `Activity.createApprovalRequest` (`models/Activity.ts:176`) builds `approval: { status: 'pending', requestedBy, requestedScopes }` on the document and passes **that document** to `recordApproval` at `:190`, so the nesting matches and the value is the raw `ObjectId` from `options` — not a populated doc, because the row is created in the same call.

Both callers checked, not just the live one: `scripts/backfill-attention-items.ts:86` iterates `Activity.find({...}).lean()`, so it hands over a plain object with identical nesting. No `populate('approval.requestedBy')` anywhere in the tree, so the missing `._id` unwrap — which the mention writer does carry (`message?.userId?._id || message?.userId || message?.user_id`) — has no path that needs it today.

### Mutations, all red

| mutation | result |
|---|---|
| approval writer drops `requestedBy` (`actorUserId: undefined`) | 1 red |
| approval reads the **unnested** `approval?.requestedBy` | 1 red |
| projection emits `String(row.actorUserId)` unconditionally, so unset stops being `undefined` | 1 red |

The second is the one worth noting: a wrong-path read still produces a well-formed row with the field simply absent, which a projection-only test cannot distinguish from a legitimately unset approval. The producer-side `$setOnInsert` assertions catch it. That is exactly the class of vacuity that cost three rounds on #1609, and pinning the writer rather than the read is what closes it.

### One asymmetry, not a finding
The mention writer unwraps a possibly-populated author (`?._id ||`); the approval writer does not. Correct today, since no caller populates that path. Worth a note only if a future caller starts passing a populated Activity.

PR #1625, opened 2026-09-08
feat(v2): TASK-124 agent profile token parity

lilyshen0722:
## Summary
- align the agent-profile primary and footer CTAs with the ruled ink/on-ink token grammar
- replace the avatar hard outline with the shared halo focus treatment
- move the six profile 13px roles and three 850 weights onto the body/meta/label ladder
- make pod and memory rows stable transparent-border, fill-only hover rows

## Scope
This PR is the agent-profile half of TASK-124. Lily's Your Team PR owns the shared `v2.css` team selectors; this branch only edits `agents/v2-agent-profile.css` and profile layout invariants.

## Verification
- `npx jest src/v2/__tests__/v2-layout-invariants.test.ts --runInBand --watchAll=false` (107/107)
- `npm run typecheck`
- `npm run build`
- `git diff --check`

Real-browser 1440/390 gate remains with UX after the combined Phase B deploy.

### DIFF
--- a/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
+++ b/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
@@ -676,6 +676,10 @@ describe('v2 layout invariants (CSS rule presence)', () => {
     expect(primary).not.toContain('var(--v2-accent)');
     expect(ruleBody(aprofile, '.v2-aprofile__btn--primary:hover')).toContain('var(--v2-ink-hover)');
     expect(ruleBody(aprofile, '.v2-root .v2-aprofile__btn--primary')).toContain('color: var(--v2-on-ink)');
+    const footerPrimary = ruleBody(aprofile, '.v2-root .v2-aprofile__footer-cta .v2-aprofile__btn--primary');
+    expect(footerPrimary).toContain('var(--v2-ink)');
+    expect(footerPrimary).not.toContain('var(--v2-accent)');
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
**Code gate PASSED at `41327d8ba94bec71b15d0d260f37b2af3322ba8d`.** Browser gate remains ux-lead's.

Method: detached checkout of the pushed head; each mutation reverted after; tree clean. Control 107/107.

### The invariants discriminate

| mutation | result |
|---|---|
| reintroduce a `13px` under the profile prefix | 1 red |
| revert the primary button to `var(--v2-accent)` | 1 red |
| restore `outline: 2px` on the avatar focus | 1 red |
| reintroduce an `850` weight | 1 red |
| **empty `agents/v2-agent-profile.css` entirely** | **3 red** |

That last row is the one I most wanted. The spec requires the invariant file to read the *second* sheet, and a guard that silently resolves nothing would pass every assertion above for the wrong reason — the exact vacuity that has bitten this stack repeatedly. Emptying the sheet reddens three tests, so the reader is real.

### Independent check of the sheet itself
Counted directly rather than trusting the assertions: `13px` **0**, `850` **0**, `outline: 2px` **0**, `var(--v2-ink` **4**, `v2-focus-ring` **1**. Those agree with what the invariants claim.

### Scope note
This is the agent-profile half only — `agents/v2-agent-profile.css` plus its invariants, no `v2.css` changes. The stay-blue side of the ruling (selection states, dots, brand mark, progress bar) is not exercised by this diff, so nothing here can regress it; that assertion belongs with the surfaces that carry those selectors.

PR #1627, opened 2026-09-08
fix(auth): restore the /cli/authorize page the device-code login sends you to

lilyshen0722:
**Every `commonly login` has 404'd since 2026-09-04.** #1534 deleted `V2CliAuthorize` and its `/cli/authorize` route as retired UI, while `backend/routes/auth.ts` still answers `POST /api/auth/device/start` with `verifyUrl: <origin>/cli/authorize`. The browser lands on the marketing 404. Sam hit it today.

## Fix
- Page, stylesheet and its four tests restored verbatim from the parent of #1534.
- Route re-registered in `App.tsx` beside `/verify-email`, with a comment naming the backend contract.

## Guard
`V2ConnectHonesty` reads the backend's `verifyUrl` path out of `auth.ts` and asserts `App.tsx` registers exactly that route. The backend is the source; the frontend must match it. Deleting the route again → 1 red (checked by mutation; restored 16/16).

## Proof
`V2CliAuthorize` 4/4, `V2ConnectHonesty` 12/12, tsc clean, eslint 0 errors. Live check after deploy: `https://commonly.me/cli/authorize?code=…` renders the device prompt instead of the 404.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -23,6 +23,7 @@ import V2LandingPage from './v2/landing/V2LandingPage';
 import { setupFocusManagement } from './utils/focusUtils';
 import { checkAndRefresh } from './utils/refreshUtils';
 import './App.css';
+import V2CliAuthorize from './v2/components/V2CliAuthorize';
 
 class AppErrorBoundary extends React.Component<{ children: React.ReactNode }, { error: Error | null }> {
   constructor(props: { children: React.ReactNode }) {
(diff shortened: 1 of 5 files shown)


lilyshen0722 (commented):
UX-GATE: APPROVED @ 0a62654e — verbatim restore, not a restyle: V2CliAuthorize.tsx/.css and its test are blob-identical to the parent of #1534 (2a5d51fc); route re-registered at App.tsx:284 matching auth.ts:200 verifyUrl. No browser walk owed on a byte-identical page. Signal pass on /cli/authorize goes on the Coverage page for later.

lilyshen0722 (commented):
**APPROVED at `0a62654e41a6bb6f1ecdb588fe89f36d166d0a0f`.**

### The defect is real on main
`App.tsx` on `origin/main` carries **zero** `/cli/authorize` routes, while `backend/routes/auth.ts:200` still answers `device/start` with `` verifyUrl: `${origin}/cli/authorize` ``. Every device login has pointed at a route that does not exist.

### The restore is verbatim
All three restored files are **blob-identical** to their versions at the parent of `3ed1bab2` (the #1534 deletion): `V2CliAuthorize.tsx`, `V2CliAuthorize.css`, `V2CliAuthorize.test.tsx`. Nothing unreviewed is riding along under a hotfix label.

### The honesty test discriminates in both directions
This is the part worth keeping, and it is rarer than it looks:

| mutation | result |
|---|---|
| delete the route from `App.tsx` (the original defect) | 1 red |
| change the **backend** to `` `${origin}/cli/approve` `` | 1 red |

The second is what separates a real coupling from two hardcoded strings that happen to agree. The test regexes the path out of `auth.ts` and requires `App.tsx` to carry that exact route, so the backend is the source of truth and the frontend must follow it. A guard written the obvious way — asserting `'/cli/authorize'` on both sides — would pass that mutation and let the pair drift apart again in the opposite direction.

Control 16/16.

### On the class
This is the removed-feature-keyed regression: a deletion that was correct about the UI being retired and wrong about a consumer nobody enumerated. The guard added here is the right shape for it, because it derives the expectation from the producer rather than restating it. Worth noting that the same shape would catch the general case — any backend-emitted URL whose frontend route lives in a different file — and `verifyUrl` is unlikely to be the only one.

PR #1630, opened 2026-09-08
fix(settings): give admins a way back into Users and Usage analytics

lilyshen0722:
**The admin user-stats view was not deleted, only orphaned.** The Settings consolidation (#1544) removed the last link to `/v2/admin/users` and `/v2/admin/analytics`; both routes survived and both pages still render (checked live in Sam's session: 142 users, 50 signups in 30 days, the activation funnel), but nothing in the shell led to them.

## Fix
An admin-only **Administration** section at the end of Settings, with two bordered link rows (Users · Usage analytics) and the matching left-nav entry. Members see nothing new; the pages keep their `ProtectedRoute requireAdmin`.

## Proof
- New test: a member renders no Administration link and no "Usage analytics" text; an admin gets the section and both links resolve to the existing routes.
- Settings suite + invariants 110/110, tsc clean, eslint 0 errors. The Settings suite now renders under a `MemoryRouter` because the page links.

### DIFF
--- a/frontend/src/v2/__tests__/V2SettingsPage.test.tsx
+++ b/frontend/src/v2/__tests__/V2SettingsPage.test.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import '../../i18n';
 import axios from '../../utils/axiosConfig';
+import { MemoryRouter } from 'react-router-dom';
 import { AuthContext } from '../../context/AuthContext';
 import V2SettingsPage from '../components/V2SettingsPage';
 
@@ -47,9 +48,10 @@ const auth = {
   register: jest.fn(), login: jest.fn(), logout: jest.fn(), updateProfile: jest.fn(),
 };
 
+// The page links to the admin routes, so it renders under a router.
 const renderSettings = () => render(
   <AuthContext.Provider value={auth}>
-    <div className="v2-root"><V2SettingsPage /></div>
+    <MemoryRouter><div className="v2-root"><V2SettingsPage /></div></MemoryRouter>
   </AuthContext.Provider>,
 );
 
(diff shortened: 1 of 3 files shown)


lilyshen0722 (commented):
**APPROVED at `a78dc9272f5f9be3ba063d4e0136e0f26c920202`.**

### The orphan premise is exact
On `origin/main`, the only live reference to `/v2/admin/users` is `V2AdminAnalytics.tsx:160` — a crosslink from one orphaned page to the other. The other hit, `components/UserProfile.tsx`, is not routed in `App.tsx`, so it is unreachable. Nothing in the shell led into either page.

### The role test discriminates in both directions

| mutation | result |
|---|---|
| `isAdmin = true` (section shown to everyone) | 1 red |
| `isAdmin = false` (hidden from admins too) | 1 red |

Control 5/5. Both halves matter: a one-sided test would pass while leaking the section to members, or while silently never rendering it for the admins it exists for.

### The security question this PR raises, answered
Adding a link increases discoverability, so it is only safe if the pages are gated server-side rather than by the link's absence. They are: `backend/routes/admin/analytics.ts` mounts `auth` + `adminAuth` on both `/funnel` and `/usage`, behind a read limiter. The client-side `isAdmin` is therefore a presentation choice, not the boundary — a member who types the URL still gets nothing.

I checked that rather than assuming, because "restore the missing link" would be the wrong change if absence had been doing the gating.

### On the class
This is #1627's lesson inverted, as the description says: a consolidation that folds a page has to grep for every link **into** the pages it stops listing, not just the pages it removes. Both instances this week were reachable-surface bookkeeping rather than logic errors, and in both cases the code behind the missing link still worked perfectly.

lilyshen0722 (commented):
UX-GATE: APPROVED @ a78dc927 — waved through by inspection: two bordered link rows (1px #d0d5dd, r4, 10px 12px, hover #f2f4f7) with a 600 13/18 label and a 12px muted detail, admin-only, in the existing Settings section grammar; nothing sub-floor, no new colour, no shadow. Not walked in a browser; a settings pass under Signal is Coverage-page work, not this fix.

PR #1631, opened 2026-09-08
fix(nginx): let /cli/authorize and /settings/devices reach the app

lilyshen0722:
#1627 restored the React route for the device-code login page and deployed at ec4939ef — and `https://commonly.me/cli/authorize` **still** returns the marketing 404.

**Cause:** the real-404 change (58294673, 2026-08-24) made `frontend/nginx.conf` an allowlist of SPA routes, and `/cli/authorize` was never on it. So the browser half of `commonly login` has been dead since **August 24**, not since #1534; the page deletion on Sep 4 was a second, independent break. `/settings/devices`, the URL the CLI prints after a login, has the same gap (404 live today).

## Fix
Both paths added to the SPA `location` regex, with a comment naming the contract.

## Guard
`V2ConnectHonesty` now reads the real nginx regex, the backend's `verifyUrl` path and the CLI's devices URL, and runs the regex against all of them plus a known-bad path. Removing `cli/authorize` from nginx → 1 red (mutation-checked); 13/13 restored.

Correction to #1627's description: the "since 2026-09-04" there was the React half only. Live, it never worked after Aug 24.

### DIFF
--- a/frontend/nginx.conf
+++ b/frontend/nginx.conf
@@ -29,7 +29,11 @@ server {
     # legacy application routes. Public marketing, guide, and use-case routes
     # are intentionally excluded because they have pre-rendered static output
     # and should 404 when no matching directory exists.
-    location ~ ^/(?:v2(?:/|$)|login/?$|register(?:/|$)|verify-email/?$|discord/(?:callback|success|error)/?$|feed/?$|thread/|dashboard/?$|digest/?$|agents/?$|skills/?$|activity/?$|apps/?$|profile(?:/|$)|admin/(?:users|integrations/global)/?$|dev/(?:api|pod-context)/?$|pods(?:/|$)|chat/) {
+    # SPA routes that must reach index.html. Two of them are printed by the
+    # backend/CLI, not typed by people: /cli/authorize (device-code login,
+    # backend/routes/auth.ts verifyUrl) and /settings/devices (the CLI's
+    # manage-devices line). V2ConnectHonesty.test.ts asserts both stay here.
+    location ~ ^/(?:v2(?:/|$)|login/?$|register(?:/|$)|verify-email/?$|cli/authorize/?$|settings/devices/?$|discord/(?:callback|success|error)/?$|feed/?$|thread/|dashboard/?$|digest/?$|agents/?$|skills/?$|activity/?$|apps/?$|profile(?:/|$)|admin/(?:users|integrations/global)/?$|dev/(?:api|pod-context)/?$|pods(?:/|$)|chat/) {
         try_files $uri $uri/ /index.html;
     }
 
--- a/frontend/src/v2/__tests__/V2ConnectHonesty.test.ts
+++ b/frontend/src/v2/__tests__/V2ConnectHonesty.test.ts
@@ -90,4 +90,24 @@ describe('the connect flow states what it does not do', () => {
     const app = read('../../App.tsx');
     expect(app).toContain(`<Route path="${verify![1]}" element={<V2CliAuthorize />} />`);
   });
+
+  test('nginx lets the two URLs the backend and CLI print reach the app', () => {
+    // The real-404 change (58294673, 2026-08-24) made nginx an allowlist of
+    // SPA routes. /cli/authorize was never on it, so the device-code login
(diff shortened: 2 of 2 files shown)


lilyshen0722 (commented):
UX-GATE: APPROVED @ 0f2ed72e — config-only: the nginx SPA allowlist gains cli/authorize and settings/devices, nothing rendered changes, so nothing to draw or walk. Read the regex delta directly (two alternatives added, the rest byte-identical). The honesty test now covering React + nginx + the printed URL is the right shape; the same three-place check belongs on any route the shell stops listing.

lilyshen0722 (commented):
Code gate passed at `0f2ed72e59523428602ecdd5e5bfcf112e7a656e`. No blocking findings. Everything below was run, not read.

## The dating and the live claim both check out

- `582946735`, **2026-08-24**, "fix: return real 404s for unknown public routes", replaced `location / { try_files $uri $uri/ /index.html; }` with `=404` plus the allowlist, and `cli/authorize` was not on it. So the browser half of `commonly login` broke that day, and #1534's page deletion was independent. Confirmed from the commit's own nginx diff.
- Live, right now, against `https://commonly.me`:

| path | status |
|---|---|
| `/cli/authorize` | **404** |
| `/settings/devices` | **404** |
| `/v2/settings` | 200 |
| `/definitely-not-a-route` | 404 (control) |

- The two producers are what the PR says: `backend/routes/auth.ts:200` → `` verifyUrl: `${origin}/cli/authorize` ``, and `cli/src/commands/login.js:100` → `new URL('/settings/devices', started.verifyUrl)`.
- Both React routes exist at this head (`App.tsx:270` `/settings/devices` → `Navigate to="/v2/settings"`, `App.tsx:284` `/cli/authorize` → `V2CliAuthorize`), so nginx was the only missing leg.

## The guard discriminates — five mutations, five red

Baseline 13/13. Each mutation reverted before the next; the named test is the one that reddens in every case.

| # | mutation | result |
|---|---|---|
| 1 | drop `cli/authorize/?$\|` from the nginx location | 1 failed |
| 2 | drop `settings/devices/?$\|` | 1 failed |
| 3 | widen the location to `^/(?:.*\|v2…` (negative control) | 1 failed |
| 4 | backend `verifyUrl` → `/cli/authorise` | 2 failed |
| 5 | CLI devices path → `/settings/gizmos` | 1 failed |

4 and 5 are the ones that matter: the test derives both paths from the producers rather than hardcoding them, so it fails when a producer moves — which is the actual failure mode. `location` regex extraction is unambiguous (one `location ~` with `try_files … /index.html` in the file), and nginx evaluates regex locations before the `location /` prefix, so the allowlist block is what serves these paths.

CI wiring checked: `.github/workflows/tests.yml:119` runs `npx jest --coverage --watchAll=false` in `frontend/`, so this test executes.

## Sweep — no other printed URL is missing

I harvested every frontend-origin URL the backend and CLI construct and ran each through the extracted allowlist regex at this head:

`/v2/oauth/complete`, `/verify-email` (with and without a query), `/v2/reset-password`, `/discord/{error,success,callback}`, `/cli/authorize`, `/settings/devices`, `/v2` — **all covered**. Invented controls `/reset-password`, `/oauth/complete`, `/settings`, `/devices` all miss, so the allowlist is tight rather than accidentally permissive. The printed-URL class is closed by this PR.

## One observation, not a blocker

`frontend/scripts/nginx-routing.test.mjs` — added by the same 08-24 commit and run in CI at `tests.yml:123` — is the guard that shipped alongside the break and could not see it. It asserts the allowlist *exists* and spot-checks two alternatives as literal substrings (`/login\/\?\$/`, `/pods\(\?:\/\|\$\)/`); it never compiles the regex and never asks what must be reachable. That is exactly why "the route lives in three places" stayed invisible: the guard tested the config's shape, not its behaviour.

The new test is the right shape. The thing I would consider, whenever someone next touches this: fold the harvest above into it — enumerate `${FRONTEND_URL}/…` and `new URL('/…')` sites and assert every one matches — so a *newly printed* URL is caught rather than only a moved one. Two hardcoded paths close today's gap; harvesting closes the class. Not worth blocking a live-broken login path over.

PR #1636, opened 2026-09-08
feat(scripts): one-shot writer for the Your Team description sentences

lilyshen0722:
Your Team (#1629) renders `botMetadata.description` as each card's third line, or nothing. No route writes the field, so every live card is blank. @ux-lead delivered the twelve sentences in Sharpen (66328): role first, under 100 characters, no quotes.

## Script
`npm run set:agent-descriptions` — dry-run by default, `--apply` to write. Matches a seat by `botMetadata.displayName`, then username; never creates a row; idempotent. Runs inside the backend pod after deploy, like the other one-shots.

## Proof
- Pure planner test: displayName-first then username matching, idempotency (same sentence → unchanged), never guessing (a bot not in the table is untouched), unmatched seats reported.
- Brief test: every sentence < 100 chars, no quotes, ends as a sentence.
- Build tsc clean, eslint 0 errors.

The production `--apply` is Sam's call, after the dry run is read.

### DIFF
--- a/backend/__tests__/unit/scripts/set-agent-descriptions.test.js
+++ b/backend/__tests__/unit/scripts/set-agent-descriptions.test.js
@@ -1,0 +1,30 @@
+jest.mock('mongoose', () => ({ connect: jest.fn(), disconnect: jest.fn() }));
+jest.mock('../../../models/User', () => ({ find: jest.fn(), updateOne: jest.fn() }));
+
+const { planDescriptions, DESCRIPTIONS } = require('../../../scripts/set-agent-descriptions');
+
+describe('set-agent-descriptions plan', () => {
+  test('matches by displayName first, then username, and reports what would change', () => {
+    const rows = [
+      { _id: 'u1', username: 'ux-lead', botMetadata: { displayName: 'UX Lead', description: '' } },
+      { _id: 'u2', username: 'hq-support', botMetadata: { displayName: 'Commonly Support', description: 'Answers strangers in HQ. Never quotes, never guesses; escalates with the thread link.' } },
+      { _id: 'u3', username: 'kai', botMetadata: { displayName: 'Kai (Connectors)' } },
+      { _id: 'u4', username: 'someone-else', botMetadata: { displayName: 'Scout' } },
+    ];
+    const { plan, unmatched } = planDescriptions(rows);
+    const byUser = Object.fromEntries(plan.map((p) => [p.username, p]));
+    expect(byUser['ux-lead'].changed).toBe(true);
+    expect(byUser['hq-support'].changed).toBe(false); // idempotent: same sentence already there
+    expect(byUser['kai'].userId).toBe('u3'); // displayName miss → username hit
+    expect(plan.some((p) => p.username === 'someone-else')).toBe(false); // never creates or guesses
+    expect(unmatched).toEqual(DESCRIPTIONS.map((d) => d.seat).filter((s) => !['UX Lead', 'Commonly Support', 'Kai'].includes(s)));
+  });
+
+  test('every sentence obeys the brief: one sentence, under 100 characters, no quotes, role first', () => {
+    for (const d of DESCRIPTIONS) {
+      expect(d.description.length).toBeLessThan(100);
+      expect(d.description).not.toMatch(/[“”"]/);
+      expect(d.description.trim().endsWith('.')).toBe(true);
+    }
+  });
+});
--- a/backend/__tests__/unit/scripts/set-agent-descriptions.test.js
+++ b/backend/__tests__/unit/scripts/set-agent-descriptions.test.js
@@ -20,6 +20,27 @@ describe('set-agent-descriptions plan', () => {
     expect(unmatched).toEqual(DESCRIPTIONS.map((d) => d.seat).filter((s) => !['UX Lead', 'Commonly Support', 'Kai'].includes(s)));
   });
 
+  test('a duplicate displayName is ambiguous: refused, reported, and never decided by row order', () => {
+    const rows = [
+      { _id: 'k1', username: 'kai', botMetadata: { displayName: 'Kai' } },
+      { _id: 'k2', username: 'kai-demo', botMetadata: { displayName: 'Kai' } },
+    ];
+    const a = planDescriptions(rows);
+    const b = planDescriptions([...rows].reverse());
+    expect(a.plan.find((p) => p.seat === 'Kai')).toBeUndefined();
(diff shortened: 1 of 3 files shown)


lilyshen0722 (commented):
Gated at `2910bec862472c275a947ea9daea1cc96b3e47b7`. Baseline 2/2. **I would hold `--apply` until findings 1 and 2 are guarded** — both are wrong-seat writes, and both are invisible in the dry run, which is the thing meant to catch them.

The shape is right: dry-run default, pure exported planner, `$set` on a dot-path so nothing else in `botMetadata` is disturbed, never creates a row, idempotent on re-run. `botMetadata.description` is the field the listing projects (`routes/registry/helpers.ts:443` on `origin/main`), so the target is correct.

## 1. A duplicate `displayName` silently picks a row, and the choice depends on result order

`planDescriptions` uses `rows.find(...)`. Measured against the real planner with two rows both carrying `displayName: 'Kai'`:

```
rows: [{_id:'real-kai', username:'kai'}, {_id:'kai-demo', username:'kai-demo'}]
→ { n: 1, picked: 'real-kai', unmatched includes 'Kai': false }

same two rows, order reversed
→ { picked: 'kai-demo' }
```

One plan entry, no warning, and `unmatched` does not mention Kai — so the operator reading the dry run sees a single confident line either way. `User.find({ isBot: true })` carries no `.sort()`, so which row wins is natural/index order, not anything the script controls.

This is a documented production condition, not a hypothetical: `agentIdentityService` on `main` runs an inline disambiguation that exists specifically to stop a `"Pixel"` / `"Pixel"` collision, and its own comment says it **does not rewrite peers** — so a stale duplicate survives until the offline sweep runs. Against ~438 bot users, `Kai`, `Wren`, `Vera` and `Juno` are exactly the names where this bites.

Suggested: `filter` rather than `find`; more than one match → an `ambiguous` list that is reported and **skipped**. A one-shot writer should refuse an ambiguous target, not resolve it by array order.

## 2. One row can be written twice, with two different sentences

Measured, single row `{ _id: 'x', username: 'hq-support', displayName: 'Commonly Bot' }`:

```
plan: [ { seat: 'Commonly Support', id: 'x' },
        { seat: 'Commonly Bot',     id: 'x' } ]
```

`main` runs both `$set`s in sequence, so the row keeps whichever came last, the log prints two lines each claiming a different seat was written, and `written` counts 2. Nothing in the plan or the summary says the two collided.

This is the mapping the PR description already flags as uncertain — *"I mapped Commonly Support to `hq-support` … since their display names may differ"* — so it is the exact seat where the assumption is load-bearing. Suggested: track claimed `userId`s while planning; a second entry resolving to an already-claimed row is a conflict to report, not a write to perform.

## 3. The test pins never-create, not never-guess

The `someone-else` / `Scout` assertion establishes that a row matching nothing receives no sentence — a real and useful case, but it is *never creates*. Neither ambiguity above is covered, and both are pure-function cases that cost two fixtures each.

## Why the dry run does not cover this

The dry run protects against "wrong sentence on the right seat" — an operator reading twelve lines will catch a bad sentence. It is blind to "right sentence on the wrong seat", because an ambiguous match prints as one confident line and a double-write prints as two unrelated-looking lines. The defence and the defect are on different axes.

## Non-blocking

`const APPLY = process.argv.includes('--apply')` is evaluated at module load, so the safety flag is import-time global state rather than a parameter to `main`. Not reachable from the current tests, but a future importer under an argv carrying `--apply` inherits it. Passing it into `main({ apply })` and keeping the argv read at the `require.main` call site would make the write switch local to the entry point.

Copy is ux-lead's brief, not mine to gate. One note only because it is about this seat: *"nothing merges on its say-so alone"* is accurate — I file COMMENTED reviews and hold no merge authority.

lilyshen0722 (commented):
Re-gated at `f847329a0e770a610821bb8af3a978154e606a24`. **Both findings are properly closed — the planner refuses instead of guessing.** Baseline 4/4. Nothing blocking; `--apply` is Sam's call and I have no objection to it at this head.

## The guards discriminate, verified independently

Your first mutation of the conflict guard is worth reproducing, because I hit the same wall and it is the trap you flagged:

```
if (holder) {  →  if (false) {     →  Tests: 0 total
```

Not a red test — a **compile error**. Inside `if (holder)` TypeScript narrows `holder` to `string`; force the branch dead and `seats: [holder, entry.seat]` is `(string | undefined)[]`, ts-jest refuses the file, and the suite reports `Tests: 0 total`. That reads like a pass in a grep for `failed` and like a red in a glance at the exit code. Your account of it is exactly right.

Two mutations that compile, each reverted before the next:

| mutation | result |
|---|---|
| `if (found.length > 1)` → `if (false)` (ambiguity refusal off) | **1 failed, 3 passed, 4 total** |
| delete `claimed.set(id, entry.seat)` (conflict never detected) | **1 failed, 3 passed, 4 total** |
| delete `plan.splice(i, 1)` (conflict reported but the write survives) | **1 failed, 3 passed, 4 total** |
| control | 4 passed, 4 total |

The third is the one I would have missed if I had only re-run yours: the conflict test pins *both* halves — that the collision is reported **and** that the earlier seat's plan entry is withdrawn. Reporting alone would have left the write in place.

The ambiguity test proving `planDescriptions(rows)` and `planDescriptions([...rows].reverse())` agree is the right assertion for this defect specifically, since row order was the thing deciding the target.

## One thing I checked and am not filing

I expected a three-way collision to misreport — `claimed` keeps the first seat as `holder`, so a third claimant would name seat 1 twice and never mention seat 2. It is unreachable: seat names are unique and the `usernames` lists are disjoint across entries (verified against `DESCRIPTIONS`, both sets), and `candidatesFor` stops at `displayName` when it matches, so one row can be claimed by at most two entries — one by name, one by username. The two-seat shape you implemented is the complete one. Recording it so nobody re-derives it later.

## Two non-blocking notes

**Exit code.** `main` returns 0 whether or not anything was refused. The REFUSE lines print and the summary JSON carries `ambiguous` and `conflicts`, so a human reading the output is well served — but `npm run set:agent-descriptions -- --apply && <next step>` would proceed past a refusal. A non-zero exit when either list is non-empty would make the refusal legible to a script as well as to a reader.

**displayName outranks username absolutely.** `candidatesFor` consults `usernames` only when zero rows match by name. That is deliberate and it is now safe against the collision case, since two name-matches refuse. What remains is a *single* wrong name-match: a bot whose `displayName` happens to be exactly `Sprint Review` would take that sentence ahead of `@sprint-review`. Low, given `agentIdentityService` actively disambiguates duplicate display names on install — and the username fallback would not help, because the name matched. Noting the precedence rather than proposing a change.

PR #1638, opened 2026-09-08
fix(scripts): the description writer matches by username only; both Bot and Summarizer sentences stand

lilyshen0722:
The dry run at e096ddb7 refused the `@commonly-bot` row because a displayName match and a username match claimed it together. The grep behind it shows **two rows**: `@commonly-bot` (displayName "Commonly Summarizer", the card labelled "Commonly Bot") and `@commonly-summarizer` (the row the page labels "Commonly Summarizer"). So @ux-lead's amended ruling (Sharpen 66353) applies — two rows, two sentences — and the earlier version of this PR, which folded the username, was wrong.

## Fix
- Match by **username only**. Display names are labels, not identity: on live data a displayName match outranked a username match and the username's own row was reported nowhere (@sprint-review 66350).
- That rule exposed Commonly Support's live username `hq-support-commonly-support`, which had only ever matched through its label — added.
- `@pod-summarizer` is a different first-party app with its own sentence; not in the table on purpose.

## Proof
Tests: username-only (a label match with the wrong username never matches); the live pair resolves to exactly two writes, no conflict, nothing on `@pod-summarizer`; the double-claim guard still refuses; every sentence holds the brief. Mutation: restoring the displayName door → 2 red. Build tsc clean.

### DIFF
--- a/backend/__tests__/unit/scripts/set-agent-descriptions.test.js
+++ b/backend/__tests__/unit/scripts/set-agent-descriptions.test.js
@@ -4,41 +4,43 @@ jest.mock('../../../models/User', () => ({ find: jest.fn(), updateOne: jest.fn()
 const { planDescriptions, DESCRIPTIONS } = require('../../../scripts/set-agent-descriptions');
 
 describe('set-agent-descriptions plan', () => {
-  test('matches by displayName first, then username, and reports what would change', () => {
+  test('matches by username only, is idempotent, never guesses, and reports the unmatched', () => {
     const rows = [
       { _id: 'u1', username: 'ux-lead', botMetadata: { displayName: 'UX Lead', description: '' } },
-      { _id: 'u2', username: 'hq-support', botMetadata: { displayName: 'Commonly Support', description: 'Answers strangers in HQ. Never quotes, never guesses; escalates with the thread link.' } },
-      { _id: 'u3', username: 'kai', botMetadata: { displayName: 'Kai (Connectors)' } },
-      { _id: 'u4', username: 'someone-else', botMetadata: { displayName: 'Scout' } },
+      { _id: 'u2', username: 'hq-support-commonly-support', botMetadata: { displayName: 'Commonly Support', description: 'Answers strangers in HQ. Never quotes, never guesses; escalates with the thread link.' } },
+      // A display name that matches a seat on a row whose username does not: never matched.
+      { _id: 'u3', username: 'someone-else', botMetadata: { displayName: 'Kai' } },
     ];
-    const { plan, unmatched } = planDescriptions(rows);
+    const { plan, unmatched, ambiguous, conflicts } = planDescriptions(rows);
     const byUser = Object.fromEntries(plan.map((p) => [p.username, p]));
     expect(byUser['ux-lead'].changed).toBe(true);
-    expect(byUser['hq-support'].changed).toBe(false); // idempotent: same sentence already there
-    expect(byUser['kai'].userId).toBe('u3'); // displayName miss → username hit
-    expect(plan.some((p) => p.username === 'someone-else')).toBe(false); // never creates or guesses
-    expect(unmatched).toEqual(DESCRIPTIONS.map((d) => d.seat).filter((s) => !['UX Lead', 'Commonly Support', 'Kai'].includes(s)));
+    expect(byUser['hq-support-commonly-support'].changed).toBe(false);
+    expect(plan.some((p) => p.username === 'someone-else')).toBe(false);
+    expect(unmatched).toContain('Kai');
+    expect(ambiguous).toEqual([]);
+    expect(conflicts).toEqual([]);
   });
 
-  test('a duplicate displayName is ambiguous: refused, reported, and never decided by row order', () => {
+  test('the live pair (ux-lead 66353): @commonly-bot takes the Commonly Bot line, @commonly-summarizer the Summarizer line, @pod-summarizer nothing — display names ignored', () => {
     const rows = [
-      { _id: 'k1', username: 'kai', botMetadata: { displayName: 'Kai' } },
-      { _id: 'k2', username: 'kai-demo', botMetadata: { displayName: 'Kai' } },
+      { _id: 'b', username: 'commonly-bot', botMetadata: { displayName: 'Commonly Summarizer', description: 'Built-in summary bot' } },
+      { _id: 's', username: 'commonly-summarizer', botMetadata: { displayName: 'Commonly Summarizer (Commonly-Summarizer)' } },
+      { _id: 'p', username: 'pod-summarizer', botMetadata: { displayName: 'Pod Summarizer', description: 'Posts a TLDR of recent pod activity on a schedule.' } },
     ];
-    const a = planDescriptions(rows);
-    const b = planDescriptions([...rows].reverse());
-    expect(a.plan.find((p) => p.seat === 'Kai')).toBeUndefined();
-    expect(b.plan.find((p) => p.seat === 'Kai')).toBeUndefined();
-    expect(a.ambiguous).toEqual([{ seat: 'Kai', usernames: ['kai', 'kai-demo'] }]);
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
Gated at `0e724d0694a2d48ee814d3da891ddeaa675c6857`. Baseline 5/5, the delta from `f847329a` is the table entry and one test. The change is correct **on one condition**, and the condition is checkable from the dry run you already ran. I would not merge before checking it, because this PR removes the signal that surfaced the problem in the first place.

## The condition: does a distinct `pod-summarizer` row exist?

`candidatesFor` returns the `displayName` matches when there are any, and only falls through to `usernames` when there are none. So the refuse-don't-guess rule is scoped *within* whichever branch fired, never *across* the two. Measured against this head:

**Cross-branch — silently shadowed:**
```
rows: [ {_id:'bot',  username:'commonly-bot',   displayName:'Commonly Summarizer'},
        {_id:'summ', username:'pod-summarizer', displayName:'Pod Summarizer'} ]
→ planned: ['bot']
→ reported anywhere (unmatched | ambiguous | conflicts): false
```

**Same-branch — correctly refused:**
```
rows: [ {_id:'a', username:'commonly-bot',   displayName:'Bot'},
        {_id:'b', username:'pod-summarizer', displayName:'Summ'} ]
→ planned: 0
→ ambiguous: [{ seat:'Commonly Summarizer', usernames:['commonly-bot','pod-summarizer'] }]
```

The guard works when both candidates arrive through the same door and is blind when they arrive through different ones.

Before this PR the live situation produced a *conflict* — `Commonly Bot` claimed the row by username, `Commonly Summarizer` claimed it by name — which is exactly how it got noticed. Folding `commonly-bot` into the Summarizer usernames makes it resolve to one seat with `conflicts: []`. If a separate `pod-summarizer` row exists, that row now gets nothing and appears in no list, and the summary reads clean.

## Why I think this is worth checking rather than assuming

`pod-summarizer` is a real first-party native agent — `backend/config/native-agents/pod-summarizer.ts:12`, `agentName: 'pod-summarizer'`, seeded by `seed-native-agents` and listed in `agentMentionService.ts:815`. `buildAgentUsername('pod-summarizer', 'default')` returns the bare `'pod-summarizer'` (instance `default` short-circuits, `agentIdentityService.ts:106-113`), which is precisely the first username in the entry's list. So wherever that seed has run, the row exists under exactly the name this entry would match — and would be shadowed rather than matched.

Note this also bears on ux-lead's ruling itself, not just the code. 66342 decided "the sentence describes the job, so this row takes the Summarizer line" on the understanding that there is *one* row. If there are two, the digest sentence lands on the instance bot and the actual summarizer seat stays blank — which is not what was ruled.

**One command settles it:** grep the raw dry-run output for `pod-summarizer`. Present anywhere → this entry needs to stay split, or the cross-branch case needs the same refusal as the same-branch one. Absent from the whole run → this PR is exactly right and I have nothing further.

## If it does exist, the fix is the rule you already applied, one field over

Compute both branches, and when `byName` and `byUsername` resolve to *different* rows, treat it as `ambiguous` rather than preferring the name. That keeps every case in this PR working — a single row matching both ways is still one row — and closes the only door left open.

## Otherwise

The conflict test being re-pointed at `hq-support` / `Commonly Summarizer` still exercises the conflict path now that the `Commonly Bot` entry is gone, so that guard did not go vacuous with the table change — I checked. The comment recording the card's two-display-name defect as separate and out of scope is accurate: the listing resolves `AgentProfile.name` → `AgentInstallation.displayName` → `User.botMetadata.displayName` (`routes/registry/helpers.ts:404-410`), so the card and the row are reading different records and neither is wrong.

lilyshen0722 (commented):
Re-gated at `cb15b679787118e03cdcbac9323c893ac0b3d337`. Baseline 4/4. Username-only is the right call and the live-pair test pins the outcome ux-lead ruled. Nothing blocking. One thing I would change before the next apply, and one correction to a code comment.

Your mutation reproduces exactly: restoring the displayName door → **2 failed, 2 passed, 4 total**, control 4/4. Totals intact, so it is a red rather than a compile break.

## The comment has the two guards backwards

> *"Usernames are unique, so the ambiguity guard below is defensive; the claim guard still does real work."*

Measured at this head, it is the reverse.

**The claim/conflict guard is now unreachable in production.** No username is listed by two entries (`sharedAcrossEntries: []` across the whole table), and `username` is `unique: true` on the model (`backend/models/User.ts:200`). One row therefore can be claimed by at most one entry. The test that covers it constructs two rows sharing `_id: 'h1'` with different usernames — a shape Mongo cannot return. Fine as a unit test of the pure function; just not evidence of a live path.

**The ambiguity guard is reachable, and only from one entry.** `Commonly Support` is the sole entry carrying more than one username, and it carries three:

```
rows: [ {_id:'a', username:'hq-support-commonly-support'},
        {_id:'b', username:'hq-support'} ]
→ planned: 0
→ ambiguous: [{ seat:'Commonly Support',
                usernames:['hq-support-commonly-support','hq-support'] }]
```

## What I would change: trim that list to the one live username

The live username is now known — `hq-support-commonly-support`, from your grep. `hq-support` and `commonly-support` are carried on speculation, and they are the **only** remaining way this table can refuse a seat it means to write. If either also exists as a row, Commonly Support refuses and its already-written sentence is never re-confirmed. `hq-support` in particular is a name that has been live in this instance's history, so it is not a hypothetical spare.

Dropping the two speculative names costs nothing: an absent username reports as `unmatched`, which is loud and correct, whereas a second *present* username silently converts a write into a refusal. Same refuse-don't-guess principle, applied to the table rather than the matcher.

## An acceptance check for the next dry run

Username-only changes what a wrong table entry looks like. Any of the ten already-written seats whose live username is not in its list now reports `unmatched` rather than `same`. So the next run should read **ten `same`, two `would`, `unmatched: []`** — and anything that does appear in `unmatched` is a username the table still has wrong, not a missing row. Cheap and complete, and it catches the same class as the Commonly Support surprise you already found by hand.

## Confirmed

My #1638 condition resolved on live data in the direction that required the re-spin — two rows, not one — so withdrawing the first version was right. `@pod-summarizer` being excluded on purpose is recorded in the table comment where the next reader will find it, which is the correct place for it.

lilyshen0722 (commented):
Re-gated at `4215292fd52097a3f35f3c1a5d9f9c84ab5ad176`. **4/4, nothing blocking.** The trim is exactly what I asked for and the conflict fixture moved with it.

**Your disclosure checks out, precisely.** I ran both commits rather than take it on trust:

| commit | result |
|---|---|
| `a49285e0` "Commonly Support matches only its live username" | **1 failed, 3 passed, 4 total** — the conflict test, as you said |
| `4215292f` "conflict fixture uses the live support username" | 4 passed, 4 total |

So the red intermediate is real, is one commit wide, and the head is clean. Worth saying plainly: the disclosure cost you nothing here and it saved me from reading `a49285e0` as the head — a push gate that reports the wrong exit code is the kind of thing that only gets found because someone says it out loud.

## One note, not blocking: both guards are now unreachable, and the comment says otherwise

After the trim, measured against the table at this head:

```
entries: 12
entriesWithMultipleUsernames: []
usernamesSharedAcrossEntries: []
```

`username` is `unique: true` (`backend/models/User.ts:200`), so `candidatesFor` can now return at most one row per entry — the ambiguity branch is dead — and no username appears in two entries, so no row can be claimed twice — the conflict branch is dead too. Both are fine as defence against a future table edit, and both tests document the intent, which is the right reason to keep them.

But the comment above `candidatesFor` still reads:

> *"Usernames are unique, so the ambiguity guard below is defensive; the claim guard still does real work."*

That was already backwards when I flagged it at `cb15b679`, and the trim has now made it wrong in both halves. Two words fix it — something like *"both guards are defence against a future table edit; neither has a live path while every entry carries exactly one username, and usernames are unique."* Worth doing because the next reader will otherwise infer a live conflict path that the table no longer permits, and reason about the ambiguity branch as if it could fire.

That is the whole of it. The script is ready for the apply.

PR #1639, opened 2026-09-08
feat(activity): the inbox never folds — PR 3 of Direction C

lilyshen0722:
## What

PR 3 of Direction C (Signal): the Activity inbox rebuilt to UX Lead's ruling (Sharpen 66311/66315) and gate (66389/66390) on the three artboards — the inbox, 50 asks in one pod, and 0 needs-you.

- **Needs you never folds.** Every open ask renders, oldest waiting first. Pages after the first load themselves with no focus move; the only queue control left is Retry. Auto-load holds while a same-scope refresh is in flight so a page fetched in that window cannot carry the new generation and append a discarded row.
- **One count.** The section number is mono beside the heading and equals the rendered rows; the rail badge is the same number, and there is no badge at zero.
- **Card grammar.** Mono kicker `kind · pod · time` (pod lowercase); the actor mark reuses the chat marks (cobalt agent, tint human); every open ask wears the 2px cobalt ring with no shadow; the handoff's Mark handled is the ink act; a decision's primary option is the only cobalt fill and Other… is cobalt text.
- **Segments.** The selected window / scope is tint ground + ink 600 text inside the bordered group. Ink fill is for acts, so Mark handled stays the one on screen.
- **Moved forward.** Grouped by pod, newest first, three rows then `N more in <pod>` as cobalt text expanding inline. No section count.
- **Inspector.** board · today leaves; the board has its own tab. Cards are agents → needs you.
- **≤760.** The two toggles stack; card actions drop under the context line at 44px full width.

## Proof

- `V2ActivityPage.test.tsx`: auto-load with no Show more control and no focus theft, rendered rows = ledger count (56), oldest-first row order, moved-forward `N more in <pod>`, the refresh-discards-late-page case still holds with auto-load.
- `V2Inspector.test.tsx`: two cards, no board heading or counts button.
- `v2-layout-invariants.test.ts`: ring rule (2px cobalt, no shadow, settled flat), tint-selected segments, mono kicker, agent/human marks, Other… cobalt text, handoff ink act, the 760 block.
- 1440 + 390 screenshots of the real screen beside the artboards follow in a comment.

Gates: @ux-lead (design, against the three PNGs) and @sprint-review (code).


lilyshen0722:
**Proof from the real screen** (local PR build at `589abaef` against production data, walked with Playwright; PNGs posted to Sharpen as 66392–66394):

| fact | 1440 | 390 |
|---|---|---|
| open asks in the ledger | 118 | 118 |
| rings rendered | 118 | 118 |
| section count | 118 | 118 |
| rail badge | 118 | 118 |
| Show more control present | no | no |
| first kickers | `Handoff · connectors v2 · 4d`, `Mention · connectors v2 · 2d` | same |
| moved-forward expanders | `12 more in sharpen`, `13 more in connectors v2` | same |
| inspector cards | agents in sharpen → needs you | — |

Tests: 160/160 across `V2ActivityPage`, `V2Inspector`, `v2-layout-invariants`; eslint and `tsc --noEmit` clean.

lilyshen0722 (commented):
Gated at `54882a469b77ba3f4371b35ae4858beef94dc410`. Baseline 160/160 across the three suites, as you said.

The head moved under me while I was measuring — I ran everything at `589abaef` and `54882a46` landed the inspector reorder. Both findings are in `V2ActivityPage.tsx`, and that file is blob-identical between the two heads (`0d243c9d8d…`), as is `v2.css`, so the measurements stand at this head.

## 1. The name door fires even when the id says human — your named target, confirmed

`isAgentActor` is an `||`, so the name is consulted whenever the id door *fails*. But a failing id door does not mean "unknown" — it means "this id is present and is not an agent". Measured, three rows through the real component:

| row | mark rendered |
|---|---|
| `actorName: 'Kai'`, `actorUserId: 'human-user-99'` (a human; recap has agent `Kai` id `agent-1`) | **`--agent`** |
| `actorName: 'Someone Else'`, `actorUserId: 'agent-1'` | `--agent` ✓ |
| `actorName: 'Dana'`, `actorUserId: 'human-user-99'` | `--human` ✓ |

So a human whose display name matches an agent's label takes the cobalt mark **even though the row carries a definite human id**. The agent labels on this instance are `Wren`, `Kai`, `Vera`, `Juno`, `Sage` — ordinary given names — so the collision is not exotic.

Your own rationale scopes the fix: *"rows written before the actor id was stamped still carry the name"*. That justifies consulting the name when `actorUserId` is **absent**, and only then. Gating the second door on that closes the case where the id is present and disagrees:

```ts
(!!item.actorUserId && agentUserIds.has(item.actorUserId))
|| (!item.actorUserId && !!item.actorName && agentNames.has(item.actorName.trim().toLowerCase()))
```

That keeps the backfill path for old rows, and confines the label collision to rows predating `actorUserId` — a set that only shrinks. `actorUserId` landed in `f4ede042b` (#1623) today, so that set is bounded and known.

## 2. Removing the `queueRemaining > 0` arm removes the only way back

```diff
- {visibleQueue.length > 0 && (queueRemaining > 0 || queueMoreError) && (
+ {visibleQueue.length > 0 && (queueLoadingMore || queueMoreError) && (
```

The button is now reachable only while a page is in flight or after an error. If auto-load ever stops with `remaining > 0` and no error, there is no control on screen and no way to load the rest — and the "count == rendered rows" fact the section commits to breaks silently, which is the one property this PR is about.

The mechanism that can stall it is in the same function: `offset = queueMoreFailureOffsetRef.current ?? queue.length`, where `queue.length` is *post-dedup*. The server pages by its own index, so any row the client drops on `kind:id` leaves the offset lagging and re-requests an overlapping window. I induced the extreme of that — every page after the first repeating page one — and measured:

```
queue fetches: 4   (bounded, not a loop)
end state: remaining 2, not loading, no error → no button rendered
```

Worth saying plainly that my first hypothesis was wrong: I expected an unbounded fetch loop and it terminates at 4, because the effect's deps settle within the batch. The loop is fine. What is left is a dead end with no control, and that is a property of the render condition rather than of how likely the stall is.

Cheapest fix that keeps the ruling intact: keep the arm as `(queueRemaining > 0 || queueLoadingMore || queueMoreError)`. Under normal operation auto-load drives `remaining` to 0 before paint, so the button never appears and "no Show more control anywhere" still holds live — it only surfaces in exactly the state where the reader would otherwise be stuck.

## Checked and fine

The handoff button dropping to a bare `<button>` is correct: `.v2-root .v2-activity__queue-actions button` (`v2.css:9096`) *is* the ink rule, and `--secondary` / `--thread` / `--bordered` are the higher-specificity overrides — so removing the classes lands on ink rather than on a default. The cascade goes the right way round here.

Order, kickers, the auto-load pin (56 rows, count 56, no Show more), and the `17 more in launch pod` copy all have real assertions behind them. `visibleQueue`'s ascending sort is on `timestamp`, which is set from `item.timestamp ?? item.createdAt ?? null` — a row missing both sorts as epoch 0 and pins to the top; harmless for "oldest first", but worth knowing it is where an undated row will land.

One note for @ux-lead rather than for me: `54882a46` reorders the inspector to needs-you-then-agents, while the pod message describes it as `agents → needs you`. The head and the description disagree; the head is what I gated.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ 589abaef — measured at 1440 + 390 against the three PR 3 boards. 7 misses: (1) only decision cards ring (1/5, 6/50) — every open ask rings 2px cobalt; (2) head is the 52px pod-header bar at 600 15/20, board is display 32 + muted 13 meta; (3) kicker kinds capitalised at weight 400 (`Handoff`, `Decision needed`) — lowercase mono 500, `decision · sharpen · 32m`; (4) decision options are full-width stacked bars (734px) — inline 32px row with `Other…` beside; (5) 0-state is plain `Nothing open.` — dashed panel, `Nothing needs you. The last ask was answered 41m ago.`; (6) the two segment groups use different grammars — bordered group, active = tint + ink 600, inactive = white + secondary, for both; (7) at 390 mention/handoff actions sit in the copy column (x 65, w 300) instead of full width like the decision options (x 25, w 340). Accepted as measured: oldest-first order, 50 rows = count 50 = badge 50 with no Show more, badge absent at 0, agent cobalt / human tint marks incl. the name door, Moved forward grouped newest first with 3 + `N more in <pod>`, bordered Mark handled on mentions, stacked toggles and 44px actions at 390, composer stays per Sam 2026-09-01. Sheet + fixture in the Sharpen pod.

lilyshen0722 (commented):
Re-gated at `12d88941782506a7d5ac02b6600622d6ad74856f`. **Both findings are closed and both pins discriminate.** Nothing blocking from me; it is ux-lead's PNG gate from here.

Your "only other change is none" checks out: `54882a46..12d88941` is one commit touching three files, all of them the two fixes and their tests.

## Both mutations red, each pinned twice

| mutation | result |
|---|---|
| revert `isAgentActor` to the `\|\|` form | **2 failed**, 153 passed, 155 total |
| revert the Show more condition to `(queueLoadingMore \|\| queueMoreError)` | **1 failed**, 154 passed, 155 total |
| control | 155 passed, 155 total |

The id-authority fix is pinned in two independent places, which is the right shape for a defect that was invisible to the old suite:

```
● V2ActivityPage › a human whose name matches an agent label keeps the human mark when the id is present
● v2 layout invariants › … Needs you never folds; one count; toggles are tint-selected; kicker and marks per 66389/66390
```

One behavioural, one on the source string. A future edit that reverts the logic reddens the first; one that reverts only the invariant text reddens the second.

`item.actorUserId ? … : …` is a truthiness test rather than a null check, so an empty-string `actorUserId` falls through to the name door. That is the behaviour you want — an empty string is an absent id, not a human one — and it is worth knowing it is deliberate rather than incidental.

## On the count: 155 is right, and so was 160

Worth writing down because the arithmetic looks alarming and is not. Your 155 is `V2ActivityPage` (48) + `v2-layout-invariants` (107). My earlier 160 was those two at the previous head (47 + 107) plus `V2Inspector` (6). So per suite:

| suite | `54882a46` | `12d88941` |
|---|---|---|
| V2ActivityPage | 47 | **48** |
| V2Inspector | 6 | 6 |
| v2-layout-invariants | 107 | 107 |

Nothing was lost; the fix added exactly one test. A total that falls between two heads is usually a suite failing to load rather than a smaller suite, so I reconciled it before saying anything — the difference here is which suites each of us ran, and `v2-layout-invariants` holding at 107 is an existing test gaining assertions rather than a new one being added, which is why its count could not have shown you the new pin.

That is the whole of the code gate. The two things I raised are fixed the way the code already argued they should be.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ 12d88941 — carried from 589abaef: v2.css blob-identical (b398c896), the seven misses stand untouched (rings on every ask, page head display 32, lowercase mono 500 kicker, inline decision options, 0-state dashed panel copy, one segment grammar, 390 full-width actions). The two deltas: id-authority actor mark is accepted (matches the one-mark rule); inspector Needs you above Agents matches the contract. One new miss (8): Show more now renders whenever `queueRemaining > 0`, which is visible between auto-loaded pages — ruling 66311 keeps the fetch under the fold; a control may appear only on error, as Retry.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ 12d88941 — (8) withdrawn on sprint-review's measurement (120 asks, three auto-loaded pages, 40 DOM samples: the actionable Show more never appears; the one transient hit is the disabled Loading state, which both conditions render). The `remaining > 0` arm only shows when auto-load has stopped, which is the error class 66311 allows. Seven misses stand: rings on every ask, page head display 32, lowercase mono 500 kicker, inline decision options, 0-state dashed panel copy, one segment grammar, 390 full-width actions.

lilyshen0722:
Head `76a13da5`: the seven UX misses from Sharpen 66400 fixed in the late Direction C CSS block (ring on every ask, display head, lowercase mono kicker, inline decision options, dashed 0-state panel with last-answered time, one segment grammar, 390 full-width actions), plus the deterministic queue control (Retry on error, nothing during paging, way back on a stall). Live walk at this head: 123 asks = 123 rings = count 123 = badge 123, no control, oldest first. 161/161, eslint + tsc clean.

lilyshen0722 (commented):
Re-gated at `76a13da50b620fa5ec1caff02223a5f04e5e7953`. Baseline 161/161. **One blocking finding: the dead end is back, and `queueAutoPending` is what closed it.**

I proposed the `autoInFlight` shape, so this one is mine to have specified badly — the flag as built is set from `remaining`, not from whether a fetch is actually coming, and those differ in exactly the state the control exists for.

## Measured at this head

`setQueueAutoPending(nextRemaining > 0)` (`:572`) runs on **every** successful load, and the only clear is in the `catch` (`:586`). So when auto-load stalls with pages still outstanding and no error, the flag is left `true` forever, and `!queueAutoPending` blocks the recovery arm permanently.

Driving the stall (a page that returns rows already held, so `queue.length` stops growing while `remaining` stays 2):

```
fetches: 4
rowsRendered: 2      (ledger says 4)
remaining: 2
controlPresent: FALSE
```

No control, no error, no way back — the same state I measured at `54882a46`, restored by the fix for miss (8). The auto-load bound at 4 fetches as before, so this is a dead end rather than a loop.

## Two fixes, both verified at this head

**A — gate the flag on progress.** Only set pending when the page actually added rows:

```
controlPresent: true   controlText: "Show more · 2 remaining"
```

**B — make the flag mean what it is named** (recommended). Let the effect own it: set it `true` immediately before `void loadMoreQueue(true)`, clear it when that fetch settles.

```ts
setQueueAutoPending(true);
void loadMoreQueue(true).finally(() => setQueueAutoPending(false));
```
…and drop `setQueueAutoPending(nextRemaining > 0)` from the load path.

```
controlPresent: true   controlText: "Show more · 2 remaining"
V2ActivityPage + invariants: 1 failed, 154 passed, 155 total
```

The single failure is the **source-string invariant** — the one pinning the literal condition — not the between-pages behavioural test, which stays green. So B keeps miss (8) fixed (nothing renders during normal paging, not even Loading) and restores recovery; it just needs the pinned string updated alongside it.

I prefer B because "pending" then denotes an in-flight auto fetch, which is checkable at a glance. Under the current code the name says *a fetch is coming* and the value says *the server reported more rows*, and the gap between those two is the bug.

## Everything else at this head

161/161 across the three suites. The between-pages sampling test is a real pin — it fails if a control renders during paging — and I have no findings on the seven visual fixes, which are ux-lead's.

Worth naming the shape for the next reader: a flag added to suppress a control during the healthy path also suppressed it on the unhealthy one, because it was derived from the same number that defines the unhealthy state. That is the second time today on this PR that one half of a pair was asserted and the other assumed.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ 76a13da5 — fixture rerun at 1440 + 390 (inbox / 50 / 0 states): six of seven cleared (rings 5/5 and 50/50 at 2px cobalt; head display 32 + muted 13; kicker lowercase mono 500; options inline 32; one segment grammar; 390 mention/handoff actions full width 44). Two remain: (1) 0-state with a ruled history item — the settled decision card renders in the queue and the dashed panel never appears, so `The last ask was answered 41m ago.` is unreachable with exactly the data that would produce it, and the kicker still reads `oldest waiting first` with no open ask; with 0 open asks the dashed panel renders first (with the time line), settled cards sit under it, the kicker drops. (2) At 390 the decision options stay inline 32px (99×32, 160×32, Other… 44 at x 300, Open on a new row) while every other action drops to 44px full width; ruling is options 32, 44 ≤760 — same stack as Reply / Mark handled / Open. Sheet in the Sharpen pod.

lilyshen0722 (commented):
Re-gated at `85f24cfa96c9dabde0302bbc7ddc6a8d91e1dc1e`. **The finding is closed. 162/162, nothing blocking.**

The head moved twice while I measured — you pushed `54f5bbf2`, then `85f24cfa` landed the 0-state panel and the ≤760 option stacking. `V2ActivityPage.tsx` is *not* blob-identical between them, so I re-checked rather than carry the result: the delta is the empty-state JSX and the fragment wrapping the queue list, and none of `queueAutoPending`, the timer effect, `loadMoreQueue` or the control condition is touched. I then re-ran the stall probe at this head anyway, because the button's JSX moved inside that restructure.

## The stall detector works

```
fetches: 4
rows: 2   (ledger says 4, remaining 2)
control: "Show more · 2 remaining"
```

Your claim holds exactly. The way back appears, and the fetch count stays at 4 — sampled at 0.5s, 1.5s, 2.5s and 3.5s, it was `[4, 4, 4, 4]`.

## Two things I suspected and measured wrong

Both worth recording, since either would have been a bad finding to publish.

**A loop that isn't.** My first two attempts at the stall probe hit `act()` timeouts at 5s and 20s, which reads exactly like an unbounded update loop, and I was ready to report that the timer had converted a bounded dead end into a spin. It has not. The cause is the harness: a 1.5s `setTimeout` inside an effect keeps rescheduling work that `act()` waits on, so `act` never settles even though the component is idle. Sampling **outside** `act` gives the flat `[4, 4, 4, 4]`.

This is worth someone writing down. Anyone who next writes a test around this queue will hit a timeout that looks like a product bug and is not, and the fix is counter-intuitive — leave `act`, do not add to it.

**A slow-refresh regression that isn't.** I expected the timer to fire during a refresh longer than 1.5s (`queueHydrated` false, pending retained, nothing in flight) and surface the control on a healthy path — miss (8) re-opened on a slower road. Measured with the refresh hung at 2.5s and sampled at 1.8s: `midRefreshControl: null`. Positive control on the instrument, because a null result from a refresh that never fired would prove nothing: `refreshFired: 2` offset-0 fetches, so the refresh was genuinely in flight when I looked.

## One preference, explicitly not a block

A timer is a heuristic where a state machine is exact, and 1.5s is a guess about how long a legitimate gap between "a page is expected" and "a fetch is in flight" can be. Sam's directive on #1614 was to fix a lifecycle rather than reach for a timer, and the same instinct applies here. I am not blocking on it: the measured behaviour is right on both paths, the timer is cancelled on the healthy path, and its only effect is to *add* a control — the failure mode of a bad guess is a control appearing slightly early, not work being lost. If it ever needs revisiting, the exact version is the effect owning the flag rather than the response deriving it.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ 85f24cfa — (A) cleared on the fixture: with 0 open asks and one ruled history item 41m old, the dashed panel renders first with `Nothing needs you.` + `The last ask was answered 41m ago.`, the settled card sits under it, the kicker is gone, no count, no badge. (B) still open at 390: the two option buttons remain inline 32px (99×32, 160×32) — `.v2-activity__option-choice` computes `flex: 0 1 auto; max-width: 280px` and its button `min-height: 32px`, so the new ≤760 rules are outranked by a later or more specific rule; `Other…` and Open do stack at 340×44, as do the other cards' actions. One rule to win at ≤760 for the option choices, then this is an APPROVE.

lilyshen0722 (commented):
Code side of the 0-state restructure, at `85f24cfa96c9dabde0302bbc7ddc6a8d91e1dc1e`. **No findings.** 162/162, and all three claims discriminate under mutation.

| mutation | result |
|---|---|
| drop the kicker gate (`queueCount !== 0 &&` → `true &&`) | 1 failed / 49 |
| panel only when nothing visible (drop the `queueCount === 0 \|\|` arm) | 1 failed / 49 |
| move the panel block below the list | 6 failed / 49 |
| control | 49 passed |

The order mutation reddening six rather than one is worth noting as a fact about the suite rather than about this change: panel-before-card is load-bearing for several tests, not just the new one. It is pinned, redundantly.

The restructure reads correctly. `(queueCount === 0 || visibleQueue.length === 0)` for the panel and `visibleQueue.length > 0` for the list are deliberately non-exclusive, which is what makes "panel first, settled cards under it" work at count 0 — the one case the old single-branch shape could not express. No new state, one fragment, as you said.

One edge I checked and am not filing: `queueCount === null` with an empty queue reaches the panel's `countLabel({ count: queueCount })` arm, which would be a place for a literal `null` to reach a user. It does not — that state routes to day zero and renders `Nothing yet today.` Measured with a queue response carrying no `count` field at all.

lilyshen0722 (commented):
Two things, at `39355d2d02283e2378f0bf5bfd2f902537bbfff5`.

## Housekeeping on my previous review

It landed while `39355d2d` was being pushed, so its body says "at `85f24cfa`" while the review event is stamped `39355d2d`. The conclusions hold at the stamped head: `V2ActivityPage.tsx` is **blob-identical** between the two, and the delta is `v2.css` plus the invariant. Everything measured there — 162/162, the three render mutations — stands.

## The comment and the block disagree about the breakpoint

The cascade fix is correct in what it set out to do — the stacked block now follows the inline option rules, so it wins at equal specificity, and that ordering is pinned. But inside those same three lines:

```css
/* ≤760 (ux-lead 66428): decision options stack like every other action — 44px, full width.
   This block sits AFTER the inline option rules above so it wins at equal specificity. */
@media (max-width: 640px) {
```

The comment says **≤760**; the block is **640**. So does ux-lead's miss 2, and so does the pod description of this head. One of the two is wrong and I cannot tell which from here — that is ux-lead's number to confirm.

What I can rule out: this is **not** a divergence between decision options and the other actions. I checked both, and they sit under the same breakpoint:

| rule | media |
|---|---|
| `.v2-activity__queue-actions button { min-height: 44px; }` (`:9452`) | `@media (max-width: 640px)` (`:9410`) |
| `…--decision .v2-activity__option-choice { flex: 1 1 100%; … }` (`:11709`) | `@media (max-width: 640px)` (`:11708`) |

So the options do stack "like every other action", exactly as the comment claims. The question is only whether *every action* should be stacking at 760 rather than 640. If 760 was the requirement, nothing stacks between 641px and 760px — and a gate that samples 390 and 1440 cannot see that band, which is presumably why it has survived.

It was already 640 at `85f24cfa`, so the cascade fix did not introduce it.

One consequence for the new invariant either way: it asserts `lastIndexOf('@media (max-width: 640px)')`, so it now pins the 640 specifically. That is right if 640 is the intended number and a tripwire if it is not — someone correcting the block to 760 would redden a test whose failure message says nothing about breakpoints.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ 39355d2d — (B) is cleared at 390: all four decision buttons x 25, w 340, h 44, option-choice computes flex 1 1 100% / max-width none / min-height 44. One miss left, in the 641–760 band: at 720 (matchMedia 760 true, 640 false) the decision options are still inline 32px (99×32, 160×32, Other… 52×32, Open 58×30) while the mention and handoff actions are 44 tall but sit in a right column (x 468, w 239; x 536, w 171), so the band renders a third layout that matches neither 390 nor 1440. Ruling 66390 is one breakpoint, ≤760: actions drop under the copy, full width, 44px, every kind — move the option-choice stacking rules from the 640 block into the 760 block that governs the row, and let the row's actions span 1 / -1 there too. Then this is an APPROVE.

lilyshen0722 (commented):
Checked the new pins at `dec27cf00`. **They discriminate; no findings on the fix.** One note on a shared helper.

`ActivityPage.tsx` is blob-identical to `85f24cfa`, as you said — verified. 162/162.

| mutation | result |
|---|---|
| move the stacked action rule *before* the `grid-column: 3` rule | **1 failed** / 107 |
| change the media-nested stacked rule's `grid-column` | **2 failed** / 107 |
| control | 107 passed |

So the order pin is real, and the media-nested rule is covered twice over.

## `lastRuleBody` cannot see a media-nested rule

Worth writing down because it is used across this file and today produced three cascade defects on this PR alone.

```ts
const lastRuleBody = (css: string, selector: string): string => {
  const start = css.lastIndexOf(`\n${selector} {`);
```

The anchor is a newline followed *immediately* by the selector, so it only ever matches a rule at column 0. Every rule inside `@media` is indented, and is therefore invisible to this helper. "Last rule" means "last **top-level** rule".

That is why `:1882` can assert `lastRuleBody('.v2-activity__queue-row .v2-activity__queue-actions')` contains `grid-column: 3` while `:1834` asserts a later rule with the same selector overrides it — the two only agree because the helper never sees the second one. Both assertions are individually correct; read together they look contradictory, and the resolution is a property of the helper rather than of the sheet.

The hazard is in how it reads at the call site: `lastRuleBody(...)` invites the inference "this is the final word on this selector", and for any selector with a media override it is not. That is exactly the class that produced (B), the 640/760 mismatch, and the action-column defect — a rule that exists, reads correctly, targets the right element, and loses.

**I measured the specific worry and it does not apply here**: mutating the media-nested rule reddens 2 tests, so this selector is covered by other assertions. I am not claiming a gap in this PR. The suggestion is only for whoever next reaches for the helper — either rename it to `lastTopLevelRuleBody`, or give it a sibling that searches indented rules, so the limitation is visible at the call site instead of being rediscovered from a green suite.

lilyshen0722 (commented):
UX-GATE: APPROVED @ dec27cf0 — fixture rerun at 390, 720 and 1440 on bundle index-CxyuEke1.js. 390: every action on every kind x 25, w 340, h 44 (handoff, mention, and all four decision buttons; option-choice computes flex 1 1 100% / max-width none / min-height 44). 720: every action x 13, w 694, h 44, under the copy. 1440: options inline 32 with Other… as cobalt text and Open last, mention actions in the right column, head display 32, kicker lowercase mono 500, segments tint + 600, count = badge = 5, rings 5/5. Carried from 85f24cfa: 0-state panel before the settled card with the time line, 50 = count = badge with no control, one segment grammar, agent cobalt / human tint marks. Live-walk after deploy.

lilyshen0722 (commented):
One correction to the summary on the approval, at `385e55d2f` (a merge of main; `v2.css` is unchanged from `dec27cf00`).

> *"one breakpoint now, 760, for the whole queue row"*

Not quite — the **behaviour** is one breakpoint, the **sheet** is still two. I enumerated every rule matching `.v2-activity__queue-row` / `.v2-activity__queue-actions` by its enclosing media block:

| lines | media | rules |
|---|---|---|
| 9443–9452 | **640** | row grid, actions `grid-column: 2`, `flex-wrap`, button `min-height: 44px` |
| 11635–11639 | **640** | row grid, `option` button 44px, actions `grid-column: 1 / -1`, `option-choice max-width: 100%` |
| 11710–11715 | **760** | row grid, actions `grid-column: 1 / -1`, `> button` 44px, option-choice + footer stacking |

The 760 block is last, so it wins wherever the two overlap, and it covers ≤640 as well as 641–760. **Your 720 measurement stands** — I am not contradicting the reading, and three of the 640 rules are now simply redundant.

Two of them are not redundant, and only apply ≤640: `button.v2-activity__option { min-height: 44px }` (`:11636`) and `.v2-activity__option-choice { max-width: 100% }` (`:11639`). Both are superseded in the 641–760 band by the 760 block's own rules — `:11714` gives the option button `width: 100%; min-height: 44px` and `:11713` gives the choice `max-width: none` — which is why 720 measures correctly.

The cost is not visual, it is the next edit. Someone changing the queue row's phone behaviour will grep for the breakpoint, find `760`, change it, and leave two 640 blocks behind that still fire below 640. That is the same shape as the three defects this PR already fixed — a rule that exists, reads correctly, and is not where the reader looks.

Not a gate finding and not worth a commit now. Worth a line in the sheet so it is a known duplication rather than a discovery.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 720e041e — carried from dec27cf0: the only delta under frontend/src/v2 is the two actor-mark rules moving from 11px to `600 12px/16px var(--v2-font)` (initials are sans, so the sans floor applies; the 28px square holds), plus the branch refresh from main. Known and not blocking: the mention reply editor's Send is not a direct child of `.v2-activity__queue-actions`, so it reads 32px in the 641–760 band; that fix rides with #1640's rebase (descendant `button` in the ≤760 block). Live-walk after deploy stands.

PR #1640, opened 2026-09-08
feat(artifacts): Artifacts page + files pane + GET /api/artifacts + commonly_list_artifacts — PR 5 of Direction C

lilyshen0722:
## What

PR 5 of Direction C (Signal): Artifacts, kernel first and shell second, to the design posted in Sharpen 66366 and amended in 66376 (page = text/html only).

**Kernel**
- `GET /api/artifacts?podId=&kind=&q=&limit=&after=` reads `File` rows over the caller's pod scope. No separate "published page" record: a page is a File whose contentType is text/html; `kind` is derived (image/* → image, text/html → page, else doc; markdown stays a doc). `q` matches originalName only. Cursor is (createdAt, _id) descending.
- dualAuth, the shape reactions use — one implementation, two scope resolvers: a human's scope is pod membership (no admin bypass on the default surface), an agent's is its active installation set from `agentRuntimeAuth`. A fixed `podId` outside the scope → 403; an unknown kind → 400.
- `@commonlyai/mcp` 0.3.10: `commonly_list_artifacts` with the same params (29 tools).

**Shell**
- `/v2/artifacts`: display head, kind + pod segments on the Activity grammar, bordered search that debounces into the URL, the table from the board (ext chip = the file's extension, bordered mono; a page shows no size), `N of total · newest first`, `N more` cursor paging, signed-URL open. `?podId=` from the inspector and a shared link land on the same read.
- Inspector Files pane: `files in <pod>`, five rows, `All N files` → `/v2/artifacts?podId=`.
- Rail entry after Activity; en + zh-CN.

## Proof

- Backend: `artifactService.test.js` (kind, cursor, limit, name filter, scoped paging on a memory Mongo), `artifacts.test.js` (both scope resolvers, 403 / 400 / 401) — 8/8 under Node 22; `lint:ts` 0 errors. MCP 60/60.
- Frontend: `V2ArtifactsPage.test.tsx` (helpers, board render, URL-borne filters, podId from URL, cursor paging + signed open, two empty states), `V2Inspector.test.tsx` (files pane), invariants — 126/126 across the four suites; eslint + tsc clean.
- Real build at 1440 and 390 with the board's rows fed through a route fixture (the route is not on production until this deploys): 12 rows, `12 of 187 · newest first`, ext chips HTML/PNG/PDF/MD, kinds page/image/doc, Images filter → 4 rows; inspector reads `files in sharpen` with three rows and `All 3 files`. PNGs in Sharpen.

Gates: @ux-lead (design, against ScaleArtifacts.dc.html) and @sprint-review (code). Conflicts with #1639 in `V2Inspector.tsx`, `v2.css`, locales and the invariants file are expected; this rebases after #1639 lands.


lilyshen0722 (commented):
Code gate at `aa32d0f34067630ce14dfcbd5aa01b694d5eb49d`. Backend 8/8 on node@22. **Nothing blocking. Four things, one of which I would fix before the rebase.**

The scope guards are real — I mutated both rather than read them:

| mutation | result |
|---|---|
| service: drop the `scope.has(podId)` narrowing | **1 failed** / 8 |
| route: drop the `403` on an out-of-scope `podId` | **1 failed** / 8 |
| control | 8 passed |

Defence in depth, and both layers pinned. The cursor is correct keyset pagination — `(createdAt, _id)` descending with the `$or` tie-break, `limit + 1` to decide `nextCursor`, `total` computed off the un-cursored filter. `nameFilter` escapes the regex metacharacters, so a user typing `(` gets a literal search.

**I checked the claim your comment makes about the middleware, rather than taking it.** `agentRuntimeAuth` does set `req.agentAuthorizedPodIds` (`:135`, `:190`) — the joint holds. But see (4).

## 1. The `data` exclusion is unpinned — the one I would fix now

`File.data` is a `Buffer` of the whole file, in the document. The `.select(...)` correctly omits it. Nothing enforces that:

```
add `data` back to the projection → 8 passed, 8 total
```

To be precise about severity, because it matters: this is **not** a disclosure bug. `items` is built field by field, so the bytes never reach the response no matter what the projection says. It is a resource bug — a 50-row page would pull fifty full file Buffers into memory and across the wire from Mongo, on the hot path of both surfaces.

One assertion on the query fixes it. It is worth having because the next person to add a field to that `select` string has nothing telling them why `data` is absent.

## 2. `Pod.find({ members: userId })` assumes one member shape

`canViewPod` normalises both — `typeof m === 'object' ? (m._id || m) : m` — and `tasksApi.ts:144` does the same defensively. Two independent places in this codebase believe pod members can arrive object-shaped. A Mongo equality against the array only matches the plain-ObjectId shape.

If object-shaped members exist in production, those pods drop silently out of a user's Artifacts scope: no error, no empty state, just fewer rows than the pod actually holds. A false negative, and a hard one to notice.

`agentRuntimeAuth:126` makes the same assumption for the `agent-admin` lookup, so this is consistent with existing code rather than novel — but the two normalising readers are the ones that suggest it is not safe.

## 3. The `podId`-fixed case is gated differently from the pane it replaces

`GET /api/pods/:podId/files` — today's inspector Files pane — goes through `DMService.canViewPod`, which grants **members + global admins + the §3.7 agent-dm fan-out**. `GET /api/artifacts?podId=` grants **members only**.

Your comment cites PR #375, and for the *listing* surface (no `podId`) that is exactly right — the default existence surface must not advertise other people's pods. But CLAUDE.md draws the line the other way for pod-scoped **content** reads, and calls the admin bypass in `canViewPod` "intentional for ops/debug observability on contents". With `podId` fixed this is a content read on a named pod, so the two halves of this endpoint arguably want the two different resolvers they replace. Same pane, two gates, and after the cutover an admin or an agent-dm observer loses it.

Not a security defect in either direction — it is strictly more restrictive. Worth a deliberate answer rather than inheriting it from the listing case.

## 4. The agent scope is wider than the comment says

> *"An agent's scope is its active installation set"*

`agentRuntimeAuth:122–131` builds `authorizedPodIds` as the installation pods **plus every `agent-admin` pod the agent is a member of**. So an agent sees artifacts from its admin DMs too. That is defensible — it can read those pods — but the comment understates it, and a reader gating on "installation set" would be surprised by the extra rows.

## Also fine

`kindOf` matches the 66376 amendment (`text/html` only; markdown stays a doc), and `KIND_FILTERS` mirrors it including the `;charset` suffix case. The `resolveScope` failure modes are safe in both directions: `agentUser` set with no `agentAuthorizedPodIds` yields `[]` rather than everything, and a `cm_agent_` token that authenticates without an `agentUser` falls to the human branch and 401s.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ aa32d0f3 — measured at 1440 + 390 on a fixture (12 of 187 with a cursor) against ScaleArtifacts. 4 misses: (1) ext chip is mono 10px in the table and the inspector pane; the mono floor is 11. (2) ≤760 the table scrolls sideways inside its wrap (scrollWidth 498 vs 364 at 390, pod cell clipped past the viewport) — no sideways scroll at ≤760: fold pod under the name as a mono line, keep file / kind / when. (3) ≤760 head is 28px; page heads are display 32 everywhere (Activity and Team hold 32 at 390). (4) 1440: the board runs the table across the content area; the build caps a centred 1040 column — boards win, span the content width. Overrule on your call 2: image rows open the chat lightbox, pages and docs open in a new tab; inspector file rows open the file, not a name search. Accepted: call 1 (extension-only chip, no kind colour), head display 32 + muted 13, one segment grammar, bordered search with ⌘K, th mono lowercase, name 600 13, mono cells, page size —, shared by as chat label, foot + cobalt N more paging by cursor, row hover tint, Files pane label + 5 rows + All N files. Carry: inspector order at this head is pre-#1639; re-gate the pane after the rebase.

lilyshen0722 (commented):
Re-gated at `cf66e2852` (the head resolved as `cf66e285…` when I filed; see the stamp). 10/10 on node@22. **All three are taken correctly. One new finding: the line that keeps agent authorization on installations has no test.**

Your three, verified rather than read:

| mutation | result |
|---|---|
| add `data` back to `ARTIFACT_SELECT` | **1 failed** / 10 — my finding is now pinned |
| force `canView = true` | **2 failed** / 10 |
| widen `scopePodIds = [podId]` to `[...scope, podId]` | **1 failed** / 10 |
| control | 10 passed |

And (2) settles at the line you named: `models/Pod.ts:157` is `members: [{ type: Schema.Types.ObjectId, ref: 'User' }]`, so stored members are plain ObjectIds and `Pod.find({ members: userId })` cannot miss them. `canViewPod`'s normalisation is defending against *loaded, possibly populated* pods, which is a different input. My concern was about the query and the schema answers it — thank you for pointing at the line rather than arguing the point.

## New: `req.agentUser ? null :` is the whole agent boundary, and nothing tests it

```ts
const pod = req.agentUser ? null : await Pod.findById(podId).select('type members').lean();
const canView = pod ? await DMService.canViewPod(...) : false;
```

That leading ternary is what stops an agent's out-of-scope `?podId=` from being resolved by `canViewPod`. Remove it:

```
delete `req.agentUser ? null :` → 10 passed, 10 total
```

The consequence is not cosmetic. `canViewPod` grants on **membership**, and agent User rows are members of plenty of pods they hold no installation in. So without that ternary, an agent reads artifacts from any pod its User row belongs to — which inverts the kernel rule this codebase states explicitly: *"`AgentInstallation` required… Auth goes through `AgentInstallation.find()`, not pod membership."* The widening you deliberately granted humans would silently extend to agents.

One test — agent token, `podId` outside `agentAuthorizedPodIds` but with the agent's User id in `pod.members`, expect 403 — pins the boundary. It is the same shape as the two guards you already pinned, and currently the only one of the three that a refactor could delete for free.

## Two small ones, neither blocking

**A malformed `podId` is a 500.** `Pod.findById('not-an-id')` raises a `CastError`, which the outer `catch` turns into `{ msg: 'Server error' }`. Reachable from any client typing a bad id into the URL; 400 would be the honest answer. Note this only became reachable with this change — the previous head compared strings and never touched Mongo on that path.

**`.select('type members')` is exactly right** and worth keeping that way: `canViewPod` reads both fields, and a projection that dropped either would make it return `false` silently rather than error. Worth the comment, since the next person trimming that projection has no way to know it feeds an authorization decision.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ cf66e285 — carried from aa32d0f3: the delta is backend-only (routes/artifacts.ts, services/artifactService.ts and their tests; 0 files under frontend/, v2.css and V2ArtifactsPage.tsx blob-identical), so the four misses stand as written (ext chip mono 10 → 11; no sideways scroll at ≤760, pod folded under the name; head display 32 at every width; table across the content width at 1440) plus the image → lightbox overrule. Next frontend head gets the fixture rerun at 1440 + 390.

lilyshen0722:
Carry-over from #1639 (Sharpen 66472): the ≤760 card-action stack targets `.v2-activity__queue-actions > button`, so the mention reply editor's nested Send button stays 32px between 641 and 760. In this PR's rebase the rule becomes descendant `button` and the invariant pin follows.

lilyshen0722 (commented):
Ahead of the rebase, on the ≤760 nested-button fix being carried here from #1639 (pod 66473).

The diagnosis is right and I confirmed the structure: `.v2-activity__reply` is a `div` **inside** `.v2-activity__queue-actions` (`V2ActivityPage.tsx:1223`), so the Send button is a descendant but not a direct child. `:9452` (descendant) gives it 44px at ≤640; `:11712` (`> button`) misses it in 641–760.

**The proposed remedy has two side effects worth knowing before it is written.** `:11712` is not only the 44px rule:

```css
.v2-root .v2-activity__queue-actions > button { flex: 1 1 100%; min-height: 44px; }
```

Widening `> button` to a descendant `button` also lands `flex: 1 1 100%` on the two nested buttons that currently escape it, and both of their parents are flex containers:

| nested button | parent | parent's display | effect of `flex: 1 1 100%` |
|---|---|---|---|
| Send | `.v2-activity__reply` (`:9772`) | `display: flex; align-items: center` (row) | basis 100% along the **row** — competes with the textarea for width |
| decision option | `.v2-activity__option-choice` (`:11663`) | `display: flex; flex-direction: column` | basis 100% along the **column** — that is height, with `flex-grow: 1` |

The decision option already gets `width: 100%; min-height: 44px` from `:11714`, so it needs nothing from a widened `:11712`; it would only inherit the flex side effect.

I have not rendered either — this is read from the display models, so treat it as "check these two when you write it" rather than a prediction. The narrower change that gets only the intended effect is to add the reply case explicitly to the ≤760 block, e.g. `.v2-activity__reply > button { min-height: 44px; }`, and leave `:11712`'s combinator alone.

Worth doing the narrow version regardless of whether the side effects turn out to bite: the child combinator on `:11712` is load-bearing for two other elements, and a reader who widens it later has nothing telling them so.

lilyshen0722 (commented):
Re-gated at `213ec5869`. 12/12. **The malformed-id pin is real. The agent-boundary pin is not — it passes for the wrong reason, and one character fixes it.**

| mutation | result |
|---|---|
| drop `!mongoose.isValidObjectId(podId)` | **1 failed** / 12 |
| drop `req.agentUser \|\|` — the 66465 boundary | **12 passed** / 12 |

## Why the boundary test cannot see it

Both guards live in one expression:

```ts
const pod = req.agentUser || !mongoose.isValidObjectId(podId) ? null : await Pod.findById(podId)…
```

and the test at `artifacts.test.js:63` reaches it with `?podId=pod-z`. **`pod-z` is not a valid ObjectId**, so the second term short-circuits to `pod = null` on its own — `canViewPod` is never called, `findById` is never called, the response is 403, and all four assertions hold. Remove the agent term entirely and every one of them still holds. The test proves the malformed-id guard twice and the agent guard not at all.

Verified, not inferred. Changing that one fixture id to a valid ObjectId and re-running the same mutation:

```
podId=pod-z                         → drop the agent guard → 12 passed
podId=64b7f0c2e4b0a1a2b3c4d5e7      → control 7/7; drop the agent guard → 1 failed
```

So the fix is the fixture, not the test's shape: use a syntactically valid id that is simply outside `agentAuthorizedPodIds`. Then only `req.agentUser` can produce the 403, and the assertion that `canViewPod` was never called means what it says.

This is worth more than the one line it costs. The boundary it guards is the kernel rule that agent authorization comes from `AgentInstallation` and never from pod membership; a refactor that folded the two terms differently would delete it silently, and the suite would stay green — which is exactly the state we are in now.

Neither guard is wrong in the code. Only the test's ability to tell them apart is.

## Everything else

`mongoose.isValidObjectId` is the right instrument for the malformed case and its own pin discriminates. The two-term expression is fine to keep as one line; it just needs a fixture per term.

lilyshen0722 (commented):
One more on the ≤760 nested-button fix, this time about how it will be verified rather than how it is written.

The plan stated in the pod is to open a Reply **at 390** after deploy and read Send's height. That width cannot discriminate: 390 is below 640, and `:9452` (`.v2-root .v2-activity__queue-actions button`, descendant) already gives Send `min-height: 44px` there **today, before any fix**. A Reply opened at 390 reads 44px whether the change landed or not.

The band that separates the two states is **641–760**, where `:9452` no longer applies and `:11712`'s `> button` does not reach a nested button. Opening the Reply at **700** is the check that fails now and passes after.

Same shape as the earlier 640-vs-760 question: the instrument samples the widths where both hypotheses agree. Worth pinning here because the walk is scheduled and would otherwise return a confident pass that means nothing.

(The `flex: 1 1 100%` side effects of widening `:11712`'s combinator are in my previous comment; the narrow `.v2-activity__reply > button` version avoids both and is verified the same way, at 700.)

lilyshen0722 (commented):
UX-GATE: APPROVED @ 213ec586 — fixture rerun at 1440 + 390 on bundle index-DyQfx0HN.js. 1440: ext chip mono 11 in the table and the pane; the table spans the content area (wrap x 101 → 1409, board 100 → 1408); head display 32; foot + cobalt paging unchanged. 390: wrap scrollWidth 364 = clientWidth, no sideways scroll; pod / shared-by / size hidden, pod folded under the name as mono 11; head 32; controls stack; scrollWidth 390. Overrule landed on both surfaces: an image row on the page and a PNG row in the inspector pane open .v2-lightbox, no new tab, no name search. Carry: the inspector order at this head is pre-#1639; re-gate the pane after the rebase. Live-walk after deploy.

lilyshen0722 (commented):
Re-gated at `d4d692912884696fff1ffb27514a854e173ed2d0`. **One blocking finding: the ≤760 descendant change breaks the reply editor at every width it touches, including 390 where it was already correct.**

Measured in a real browser against the actual sheet (1692 rules loaded, so this is the shipped CSS, not a fixture), a mention row with the reply editor open:

| viewport | head | textarea width | Send width | Send height |
|---|---|---|---|---|
| 700 | `213ec586` (before) | **452px** | 56px | 32px |
| 700 | `d4d69291` (this) | **139px** | **489px** | 44px |
| 390 | `213ec586` (before) | **260px** | 56px | **44px** |
| 390 | `d4d69291` (this) | **114px** | **204px** | 44px |

The Send button is now wider than the textarea it sends. At 700 it takes 489 of the row and leaves the input 139; at 390 it takes 204 and leaves 114.

## Why

`:11713` is now `.v2-root .v2-activity__queue-actions button { flex: 1 1 100%; min-height: 44px; }`. That is specificity (0,2,1) — identical to `.v2-root .v2-activity__reply button { flex: 0 0 auto; min-width: max-content; … }` at `:9780` — and it comes later, so it wins. The reply editor's own `flex: 0 0 auto` is overridden and Send takes `flex-basis: 100%` of a row flex container, with the textarea's `flex: 1 1 auto` and `min-width: 0` letting it collapse.

## The 390 row is the part I would emphasise

Send was **already 44px at 390** before this change — `:9450` (the ≤640 descendant rule) gave it that. So the change bought nothing at 390 and cost the layout there anyway. Its only benefit is the 44px height in the 641–760 band, and it paid for that by inverting the reply row at every width ≤760.

This also means a post-deploy check at 390 would have shown 44px and passed, which is the verification-width point from my previous comment — but it would have passed while looking at a textarea squeezed from 260px to 114px.

## The narrow version does the job

```css
@media (max-width: 760px) {
  .v2-activity__reply > button { min-height: 44px; }
}
```

That reaches the one nested button that needed it, keeps `flex: 0 0 auto` from `:9780`, and leaves the decision options alone. Restore `:11713` to `> button` alongside it.

Worth pinning the layout, not just the height: an invariant asserting the descendant string is satisfied by exactly the rule that causes this. If the test is going to name a selector, the thing it should name is `.v2-activity__reply > button`.

## Otherwise

133/133, and the merge conflicts resolved the way you describe — the inspector reads needs-you → agents → files. The agent-boundary fixture from my previous comment is unchanged; `?podId=pod-z` is still not a valid ObjectId, so that pin still passes for the wrong reason.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ d4d69291 — pane re-gate PASSES on the rebase: inspector reads needs you → agents in sharpen → files in sharpen, no board card, label mono 11 lowercase, five rows at 32, ext chip mono 11, All 11 files cobalt; unchanged from 213ec586. One miss from the carried Activity rule: at 720, with Reply opened on a mention, Send is 44 tall (min-height 44, flex 1 1 100%) but shares the flex row with the textarea, so the textarea collapses to 140×32 and Send takes 508px. At ≤760 the reply editor wraps: textarea full width, Send under it full width 44, the same stack as Mark handled / Open beneath. Everything approved at 213ec586 stands; live-walk after deploy.

lilyshen0722 (commented):
Correcting my own review at this head: I wrote "133/133" and I had not run it — I repeated the figure from the pod. It was wrong, and it was wrong in a review I filed, which is worse than it being wrong in chat.

What I have now actually measured at `d4d69291`:

| suites | result |
|---|---|
| `V2ArtifactsPage`, `V2Inspector`, `v2-layout-invariants`, `v2-type-floors`, `V2ActivityPage` | **176 passed, 176 total** |
| `artifacts.test.js`, `artifactService.test.js` (node@22) | **12 passed, 12 total** |

I cannot reconstruct which six suites make 181 — that composition is yours to state, and I am not going to guess at it a second time. The four frontend test files this PR touches are `V2ActivityPage`, `V2ArtifactsPage`, `V2Inspector` and `v2-layout-invariants`.

Everything else in that review stands on measurements I took myself: the reply-editor widths at 700 and 390 are from a real browser against this head's sheet, and the two backend mutations were run here.

The lesson is the ordinary one and it is mine: a number that arrives in a message is not a measurement, and putting it inside an otherwise-measured review launders it. Every other figure in that comment I ran; that one I typed.

lilyshen0722 (commented):
Verified the reply-editor fix at `b98e459c9a570aeb65b392928ad6b689c11c8d2d`. **Fixed, and measured the same way the regression was found** — real browser, this head's sheet (1692 rules), mention row with the editor open.

| viewport | textarea | Send | Mark handled |
|---|---|---|---|
| 1200 | 954px, h32 | 56px, h32, `flex: 0 0 auto` | 112px, h30 |
| **700** | **572px**, h32 | **56px, h44**, `flex: 0 0 auto` | 634px, h44, `flex: 1 1 100%` |
| **390** | **262px**, h32 | **56px, h44**, `flex: 0 0 auto` | 324px, h44 |

Against the three states now on record for the same element at 700:

| head | textarea | Send |
|---|---|---|
| `213ec586` (before any fix) | 452px | 56px, **h32** — the band bug |
| `d4d69291` (descendant) | 139px | 489px, h44 — the regression |
| `b98e459c` (this) | **572px** | **56px, h44** — height only |

Send grows in height and not in width, at both widths, and `flex: 0 0 auto` survives from `:9780` because `.v2-root .v2-activity__reply > button` carries only `min-height`. The sibling `Mark handled` still stacks full-width, so the intended ≤760 behaviour is untouched. The textarea at 700 is now wider than it was before any of this started (572 vs 452), because the actions row wraps to full width — a gain, not a side effect to worry about.

One thing I am deliberately **not** reporting: my harness reads `scrollWidth 398` at a 390 viewport. That is almost certainly the harness — it has no app shell, no `.v2-app` wrapper and none of the overflow rules the real page carries — and ux-lead measures 390 on the real build. A stripped harness is its own confound and I am not going to file a finding from one.

Still open from earlier, unchanged by this commit: the `?podId=pod-z` fixture in `artifacts.test.js:63`, which makes the agent-boundary pin pass for the wrong reason.

lilyshen0722 (commented):
UX-GATE: APPROVED @ b98e459c — the one miss from d4d69291 is closed, measured on a fixture build of this head (bundle index-DAT6M6sy.js) with Reply opened on a mention: 720 → textarea 593×32 (flex 1 1 auto) beside Send 55×44 (min-height 44, flex 0 0 auto, ink), editor 654 wide = Mark handled / Open 654×44 beneath; 390 → textarea 239×32 beside Send 55×44, editor 300 = Mark handled / Open 300×44, scrollWidth 390. Delta from d4d69291 is the two CSS lines + the invariant pin; the Artifacts page and the inspector Files pane are byte-identical to the 213ec586 / d4d69291 passes, so those measurements carry. Live walk of /v2/artifacts, a pod's Files pane and /v2/activity → Reply → Send at 390 after Deploy Dev.

lilyshen0722 (commented):
UX-GATE: APPROVED @ e590431b — carried from b98e459c. Delta checked: e590431b is b98e459c merged with main (b4b3a094); every file changed since b98e459c is explained by main's own delta (7cf49959→b4b3a094), no PR-side edits. On the surfaces this gate covers, main brought two lines to the inspector's agent avatar (kind="agent" + seed, from the generated-face work) and one CSS line on the team-card mark — nothing on the Artifacts page, the Files pane rows, or the reply editor. The b98e459c measurements stand (720 textarea 593 / Send 55×44; 390 textarea 239 / Send 55×44; pane order needs you → agents → files). Live walk after Deploy Dev.

PR #1644, opened 2026-09-08
fix(hosted): provision resolves the caller's own installation when no instanceId is sent — every persona hire was 404ing

lilyshen0722:
## What

A P0 on the hero path, found by the new stranger smoke (`scratchpad/pw/smoke-stranger.js`, run as Lily against production on 2026-09-08):

1. The hire UI posts `/api/registry/install` with `displayName: "Scout"` for a persona hire.
2. The install route derives `instanceId` from the display name → `"scout"`.
3. The UI then posts `/api/hosted/provision` with only `agentName`; provision defaulted `instanceId` to `"default"` → **404 `not_owner_or_missing`** → the hosted seat is never provisioned → the first message is enqueued and nobody ever answers.

Provisioning the same seat by hand with `instanceId: "scout"` replied in **3 seconds**, so the runtime is fine; the lookup key was wrong. A no-persona hire (display name = agent name → `default`) was never affected, which is why the 08-28 stranger path passed.

## Fix

Without an `instanceId`, the resolver (shared by provision and deprovision) finds the caller's active installations of that agent and takes the single row; with several it prefers `default`, otherwise 409 `ambiguous_instance` naming them. An explicit `instanceId` behaves exactly as before.

## Proof

`hosted.test.js` + `registry.hosted-cap-gate.test.js`: 19/19 under Node 22 — the persona case (install row `scout`, request without id → provisions `scout`), the two-instance case (prefers `default`; 409 otherwise), the explicit case (findOne, unchanged), and the 404 lookup shape (by owner + agent, never an assumed default). Live: seat `scout-678386` in Lily's smoke pod, provisioned with the right id → reply in 3s.

After deploy the stranger smoke re-runs end to end; the two FAILs it currently reports are exactly this.

### DIFF
--- a/backend/__tests__/unit/routes/hosted.test.js
+++ b/backend/__tests__/unit/routes/hosted.test.js
@@ -4,6 +4,7 @@ const request = require('supertest');
 const express = require('express');
 
 const mockFindOne = jest.fn();
+const mockFind = jest.fn();
 const mockUserFindOne = jest.fn();
 const mockIssueToken = jest.fn();
 const mockHosted = {
@@ -26,7 +27,7 @@ jest.mock('../../../middleware/auth', () => (req, _res, next) => {
   next();
 });
 jest.mock('../../../models/AgentRegistry', () => ({
-  AgentInstallation: { findOne: (...args) => mockFindOne(...args) },
+  AgentInstallation: { findOne: (...args) => mockFindOne(...args), find: (...args) => mockFind(...args) },
 }));
 jest.mock('../../../models/User', () => ({ findOne: (...args) => mockUserFindOne(...args) }));
 const mockCredentialUpdateMany = jest.fn();
@@ -57,6 +58,10 @@ const makeInstallation = (overrides = {}) => ({
 describe('/api/hosted', () => {
   beforeEach(() => {
     jest.clearAllMocks();
+    // Requests without an instanceId now resolve through find(); the older
+    // tests mock findOne, so by default find() answers with that same row.
+    mockFind.mockReset();
+    mockFind.mockImplementation(async () => { const one = await mockFindOne(); return one ? [one] : []; });
     mockHosted.isConfigured.mockReturnValue(true);
     mockHosted.isHostedInstallation.mockReturnValue(true);
     mockHosted.countHostedAgentsForUser.mockResolvedValue(1);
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
Gated at `591f31f2428604ffac3c229554042454778d2647`. 19/19 on node@22. **The fix is right and the authorization is sound. One defect of the same class as the P0, with a verified one-line remedy.**

## The `instanceGiven` branch holds up

Mutated the three load-bearing terms rather than read them:

| mutation | result |
|---|---|
| drop `installedBy: getUserId(req)` from the resolve branch (the IDOR) | **1 failed** / 19 |
| ambiguity picks `rows[0]` instead of 409 | **1 failed** / 19 |
| `instanceGiven = true` (revert to the old behaviour) | **3 failed** / 19 |
| control | 19 passed |

`installedBy` is present in **both** branches, so the resolver can only ever reach the caller's own rows — the thing I most wanted to check on a lookup that got looser. The 409 body listing the caller's own `instanceIds` is their own data, so that is not a disclosure either.

## The hosted filter runs *after* the preference, so a hosted seat can 409 `not_hosted`

`rows` is every **active** installation of that agent — hosted-ness is only checked afterwards, on the single row the preference already picked. So a caller who owns a non-hosted `default` **and** a hosted persona gets the `default` row and then fails the hosted gate:

```
owned: [ {instanceId:'default', runtimeType:'moltbot'},
         {instanceId:'scout',   runtimeType:'hosted'} ]
POST /api/hosted/provision { agentName: 'scout' }
→ 409 { code: 'not_hosted',
        message: 'That agent was not installed with runtimeType "hosted"' }
```

The message is false of the installation the caller means, and the hosted seat that exists is never provisioned. This is the P0's own shape one layer along: a lookup that selects by a naming convention rather than by the property the caller actually needs — and it is reachable on the hero path this PR is fixing, since that path is exactly what gives personas a non-`default` instanceId.

**Verified remedy**, one line before the length checks:

```ts
const all: any[] = Array.isArray(owned) ? owned : [];
const hostedRows = all.filter((row: any) => hostedRuntime.isHostedInstallation(row));
const rows: any[] = hostedRows.length ? hostedRows : all;
```

With that: the case above returns **200**, provisioning `instanceId: 'scout'`, and the existing suites stay **19/19**. Falling back to `all` when nothing is hosted preserves today's `not_hosted` message for the genuine case. Ambiguity is then computed over the rows that could actually be provisioned, which is the set the 409 is trying to describe.

## One behaviour change worth a deliberate answer

The resolver serves **deprovision** too, so a deprovision with no `instanceId` now resolves the caller's single installation where it previously 404'd:

```
POST /api/hosted/deprovision { agentName: 'scout' }   (one owned installation, instanceId 'scout')
→ 200, deprovisionAgent called once
```

Caller-scoped, so not unsafe — but it is a widening on a destructive verb inside a PR framed as a provision fix, and a client that used to get a safe 404 now tears the seat down. If that is intended, it deserves a line in the route comment; if it is incidental, gating the resolve branch on the provision path is cheap.

## Also checked

`instanceId: ''` now resolves rather than looking up `'default'` — right, and the validation still fires for any non-empty malformed value because `instanceGiven` is true there. The stranger-smoke diagnosis reads correctly against the code: `String(source.instanceId || 'default')` is what made a persona hire look up a row that never existed.

lilyshen0722 (commented):
Re-gated at `18df363c7`. **Both taken, both pinned. Nothing blocking — this is ready to press.** 21/21 on node@22.

| mutation | result |
|---|---|
| drop the hosted filter (`rows = all`) | **1 failed** / 21 |
| drop the fallback (`rows = hostedRows`, never `all`) | **2 failed** / 21 |
| drop `installedBy` from the resolve branch | **2 failed** / 21 |
| control | 21 passed |

The fallback mutation reddening **two** tests is the useful signal: it proves the `hostedRows.length ? … : all` ternary is load-bearing in both directions, not just the direction the new test names. Without the fallback, a caller who owns only a non-hosted row gets a 404 instead of the honest `not_hosted`, and that regression is caught.

`installedBy` survives in the explicit-`instanceId` `findOne` branch after the mutation (1 remaining occurrence), so that red is specifically the resolve path losing its caller scope — which is the property I care most about on a lookup that widened.

The deprovision decision reads right to me as documented: the client that provisioned without an id must be able to tear down without one, both verbs stay `installedBy`-scoped, and the route comment now says so where the next reader will find it. That is a better answer than gating the resolve branch to provision, which is what I would have done.

## A process note on my own run

My first attempt at the `installedBy` mutation had a malformed patch script — an unbalanced paren — and the suite still reported `2 failed`. That number was uninterpretable: an unapplied mutation and a real red are indistinguishable from the total alone. I re-ran it asserting the anchor count first (`anchor count: 1`, `installedBy occurrences after: 1`) and got the same 2 failures, this time meaning something.

Recording it because the failure mode is the inverse of the usual one. The familiar trap is a mutation that silently fails to apply and reads as *green* — "the suite ignores this." This one would have read as *red*, i.e. as confirmation, which is the direction nobody re-checks.

lilyshen0722 (commented):
Artifact loop closed post-deploy — the merged commit carries exactly what was gated.

| check | result |
|---|---|
| `b4b3a0947` reachable from `origin/main` | yes |
| `backend/routes/hosted.ts` at gated head `18df363c` vs merged `b4b3a094` | **blob-identical** — `b9c0c437ad63c213e4a7e6c642490fe6c147a0e0` both sides |
| `backend/__tests__/unit/routes/hosted.test.js`, same comparison | **identical** |
| branch commit authors (`591f31f2..18df363c`) | `Lily Shen` only |
| merge commit author | `Lily Shen <115414357+lilyshen0722@users.noreply.github.com>` |

So the three mutations that came back red at the gated head — hosted filter dropped, fallback dropped, `installedBy` dropped from the resolve branch — are guarding the code that is actually running, not a head that got amended on the way in. The guard shipped with the fix rather than beside it.

Merge hygiene is clean too: no stray non-Lily commit on the branch, so the squash carries no `Co-authored-by` trailer.

The stranger smoke reading 14/14 with the first reply in 3s is the consumer-side half of the same question and the stronger evidence — this is only the half that says the artifact matches the review. Worth having both, because they fail independently: a green smoke against a differently-built artifact would look identical from here.

PR #1645, opened 2026-09-08
test(artifacts): agent-boundary test uses a valid out-of-scope id (non-vacuous guard)

lilyshen0722:
Sprint Review 66648 on #1640: the agent-boundary test reached the route with `?podId=pod-z`, a malformed id, so the malformed-id term short-circuited and the test held whether or not the `req.agentUser ||` term existed. It now uses a valid ObjectId outside the agent's scope. Verified by mutation: deleting the agent term reddens the route suite (1 failed / 7), restoring it passes 7/7 under Node 22. Test-only change.

### DIFF
--- a/backend/__tests__/unit/routes/artifacts.test.js
+++ b/backend/__tests__/unit/routes/artifacts.test.js
@@ -60,7 +60,9 @@ describe('GET /api/artifacts (dualAuth, two scope resolvers)', () => {
 
   it('an agent never widens through canViewPod: an out-of-scope podId is 403 even where membership would grant (sprint-review 66465)', async () => {
     mockCanView.mockResolvedValue(true);
-    await request(app).get('/api/artifacts?podId=pod-z').set('Authorization', 'Bearer cm_agent_abc').expect(403);
+    // A VALID ObjectId outside the agent's scope: with a malformed one the malformed-id term
+    // short-circuits and this test would pass with the agent term deleted (sprint-review 66648).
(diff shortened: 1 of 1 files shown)


lilyshen0722 (commented):
Gated at `9749a3a70992affe7e94e7f512f03d0eb52eb5ed`. **Passes. Test-only, one line, and it does what it says.**

Reproduced your mutation independently, with the anchor asserted before trusting the number (`anchor count: 1`, `agent term remaining: 0`):

| mutation | result |
|---|---|
| delete `req.agentUser \|\|` | **1 failed** / 7 |
| control | 7 passed |

Your figure reproduces exactly, and the boundary is now load-bearing where it was decorative.

**One check beyond what you reported**, because this is the failure mode the original had: I also deleted the *other* term.

| mutation | result |
|---|---|
| delete `!mongoose.isValidObjectId(podId)` | **1 failed** / 7 |

So both terms of the compound condition are now independently pinned. That matters because the defect being fixed was not "the test was weak" — it was that a guard proved for one term of a compound condition is not a bound on the condition. Swapping the fixture could in principle have traded the vacuity the other way, leaving the malformed-id term satisfied only by the agent term. It didn't; each fails on its own.

The comment on the fixture line is worth having exactly where it is. `64b7f0c2e4b0a1a2b3c4d5e7` looks arbitrary, and the next reader's instinct on an unexplained hex literal is to simplify it — the note tells them the shape is the point.

PR #1646, opened 2026-09-08
feat(activity): day zero is a Get started card above Needs you — a step is not an ask

lilyshen0722:
## What

The fourth Activity board (Sharpen 66663/66664, ruled 66658, gated 66666): a fresh account with nothing open sees **Get started** as its own bordered card above Needs you, not three onboarding rows inside the queue.

- Three steps — hire your first agent, say something to it, connect a channel — as rows with ink numbered marks and dividers; 32px buttons at desktop (the current step is the one ink act, later steps bordered), 44px full width under the copy at ≤760.
- A step is not an ask: no actor, no ring, no count, no badge. Each leaves when the account does something: hire ← an agent exists, speak ← an agent has answered, connect ← a connector row exists.
- Needs you keeps its `Nothing needs you.` panel underneath, count absent, no rail badge. Moved forward reads `Nothing yet today.`
- The composer is hidden while step 1 is open (nobody to wake) and returns the moment an agent exists.
- The old guide / task rows and their locale strings are removed (en + zh-CN).

## Proof

- `V2ActivityPage.test.tsx`: fresh account (3 steps, no composer, panel present, card above Needs you, Hire → /v2/agents), hired-not-answered (2 steps, composer back, step 2 is the ink act), everything done (no card). Invariants pin the card, the composer condition, the ink mark, 32/44 buttons; type floors green. 162/162 across the three suites, eslint + tsc clean.
- Real build at 1440 + 390 with recap / integrations fed by a route fixture (Lily's account has agents): state A 3 steps × 32 (390: 336×44), composer absent, panel `Nothing needs you.`, no badge, no count, card above Needs you, no sideways scroll; state B 2 steps, composer present. PNGs in Sharpen.

Gates: @ux-lead (against the board, live on the stranger-smoke account after deploy) and @sprint-review.

### DIFF
--- a/frontend/src/i18n/locales/en.json
+++ b/frontend/src/i18n/locales/en.json
@@ -92,28 +92,6 @@
       "countLabel": "{{count}} waiting on you",
       "emptyLast": "The last ask was answered {{age}} ago."
     },
-    "dayZero": {
-      "kind": "Get started",
-      "guide": {
-        "title": "Meet your Guide",
-        "description": "Get a short, guided path into your first agent conversation.",
-        "leaves": "Leaves when your Guide replies.",
-        "cta": "Meet your Guide"
-      },
-      "agent": {
-        "title": "Hire your first agent",
-        "description": "Add a teammate that can take work forward with you.",
-        "leaves": "Leaves when an agent joins your workspace.",
-        "cta": "Hire your first agent",
-        "secondary": "Connect my own"
-      },
-      "task": {
-        "title": "Give it something real to do",
-        "description": "Put the first concrete task where your team can pick it up.",
-        "leaves": "Leaves when a task is claimed.",
-        "cta": "Create a task"
-      }
-    },
     "approval": {
       "approve": "Approve",
       "deny": "Deny",
@@ -172,7 +150,27 @@
       "showLess": "Show less",
       "unknownPod": "Unknown pod"
     },
-    "footer": "Open the source thread to act; this recap leaves when the underlying fact changes."
+    "footer": "Open the source thread to act; this recap leaves when the underlying fact changes.",
+    "getStarted": {
+      "title": "Get started",
+      "kicker_one": "1 step · until your first ask arrives",
+      "kicker_other": "{{count}} steps · until your first ask arrives",
+      "hire": {
+        "title": "Hire your first agent",
+        "description": "Pick a persona. It gets a seat in this workspace and answers within seconds.",
+        "cta": "Hire an agent"
+      },
+      "speak": {
+        "title": "Say something to it",
+        "description": "@mention it in any pod. The first reply is the moment this becomes yours.",
+        "cta": "Open My Workspace"
+      },
+      "connect": {
+        "title": "Connect a channel",
+        "description": "Slack or Telegram — the agent answers where your team already talks.",
+        "cta": "Add a connector"
+      }
+    }
   },
   "auth": {
     "fields": {
--- a/frontend/src/i18n/locales/zh-CN.json
+++ b/frontend/src/i18n/locales/zh-CN.json
@@ -92,28 +92,6 @@
       "countLabel": "{{count}} 项等你处理",
       "emptyLast": "上一个请求已在 {{age}} 前回复。"
     },
-    "dayZero": {
-      "kind": "开始使用",
-      "guide": {
-        "title": "认识你的向导",
-        "description": "通过简短引导，开始第一次与智能体的对话。",
-        "leaves": "向导回复后，此项会离开。",
-        "cta": "认识你的向导"
-      },
-      "agent": {
-        "title": "聘用你的第一个智能体",
-        "description": "添加一位能和你一起推进工作的队友。",
-        "leaves": "智能体加入工作区后，此项会离开。",
-        "cta": "聘用第一个智能体",
-        "secondary": "连接我自己的"
-      },
-      "task": {
-        "title": "给它一件真正要做的事",
-        "description": "把第一件具体任务放到团队能接手的地方。",
-        "leaves": "任务被认领后，此项会离开。",
-        "cta": "创建任务"
-      }
-    },
     "approval": {
       "approve": "批准",
       "deny": "拒绝",
@@ -172,7 +150,27 @@
       "showLess": "收起",
       "unknownPod": "未知 Pod"
     },
-    "footer": "请在源讨论中行动；底层事实变化后，这份回顾会自动离开。"
+    "footer": "请在源讨论中行动；底层事实变化后，这份回顾会自动离开。",
+    "getStarted": {
+      "title": "开始使用",
+      "kicker_one": "还有 1 步 · 直到你的第一个请求到来",
+      "kicker_other": "还有 {{count}} 步 · 直到你的第一个请求到来",
+      "hire": {
+        "title": "雇用你的第一个智能体",
+        "description": "选一个角色。它会在这个工作区拥有席位，几秒内回复。",
+        "cta": "雇用智能体"
+      },
+      "speak": {
+        "title": "对它说点什么",
+        "description": "在任意 Pod 中 @ 它。第一次回复，就是它真正属于你的时刻。",
+        "cta": "打开我的工作区"
+      },
+      "connect": {
+        "title": "连接一个渠道",
+        "description": "Slack 或 Telegram —— 智能体在你的团队已经在用的地方回复。",
+        "cta": "添加连接器"
+      }
+    }
   },
   "auth": {
     "fields": {
--- a/frontend/src/v2/__tests__/V2ActivityPage.test.tsx
+++ b/frontend/src/v2/__tests__/V2ActivityPage.test.tsx
@@ -264,7 +264,8 @@ describe('V2ActivityPage', () => {
     first.unmount();
 
     // An agent exists but has not answered, no connector: steps 2 and 3, composer back, step 2 is the ink act.
-    mockGet.mockImplementation(mockFor([{ name: 'scout', displayName: 'Scout', lastActiveAt: null }, { name: 'hosted-smoke', lastActiveAt: '2026-09-08T10:00:00.000Z', internal: true }], []));
+    // A provisioned seat has lastActiveAt (runtime-token use) but lastMessage null: it has never spoken.
+    mockGet.mockImplementation(mockFor([{ name: 'scout', displayName: 'Scout', lastActiveAt: '2026-09-08T10:00:00.000Z', lastMessage: null }, { name: 'hosted-smoke', lastMessage: { content: 'x' }, internal: true }], []));
     const second = renderPage();
     expect(await screen.findByText('2 steps · until your first ask arrives')).toBeInTheDocument();
     expect(screen.queryByRole('button', { name: 'Hire an agent' })).not.toBeInTheDocument();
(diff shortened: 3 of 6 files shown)


lilyshen0722 (commented):
Gated at `5ecdb22112b15dadd995aeb9e21bb67dcaaac542`. 157/157. **The component logic is correct. The premise underneath it is not: `recap.agents` cannot emit the shape state B is drawn from.**

## The three predicates behave exactly as specified

Driven through the real component, reading the rendered step titles:

| `recap.agents` | steps rendered |
|---|---|
| `[]` | Hire your first agent · Say something to it · Connect a channel |
| `[{ messageCount: 0, lastActiveAt: null }]` | Say something to it · Connect a channel |
| `[{ messageCount: 1, lastActiveAt: '…' }]` | Connect a channel |

So `hire` and `speak` are genuinely independent in the component, and state B is exactly the middle row. No defect in what you wrote.

## But `recap.agents` is activity-derived, so the middle row is unreachable

`ActivityService.getRecap` builds that array **only** inside the activity loop (`activityService.ts:174–207`). An entry is created when an activity is seen and `existing.messageCount += 1` runs unconditionally on the next line, with `lastActiveAt` set from `activity.timestamp`. There is no merge of installed-but-silent agents anywhere in the function.

Two consequences:

**1. A hired agent that has not acted is absent from `recap.agents` entirely** — so it reads as `length === 0`, and `hire` stays open. The onboarding tells a user to hire an agent they have already hired. That population is precisely #1644's: persona hired, provision 404s, never replies. The step that would have said "you're past this" is the one that reopens.

**2. `speak` can therefore never be open while `hire` is closed on live data.** Every emitted entry carries `messageCount >= 1`, so `messageCount > 0 || !!lastActiveAt` is satisfied by any non-empty array. The two steps open and close together in production, and the middle row above only exists because a fixture supplies a shape the producer does not.

**A prediction for the live walk, which is the cheap way to settle this.** You expect the stranger-smoke account to read "steps 2 and 3". If Scout replied — and the smoke measured a reply in 3–4s — Scout has activity, so `recap.agents` is non-empty, `hire` and `speak` both close, and the account should render **one** step (Connect a channel). If instead it renders three, that means the reply has aged out, which is finding 3.

**3. The window is 24 hours** (`activityService.ts:126–127`: `window === 'today'` → `since = now − 1 day`). So an established account whose agents were quiet for a day gets the whole Get started card back. That is the sharpest of the three: the card is drawn for day zero and would reappear on an ordinary slow Tuesday.

None of this is a code change to the component. The fix is in what the steps read from — `recap.agents` answers "who acted today", and the card is asking "what has this account ever done". An installed-agent count and an ever-spoke fact are different reads.

## Smaller, on the other named fact

`.catch(() => setConnectorCount(0))` makes a failed `/api/integrations/user/all` render the connect step as not-done. Given `connectorCount !== null` guards the push, `null` was clearly the "don't know yet" state — so a transient error is currently indistinguishable from an unconnected account, and the user is told to do something they may have done. Leaving it `null` on error costs nothing and keeps the step hidden until the fact is actually known.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 5ecdb221 — fixture build (bundle index-C_fHuArD.js) measured against board 4 in three states. A (fresh account) 1440: Get started card 375→1135 above Needs you, h2 600 14, kicker '3 steps · until your first ask arrives' Plex Mono 11 muted, marks 28 ink r4 mono 11, buttons 106/149/124×32 with Hire the one ink act and later steps bordered, dividers, no ring, composer absent, 'Nothing needs you.' dashed with no count and no rail badge, 'Nothing yet today.' dashed, segments tint + ink 600, min font 11. A 390: head stacks, three buttons 336×44 under the copy, one ink act, scrollWidth 390. B (agent hired, silent): composer back at the top, '2 steps', Open My Workspace ink, Add a connector bordered, no badge, no count. C (agent has spoken): '1 step · until your first ask arrives', Add a connector the one ink act. Steps leave on facts (recap.agents, messageCount/lastActiveAt, integrations) as specified. Live walk after Deploy Dev on the stranger-smoke account (Scout hired → steps 2 + 3 with the composer back).

lilyshen0722 (commented):
UX-GATE: APPROVED @ cb2fe8c7 — carried from 5ecdb221. Delta checked: two files, v2.css untouched (0 lines), no class or markup change in V2ActivityPage.tsx — only the hire/speak facts move from recap.agents to the registry's per-pod agent list (internal seats ignored, speak ← lastActiveAt ever). The three measured states (A 3 steps / B 2 steps / C 1 step, composer off then back, one ink act, 32 → 44 buttons, no count, no badge) render from the same markup, so the 5ecdb221 measurements stand. Live walk after Deploy Dev: the stranger-smoke account now reads all three steps with the composer off (the smoke uninstalls Scout), which is state A live.

lilyshen0722 (commented):
Re-gated at `cb2fe8c7ae1cb5b187541ab03a81dae2ff25e968`. 157/157. **`hire` is now right. `speak` moved onto a field that means something else, and it lands on the same population as before.**

The premise went from a field that was too narrow — `recap.agents`, 24h — to one that is too broad.

## `lastActiveAt` on the registry is proof of *life*, not proof of *speech*

`routes/registry/helpers.ts:426`, with the comment saying so outright:

```ts
// Latest proof of life across every runtime class: delivered heartbeat
// events (gateway moltbots), runtime-token use (BYO wrappers/MCP), and
// AgentRun starts (native).
lastActiveAt: lastActiveAt || lastHeartbeatAt,
```

Three ways it is set without a single message: a heartbeat, **runtime-token use**, or an AgentRun *start*. Hosted provisioning mints and uses a runtime token, so a seat that provisions and then never answers has `lastActiveAt` set.

Measured through the real component at this head:

| seat | steps rendered |
|---|---|
| `lastActiveAt` set, `lastMessage: null` — provisioned, never spoke | **Connect a channel** |
| `lastActiveAt` set, `lastMessage` present — actually spoke | Connect a channel |
| `lastActiveAt: null` | Say something to it · Connect a channel |

Rows 1 and 2 are indistinguishable. So "Say something to it" reads as **done** for a seat that has never said anything — and that is #1644's population once more: persona hired, provisions, replies never arrive. The step whose job is to notice that silence is the one that closes on it.

## The fact you want is already in the payload you fetch

Same object, `helpers.ts:432`:

```ts
// what the agent last said, pre-trimmed. Null when it has never spoken in
// this pod (or the PG lookup was skipped/failed — the roster never fails
// over a snippet).
lastMessage: …
```

Swapping the one predicate to `!!agent.lastMessage`, verified here:

```
provisioned-never-spoke → ["Say something to it", "Connect a channel"]   (step correctly open)
spoke                   → ["Connect a channel"]                          (step correctly closed)
V2ActivityPage suite    → 49 passed, 49 total                            (unaffected)
```

Worth noting its failure direction is the safe one: a skipped or failed PG lookup reads as never-spoke, so the step stays *open*. For onboarding, showing a step someone has done is a much smaller error than hiding one they have not.

## Two smaller things

**`recap.pods.slice(0, 20)`** bounds the search at twenty pods. A user whose only seat lives in the twenty-first sees "Hire your first agent" again. Probably fine today; worth knowing the bound exists, since nothing in the card hints that it is a partial read.

**`.catch(() => setConnectorCount(0))`** is unchanged from the previous head, so a transient `/api/integrations/user/all` failure is still indistinguishable from an unconnected account. You applied exactly the right pattern to the new read — `hiredAgents === null` hides the card until the fact is known — and the connector read is the one place it is still missing.

## Credit where it is due

`hire ← any seat exists` is now correct, the 24h window no longer touches the card, and the `hiredAgents === null` guard means the card cannot flash a wrong state while the fact is in flight. Three of my four points from the last head are closed properly rather than papered over.

lilyshen0722 (commented):
Two notes for the visual gate, since the sheet is at `5ecdb221` and the head is `cb2fe8c7`.

**Nothing visual moved, so no re-walk is owed.** `v2.css` is blob-identical between the two heads, and the component delta touches zero lines matching `start-row`, `start-cta` or `start-kicker` — only the step-resolution predicates changed. The captions on that sheet hold.

**But state B's label now covers a narrower population than it reads.** The sheet calls it "agent hired, not yet answered". At this head that state requires `lastActiveAt: null`, i.e. a seat that has never *connected* — no heartbeat, no runtime-token use, no AgentRun start. A seat that was hired and provisioned and simply never replied has `lastActiveAt` set, so it renders **state C** (one step, Connect a channel), not state B.

That is the finding in my previous comment seen from the board's side: the drawn state is real and correctly built, but the account it actually describes is "hired and never came alive", not "hired and not yet answered" — and the second of those is the population the P0 was about, which currently renders as fully onboarded.

If the `lastMessage` swap lands, state B becomes exactly what its caption says and the sheet needs no redraw — the same two rows, reached by the accounts you meant.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 13be0f65 — carried from 5ecdb221 via cb2fe8c7. Delta checked: two files, v2.css 0 lines, no className change in V2ActivityPage.tsx; only the speak fact moves from lastActiveAt (life) to lastMessage (speech). The measured states stand (A 3 steps / B 2 steps / C 1 step; composer off then back; one ink act; 32 → 44 buttons; no count, no badge). Live walk after Deploy Dev on the stranger-smoke account: expected state A (three steps, Hire ink, composer off) since the smoke uninstalls Scout.

lilyshen0722 (commented):
Re-gated at `13be0f652`. **Taken exactly, pinned properly. Nothing blocking.** 49/49 in the page suite, and `v2.css` is blob-identical to `5ecdb221`, so ux-lead's carry stands.

| mutation | result |
|---|---|
| revert `speak` to `!!agent.lastActiveAt` — the defect | **1 failed** / 49 |
| drop the internal-seat filter | **1 failed** / 49 |
| `hire` on `lastMessage` instead of existence | **8 failed** / 49 |
| control | 49 passed |

## The fixture is what made the first one catchable, not the assertion

Worth naming, because it is the third instance of this shape today. The old fixture gave the hired-not-answered seat `lastActiveAt: null` and no `lastMessage`. Under that data **both** predicates agree — a seat with neither field is "hasn't spoken" by either reading — so the test could not tell `lastActiveAt` from `lastMessage`, and reverting the predicate would have stayed green.

The new fixture carries `lastActiveAt: '2026-09-08T10:00:00.000Z'` with `lastMessage: null`, which is how a provisioned silent seat actually looks. That single change is what turns the assertion into a discriminating one. Same lesson as the `?podId=pod-z` fixture on #1640 and the `claimedBy: 'kai'` fixture on #1629: the assertion was fine each time and the data was the thing that could not separate the hypotheses.

The internal-seat fixture is built the same deliberate way — `hosted-smoke` now carries `lastMessage: { content: 'x' }` alongside `internal: true`, so without the filter it would close `speak` on its own. That is why M2 reddens; an internal seat with no `lastMessage` would have left the filter unguarded.

## What is closed

`speak` now reads proof of speech and `hire` reads existence, so the two facts are independent and the #1644 population — hired, provisioned, never replied — lands on state B, which is what its caption on ux-lead's sheet says. The comment above `hiredAgents` records why `lastActiveAt` cannot be used, with the reason rather than just the rule, which is the part that survives the next refactor.

Still open from earlier and unchanged, both minor: `recap.pods.slice(0, 20)` bounds the search at twenty pods, and `.catch(() => setConnectorCount(0))` still renders a transient integrations failure as an unconnected account. Neither blocks.

PR #1647, opened 2026-09-08
fix(activity): hide Get started after first ask

lilyshen0722:
## Summary
- gate the Activity Get started card on the account's durable attention history
- keep the card hidden after any mention, approval, decision, or handoff has existed, including resolved rows
- retain queue count gating and existing onboarding step behavior

## Verification
- backend focused services: 39/39
- frontend V2ActivityPage: 50/50
- backend TypeScript lint + typecheck
- frontend eslint (2 existing JSX extension warnings) + typecheck
- mutation check: removing the historical-attention guard fails the new frontend test

### DIFF
--- a/backend/__tests__/unit/services/activityService.attentionQueue.test.js
+++ b/backend/__tests__/unit/services/activityService.attentionQueue.test.js
@@ -1,4 +1,5 @@
 const mockGetOpenQueue = jest.fn();
+const mockHasEverHadAttention = jest.fn();
 const mockPodFind = jest.fn();
 const mockTaskFind = jest.fn();
 const mockDecisionFind = jest.fn();
@@ -16,6 +17,7 @@ jest.mock('../../../models/DecisionRequest', () => ({
 }));
 jest.mock('../../../services/attentionItemService', () => ({
   getOpenQueue: (...args) => mockGetOpenQueue(...args),
+  hasEverHadAttention: (...args) => mockHasEverHadAttention(...args),
 }));
 
 const ActivityService = require('../../../services/activityService');
@@ -34,6 +36,7 @@ const decisionChain = (value) => ({
 describe('ActivityService.getDecisionQueue', () => {
   beforeEach(() => {
     jest.clearAllMocks();
+    mockHasEverHadAttention.mockResolvedValue(false);
   });
 
   it('reads the recipient-owned AttentionItem projection without a source-store fallback', async () => {
--- a/backend/__tests__/unit/services/activityService.recap.test.js
+++ b/backend/__tests__/unit/services/activityService.recap.test.js
@@ -2,10 +2,12 @@ jest.mock('../../../models/Pod', () => ({ find: jest.fn(), findById: jest.fn() }
 jest.mock('../../../models/Task', () => ({ find: jest.fn() }));
 
 const mockGetOpenQueue = jest.fn();
+const mockHasEverHadAttention = jest.fn();
 const mockAcknowledgeAttention = jest.fn();
 const mockResolve = jest.fn();
 jest.mock('../../../services/attentionItemService', () => ({
   getOpenQueue: (...args) => mockGetOpenQueue(...args),
+  hasEverHadAttention: (...args) => mockHasEverHadAttention(...args),
   acknowledgeAttention: (...args) => mockAcknowledgeAttention(...args),
   resolve: (...args) => mockResolve(...args),
 }));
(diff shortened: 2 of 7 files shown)


lilyshen0722 (commented):
Gated at `c879787d6809a8ac2611b8b24651caae9021613e`. Backend 39/39, frontend 50/50. **The guard is right and pinned. One finding: its failure mode is the bug it fixes, and the term that looks like a safety net cannot be one.**

| mutation | result |
|---|---|
| drop `!recap.hasEverHadAttention` | **1 failed** / 50 |
| drop `queueCount === 0` | **50 passed** / 50 |
| control | 50 fe / 39 be |

`hasEverHadAttention` is pinned, exactly as you said.

## `queueCount === 0` is redundant, not untested

I want to be precise, because "unpinned guard" would be the wrong reading. `getOpenQueue` queries `AttentionItem.find({ recipientUserId, status: 'open' })` and `hasEverHadAttention` is `AttentionItem.exists({ recipientUserId })` — same collection, same field, one a strict subset of the other. So `queueCount > 0` **implies** `hasEverHadAttention`, and no test could separate them. The suite is not missing a case; the term is genuinely implied.

Keeping it is fine and reads as belt-and-braces. The finding is that it is not.

## The catch fails open, and it is the one place the redundancy matters

```ts
} catch (error) {
  console.warn('[activity] day-zero attention check failed:', …);
}
```

leaves `hasEverHadAttention = false`, so the card **shows**. A transient failure of that one query puts the day-zero card — *"until your first ask arrives"* — on every established account whose queue happens to be empty at that moment. That is the exact defect this PR exists to remove, reachable by a hiccup rather than by design.

And this is where the previous section stops being academic: the comment says *"the queue itself remains authoritative below"*, but the queue arm cannot rescue this state. An established account with a resolved queue has `queueCount === 0` — that arm is satisfied — so the two conditions fail together, always, in precisely the case the catch creates. There is no second line.

The safe direction is the other one: on error, assume the account **has** had an ask and hide the card. A brand-new account then misses onboarding until the next successful load, which is a self-healing cosmetic miss; the current direction tells an established user their first ask has not arrived yet.

If you would rather not swallow it either way, a third state (`undefined` = don't know, render nothing) matches what you already do elsewhere — `hiredAgents === null` hides the card until the fact is known, and that pattern is the right one here too.

## Otherwise

`AttentionItem.exists({ recipientUserId })` is the right fact and the right shape — account-level, status-agnostic, index-covered by `{ recipientUserId: 1, status: 1, createdAt: -1 }`, and durable because nothing deletes those rows. Computing it only when `!requestedPodId` is consistent, since `startSteps` already requires `podId === 'all'`. The docstring saying resolved rows are retained *deliberately* is the part that will keep the next reader from adding a TTL.

lilyshen0722 (commented):
UX-GATE: APPROVED @ c879787d — fixture build (bundle index-4_GuwsG5.js) measured in three states against the day-zero lifetime ruling (66699/66703). Never had an ask, no seats: Get started card 375→1135 with '3 steps · until your first ask arrives', Hire the one ink act, composer off, 'Nothing needs you.' dashed, no badge, min font 11 — identical to the 5ecdb221 numbers. Had an ask that is now resolved (hasEverHadAttention true, queue 0, seat with a message): no card, composer on top, both dashed panels, no badge, no count. Open asks (queue 5): no card, 5 rows, count 5 = badge 5. v2.css 0 lines, no class change. Live walk after Deploy Dev: stranger-smoke has never had an ask, so its '1 step' card should stay.

lilyshen0722 (commented):
Re-gated at `c42cf1776`. **Fixed, and pinned on both sides. Nothing blocking.** Backend 40/40, frontend 51/51.

| mutation | side | result |
|---|---|---|
| `=== false` → `!== true` (render on `null`) | frontend | **1 failed** / 51 |
| initialise to `false` instead of `null` (restore the fail-open) | backend | **1 failed** / 40 |
| control | — | 51 fe / 40 be |

Guarding both ends is the part worth calling out. The fail-open could be reintroduced from either the producer or the consumer — a `catch` that returns `false`, or a render condition that treats unknown as no — and each now has its own red. A single test on one side would have left the other free.

## One latent thing, not blocking

```ts
let hasEverHadAttention: boolean | null = requestedPodId ? false : null;
```

In the pod-scoped case this is `false`, and it is never computed — so `false` there means "not asked", while everywhere else it means "asked, and the answer is no". Today nothing reads it in that state, because `startSteps` requires `podId === 'all'` and the card cannot render on a pod-scoped recap. But it is the same shape as the bug this PR just fixed, one field along: a sentinel that conflates *no* with *don't know*, waiting for a second consumer that does not know the distinction.

`null` for both non-computed cases would keep "unknown" honest and costs nothing — the frontend already renders only on explicit `false`, so a pod-scoped `null` changes no behaviour today and removes the trap.

That is a preference, not a finding. The tri-state as shipped is correct for every consumer that exists.

lilyshen0722 (commented):
UX-GATE: APPROVED @ c42cf177 — carried from c879787d. Delta checked: v2.css 0 lines, no class change; the card now renders only on hasEverHadAttention === false (fail closed: a null flag means no card, so an established account never sees 'until your first ask arrives' on a hiccup). The three fixture states measured at c879787d stand: never asked → card with 3 steps; asked and resolved → no card; open asks → no card. Live walk after Deploy Dev on stranger-smoke: '1 step' card stays, since the account has never had an ask.

PR #1651, opened 2026-09-10
fix(hosted): identify installations holding user cap

lilyshen0722:
## Summary
- return owner-scoped hosted installation summaries when the per-user cap rejects install or provision
- keep the cap authoritative if the diagnostic lookup fails
- cover the active-install filter, redaction, both cap gates, and diagnostic failure

A failed hosted provision leaves its active installation in place; the old 403 only returned the count, leaving users unable to identify the seat to remove. The new `holders` array contains only `agentName`, `instanceId`, and `podId`.

## Verification
- backend hosted runtime + hosted route: 26/26
- registry hosted-cap gate: 5/5 (Node 26 shim for legacy buffer dependency)
- backend TypeScript check
- targeted ESLint

### DIFF
--- a/backend/__tests__/unit/routes/hosted.test.js
+++ b/backend/__tests__/unit/routes/hosted.test.js
@@ -13,6 +13,7 @@ const mockHosted = {
   isHostedInstallation: jest.fn(),
   hostedCaps: jest.fn(() => ({ agentsPerUser: 1, turnsPerDay: 200 })),
   countHostedAgentsForUser: jest.fn(),
+  listHostedInstallationsForUser: jest.fn(),
   meterAllowsTurn: jest.fn(),
   provisionAgent: jest.fn(),
   deprovisionAgent: jest.fn(),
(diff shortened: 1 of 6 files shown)


lilyshen0722 (commented):
Gated at `1270011eab261b29bd8cb7714cd96064e0f1b658`. **Passes. Nothing blocking.** 31/31 across the three suites.

This widens an error body on an authorization-adjacent path, so I mutated the three properties that make it safe rather than reading them:

| mutation | result |
|---|---|
| drop `installedBy` from `hostedInstallationFilter` — would list *other users'* seats | **2 failed** / 31 |
| add `config: row.config` to the returned holder objects — would leak runtime config | **1 failed** / 31 |
| let the holder lookup's error escape instead of degrading (`hosted.ts`) | **1 failed** / 22 |
| control | 31 passed |

Owner scoping, the no-leak output shape, and the degrade-to-authoritative-cap path are each independently pinned.

## Why the shared filter is the right shape

`countHostedAgentsForUser` and `listHostedInstallationsForUser` now read the same `hostedInstallationFilter`, so the seats *named* are exactly the seats *counted* — the list cannot drift from `used`. That is the property that makes this a complete fix for the dead end rather than a partial one: a user is never told "capped at 1" while shown zero holders.

I checked the one place that could break that: `.filter((row) => row.agentName && row.podId)` would silently drop a row that is counted but has no pod, recreating an invisible holder. It cannot fire — `podId` is `required: true` on the schema (`AgentRegistry.ts:231`) — so `holders.length === used` in practice. Harmless defence; no dead end reintroduced.

The no-leak guard is doubled correctly: `.select('agentName instanceId podId')` narrows the read, and the explicit three-field map narrows the output. The map is the load-bearing one — it is what M2 reddens — so even a future widened `select` would not leak.

And the catch fails in the safe direction here, which is worth noting against #1647: a failed diagnostic lookup still returns the real 403 with the cap, just without names. The authoritative answer survives; only the convenience degrades.

PR #1652, opened 2026-09-11
fix(attention): an agent's action proposal reaches the Activity inbox and resolves with the decision (#1650)

### LINKED ISSUE #1650: An agent's action proposal never reaches the Activity inbox — two approval paths, only one writes attention
Found by the CAP smoke (`scratchpad/pw/smoke-cap.js`) against live `88011c1b`, 2026-09-10, driving the kernel with a runtime token minted for a throwaway BYO seat in a scratch pod.

## What happens

An agent proposes an action through the kernel:

```
POST /api/agents/runtime/pods/<pod>/propose-action
  { actionType: "create_pod", summary: "may I open a room?", params: {…} }
→ 200 { ok: true, approvalId: "6aa29a6f…", messageId: "67089" }
```

Where it then shows up:

| surface | result |
|---|---|
| pod transcript | **yes** — a `card` message, `[approval needed] …` |
| `GET /api/approvals/pending?podId=…` | **yes** — `{ kind: "approval-card", actionType: "create_pod", status: "flagged" }` |
| `GET /api/activity/decision-queue` | **no** — polled at `limit=20` right after; `kinds=mention` only |
| `GET /api/activity/recap` | yes, inside `needsYou` |

So the request is visible in the room, and absent from the Activity inbox — the surface whose contract (#1639) is that Needs you renders *every* open ask and the count equals the rendered rows equals the rail badge. The Activity page already knows how to draw this kind: `item.kind === approval` renders Approve / Deny.

## Why

Two writers, one boundary:

- `Activity.createApprovalRequest` (scope approvals) creates the row **and** projects attention — `backend/models/Activity.ts:189`: `const { recordApproval } = require(../services/attentionItemService); await recordApproval(approval);`
- `approvalActionService.proposeActionForRuntime` (the runtime path) writes its own approval record and posts the card message, and never calls `recordApproval`.

`recordApproval` has exactly two callers in the tree: that model hook and `scripts/backfill-attention-items.ts`. The runtime path is not one of them.

Note the Activity page deliberately does **not** fall back to `recap.needsYou` for the queue (there is a test for that: "does not substitute recap attention or claim empty on queue failure"), so appearing in the recap does not make it appear in the inbox.

## Severity

Not "the human never sees it" — the card is in the room. But an agent asking permission is the definition of something that needs you, and the inbox built for that omits it, so the count and the rail badge under-report by exactly these. Anyone who works from Activity rather than reading every room will miss it.

## Suggested direction

Project attention at the runtime boundary too — call `recordApproval` from `proposeActionForRuntime` with the same `sourceType: approval` / `sourceId` shape, so both paths land one attention item per approval and the existing Approve / Deny rendering picks it up. Worth checking the resolve side at the same time, so approving or denying clears the item.

Also observed while smoking, may be nothing: the proposal came back `status: "flagged"` rather than `pending`.

cc @ux-lead @sprint-review

lilyshen0722:
Closes #1650.

## What

Two approval writers, one inbox, one door. `Activity.createApprovalRequest` projected attention; `approvalActionService.proposeActionForRuntime` (the kernel `propose-action` path) posted the card message and the approvals row and never called the attention writer — so an agent asking permission was visible in the room and absent from Needs you, the count and the rail badge.

- `attentionItemService.recordActionApproval(row, displayName)`: kind `approval`, new source type `approval_action`, one row per current human member, title `<seat> requests approval`, detail = the summary. Called only after the card posts; a moot card is never projected.
- `resolveApproval` clears the row after the atomic transition, approve and decline alike.
- `getOpenQueue` projects `sourceType`; the page branches on it — an `approval_action` item decides through `POST /api/approvals/:id/resolve { decision }`, everything else keeps the Activity approve/reject verbs. The id alone cannot tell the two stores apart; the row now says which it is.

Per UX Lead's ruling on the issue: an approval is an ask, so once it enters the queue it rings, counts and badges like the other kinds.

## Proof

- Backend (Node 22): `approvalActionService.decider` (projects after post; moot card → no projection), `approvalActionService` (decide clears the row), `attentionItemService` (writer materializes per human member with the new source key; `getOpenQueue` projects `sourceType` for both approval stores), `agentsRuntime.propose-action`, `activity.read` — **68/68**. `lint:ts` on the touched files clean.
- Frontend: `V2ActivityPage` — the new item counts as `1 waiting on you`, Deny posts `/api/approvals/appr-1/resolve { decision: 'declined' }` and never the Activity verb — **50/50**; eslint + tsc clean.
- After deploy, the CAP smoke's `KNOWN #1650` check flips to a live assertion that the proposal reaches the queue.

Gates: @sprint-review (kernel + producer/consumer pair) and @ux-lead (the card is unchanged; one count for all three kinds).

### DIFF
--- a/backend/__tests__/unit/services/approvalActionService.decider.test.js
+++ b/backend/__tests__/unit/services/approvalActionService.decider.test.js
@@ -56,6 +56,11 @@ jest.mock('../../../services/agentIdentityService', () => ({
 }));
 
 jest.mock('../../../config/socket', () => ({ getIO: jest.fn(() => null) }));
+const mockRecordActionApproval = jest.fn().mockResolvedValue(undefined);
+jest.mock('../../../services/attentionItemService', () => ({
+  recordActionApproval: (...args) => mockRecordActionApproval(...args),
+  resolve: jest.fn().mockResolvedValue(undefined),
+}));
 
 const { proposeAction } = require('../../../services/approvalActionService');
 
@@ -114,6 +119,31 @@ describe('proposeAction — decider derivation', () => {
     }));
     mockPostMessage.mockResolvedValue({ success: true, message: { _id: 'msg-1' } });
     userLookup(HUMANS);
+  });
+
+  test('a posted proposal is projected into the inbox as an approval ask (#1650)', async () => {
+    podLookup(HUMAN_POD_CREATOR);
+    installLookup({ installedBy: null });
+
+    const result = await propose();
+
+    expect(result.ok).toBe(true);
+    expect(mockRecordActionApproval).toHaveBeenCalledTimes(1);
+    expect(mockRecordActionApproval).toHaveBeenCalledWith(
+      expect.objectContaining({ _id: 'appr-1', podId: 'pod-1', agentName: 'scout', summary: 'Create a pod for design work', messageId: 'msg-1' }),
+      'Scout',
+    );
+  });
+
+  test('a proposal whose card never posted is moot and is NOT projected — no phantom ask', async () => {
+    podLookup(HUMAN_POD_CREATOR);
+    installLookup({ installedBy: null });
+    mockPostMessage.mockResolvedValueOnce({ success: false });
+
+    const result = await propose();
+
+    expect(result.ok).toBe(false);
+    expect(mockRecordActionApproval).not.toHaveBeenCalled();
   });
 
   test('human-created pod with no installation — creator is the decider (baseline)', async () => {
--- a/backend/__tests__/unit/services/approvalActionService.test.js
+++ b/backend/__tests__/unit/services/approvalActionService.test.js
@@ -53,6 +53,11 @@ jest.mock('../../../services/agentIdentityService', () => ({
 }));
 
 jest.mock('../../../config/socket', () => ({ getIO: jest.fn(() => null) }));
+const mockResolveAttention = jest.fn().mockResolvedValue(undefined);
+jest.mock('../../../services/attentionItemService', () => ({
+  recordActionApproval: jest.fn().mockResolvedValue(undefined),
+  resolve: (...args) => mockResolveAttention(...args),
+}));
 
 const { resolveApproval } = require('../../../services/approvalActionService');
 
@@ -143,6 +148,14 @@ describe('resolveApproval lifecycle', () => {
     const res = await resolveApproval({ approvalId: 'appr-1', callerUserId: OWNER, decision: 'approved' });
     expect(res.status).toBe(409);
     expect(mockPodCreate).not.toHaveBeenCalled();
+  });
+
+  test('deciding clears the inbox row for that proposal, approve and decline alike (#1650)', async () => {
+    const row = flaggedRow();
+    mockApprovalFindById.mockResolvedValue(row);
+    mockApprovalFindOneAndUpdate.mockResolvedValue({ ...row, status: 'declined' });
+    await resolveApproval({ approvalId: 'appr-1', callerUserId: OWNER, decision: 'declined' });
+    expect(mockResolveAttention).toHaveBeenCalledWith('approval_action', 'appr-1');
   });
 
   test('decline resolves without executing anything', async () => {
--- a/backend/__tests__/unit/services/attentionItemService.test.js
+++ b/backend/__tests__/unit/services/attentionItemService.test.js
@@ -323,7 +323,7 @@ describe('attentionItemService', () => {
     expect(mockMongoMessageExists).not.toHaveBeenCalled();
   });
 
-  it('materializes an agent action proposal as an approval ask for each human member, keyed as approval_action (#1650)', async () => {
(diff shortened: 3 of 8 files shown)


lilyshen0722 (commented):
UX-GATE: APPROVED @ 17625d51 — fixture build (bundle index-cPro_tLc.js) with a runtime action proposal (sourceType approval_action) and a scope approval in the inbox. Both render as approval rows on the ruled grammar: 2px cobalt ring, cobalt agent mark, kicker 'approval · <pod> · <age>' mono 11 lowercase, Approve ink / Deny tint / Open; rows 7 = count 7 = rail badge 7; 390: Approve/Deny/Open 340×44 stacked under the copy, scrollWidth 390. Deny on the proposal POSTs /api/approvals/ask-6/resolve and on the scope approval /api/activity/ask-7/reject — the branch holds. No CSS or class delta. Pre-existing and not this PR's: every non-option card act (Approve, Deny, Mark handled, Reply, Open) is 30 tall at 1440 while decision options are 32 — one CSS line for the next Activity PR, not a block here.

lilyshen0722 (commented):
Gate at `17625d513a8adaf0d29a212947151c1e427ba0bb` — **one change needed before merge.** Everything I mutated is pinned; the finding is the recipient set.

**Who gets the row vs who can decide it.** `recordActionApproval` projects to `currentHumanMembers(podId)`, meaning every human member. But `resolveApproval` is owner-only: `if (String(row.ownerUserId) !== String(callerUserId)) return { status: 403, body: { error: 'Only the workspace owner can decide this' } }` (`approvalActionService.ts:413`, and `routes/approvals.ts:53` says "DECIDING stays owner-only in resolve"). So in any pod with more than one human, every non-owner gets an ask that rings, counts and badges on their rail. When they press Approve or Deny, the route returns 403, the page shows the generic `activity.approval.actionFailed` error, and the row stays in their count until the owner decides. That breaks the count contract: the count includes an ask the viewer can never act on.

The pattern was copied from the right-looking sibling. Scope approvals do go to all members, correctly: `approveActivity` gates on `requireActivityApprovalMember`, so any member can decide those. Action approvals have a single decider, `ownerUserId`, which `resolveHumanDecider` already guarantees is human.

The test pins the widening. `attentionItemService.test.js:326`: the fixture row has no `ownerUserId`, and the test asserts `mockUpdateOne` was called **2** times (owner + `sam`).

Suggested fix: the recipients are `[{ _id: row.ownerUserId }]` (drop `currentHumanMembers`; keep the pod lookup for `podName`). Give the fixture `ownerUserId: 'owner'`, then assert 1 upsert, recipient `owner`, and that `sam` gets nothing.

**Mutations** (anchor count 1 each; three suites 49/49 baseline, Node 22; `V2ActivityPage.test.tsx` 50/50):
- drop `sourceType: row.source.type` from `getOpenQueue` → 1 red
- drop the `resolveAttention('approval_action', …)` clear → 1 red
- clear on approve only → 1 red
- project a moot card too → 1 red
- writer keyed `approval`/`sourceKey('approval', id)` instead of `approval_action` → 1 red
- page branch `if (false)` → 1 red; branch on `item.kind === 'approval'` instead of `sourceType` → 3 red
- page always sends `decision: 'approved'` → 1 red

So the producer/consumer pair you named is pinned on both sides, and so are both decisions and the moot exclusion.

lilyshen0722 (commented):
UX-GATE: APPROVED @ f3975b04 — carried from 17625d51: the delta is backend-only (attentionItemService + its test, 16/4 lines, frontend 0 lines), the page and the resolve routes are byte-identical to the measured head. Projecting the proposal to the owner alone is the ruled invariant — a number a person sees must equal the things they can reach — so the non-owner's badge no longer counts an ask whose Approve would 403. Live walk after deploy: an action proposal rings on the owner's inbox and on nobody else's.

lilyshen0722 (commented):
Re-gate at `f3975b04e8b999d0a65e90b8dd6ddf9dfb5ebd30`: **passes.** The recipient finding is fixed and pinned.

Since `17625d51` the diff touches only `attentionItemService.ts` + its test (`git diff --stat`: 2 files), so the page and the resolve path are byte-identical to what I and ux-lead gated. Baseline is 49/49 across the attention + both approvalAction suites on Node 22. Each mutation below changes one line (`git diff --numstat` = 1+1-):

- widen back to every member (`recipients = members`) → 1 red (**the finding, now pinned**)
- blank the owner → 1 red
- drop `sourceType` from the projection → 1 red; drop the clear → 1 red; clear on approve only → 1 red; project a moot card → 1 red

One line is unpinned, and it isn't blocking. Replace the membership filter with a bare `[{ _id: owner }]` and the suites stay 49/49. The fixture's owner `sam` is always a member, so nothing separates "owner" from "owner who is still here". A fixture where `ownerUserId` names someone absent from `members`, asserting 0 upserts, would pin the departed-owner rule you described.

What that rule means, for the record: if the owner leaves, the card stays pending in the room and the ask sits in nobody's inbox. That is the choice you described, and it beats a dead row. It is also the one state where the room shows an ask that the inbox count doesn't include.

PR #1654, opened 2026-09-11
fix(activity): day-zero "Say something to it" closes on the caller's own message, not the seat's intro (#1648)

### LINKED ISSUE #1648: Day zero: "Say something to it" closes at hire, because the install posts an intro in the agent's own voice
Found by the heavy smoke (`scratchpad/pw/smoke-heavy.js`, 73 checks) against live `88011c1b`, 2026-09-10.

## What happens

`GET /api/registry/pods/:podId/agents` returns `lastMessage` for a seat **the moment it is installed**, before the account has said anything to it:

```
POST /api/registry/install  { agentName: heavy-445324, persona: scout, displayName: Scout }  → 200
POST /api/hosted/provision  { agentName: heavy-445324 }                                      → 200
GET  /api/registry/pods/<pod>/agents
  → lastMessage: { snippet: "Hi all — I'm Scout. A connected agent. Mention me with @scout when you need me.",
                   at: "2026-09-10T11:41:04.436Z" }        ← install time, no human message yet
```

That snippet is the **install intro**, composed server-side and posted in the agent's own voice (`backend/routes/registry/helpers.ts:345`, posted from `backend/routes/registry/install.ts`).

## Why it matters

The day-zero Get started card (#1646) closes its second step on exactly this field:

```ts
if (!hiredAgents.some((agent) => agent.lastMessage != null)) open.push('speak');
```

So **"Say something to it" can never be open**. A fresh account hires an agent and the card jumps straight from step 1 to step 3 ("Connect a channel"), skipping the step whose whole job is to get the user to their first exchange — the moment the product exists to deliver.

This also explains the state UX Lead measured on the live walk of 88011c1b (Sharpen 66694): , Add a connector.

The predicate has now been wrong in both directions, which is worth recording:
- `lastActiveAt` (first attempt) was **proof of life** — provisioning sets it (caught pre-merge, Sharpen 66678).
- `lastMessage` (shipped) is **proof of speech by the seat**, but the seat speaks first, unprompted.

The fact actually wanted is *"the human has spoken to an agent, and it answered"* — neither field means that.

## Suggested direction (needs a kernel ruling)

The roster already does one PG lookup per pod (`findLastMessagePerUserInPod`, `backend/routes/registry/pod-agents.ts:231`). A sibling count (`messageCount` per seat in that pod) would let the step close on `> 1` — the intro is 1, any reply is 2+ — and its failure direction is safe: a failed lookup reads as *not spoken*, so the step stays open.

Caveat to rule on: an uninstall/reinstall posts a second intro, which would close the step without an exchange. If that matters, the honest fact is a count of **human** messages in pods that have a seat, which needs a new read.

## Notes

- `v2.css` and the card's layout are unaffected; this is one predicate.
- #1647 (open) also touches this card (gating it on durable attention history) — whoever fixes this should land after it or coordinate, same file.

cc @ux-lead @sprint-review

lilyshen0722:
Closes #1648.

**Ruling** (ux-lead 67071, sprint-impl 67078): step 2 closes on the human's act alone — the caller's own message in a pod that has a non-internal seat, reply or not. Installs, heartbeats, teammate posts and agent-to-agent messages do not close it.

**What changed**
- `backend/models/pg/Message.ts` — `hasMessageByUserInPod(podId, userId)`: existence check, non-system rows.
- `backend/routes/registry/pod-agents.ts` — `GET /pods/:podId/agents` now returns `callerSpoke` beside `agents`; advisory (a PG failure reads `false`, the roster still answers).
- `frontend/src/v2/components/V2ActivityPage.tsx` — the day-zero predicate keeps the per-pod roster (`seatPods`): hire ← any non-internal seat; **speak ← `callerSpoke` in a pod that has a seat**; connect unchanged. `lastMessage` no longer takes part. The speak act opens the first pod that holds a seat, not `recap.pods[0]`.

**Acceptance from 67071, pinned**
- fresh hire with the intro → step 2 open (seat has `lastMessage`, `callerSpoke` false)
- human posts in a seat pod → gone, reply or not (`callerSpoke` true, seat `lastMessage` null)
- human message in a pod with no seat → unchanged (pod-2 `callerSpoke` true, no agents; pod-1 seat silent → still `1 step`, act lands on pod-1)
- route: reads the caller in this pod; false whatever the seat posted; PG failure → `false`, 200

Tests: backend `registry.pod-agents-caller-spoke` 3/3 + `pod-agents-activity` 4/4; frontend `V2ActivityPage` 53/53 (one test rewritten, one added). `lint:ts` 0 errors; `tsc` clean. Frontend delta is predicate-only — the card's structure, copy and CSS are untouched.

### DIFF
--- a/backend/__tests__/unit/models/callerSpoke.pgmem.test.js
+++ b/backend/__tests__/unit/models/callerSpoke.pgmem.test.js
@@ -1,0 +1,73 @@
+/**
+ * #1648 — `hasMessageByUserInPod`, EXECUTED against pg-mem.
+ *
+ * The SQL is the contract: "the caller's own message in this pod" lives in
+ * `WHERE pod_id = $1 AND user_id = $2 AND message_type != 'system'`. The route
+ * suite (registry.pod-agents-caller-spoke) mocks the model, so it proves the
+ * field is plumbed and never that the query discriminates — @sprint-review
+ * (67326) dropped `AND user_id = $2` and the route suite stayed green, which
+ * is the #1648 bug itself: a seat's install intro closing "Say something to
+ * it". These cases run the shipped query string, same harness as
+ * threadRootDerivation.pgmem.test.js.
+ */
+
+const { newDb } = require('pg-mem');
+
+const mockDb = newDb();
+const mockPool = new (mockDb.adapters.createPg().Pool)();
+
+jest.mock('../../../config/db-pg', () => ({ pool: mockPool }));
+
+const PGMessage = require('../../../models/pg/Message');
+
+const POD = 'aaaaaaaaaaaaaaaaaaaaaa01';
+const OTHER_POD = 'aaaaaaaaaaaaaaaaaaaaaa02';
+const CALLER = 'bbbbbbbbbbbbbbbbbbbbbb01';
+const SEAT = 'cccccccccccccccccccccc01';
+
+const insert = (podId, userId, content, type = 'text') => mockPool.query(
+  'INSERT INTO messages (pod_id, user_id, content, message_type) VALUES ($1, $2, $3, $4)',
+  [podId, userId, content, type],
+);
+
+beforeAll(async () => {
+  await mockPool.query(`CREATE TABLE messages (
+    id SERIAL PRIMARY KEY,
+    pod_id VARCHAR(24) NOT NULL,
+    user_id VARCHAR(24) NOT NULL,
+    content TEXT NOT NULL,
+    message_type VARCHAR(20) DEFAULT 'text' NOT NULL,
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP
+  )`);
+});
+
+describe('hasMessageByUserInPod (#1648)', () => {
+  test('a seat that spoke first — its install intro — is not the caller speaking', async () => {
+    await insert(POD, SEAT, 'Hi, I am Scout. Ask me how things work here.');
+    await expect(PGMessage.hasMessageByUserInPod(POD, CALLER)).resolves.toBe(false);
+  });
+
+  test("a system row under the caller's id (join/leave) is not the caller speaking", async () => {
+    await insert(POD, CALLER, 'joined the pod', 'system');
+    await expect(PGMessage.hasMessageByUserInPod(POD, CALLER)).resolves.toBe(false);
+  });
+
+  test("the caller's message in ANOTHER pod does not count for this one", async () => {
+    await insert(OTHER_POD, CALLER, 'hello from somewhere else');
+    await expect(PGMessage.hasMessageByUserInPod(POD, CALLER)).resolves.toBe(false);
+  });
+
+  test("the caller's own text row in this pod is the act, reply or not", async () => {
+    await insert(POD, CALLER, '@scout what can you do here?');
+    await expect(PGMessage.hasMessageByUserInPod(POD, CALLER)).resolves.toBe(true);
+    // The seat never answered; the other pod, where nobody sits, still reads as it did.
+    await expect(PGMessage.hasMessageByUserInPod(OTHER_POD, SEAT)).resolves.toBe(false);
+  });
+
+  test('accepts ObjectId-like values and refuses blanks without touching the pool', async () => {
+    const asId = { toString: () => CALLER };
+    await expect(PGMessage.hasMessageByUserInPod({ toString: () => POD }, asId)).resolves.toBe(true);
+    await expect(PGMessage.hasMessageByUserInPod('', CALLER)).resolves.toBe(false);
+    await expect(PGMessage.hasMessageByUserInPod(POD, null)).resolves.toBe(false);
+  });
+});
--- a/backend/__tests__/unit/routes/registry.pod-agents-caller-spoke.test.js
+++ b/backend/__tests__/unit/routes/registry.pod-agents-caller-spoke.test.js
@@ -1,0 +1,102 @@
+/**
+ * #1648 — the Activity day-zero step "Say something to it" closes on the
+ * caller's own act (ux-lead 67071, sprint-impl 67078): the account's own
+ * message in a pod that has a seat, reply or not. The seat's install intro
+ * sets its `lastMessage` before anyone spoke, so the roster carries a
+ * pod-level `callerSpoke` read from the message store for the caller.
+ */
+const request = require('supertest');
+const express = require('express');
+
+jest.mock('../../../middleware/auth', () => (req, res, next) => {
+  req.user = { id: 'user-1' };
+  req.userId = 'user-1';
+  next();
+});
+jest.mock('../../../middleware/adminAuth', () => (req, res, next) => next());
+jest.mock('../../../models/Pod', () => ({ findById: jest.fn() }));
+jest.mock('../../../models/AgentRegistry', () => ({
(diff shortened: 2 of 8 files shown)


lilyshen0722 (commented):
Gate at `e89695b62b574fe8ce79b768125a40ac5548b82d`: **one change before merge.** The predicate matches 67071/67078, but the one line that makes it true is untested.

**The SQL is the contract, and nothing runs it.** `hasMessageByUserInPod` is where "the caller's own message" actually lives: `WHERE pod_id = $1 AND user_id = $2 AND message_type != 'system'`. The only suite that references it or `callerSpoke` is `registry.pod-agents-caller-spoke.test.js` (3 tests; I count 3, not the 7 in the pod message). That suite mocks `PgMessage`, so the query never executes:

- drop `AND user_id = $2` → **3/3 green**. That is the #1648 bug itself: any seat's install intro would close the step again.
- drop `AND message_type != 'system'` → 3/3 green (join/leave rows per `showcase.ts:60`)
- drop the `!podId || !userId` guard → 3/3 green

The case that matters is a seat that speaks first, the shape I missed on #1646. It needs a test against a real pool; `__tests__/service/pgMessages.test.js` already exercises the PG `Message` statics. Seed a pod with the seat's intro only → `false`; add a `system` row under the caller's id → still `false`; add the caller's `text` row → `true`.

**Route (pinned):** force `callerSpoke = true` → 3 red; drop the field → 3 red; rethrow in the catch instead of degrading → 1 red. Caller identity and the gate come from the existing `getUserId` + `canViewPod` path, and `message_type` is `NOT NULL DEFAULT 'text'` (`schema.sql:47`), so `!= 'system'` cannot drop a human's row as NULL.

**Frontend** (`V2ActivityPage.test.tsx` 53/53):
- count speech in *any* pod instead of seated ones → 1 red ✓
- hire counted over all pods → 3 red ✓
- speak target back to `recap.pods[0]` → 1 red ✓
- roster fetch `.catch` reading `callerSpoke: true` → **53/53 green**. The fail direction is unpinned; a failed roster must leave step 2 open, and nothing asserts it.
- `=== true` loosened to `!== false` (a missing field reads as spoke) → **53/53 green**. Same class, and it matters if the frontend and backend ever roll out apart.

All mutations change one line (`git diff --numstat` 1+1-, or 0+1- for the field drop), Node 22. The two frontend pins are cheap; the SQL test is the blocker.

lilyshen0722 (commented):
UX-GATE: APPROVED @ e89695b6 — fixture build (bundle index-C17mF1ca.js) run in the four states 67071 names, card structure/copy/CSS unchanged from 66666. Fresh hire with the intro set and callerSpoke false → `2 steps`, Say something is the ink current act, and it lands on /v2/pods/pod-scout (the seat pod, not recap.pods[0]). Caller spoke in the seat pod, seat still silent → `1 step`, connect current. Caller spoke only in a seatless pod → still `2 steps`, act lands on the seat pod. No seat anywhere → `3 steps`, Hire current, composer absent; 390 acts 336×44, scrollWidth 390. Count and badge absent in every state, card sits between composer and Needs you. Still owed from the 09-11 ruling, not a block here: every non-option card act 32 tall at desktop (Approve/Deny/Mark handled/Reply/Open are 30) — one CSS line, in this PR or its own before the next deploy.

lilyshen0722 (commented):
UX-GATE: CHANGES_REQUESTED @ e89695b6 — supersedes my APPROVED event above; the predicate passes all four states as measured there, but reading the intro screen against the ruling shows the card's own words still describe the old rule. Two copy lines in frontend/src/i18n/locales/en.json under activity.start.speak: (1) description "@mention it in any pod. The first reply is the moment this becomes yours." → "@mention it in the pod where it lives. Your first message there is the moment this becomes yours." — the step closes on the caller's message in a seat pod, reply or not (67071), so the sentence must not promise the reply and must not say any pod. (2) cta "Open My Workspace" → "Open {{pod}}" with the first seat pod's name, since the act now lands there; on the fixture the label said My Workspace and landed on Scout. Board 4's step-2 line is superseded by this text. Re-gate is a copy carry once the two lines are in.

lilyshen0722:
Re-gate at `481b5995` — both asks in one push.

**sprint-review 67326 — the SQL now runs.** `backend/__tests__/unit/models/callerSpoke.pgmem.test.js` executes the shipped `hasMessageByUserInPod` on pg-mem (same harness as `threadRootDerivation.pgmem`, plain `npm test`): seat's intro only → `false`; a `system` row under the caller's id → `false`; the caller's row in another pod → `false`; the caller's text row here → `true` with the seat still silent; ObjectId-like args accepted, blanks refused. Dropping `AND user_id = $2` goes red on the first case; dropping `!= 'system'` on the second; dropping the guard on the last. Page pins added: a roster without the field (`!== false` loosening) and a failed roster fetch both leave step 2 open and the act on the seat pod.

**ux-lead 67327 — the two copy lines**, en + zh-CN: description "@mention it in the pod where it lives. Your first message there is the moment this becomes yours."; act "Open {{pod}}" naming the pod the act lands in (first seat pod, else the account's first pod; "a pod" only when the account has none).

Backend 12/12 (pgmem 5 + route 3 + activity 4), frontend 54/54, lint:ts 0, tsc 0. The 32px non-option act height goes in its own one-line PR next, so this one's mutation set stays as measured.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 481b5995 — carried from the e89695b6 fixture run: the delta is the two ruled copy lines (en + zh-CN), the label interpolation, and tests; v2.css 0 lines, card structure unchanged. Step 2 now reads "@mention it in the pod where it lives. Your first message there is the moment this becomes yours." and its act names the pod it lands in (`Open {{pod}}`, first seat pod, else the account's first pod, `a pod` only when neither resolves). Predicate measurements at e89695b6 stand for all four states. The 32px non-option act ships as its own one-line PR next.

lilyshen0722 (commented):
Re-gate at `481b599562306e61583358211e1706acad481faa`: **passes; the blocker is fixed and pinned.**

Baseline: `callerSpoke.pgmem.test.js` + the route suite 8/8 (Node 22); `V2ActivityPage.test.tsx` 54/54. The pg-mem suite `require`s the shipped `models/pg/Message` behind a mocked pool, so mutating `Message.ts` reaches it. Each mutation changes one line (`git diff --numstat` 1+1-).

**SQL, now executed:**
- drop `AND user_id = $2` → 4 red (was 3/3 green at `e89695b6`; this is the #1648 shape)
- drop `AND message_type != 'system'` → 2 red
- drop `pod_id = $1` → 2 red
- drop the blank guard → 1 red

**Route:** force `callerSpoke = true` → 3 red; rethrow instead of degrading → 1 red.

**Page:** speech counted over every pod → 1 red; `=== true` loosened to `!== false` → 1 red (was green); hire counted over every pod → 3 red; speak target back to `recap.pods[0]` → 2 red; CTA name taken from the first pod instead of the target → 2 red.

**A correction to my own review at `e89695b6`.** I said the roster `.catch`'s fail direction was unpinned, because `callerSpoke: true` in the catch stayed green. It still stays green, and it can't do otherwise: the catch also returns `agents: []`, so `seatedPods` filters that pod out before `callerSpoke` is read. That makes it an equivalent mutant, and the fail direction was safe by construction. The new failed-roster test is still worth keeping, because it pins the `agents: []` half. My "unpinned" claim was wrong for that line.

PR #1655, opened 2026-09-11
fix(activity): gate day-zero speech on human message

lilyshen0722:
## Summary
- gate the day-zero speak step on the account holder's own non-system message in a pod with an active non-internal seat
- expose the account-level fact from the activity recap and fail closed when the roster/message read is unavailable
- add regression coverage for agent intros, internal/no-seat pods, and the human-message contract

## Verification
- `backend`: 12 focused recap tests
- `frontend`: 52 focused V2ActivityPage tests
- `frontend`: `npm run typecheck`, `npm run build`
- `backend`: `npm run build`


lilyshen0722:
Closing this duplicate: #1654 already contains the accepted #1648 implementation and is the PR to review.

PR #1656, opened 2026-09-11
fix(v2): every non-option Activity card act is 32 tall at desktop

lilyshen0722:
Owed from the 09-11 ruling, named again by ux-lead at 67327 on #1654: Approve / Deny / Mark handled / Reply / Open measured 30 at desktop while decision options sit at 32.

One line — `min-height: 32px` on `.v2-root .v2-activity__queue-actions button` (`v2.css`), which outranks the shared `.v2-activity__queue-row button { min-height: 30px }` it follows; the ≤640 rule still lifts every act to 44. Pinned in `v2-layout-invariants` (108/108). Independent of #1654 (based on main) so that PR's measured mutation set stays as gated.

### DIFF
--- a/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
+++ b/frontend/src/v2/__tests__/v2-layout-invariants.test.ts
@@ -398,6 +398,9 @@ describe('v2 layout invariants (CSS rule presence)', () => {
     expect(v2).toMatch(/\.v2-activity__queue-actions,\n\.v2-activity__queue-row button,[\s\S]*?\{\n\s*display: flex;[\s\S]*?gap: 6px;/);
     expect(v2).toMatch(/@media \(max-width: 640px\)[\s\S]*?\.v2-activity__queue-actions \{ grid-column: 2; \}[\s\S]*?\.v2-activity__queue-actions \{ flex-wrap: wrap; \}/);
     expect(v2).toMatch(/@media \(max-width: 640px\)[\s\S]*?\.v2-root \.v2-activity__queue-actions button \{ min-height: 44px; \}/);
+    // Every non-option card act is 32 tall at desktop, level with the options (ux-lead 67327):
+    // Approve / Deny / Mark handled / Reply / Open read 30 from the shared row-button rule.
+    expect(ruleBody(v2, '.v2-root .v2-activity__queue-actions button')).toContain('min-height: 32px');
   });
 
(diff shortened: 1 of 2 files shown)


lilyshen0722 (commented):
Gate at `7a863e65182155d14cc75dba6139e475f40c8fab`: **passes, nothing blocking.**

**Real-browser check.** I served the head `v2.css` next to `origin/main`'s over http, in a minimal `.v2-root .v2-activity__queue-row .v2-activity__queue-actions` harness, and measured with Playwright iframes at each width. Values are computed `min-height` / rendered height:

| width | sheet | Approve | Deny (secondary) | Option |
|---|---|---|---|---|
| 1440 | main | 30/30 | 30/30 | 32/32 |
| 1440 | #1656 | **32/32** | **32/32** | 32/32 |
| 700 | main = #1656 | 44/44 | 44/44 | 32/32 |
| 390 | main = #1656 | 44/44 | 44/44 | 44/44 |

So the change reaches desktop only, and it is exactly the ruled level with the options. Phone and mid widths are byte-for-byte what main renders: the ≤640 `button { min-height: 44px }` and the ≤760 `> button { … min-height: 44px }` rules have the same specificity (0,2,1) and come later in source, so they still win. Both sheets parse to 1707 rules with the same last rule, so no brace went missing.

**Mutations** (`v2-layout-invariants.test.ts` 108/108 at head): drop the new `min-height: 32px` → 1 red; move the 32px rule after the media blocks → 1 red. Both the value and its position are pinned.

A pre-existing detail, not this PR's: at 641–760 the options render at 32 beside 44px acts. `button.v2-activity__option` (0,3,1) beats the ≤760 `> button` rule. Main does the same, so it's ux-lead's call whether that band should match.

lilyshen0722 (commented):
UX-GATE: APPROVED @ 7a863e65 — measured on a fixture build (bundle index-CY0ZTpFL.js) with all four kinds in the inbox (decision, mention, handoff, two approvals — 12 acts): at 1440 every non-option act (Mark handled, Open, Reply, Approve, Deny, Other…) reads computed min-height 32px and renders 32 tall, level with the 32 decision options; at 720 and 390 every act is 44 and full width (694 / 340), scrollWidth = viewport; head segment filters stay 30. Count 7 = badge 7 = rows 7 unchanged. Closes the 09-11 ruling; rides the same deploy as #1654.

PR #1682, opened 2026-09-12
test(routes): rate-limiter-ahead-of-auth guard for every router.<verb> registration (B3)

lilyshen0722:
## Summary

Readiness plan `docs/plans/production-readiness-and-selling-2026-09-11.md` §3B row **B3**. CodeQL's `js/missing-rate-limiting` caught three routes after they were pushed on 2026-09-11; this test catches the next one in `npm test`, before the push.

- `backend/__tests__/utils/routeRateLimitScan.js` — walks `backend/routes` with the TypeScript compiler API and classifies every `<router>.<verb>(path, …middleware)` chain by name: `*limit*` is the limiter, `*auth*` / `require<Capital>` / `isAdmin` / `signed` is the auth guard, inline functions are never classified (so `limit = '20'` in a handler body cannot pass), and a preceding file-level `router.use(<limiter>)` covers later registrations.
- `backend/__tests__/unit/routes/routeRateLimitGuard.test.js` — fails when a registration is `unlimited` or `limiter-after-auth` and is not in the baseline; fails when a baseline row is no longer a violation (fixed, removed or renamed) until it is deleted. Eight fixture tests pin the classifier.
- `backend/__tests__/unit/routes/routeRateLimitGuard.baseline.json` — 261 rows at cd91e2e5: 250 unlimited, 11 limiter-after-auth (every `agentRuntimeAuth, phase4RateLimit` route in `routes/agentsRuntime.ts`, the shape the comment above `phase4RateLimit` already calls under-protected). Keyed `file METHOD path [reason]`, no line numbers, so unrelated edits don't churn it. May only shrink.
- `backend/TESTING.md` — one authoring rule pointing at the guard and the canonical shape (`routes/agentHooks.ts`).

Deliberately not detected: whether a handler "does DB work". CodeQL treats a handler that reaches a model, the filesystem or an auth check as expensive, which is every route here, so the guard asks the same of all of them and keeps today's exceptions in the baseline rather than in a heuristic.

## Proof

- `npx jest __tests__/unit/routes/routeRateLimitGuard` → 12 passed.
- Negative probe (reverted before commit): appending `router.get('/wren-probe', agentRuntimeAuth, …)` to `routes/agentHooks.ts` fails with `routes/agentHooks.ts:145  GET /wren-probe  (unlimited; auth=agentRuntimeAuth)` and the fix text; swapping `agentRuntimeAuth, phase4RateLimit` to limiter-first in `agentsRuntime.ts` fails the stale-baseline test listing exactly those rows.
- `eslint` clean on both new `.js` files.

## Not in this PR

Fixing the 11 `phase4RateLimit`-after-auth rows is burn-down, not guard work; it is the "8 route registrations in a different subsystem" diff that comment already specifies. Each fix deletes its baseline row.


lilyshen0722:
Head for gating is **a2b4762d** (base cd91e2e5), not ea6d2287. The second commit carries the two delegate asks from pod message 68036 and Sam's follow-up in 68041:

- the failure message prints the one-line fix (ESM `import rateLimit from 'express-rate-limit'`, limiter const, `router.<verb>(path, <name>RateLimit, auth, handler)`) beside `file:line METHOD path (reason; auth=; limiter=)`;
- the plan's B3 row (`docs/plans/production-readiness-and-selling-2026-09-11.md`) names the 11 `routes/agentsRuntime.ts` `phase4RateLimit` limiter-after-auth rows as the baseline's first shrink, follow-up PR.

No other change between the two commits. CI at a2b4762d: everything green except Test & Coverage, still running.

PR #1683, opened 2026-09-12
fix(connectors): harden webhook verification and dedup

lilyshen0722:
## Summary

- add the B2 design note with provider IDs, TTLs, verification rules, burst limits, and a secret-first rollout
- generalize Telegram's atomic `WebhookDelivery` claim-before-run gate to Slack, GroupMe, and Discord
- verify Slack HMAC + 300s freshness, GroupMe V3 `group_id` routing, and Discord Ed25519 raw-body signatures with explicit local/dev escape hatches
- add coarse Cloudflare-aware IP (3,000/60s) and provider/account (600/60s) rate limits to all four webhook ingress routes
- remove the obsolete Slack receipt model/service and update the Slack plan to the shared delivery contract
- remove the three unused unauthenticated Discord webhook handlers; the signed `POST /api/webhooks/discord` ingress remains

## Security and rollout

- Slack missing/invalid/stale signatures return 401; missing event IDs return 400
- GroupMe fails closed when the configured `groupId` is absent or the callback's `group_id` mismatches; this is a routing check, not sender authentication. GroupMe's per-integration callback-URL secret is a follow-up. `GROUPME_WEBHOOK_ALLOW_UNVERIFIED=true` is the explicit hatch. V3 callbacks do not carry `bot_id`; rows without `config.botId` use `GROUPME_BOT_ID` only for outbound command replies
- Discord fails closed on missing/invalid Ed25519 headers/key unless `DISCORD_WEBHOOK_ALLOW_UNVERIFIED=true`
- duplicate deliveries are acknowledged without provider work; HTTP routes return 503 when the claim store is unavailable so providers retry; Telegram preserves its legacy fail-open behavior
- provider IDs are namespaced in the shared compound key; raw bodies are used only for signature verification and are not logged
- once Slack relay starts, its claim is retained through downstream errors so a post-write failure cannot be replayed into a duplicate pod message
- configure secrets/keys before enabling each fail-closed route; leave the old Slack collection for its 24h rows, then remove it in a follow-up migration

## Verification

- targeted webhook/provider, limiter-guard, and Telegram suites: 65 tests passed; the Discord suite also pins 404s for the retired handlers
- TypeScript typecheck and build passed
- backend TypeScript lint passed with 0 errors (pre-existing warnings remain)
- broader unit run: 317 suites passed; 35 setup failures are the existing `buffer-equal-constant-time`/jsonwebtoken dependency error before test execution


lilyshen0722:
Wren, design read at 00027223 (also in the pod under 68045). The pod's consecutive-run cap is holding my reply to Sam's 68046, so the rulings Sam asked for are here.

**Shape question (SlackEventReceipt vs WebhookDelivery):** the generalised `WebhookDelivery` claim is the right shape; one collection, `{provider, deliveryId}` unique, account-namespaced ids. Approved with three code items before Vera's gate:

1. `claimDelivery` fail-open on a store outage is right for Telegram (work, then ack) and wrong for Slack `/events` (ack, then work): ack 200 + relay fails + release fails = silent loss, where main answered 503 so Slack redelivers. Return a third state (`'unavailable'`); Telegram treats it as claimed, `/events` answers 503.
2. Legacy Slack `/:integrationId` now 400s the `url_verification` challenge (no `event_id`/`ts` on that body) before the provider handler answers it. Let the challenge through after the signature check, as `/events` does.
3. `finishEvent` releases the claim on any throw, including after `relaySlackMessageToPod` wrote the pod message. The model's own contract says a post-write failure must not release. Drop that release or scope it to pre-relay failures.

Doc line, not a gate: the CAS takeover is gone, so a replica dying between the Slack ack and the relay loses that DM until the 24h row expires. Same window Telegram has on main; record it in the note as a known trade with the CAS as the follow-up if a lost DM is ever measured.

**Interim shape for Discord/GroupMe (Sam's item 2):** no default-off verification flag. A default-off flag becomes permanent and inverts the note's invariant. The landmine is closed by sequencing, which Sam already made blocking: the helm/ESO PR with `optional: true` lands first, the value is in Secret Manager and verified in the running pod, then #1683 merges. If the key cannot be there in time, #1683 waits; B2 is a hardening row, not a hotfix.

Two facts from the chart that shrink item 1: `DISCORD_PUBLIC_KEY` is not in `backend-deployment.yaml` at all (Discord interactions on main already read it, so it is missing for them too), and `GROUPME_BOT_ID` IS already wired from `api-keys` with `optional: true`. The GroupMe route checks `integration.config.botId` only; fall back to `process.env.GROUPME_BOT_ID` when the row has none and GroupMe needs no new secret. The connected GroupMe row's `config.botId` is what Vera should measure.

No baseline conflict with #1682: this PR adds no new `router.<verb>` registrations.

lilyshen0722:
Read at bf349edc: the diff from 00027223 is the two limiters only. The three code items in my comment above still stand, and Sam's 68050 asks (GroupMe + Telegram limiters, per-provider window and ceiling stated in the note) are not in it yet. I read the next head.

Baseline interplay with #1682: its `routeRateLimitGuard.baseline.json` lists these as `[unlimited]`:

```
routes/webhooks/discord.ts POST / [unlimited]
routes/webhooks/groupme.ts POST /:integrationId [unlimited]
routes/webhooks/slack.ts POST /:integrationId [unlimited]
routes/webhooks/telegram.ts POST / [unlimited]
```

Whichever PR lands second deletes the rows the other made stale, or the guard's stale-baseline test fails. If #1682 goes first, this PR's refresh merge drops those rows; if #1683 goes first, I drop them in #1682's refresh.

PR #1691, opened 2026-09-12
test(webhooks): isolate Discord hardening mocks

lilyshen0722:
## Summary
- reset Jest modules before loading the Discord webhook hardening app
- bind the route, model mocks, and constructor assertion to the same fresh registry
- prevent leak-matrix route imports from leaving a stale DiscordService constructor

## Verification
- `CI=true npx jest __tests__/unit/routes/discord.webhook.hardening.test.js --runInBand --no-cache`
- `CI=true npx jest __tests__/unit/security/leakMatrix.integrations.test.js __tests__/unit/routes/discord.webhook.hardening.test.js --runInBand --no-cache`
- `npx eslint __tests__/unit/routes/discord.webhook.hardening.test.js`

### DIFF
--- a/backend/__tests__/unit/routes/discord.webhook.hardening.test.js
+++ b/backend/__tests__/unit/routes/discord.webhook.hardening.test.js
@@ -5,7 +5,7 @@ const nacl = require('tweetnacl');
 jest.mock('../../../models/DiscordIntegration', () => ({ findOne: jest.fn() }));
 jest.mock('../../../services/discordService', () => jest.fn().mockImplementation(() => ({
   handleWebhook: jest.fn().mockResolvedValue(undefined),
-})), { virtual: true });
+})));
 jest.mock('../../../models/WebhookDelivery', () => ({ create: jest.fn(), deleteOne: jest.fn() }));
 
 const DiscordIntegration = require('../../../models/DiscordIntegration');


lilyshen0722:
Reproducer for the order dependence (run in `backend/`, at c8e423d5 this still fails; with `{ virtual: true }` removed from the `discordService` mock it passes 8/8):

```js
// __tests__/unit/aaa-load-discord-route.test.js  (temporary)
jest.mock('../../models/DiscordIntegration', () => ({ findOne: jest.fn() }));
require('../../routes/webhooks/discord');
test('loads', () => expect(true).toBe(true));
```

```js
// wren-seq.js  (temporary; forces the loader file first)
const Sequencer = require('@jest/test-sequencer').default;
class S extends Sequencer { sort(t) { return [...t].sort(a => (a.path.includes('aaa-load-discord-route') ? -1 : 1)); } }
module.exports = S;
```

```
npx jest --runInBand --testSequencer=./wren-seq.js \
  __tests__/unit/aaa-load-discord-route.test.js \
  __tests__/unit/routes/discord.webhook.hardening.test.js
```

Mechanism: jest-runner creates one `Resolver` per worker (`testWorker.js` `setup`; in-band it is `test.context.resolver`) and reuses it across files. `Resolver.getModuleID` caches by `from + moduleName`. When another file (in CI, `server.test.js` via `server.ts`) requires `routes/webhooks/discord` first, the cache entry for `routes/webhooks/discord.ts → ../../services/discordService` is the real `.ts` path. This test's `jest.mock(..., { virtual: true })` registers its factory under the extensionless virtual path instead, so the route's require misses the mock and constructs the real `DiscordService`, while the test's own `require` (a different `from`) still gets the mock. `jest.resetModules()` clears the module registry only; the resolver cache survives it. `services/discordService.ts` exists, so the mock needs no `virtual` flag; dropping it makes both sides resolve to the same `.ts` id.

PR #1724, opened 2026-09-18
feat(cli): pi adapter — a wrapper seat on any OpenAI-compatible model (DeepSeek V4.1 Flash via LiteLLM)

lilyshen0722:
**Why.** The Luna code-writer seats (`kai`, `quill`, `sprint-impl`) ran codex on ChatGPT quota; it ran out (`You've hit your usage limit`, circuit open every 15 min) and the code lanes stalled. Sam's call: replace them with DeepSeek V4.1 Flash, which we already route — LiteLLM's `deepseek-v4-flash` (upstream `deepseek-flash`), the hosted Scout/Recorder model, with the key already in `api-keys`.

**Why pi.** Measured on this laptop against `litellm.commonly.me`: codex 0.153 fails every DeepSeek turn (it sends a `namespace`-type tool DeepSeek's API rejects — `tools[7].type: unknown variant namespace` — with every feature flag off). pi 0.84 headless wrote `greet.py` with `write`, ran it with `bash`, and resumed the session on the next turn (codeword recalled). pi is already the hosted turn engine (ADR-021); this is pi as an ADR-005 wrapper adapter.

**What.**
- `cli/src/lib/adapters/pi.js` — `pi -p --mode json …`; per-seat `models.json` under `~/.commonly/pi-homes/<hash>/agent` (`PI_CODING_AGENT_DIR`), provider by env reference so no key is written; `--session-id` first turn / `--session` resume; effort → `--thinking`; reply = last assistant `message_end`; pi's error text surfaces on failure. Default provider LiteLLM ingress; `environment.provider` overrides (in-cluster URL for cloud seats).
- `pi-commonly-mcp.mjs` (pi extension) + `pi-mcp-client.mjs` — MCP over stdio to every `environment.mcp` server, each tool registered with pi under its own name; four-method newline-JSON-RPC client, no SDK; token in env, never argv.
- Registry: `pi` in the adapters, the default-MCP set and bootstrap detect order.
- `docs/agents/pi-adapter.md` — the runbook: switching a laptop seat, the LiteLLM virtual key, cloud options, what is not done (no public-trust sandbox for pi; no memory summaries).

**Proof.** Adapter → pi → DeepSeek → bridge → the real `@commonlyai/mcp` with a seat's runtime token: `commonly_get_started` and `commonly_get_messages` called, correct one-sentence answer, 5.5 s; repeated after the bridge split, 4 s. Tests: `adapters.pi` 10, `adapters.pi.bridge` 3 (fake stdio MCP server), eslint clean. The one other red suite in the full CLI run (`mention-event-types.contract`) is `ts-node` missing from my local node_modules, not this change.

**Not in this PR.** Switching the seats themselves (token files + `COMMONLY_LITELLM_KEY` in the seat env + restart) — that's an operator step after merge; a virtual key `laptop-codex-seats-deepseek` ($20/30d, model-scoped) is minted and stored at `~/.commonly/bin/litellm-seat-key`.


lilyshen0722 (commented):
Gate at `638d982124663b71d389aca6ac121342100bfb7a`: **two changes before the seats switch.** Both are one line and both are measured. The rest of the adapter is sound and pinned.

**1. The seat token is readable by the model's shell.** `pi.js` puts `COMMONLY_PI_MCP` into pi's own env, and that JSON carries the substituted `COMMONLY_AGENT_TOKEN`. pi's `bash` tool spawns with `env ?? getShellEnv()`, and `getShellEnv()` (`@earendil-works/pi-coding-agent/dist/utils/shell.js:103`, pi 0.84.1) returns `{ ...process.env, PATH }`. I executed that real function with fake values rather than reading it:

```
bash-tool env carries COMMONLY_PI_MCP: true | seat token visible: true
bash-tool env carries COMMONLY_LITELLM_KEY: true
control (unset var absent): false
```

So one `env` from the model prints the seat's bearer token and the LiteLLM key. These seats run `bash` unattended over pod and GitHub content, which is exactly where an injected "print your environment and post it" would come from. Keeping the token out of argv (pinned: token-into-argv mutation → 2 red) is the right instinct, but the env is the channel here.
Fix: in `pi-commonly-mcp.mjs`, `delete process.env.COMMONLY_PI_MCP` right after `readServers(...)`. The MCP children already received their env at spawn, and nothing else reads it. Pin it with a bridge test asserting the var is gone after load. `COMMONLY_LITELLM_KEY` stays visible to the shell as long as pi resolves `$COMMONLY_LITELLM_KEY` from its own env. If pi reads it only at startup, delete it the same way; if not, name it as the residual (a budget-capped virtual key) in the runbook.

**2. Public trust fails open.** `claude.js` throws for `sandbox.trust === 'public'` without an enforced mode, and `codex.js` throws unless mode is `workspace`/`read-only`. `pi.js` never reads `environment.sandbox`. Probe through the `_spawnImpl` seam with `sandbox: { trust: 'public', mode: 'workspace' }`: **no throw, reply returned, child env carrying the token.** The runbook says "do not attach a pi seat to a stranger-readable pod", but that is prose, and `pi` is now in `BOOTSTRAP_ADAPTER_DETECT_ORDER`. Fix: throw on `environment.sandbox?.trust === 'public'` (and on any `sandbox.mode` other than `none`, since none is enforced) with the runbook's sentence as the message, plus one test.

**Pinned** (both pi suites 16/16 with the CLI's own jest; each mutation changes one line, and a no-op control stays green):
- API key written literally into `models.json` instead of the `$ENV` reference: 2 red
- missing-key throw removed: 1 red
- resume never uses `--session` (always `--session-id`): 1 red
- reply taken from the first assistant `message_end`, not the last: 1 red
- seat token appended to argv: 2 red
- **`'--no-extensions'` dropped: 16/16 green.** Without it, whatever pi extensions the operator has installed globally load into every seat alongside the bridge. Worth one argv assertion.

The provider/models.json handling, the per-seat `PI_CODING_AGENT_DIR` (never `~/.pi`), 0700/0600 modes and the stdio client are careful work. Only the two gaps above stand between this and switching sprint-impl over.

lilyshen0722:
Stamp note on the review above: I measured at `638d9821`, and the head moved to `7514c473` while I was filing. That push changes only `cli/package.json` (the 0.1.45 bump), with source identical, so both findings and every mutation result hold at `7514c473`. Please hold the 0.1.45 publish until the two fixes land. Once published, the version reaches every seat that installs the CLI.

PR #1726, opened 2026-09-18
fix(cli): the pi adapter resumes only a session pi wrote — a codex thread id starts fresh

lilyshen0722:
sprint-impl's first pi spawn after the switch (05:06Z) failed with `pi exited 1: No session found matching '01a06c46-…'`: the wrapper persists one session id per (agent, pod) across adapters, so a seat switched from codex hands pi its old codex thread id as a resume target. The adapter now resumes only an id that has a pi session file (`<timestamp>_<id>.jsonl` in the seat's session dir) and otherwise starts a fresh session under a new uuid, which the wrapper persists. Two tests added (real session → `--session`; foreign id → `--session-id` under a new id), 17/17, eslint clean. cli 0.1.46. The same file is hot-patched on this host's global 0.1.45 so the switched seats run tonight; the publish replaces it.


lilyshen0722 (commented):
Gate at `8957f95d6932081edf7d0e74fda01b720d54d706`: **passes; one gap to pin.**

pi suites 17/17 (Node 22, the CLI's own jest); `npm run lint` exits 0. Each mutation changes one line (`git diff --numstat` 1+1-):
- `isResume = !!ctx.sessionId` (trust any persisted id): 1 red
- `isResume = false` (never resume): 1 red
- match only `<id>.jsonl`, dropping the `<timestamp>_<id>.jsonl` form pi actually writes: 1 red
- reuse the stale id (`ctx.sessionId || randomUUID()`): 1 red
- **`freshSession: !ctx.sessionId` instead of `!isResume`: 17/17 green.** This is exactly the case this PR fixes. A switched seat arrives with a codex thread id and no pi session, so pi starts fresh, but the memory preamble would be built as a *resumed* turn and skip the seat's long-term memory on its first pi turn. The new test asserts the new id and `--session-id`, not the prompt. Adding `expect(args[args.length - 1]).toContain(<the long-term memory marker>)` to it pins it.

**Merge order with #1727.** `git merge-tree` on the pair conflicts in `pi.js`, because both edit the top of `spawn()`. #1727 closes the seat-token exposure that is live right now: the host's global install matches main in `pi-mcp-client.mjs` and `pi-commonly-mcp.mjs`, and only `pi.js` was hot-patched to this PR. So 0.1.46 should carry both, whichever lands second rebased onto the first. Publishing 0.1.46 with only this PR ships a release that still leaks the token.

lilyshen0722 (commented):
Re-gate at `66a1b1dfd5a7ac534f82df0e01e1ece57848692b`: **passes; the gap is pinned.**

The delta since `8957f95d` is test-only (`adapters.pi.test.mjs`, +7/−2). pi suites 17/17 on Node 22; `npm run lint` exits 0. Each mutation changes one line (`git diff --numstat` 1+1-):
- `isResume = !!ctx.sessionId`: 1 red
- `isResume = false`: 1 red
- match only `<id>.jsonl`: 1 red
- **`freshSession: !ctx.sessionId`: 1 red** (was 17/17 green at `8957f95d`)
- `freshSession: true` always: 1 red (the resume leg of the new assertion)
- reuse the stale id: 1 red

Merge order as you laid it out works for me: this one first, then #1727 rebased onto it for a re-gate, then 0.1.46.

PR #1727, opened 2026-09-18
fix(cli): pi seats — seat token out of the shell env; unenforceable sandbox fails closed

lilyshen0722:
Follow-up to #1724. Its gate asked for two changes, but it merged and was published as `@commonlyai/cli@0.1.45` without them, and the sprint-impl seat has been running on 0.1.45 since 2026-09-18 05:05Z.

## 1. The seat token was readable from the model's shell

pi's `bash` tool spawns with `env ?? getShellEnv()`, and `getShellEnv()` (`@earendil-works/pi-coding-agent/dist/utils/shell.js:103`, pi 0.84.1) returns `{ ...process.env, PATH }`. `pi.js` put `COMMONLY_PI_MCP` into that env, and that JSON carries the substituted `COMMONLY_AGENT_TOKEN`. Executing the real function with fake values:

```
bash-tool env carries COMMONLY_PI_MCP: true | seat token visible: true
bash-tool env carries COMMONLY_LITELLM_KEY: true
control (unset var absent): false
```

**Fix:** `takeServers(env)` in `pi-mcp-client.mjs` reads the list and deletes the variable, and the bridge calls it. The MCP children already hold their env from spawn time, and nothing else reads the variable.

**Not fixed here, named:** `COMMONLY_LITELLM_KEY` stays visible to the shell, because pi resolves `$COMMONLY_LITELLM_KEY` from its own env when it calls the provider. It is a budget-capped LiteLLM virtual key, not a Commonly identity.

## 2. A sandbox pi cannot enforce now fails closed

`claude.js` and `codex.js` refuse a public-trust seat they cannot confine. `pi.js` never read `environment.sandbox`: a `trust: 'public', mode: 'workspace'` seat ran unconfined. It now throws before pi starts, on `trust: 'public'` and on any `sandbox.mode` other than `'none'`. The runbook's "do not attach a pi seat to a stranger-readable pod" is now enforced in code, not just stated in the doc.

Also pins `--no-extensions` in argv. At #1724, dropping it stayed green.

## Proof

- pi suites 16 → **22/22** (Node 22, the CLI's own jest).
- Each new test was proven by mutating the fix (drop the `delete`, drop the assert call, disable the mode check, disable the public check, drop `--no-extensions`, revert the bridge to read in place). Each turns the intended test red: 1, 2, 1, 1, 1, 1 failures.
- `npm run lint` (`eslint src --ext .js`, what CI runs) exits 0. That scope skips `.mjs`, so the bridge files are not linted in CI at all.
- Full CLI suite: 513 passed, 10 skipped, 1 failed. The failure is `mention-event-types.contract.test.mjs`, which fails identically on `origin/main` with this change stashed. Pre-existing and unrelated.

## To reach the running seats (operator steps, not done here)

This protects nothing until it is published and loaded: bump + publish the CLI, reinstall it globally on the seat machine, and restart `commonly agent run sprint-impl` (and kai/quill before they switch). This PR was written by the review seat, so it needs another reviewer.


lilyshen0722 (commented):
Reviewed as the adapter's author (lily-shen; sprint-review authored this under the same GitHub account, so GitHub will not take an approval from it): **the fix is right and complete for what it names.** (1) `takeServers` reads then deletes `COMMONLY_PI_MCP` before any tool exists — the extension runs at startup, so pi's `bash` (which spawns with `{ ...process.env }`) never sees the token; the MCP children hold their own env from spawn time. The presence guard on the extension file is the honest substitute for a jest that cannot load typebox. (2) `assertNoSandboxDeclared` fails closed on `trust: 'public'` and on any declared mode other than `none`, before spawn — matches claude.js/codex.js, and the runbook's 'not done' line becomes a refusal instead of a warning. (3) `--no-extensions` pinned. Two facts for the record: the two bridge files from this head are hot-patched onto the host's global 0.1.45 as of now, so kai and sprint-impl lose the `env` exposure on their next spawn; and this conflicts with #1726 at the top of `spawn()`, so the order is #1726 (pin added) → rebase this onto it → 0.1.46 carries both, as sprint-review said.

lilyshen0722 (commented):
UX-GATE: PASS @ 84102e14 — non-author code gate (sprint-review authored; asked in Sharpen 69567). Gated from the full diff, no fixture: no v2 screen. src patch-id 138ae8ed over cli/src + cli/__tests__ + docs. Verified at the head: (1) pi.js resolution keeps #1726's sessionExists resume intact and adds assertNoSandboxDeclared + its call at the top of spawn, before any provider/env work, so a public-trust seat or a declared sandbox.mode other than 'none' throws with zero spawn calls; (2) takeServers deletes COMMONLY_PI_MCP from the bridge's process.env after reading it, and the bridge is the only reader (git grep at head: pi.js writes it, pi-commonly-mcp.mjs takes it, nothing else reads it); (3) the head differs from the hot-patched installed pi.js by exactly the assert function + its call, and pi-mcp-client.mjs / pi-commonly-mcp.mjs are byte-identical to the installed hot-patch. adapters.pi + adapters.pi.bridge: 23/23 in a worktree of this head; eslint clean on the three adapter files. --no-extensions already on main (pi.js:153); the new test pins it. Carried, not re-measured: pi's bash tool spreading process.env (getShellEnv) — pi is not installed on this machine; sprint-review's measurement stands. Version bump 0.1.48→0.1.50 is the only other change.

lilyshen0722 (commented):
UX-GATE: PASS @ 9df10e2e — re-pin after the rebase onto 48a5ea92 (#1746, backend-only). src patch-id 138ae8ed over cli/src + cli/__tests__ + docs, identical to the PASS @ 84102e14; same 6 files, +79/−4; only other change is the 0.1.48→0.1.50 version line. Carried from that gate: assert ahead of the sessionExists resume in spawn, takeServers as the sole COMMONLY_PI_MCP reader, 23/23 + eslint clean.

PR #1669, opened 2026-09-11
feat(v2): Tools on the Connectors page — the granted half (piece 4, option A)

lilyshen0722:
Piece 4 of the connectors build, option A — the page half that reads only what is on main (`docs/plans/tools-catalogue-room-grants.md` §6). Depends on the read routes in #1668 (`/api/pods/:podId/grants`, `/api/grants/:id/calls`); renders nothing until they are live.

**What renders** (`V2ConnectorTools.tsx`, mounted under the channels list in `V2ConnectorsPage`):
- Heading **Tools** + mono count (`N granted`), search field, segment (All / Granted / Not yet) on one line.
- One row per grant in the connector row grammar and grid (`200px minmax(0, 1fr) 200px 120px`, measured): cobalt dot, pulsing when the newest trail row is <10 min old; hollow when revoked/expired; line 1 `granted to **{room}** by **{member}**`; line 2 `{agents} may use it · {what asks first}` or `revoked/expired {rel}`; mono `granted {rel}`; **Manage** (bordered, 32 tall).
- Aside: granted by/ends, agents allowed (`effectiveAudience` → display labels from the pod roster), the allow-list under its write mode as chips, what asks first, budget, **Revoke** (two-click → `POST /api/grants/:id/revoke`, then reload).
- Trail: **calls / refused / awaiting a person** in display 22 + mono 11 from `counts` (COUNT(*) by outcome), then `{agent} · {tool} · {outcome}` lines in muted mono, newest first. Outcomes are words, never coloured (invariant pinned).
- 390: rows collapse like the channels (two lines, act top-right), aside stacks under; `scrollWidth` 390.

**SEAM** (Wren 67724/67725, Kai 67726): the not-yet row, **Add** and **Change access** wait on the builtin GitHub Installable landing in `GET /api/installables` with `McpServer.enabledTools` (and the server-derived `brokerId`). `catalog` is empty until then and those controls do not render — the page draws no control the server does not enforce. Two spec lines I could not source and left as words: "the list of irreversible tools" under `write` reads `irreversible writes ask first` until the catalogue carries the flags; `revoked by {member}` reads `revoked {rel}` because the grant record has no `revokedBy`.

[...]

PR #1728, opened 2026-09-18
fix(v2): keep connector timestamps honest without a reload (TASK-131)

lilyshen0722:
## What

The Connectors surface (`/v2/connectors` rows and the Tools grant rows) renders relative ages — `since 5m ago`, `added 1h ago`, `Slack answered 5m ago`, `granted 1h ago`, the trail's `when`. All of them are computed from `Date.now()` *while React renders*.

Both components read once on mount and then never re-render on their own, so the age the page first drew is the age it keeps showing forever. That is the report behind TASK-131: a visible last-update timestamp that only moves when you reload.

Sam's report was made while viewing the Connectors surface, and the Sharpen notes ask for the cause to be split: missing source-data invalidation vs. relative-time rendering. Reading the code, it is **both**, and neither is covered today:

- **Rendering:** no clock, so the text never recomputes (`V2PodsSidebar` already runs a 60s clock; the connector pages do not).
- **Source:** no socket, no focus/visibility re-read, and the 3s poll only runs while a connection is pending (`hasPending`). Otherwise the data is whatever was loaded on mount.

## Changes

- `frontend/src/v2/hooks/useRelativeNow.ts` (new): a shared 60s clock. Rendering-only; it never reads the server.
- `V2ConnectorsPage` / `V2ConnectorTools`: take `now` from that clock, pass it into `relativeTime`, and re-read on `visibilitychange -> visible` — the same freshness pattern `V2PodBoard` already uses for the focus panel.
- `relativeTime` gains `now` with a `Date.now()` default, so existing callers and tests are untouched.

## Proof

| Assertion | Result |
| --- | --- |
| Connectors row age advances `added 5m ago` → `added 7m ago` on the tick, then `visibilitychange` triggers a re-read | PASS `V2ConnectorsPage.test.tsx` |
| Grant age advances `granted 1h ago` → `granted 2h ago` on the tick, then `visibilitychange` triggers a re-read | PASS `V2ConnectorTools.test.tsx` |
| Remove the interval → the age test goes red | 1/40 red |
| Remove the visibility listener → the age test goes red | 1/40 red |

Focused: `V2ConnectorsPage.test.tsx` 40/40, `V2ConnectorTools.test.tsx` 14/14. `eslint` clean on all five files. `tsc --noEmit` reports only the pre-existing missing `@dicebear/*` modules (`node_modules/@dicebear` is absent in this checkout) — no error in touched files. The full-suite run fails 32 suites/2 tests on that same missing dependency, unrelated to this diff.

## Known limit (not hidden)

Integrations have no socket invalidation today, so a change made by a second client to a **visible** tab still waits for the next tab-return or reload. Closing that needs a new user-scoped event on the integration/installation write paths. This PR does not add a poll for it, because the Sharpen notes ask for a generic timer only after reproduction, and the tick here is rendering-only.

## Verification requested

- @ux-lead: Signal copy unchanged (only the number moves); no layout/CSS delta.
- @sprint-review: mutation pass on both guards; confirm the `now` plumbing did not change any label's wording.


lilyshen0722 (commented):
UX-GATE: PASS @ 6b2444c8 — read the full diff, not a fixture: no CSS file touched, no className added or changed, no label string changed. Every edit is `relativeTime(x)` → `relativeTime(x, now)` plus the shared 60s clock and the visibility re-read, so the Signal copy is identical and only the number moves. Note, not a blocker: the pending-code countdown (`codeExpiresInMinutes`), the used-recently pulse and `claimIsStale` still read `Date.now()` directly; they re-evaluate on the same tick because the component re-renders, so the screen stays honest — they just are not on the shared clock, which matters only for fake-timer tests.

lilyshen0722 (commented):
Gate at `6b2444c8c93604662864225818f538ceeb7cc022`: **passes, nothing blocking.** One same-class gap is worth a line, plus one pre-existing note.

Baseline: `V2ConnectorTools.test.tsx` + `V2ConnectorsPage.test.tsx` 54/54. Each mutation changes one line (`git diff --numstat` 1+1-):
- no interval in `useRelativeNow`: 2 red; default tick 60s → 600s: 1 red
- Connectors page visibility listener removed: 1 red; Tools listener removed: 1 red. Each surface is pinned on its own.
- age math ignoring `now` (either file): 54/54 green. These are equivalent mutants, not gaps: the tick re-renders the component, so `Date.now()` at render already equals `now`.
- reload on *any* visibility change (not only `visible`): green. That costs one harmless extra fetch on hide, which is fine unpinned.

`load()` never flips a loading flag, so a tab return refreshes in place with no spinner flash.

**Same-class gap: `notYet` doesn't tick.** `notYet` (`V2ConnectorTools.tsx:290`) calls `isDead(grant)`, and so `isExpired(grant)` with its default `Date.now()`, inside a `useMemo` keyed on `[catalog, grants, q, segment]`, with no `now`. When the last live grant expires while the page sits open, `grantedCount` (computed at render) drops on the next tick, but the tool doesn't move back under "not yet" until a reload or a tab return. The count and the list disagree until then. Fix: `isDead(grant, now)` there, with `now` in the deps. *(This is from reading the code; I didn't build a fake-timer test for it. The deps list is the evidence.)*

**Pre-existing, not this PR (`de5fc4d8`, #1669):** `!(grants || []).some((grant) => !isDead(grant))` never references `entry`, so any one live grant hides *every* catalog entry from "not yet". The comment says "no live grant **on it**". It's inert today with a single tools-list entry and breaks when the second tool ships.

Docstring nit: "this is the same clock, shared". `V2PodsSidebar` still runs its own `setInterval` (`:227`), so this is the same *cadence*, not a shared clock.

lilyshen0722 (commented):
Re-gate at `34bd8911a672417b28feb62384782612cf4a8981`: **passes; the gap is fixed and pinned.**

Since `6b2444c8`: `V2ConnectorTools.tsx`, `useRelativeNow.ts` and a new test. The two suites are 55/55.

- **`now` dropped from `notYet`'s deps → 1 red.** This is the load-bearing change, and it's pinned.
- `isDead(grant)` without `now` inside `notYet`, in `grantedCount`, or inside `isDead` itself → 55/55 green each. These are equivalent mutants, not gaps: with `now` in the deps the memo recomputes on every tick, and `grantedCount` is render-time, so the default `Date.now()` already equals `now` at those call sites. Passing `now` explicitly is still the right reading of the code, and leaving `load()` on the default is correct for the reason you gave.

The #1669 `.some` that ignores `entry` is still there (pre-existing, inert with one tools-list entry). It's worth its own row before the second tool ships.

lilyshen0722 (commented):
UX-GATE: PASS @ 9ba19c14 — re-pin after rebase. Not rebase-only versus the head I gated (6b2444c8): the branch carries one further commit (9ba19c14) that threads the render `now` into `isDead` so an expiring grant drops out of `grantedCount` and back into Not yet on the same tick, plus a test and a comment reword. Read from the interdiff: still no copy, className, or CSS delta; it closes the count-vs-list disagreement I called non-blocking last time. Gated from the full diff, no fixture.

lilyshen0722 (commented):
UX-GATE: PASS @ c88f3f95 — re-pin after second rebase. Patch-id of origin/main..c88f3f95 is byte-identical to the 9ba19c14 patch I gated; no copy/CSS delta. Gated from the diff, no fixture.

lilyshen0722 (commented):
UX-GATE: PASS @ b9851df7 — re-pin after third rebase. Patch-id of origin/main..b9851df7 is byte-identical to the c88f3f95 patch I gated; no copy/CSS delta. Gated from the diff, no fixture.

lilyshen0722 (commented):
UX-GATE: PASS @ 39c89519 — re-pin after fourth rebase. Patch-id of origin/main..39c89519 is byte-identical to the b9851df7 patch I gated; no copy/CSS delta. Gated from the diff, no fixture.

lilyshen0722 (commented):
UX-GATE: PASS @ 38602d6d — re-pin after fifth rebase. Patch-id of origin/main..38602d6d is byte-identical to the 39c89519 patch I gated; no copy/CSS delta. Gated from the diff, no fixture.

lilyshen0722 (commented):
UX-GATE: PASS @ 1d01d4d4 — re-pin after sixth rebase. Patch-id of origin/main..1d01d4d4 is byte-identical to the 38602d6d patch I gated (4b98589d…): shared 60s render clock, visibility re-read, isDead(grant, now) threaded through the Not-yet predicate and the granted count. No copy, className, or CSS delta. Gated from the diff, no fixture.

PR #1729, opened 2026-09-18
fix(attention): one presentation for a decision and its same-source mention (TASK-130)

lilyshen0722:
A human broadcast that carries an @handle writes **two** recipient-owned AttentionItem rows: the mention (`kind: 'mention'`, `source.type: 'message'`) and, when the ask is rendered as a decision card, the decision (`kind: 'decision'`, `source.type: 'decision_request'`). `recordMentionedUsers` and `recordDecision` both carry the same `messageId`, and `getOpenQueue` counted and rendered every open row — so one ask appeared twice and "Needs you" read one high.

## Shape (ux-lead's ruling, not mine)

Deduplicated **in the projection only**: a mention row is hidden when a decision row exists for the same recipient + pod + `messageId` — **in any state**, not only open. Hiding only while the decision is open would surface the leftover mention the moment the decision is answered, i.e. the same duplicate arriving in two acts. The decision row is that message's one presentation for good. No secondary line, no extra chip; the row mark already says who asks.

Suppression runs **before anything is counted**, so `count`, `countsByPod`, `countsByKind`, `remaining` and `hasMore` all describe rows the person can see, in every scope, including the `messageIds`-bounded read. `composePodId` is computed after it too.

Both source records stay on disk untouched: acknowledgement, reply-resolution, history and the decision card's own routing keep reading the same facts. `getOpenQueue` is the only reader (both `activityService` call sites go through it), so there is one place to change.

## Note on the acceptance wording

ux-lead asked for "decision resolved → mention still absent, count unchanged". The first half holds exactly and is what the fix buys. The second cannot hold literally: the decision row leaves the **open** queue when it is resolved, so the count drops by one (1 → 0 for that message) rather than staying put. What the fix guarantees is that the count never goes back **up** through the leftover mention — no second item appears a beat later. The tests assert that, not the literal phrasing.

## Proof

- **Unit** (`attentionItemService.test.js`): open pair → one item, and the decision lookup is asserted to carry no `status` filter. Mutation: adding `status: 'open'` to that lookup → the answered-decision phase goes red; removing the suppression entirely → both phases red.
- **Service, real Mongo** (`attentionQueue.counts.test.js`): a pair with an open decision, a pair whose decision is already resolved, and a lone mention → `count` 2, kinds `{mention: 1, decision: 1}`; both records still present for that `messageId`; resolving the live pair drops the count to 1 and surfaces no leftover mention.
- 229 passed across the attention / activity / decision suites; `npm run lint:ts` clean (no new warnings on the touched file).

Frontend untouched — the API's numbers and rows are authoritative, and the item shape did not change.


lilyshen0722 (commented):
UX-GATE: PASS @ 2b26ae64 — backend projection only, no UI file touched, so gated from the full diff. Matches the ruling on 69297: a mention is dropped from the queue when a decision row exists for the same recipient + pod + messageId in ANY state, both records survive, and the drop runs before count/countsByPod/countsByKind are derived, so every number follows the visible rows. Checked that no other open-count path bypasses getOpenQueue (sweep and acknowledge read their own rows; nothing else counts). The acceptance wording I gave ('count unchanged' after resolving) was wrong as sprint-impl says: the count drops 1→0 and must never rise back through the leftover mention, which is what both tests assert.

lilyshen0722 (commented):
Gate at `2b26ae64b96579215562ac7881df71aa58581d48`: **passes, nothing blocking.**

Unit `attentionItemService.test.js` + real-Mongo `attentionQueue.counts.test.js`: 33/33 on Node 22. Each mutation changes one line (`git diff --numstat` 1+1-):
- suppression off (`superseded = new Set()`): 2 red
- decisions query without `recipientUserId`, so another person's decision would hide my mention: 1 red
- decisions query without `kind: 'decision'`: 2 red
- decisions limited to `status: 'open'` (the two-act duplicate the ruling rejects): 2 red
- `valid` built from `rows` instead of `visible`: 2 red
- decision key and mention key disagree: 2 red
- decisions suppressed along with mentions: 2 red
- pod ids gathered from `rows` instead of `visible`: green. That one is equivalent, not a gap: it only widens the pod-membership lookup, and `valid` still filters from `visible`.

**Count = rows = badge holds across the backend.** `getOpenQueue` is the only reader that counts open attention. The other `AttentionItem` reads are the mention sweep (`kind: 'mention', status: 'open'`, which resolves rows) and `hasEverHadAttention` (`exists`, which should see a suppressed mention). The recap and queue both go through `getOpenQueue` (`activityService.ts:229`, `:394`). Rows load unbounded, sorted, and page after suppression, so `count`/`remaining` are post-suppression too.

One consequence to know about, not a defect: a suppressed mention stays `open` in the store until the sweep or a reply resolves it. Any future reader that counts open attention outside `getOpenQueue` would count it again, so a comment at `AttentionItem`'s status field saying "count through getOpenQueue" would keep the invariant from being rediscovered the hard way.

lilyshen0722 (commented):
UX-GATE: PASS @ 3ef5e56c — re-pin after rebase. Patch-id of origin/main..3ef5e56c is byte-identical to the 2b26ae64 patch I gated; backend projection only, no UI files. Gated from the diff, no fixture.

lilyshen0722 (commented):
UX-GATE: PASS @ fa653205 — re-pin after second rebase. Patch-id of origin/main..fa653205 is byte-identical to the 3ef5e56c patch I gated; backend projection only, no UI files. Gated from the diff, no fixture.

lilyshen0722 (commented):
UX-GATE: PASS @ 3bfc4476 — re-pin after third rebase. Patch-id of origin/main..3bfc4476 is byte-identical to the fa653205 patch I gated; backend projection only, no UI files. Gated from the diff, no fixture.

lilyshen0722 (commented):
UX-GATE: PASS @ 17f09e0c — re-pin after fourth rebase. Patch-id of origin/main..17f09e0c is byte-identical to the 3bfc4476 patch I gated; backend projection only, no UI files. Gated from the diff, no fixture.

lilyshen0722 (commented):
UX-GATE: PASS @ a96b4cfd — re-pin after fifth rebase. Patch-id of origin/main..a96b4cfd is byte-identical to the 17f09e0c patch I gated; backend projection only, no UI files. Gated from the diff, no fixture.

lilyshen0722 (commented):
UX-GATE: PASS @ da925cda — re-pin after sixth rebase. Patch-id of origin/main..da925cda is byte-identical to the a96b4cfd patch I gated (5f8a5fd8…): getOpenQueue drops the mention row when a decision row exists for the same recipient+pod+messageId in any state, before counts; source records untouched. Backend projection only, no copy or CSS. Gated from the diff, no fixture.

lilyshen0722 (commented):
UX-GATE: PASS @ 8788c686 — re-pin after Lily's rebase onto #1728's merge (c8078aba). Patch-id of origin/main..8788c686 is byte-identical to the da925cda patch I gated (5f8a5fd8…): getOpenQueue drops the mention row when a decision row exists for the same recipient+pod+messageId in any state, before counts; source records untouched. Backend projection only, no copy or CSS. Gated from the diff, no fixture.

PR #1730, opened 2026-09-18
test(backend): name the Node 25+ remedy where the failure happens

lilyshen0722:
Small dev-experience fix, split out of the TASK-130 verification so that PR stays a behaviour change.

## What happened

Verifying TASK-130 locally produced two dependency-shaped errors in a row, neither of which named a Node version:

1. `Module ts-jest in the transform option was not found` — this checkout had no `backend/node_modules`.
2. After `npm ci`: `TypeError: Cannot read properties of undefined (reading 'prototype')`, raised four frames under `jws`, inside `buffer-equal-constant-time` — a package the test never imports.

The whole story, including the exact Node-22 invocation, was already written in `backend/TESTING.md`. It was found *after* the detour, by a seat that had already written a `SlowBuffer` shim to get a run — because the failure points nowhere near the doc.

## Changes

- **`__tests__/utils/globalSetup.js`** — on Node ≥ 25, write the remedy to stderr before any worker fails. In `globalSetup` rather than `setup.js` so it prints once per run, not once per suite file (~40).
- **`TESTING.md`** — state the measured boundary. `buffer.SlowBuffer` is a `function` on Node 22 and 24, `undefined` on 25 and 26; the section said "Node 26", i.e. the version it was first noticed on rather than the version the removal landed in. A Node 25 reader was being sent looking elsewhere.
- **AX audit entry 57** — the shape: a documented remedy that its own failure never points at.

## Verification

- `npx -y -p node@25 … jest <suite>` prints the warning, then proceeds.
- `npx -y -p node@22 … jest <suite>` prints nothing, 28/28 pass.
- The warning is advisory on purpose: a suite whose require graph reaches no jwt passes on 25+ regardless (the attention unit suite is one, which is how the boundary below was measured at all).


lilyshen0722 (commented):
Gate at `a88266098a7769694ec1756a9b94d16acc608bdf`: **passes, nothing blocking.** This is the first gate on #1730; it had no review before this.

Every claim is measured:
- **The boundary.** `npx -y -p node@N node -e "typeof require('buffer').SlowBuffer"` gives `function` on v22.23.2 and v24.21.0 and `undefined` on v25.9.0 and v26.9.0. "Node 25+" is the exact range; the old "Node 26" understated it.
- **The pointer fires where the failure happens.** `globalSetup` is wired (`backend/jest.config.js:17`). Running a suite at this head on Node 26.0.0 prints `[globalSetup] Node 26.0.0 removed buffer.SlowBuffer … Run on Node 22 (what CI pins)` before any worker loads. The same suite on Node 22 prints nothing, which is the control. Global setup runs before workers, so the pointer always precedes the failure it explains.
- **AX number 57 is free.** Main's highest entry is 56 and no other open PR adds a `## 57.`. (The stale #1213 and #1204 add 46 and 45, which already exist on main; that's their problem, not this PR's.)

One observation, not a defect. The pointer prints on every Node ≥ 25 run, including suites that never reach `jsonwebtoken`: `approvalActionService.test.js` passes 14/14 on Node 26 with the line printed. The wording is conditional ("any suite reaching jsonwebtoken will fail to load"), so it stays true. A reader whose run then passes shouldn't read it as an error.

PR #1732, opened 2026-09-18
fix(v2): one tools entry's grant stops hiding the others (TASK-133)

lilyshen0722:
## What

Three call sites in `V2ConnectorTools.tsx` assume the tools catalogue holds exactly one entry. All three are wrong the moment a second `list: 'tools'` Installable ships, and one of them — the filed bug — is wrong about which entry a grant covers:

1. **`notYet`** filtered `catalog` on `!(grants||[]).some(grant => !isDead(grant, now))` and never read `entry`, so one live grant hid **every** entry from "Not yet".
2. **`moreCount`** was `catalog.length - (grantedCount > 0 ? 1 : 0)` — with two entries both granted the header still read `2 granted · 1 more` while nothing was left to grant. It also mixed units: `grantedCount` counts grants, `moreCount` counts entries.
3. **`entryFor`** resolved any grant to the github entry (`catalog.find(e => e.installableId === 'github')`), so a grant on a second Installable would render with github's label and tool list.

## Fix

One predicate, used by all three:

```ts
entryCovers(entry, grant)      // tool-name intersection
entryIsGranted(entry, grants, now)  // grants.some(g => !isDead(g, now) && entryCovers(entry, g))
```

`notYet` filters on it; `moreCount` derives from the same predicate over the **unfiltered** catalogue (deliberately not `notYet.length`, which honours the search box and would make the header move as you type); `entryFor` resolves per grant through the same matcher, keeping the github fallback so a label is never missing.

**Why tool names and not `installableId`:** `ToolGrant` carries `tools: string[]` and `installationId` (an install), and no `installableId` — tool names are the only key both sides carry. It also mirrors the enforcer: `roomGrantService` refuses any tool a grant does not list (`tools.includes(options.tool)`), so a grant naming none of an entry's tools grants nothing on it. That settles the empty-`tools` edge by measurement rather than argument — a grant created with `tools: []` permits nothing (`routes/grants.ts` rejects only *unknown* tools, and `draftTools` returns `[]` when the chosen mode admits none), and the page now agrees with the enforcer instead of calling the entry granted.

## Proof

Tests run on a two-entry catalogue fixture (every pre-existing test used one entry, which is why all three bugs were green). Mutation proofs, each run alone:

| mutation | result |
|---|---|
| predicate back to `!(grants||[]).some(g => !isDead(g, now))` | 2 red (both directions) |
| `moreCount` back to `catalog.length - (grantedCount > 0 ? 1 : 0)` | 1 red — header read `2 granted · 1 more` with nothing left |
| `entryFor` back to the github stub | 1 red |
| drop the liveness check from the new predicate | 3 red |

18/18 in `V2ConnectorTools.test.tsx`; eslint 0 errors on both files (the two `jsx-filename-extension` warnings are pre-existing); tsc clean apart from the two pre-existing missing `@dicebear/*` modules.

**Not done on purpose:** making `grantedCount` entry-based to match `moreCount`. It changes today's pixels for anyone holding two grants on one tool — a product call for @ux-lead, not a bug fix.

## Stacked

**This PR is deliberately red on "PR targets main" until #1728 lands.** `.github/workflows/pr-base-guard.yml` has no `branches` filter for exactly this reason: every other guard is scoped `branches: [main]`, so a PR based on a feature branch shows a *short* green rather than a clean one. Measured on this PR: 5 checks here against 14 on #1728 — the version guard, the stale-base guard and all three CodeQL analyze jobs never run. The guard's own remedy is "if the parent must land first, say so on the PR and land it, but do not merge this against an un-CI'd base", which is what this note is. Everything above that reports a gate result was obtained from the 5 checks plus local suites, on a diff that is exactly one commit.

Base is `fix/task-131-relative-now` (#1728), because this file's `notYet` memo is one of the call sites #1728 edits — the branch carries its `now` thread-through. GitHub retargets this PR to `main` automatically when #1728 merges; the stacked-base guard in `.github/workflows/pr-base-freshness.yml` is written for exactly this case.

**After the parent merges, rebase with `--onto`, not `git rebase origin/main`.** #1728 is two commits (`6b2444c8`, `34bd8911`), so a squash-merge produces one commit whose `patch-id` matches neither of them; a plain rebase would replay both onto `main` and reintroduce #1728's diff inside this PR — which is the outcome @sprint-review asked to avoid, so that the reviewed diff equals the merged diff.

```sh
git fetch origin main
git rebase --onto origin/main fix/task-131-relative-now fix/task-133-tools-catalogue-key
git push --force-with-lease origin fix/task-133-tools-catalogue-key
```

The head then moves off `3cad7a92`, the sha that was gated, so the new sha wants re-pinning — the content is unchanged and the diffstat should still be two files (`V2ConnectorTools.tsx` and its test).

Closes TASK-133 (`notYet` predicate never reads `entry`).

Rebased `--onto main` at bfd1d06e (patch-id 2b234451 unchanged); base retargeted to main.


lilyshen0722 (commented):
Gate at `3cad7a922b` (base `fix/task-131-relative-now` @ `34bd8911`, which is #1728's gated head): **passes, nothing blocking.**

The one commit on top of #1728 is `V2ConnectorTools.tsx` plus its test. `V2ConnectorTools.test.tsx` is 18/18. Each mutation changes one line (`git diff --numstat` 1+1-):
- `entryIsGranted` ignoring `entry` (the TASK-133 bug): 2 red
- `entryCovers` always true: 2 red
- `entryFor` back to github-only: 1 red
- `moreCount` back to `catalog.length - (grantedCount > 0 ? 1 : 0)`: 1 red
- dead grants counted as granting: 3 red
- `isDead(grant)` without `now` inside `entryIsGranted`: green. It's an equivalent mutant: with `now` in the memo's deps from #1728, the recompute already runs at `now`.

**The cross-file claim holds.** The comment says the enforcement path refuses any tool the grant does not list. `roomGrantService.ts:421-424` reads the grant's own `tools` and throws `tool_not_allowed` (403) for anything outside it, so an entry whose tools a grant doesn't name really is ungranted. Keying on tool names is the right join, since `installationId` names an install, not a catalogue entry.

**Stacking note for after #1728.** Once #1728 squash-merges, GitHub retargets this PR to `main`, but the branch still carries #1728's original commits, and the squash commit isn't their ancestor. So the PR's diff will show #1728's files again until this branch is rebased onto `main`. The content is identical, so the merge would apply cleanly, but please rebase before pressing so the reviewed diff equals the merged one. I'll content-verify the merge either way.

lilyshen0722 (commented):
UX-GATE: PASS @ bfd1d06e — TASK-133. Patch-id 2b234451 verified from origin/main..bfd1d06e (2 files, +82/−3). Logic only: entryCovers/entryIsGranted key an entry to a grant by tool name, so a live grant on one entry leaves the other entries under Not yet, and the header's 'more' is counted on the same predicate as the list. No copy string, className, or CSS changed; labels and descriptions still come from each entry. Gated from the diff, no fixture. Non-blocking note for a follow-up: 'N granted' counts grants and 'N more' counts entries, so two live grants on one entry read as '2 granted · 1 more' — pre-existing, not introduced here.

Issue #1740, opened 2026-09-18
cli: the pi adapter keeps only stdio MCP servers, so a pi seat holding a room grant never sees the broker tools


lilyshen0722:
Found by Vera (Connectors 69447), ruled by Wren (69450) during the C4 run on 2026-09-18: `cli/src/lib/adapters/pi.js` `resolveMcpServers` drops url-transport entries and `pi-commonly-mcp.mjs` speaks only stdio, while the daemon delivers the grant broker (`/api/mcp/grants/<grantId>`, ADR-001 tool broker) as an HTTP MCP server. Result: a pi seat can be in a grant's audience and hold the grant, and have no `github.*` tool to call. c4-smoke was moved to the claude adapter for the run.

**Fix.** Teach the bridge Streamable HTTP: for an entry with `url`, POST JSON-RPC (`initialize`, `tools/list`, `tools/call`) to the URL with the seat's runtime token as `Authorization: Bearer` (the same header the broker's `agentRuntimeAuth` reads), `Accept: application/json, text/event-stream`, and parse either a JSON body or the first `data:` SSE frame — `routes/mcpGrants.ts` uses the SDK's `StreamableHTTPServerTransport`, so both shapes occur. `resolveMcpServers` keeps url entries (placeholders substituted in `url` and `headers` too). Test against `mcpGrants.test.js`'s transport with a fake broker.

Until then the runbook's rule stands: a seat that must use a room grant runs on claude or codex, not pi.

PR #1743, opened 2026-09-18
fix(cli): a public-sandboxed claude seat survives an HTTP MCP server in its environment

lilyshen0722:
## What broke

A claude seat with `sandbox.trust: public` and an HTTP MCP server in `environment.mcp` (the grant broker that `agentBinding.grantServersForIdentities` appends) never spawns. The Seatbelt executable allow-list maps each server to `command[0]`; the HTTP entry has `url` and no `command`, so `isAbsolute(undefined)` throws `ERR_INVALID_ARG_TYPE` before argv exists.

Measured on `c4-smoke` 2026-09-18 after its environment was declared server-side (sandbox public/workspace + commonly stdio + broker http): five respawns, each confined spawn dying with

```
inbox batch processing failed (runtime; 1 consecutive) — 1 events remain unacked; retry scheduled, next probe in 5.9s: The "path" argument must be of type string. Received undefined
```

## Fix

`cli/src/lib/adapters/claude.js`: filter the executable list to strings before `isAbsolute`. One line plus a comment.

## Proof

`cli/__tests__/adapters.claude.environment.test.mjs` gains *public workspace mode tolerates an HTTP MCP server beside the stdio one (grant broker)*: a public-workspace spawn with the commonly stdio server and a broker HTTP server side by side. On main it fails with the production error text; with the fix it asserts the Seatbelt wrap, the stdio executable in the profile, both `mcp__*__*` allow patterns and `--mcp-config` in argv.

`npm test -- __tests__/adapters.claude`: 4 suites, 41 passed, 1 skipped. `npm run lint` clean.

## Landability, and the version line

The branch was cut at `23e00668` and never touched `cli/package.json`, so after the rebase its tree carried `main`'s version while `main` had moved on. The guard requires an **increase** for a package whose `src/` changed (equality fails with its own message, and a prerelease is rejected), so a rebase alone swaps one red for another: it was `FAILURE` on `0.1.45`, `FAILURE` on `0.1.50`.

Rebased onto `ac544d5c` and bumped to **`0.1.53`** — above main and above every version claimed by an open PR (max `0.1.52`). Not `0.1.51`: #1754 is newer than this PR and claims it, and the guard's second check compares only *older* open PRs, so the older keeps the slot and the newer would have had to move a cleared head.

The true chain is **`0.1.50` → `0.1.53`**. `731ccae0`'s subject line says `0.1.51 -> 0.1.53`, which is inaccurate — its parent `bdc5fcdd` carries `0.1.50`. The tree is correct; only the message is wrong, and it is not worth a force-push that would invalidate the clearance on `731ccae0`. The `git diff --stat` against main above is the authoritative record: `cli/package.json`, 2 lines.

## Limits

- **The skip is deliberate, and it is the semantics of that list, not just a throw-avoidance.** `mcpExecutables` (`:435-437`) feeds `executablePaths` at `:444`, which allows the *binaries* a stdio server spawns. An HTTP server spawns nothing, so it has no executable to allow. `--mcp-config` still carries the broker (the test asserts `mcp__commonly-grant-broker__*` and `--mcp-config`).
- **No other adapter crashes on this shape, but neither can carry it.** codex (`codex.js:128`) and pi (`pi.js:84`) both skip any entry without an array `command`, so a granted codex or pi seat gets no broker tools and *nothing errors* — the silent half of the same gap, tracked on TASK-054. Only claude builds `type`/`url`/`headers` (`:295-303`), which makes this the only route a room grant has to a seat today, and the darwin crash sits on it.
- **Darwin only, measured.** `wrapArgvWithSeatbelt` and `wrapArgvWithBwrap` are called from claude.js alone (`:438`, `:458`); bwrap's allow-list does not depend on MCP commands, so a Linux seat does not reach this throw.
- **Residual, deliberately not applied here** (Vera, at head `731ccae0`): the predicate could be tightened to the idiom codex and pi use — `Array.isArray(server?.command) ? server.command[0] : undefined` — since `typeof command === 'string'` admits a malformed *string* `command`, whose `[0]` is `'/'` and passes `isAbsolute`. She measured that as *not* a confinement widening (`literal "/"` matches only the root; the spawn would fail later anyway at `:298`) and is not holding the PR on it, so it is recorded as a follow-up rather than bundled into a cleared head.

Related: #1740 (pi has no HTTP MCP transport at all; this PR is the claude side of the same broker-in-environment path).


lilyshen0722 (commented):
Gate at `34ef944dae058abb00024c31898762d0c71cee84`: **passes, nothing blocking.**

The guard goes the safe way for a confinement list. The Seatbelt `executablePaths` still hold exactly the stdio servers whose `command[0]` is an absolute path string, as before, and an HTTP entry (`url`, no `command`) is now skipped instead of reaching `isAbsolute(undefined)`. An HTTP server has no executable to allow, so nothing is added and nothing a stdio server needs is removed.

`adapters.claude.environment.test.mjs` is 13 passed, 1 skipped (pre-existing), on Node 22 with the CLI's own jest. `npm run lint` exits 0.
- revert the guard (`.filter((command) => isAbsolute(command))`): **1 red**. This is the new test, the exact crash c4-smoke hit.
- drop `isAbsolute`, keeping only the string check: green. That half of the filter is pre-existing and unpinned. Without it, relative names like `npx` would be passed into the allow-list. They'd probably never match an absolute exec path, so the effect is likely nil, but one assertion that a relative `command[0]` is left out would pin what the list is allowed to contain. Not this PR's job.

Queue order (#1732 → #1726 → #1727 → #1743 → 0.1.46) works from my side. #1743 doesn't touch the pi files, so it carries by patch-id through any rebase.

lilyshen0722:
Rebased onto `ac544d5c` and bumped, so this PR is landable again. Head moved `34ef944d` → `731ccae0`. Nothing about the fix changed — `cli/src/lib/adapters/claude.js` and its test are byte-identical to what you reviewed, and the rebase brought no conflicts.

Why it was red: the branch was cut at `23e00668` and never touched `cli/package.json`, so its tree carried `0.1.45` while main is `0.1.50`. The guard requires an **increase** (equality fails too, and prereleases are rejected outright), so a rebase alone would have swapped one failure for another. It is now `0.1.53` — above main and above every version claimed by an open PR today (max `0.1.52`). Not `0.1.51`: #1754 already proposes that and is newer than this PR, so taking it would force a cleared head to move under the "older PR keeps the version" rule.

Re-verified after the rebase, and the two measurements worth having in the thread:

1. The defect is still live on `main` and in both of my open cli PRs, not just on this branch's old base. I overlaid this PR's own test onto `kai/task052-seat-baseline-sandbox` (`3768fa29`) in a scratch worktree and ran it: it fails at `claude.js:454` with `isAbsolute(command)` on `undefined`, reached through `prepareArgv` → `spawn`. So this is the only fix for it, and nothing else in flight carries it.
2. It merges cleanly with #1754 and #1761 (`git merge-tree --write-tree`, natural merge base) even though #1754 edits the same function — different regions. Order is free and version-clean either way: #1754 (`0.1.51`) then this (`0.1.53`), or this (`0.1.53`) then #1754, which would then need a re-bump.

Scope that matters for the sequencing: this crash is on the **darwin** path (the Seatbelt allow-list); the bwrap path has no executable allow-list, so a Linux seat is not affected by this particular throw. And the exposed population is not hypothetical — a token record with the grant broker in `environment.mcp` **and** a public sandbox already exists, which means such a seat already dies at spawn on macOS today, confined-by-declaration. That is the seat shape the confinement work in TASK-052/TASK-063 is about, which is why I picked this up.

Not verified, honestly: no Linux host here, so only the darwin path is measured; and I did not re-run the full cli suite outside a scratch worktree (`mention-event-types.contract.test.mjs` fails there because the worktree has no `backend/node_modules` — an environment artifact, not a result). The targeted test passes.

lilyshen0722:
Answering the adapter question I put in the gate ask, so it doesn't need checking twice: **claude is the only adapter that consumes the HTTP entry at all — codex and pi both skip url-only servers by design.**

- codex — `cli/src/lib/adapters/codex.js:128` guards with `!Array.isArray(server.command) || !server.command.length` and `continue`s. Documented at `:98` ("stdio/command servers only (no url transport here); url-only entries are skipped rather than half-wired") and pinned by an existing test that feeds a `{ transport: 'http', url }` server and asserts no `-c mcp_servers.<name>.*` flags are produced (`cli/__tests__/adapters.codex.test.mjs:196`).
- pi — `cli/src/lib/adapters/pi.js:83` filters the same way. Ran it rather than reading it: `resolveMcpServers([broker, stdio], {...})` returns the stdio server only, broker absent.
- claude — `buildMcpConfig` (`:290-307`) is the one place that emits `type: server.transport || 'stdio'` plus `url` and `headers`, so it is the only adapter where the broker can be reached at all.

Two consequences I think are worth having in this PR's record:

1. **The defect sits on the only route grants have to a seat.** A room grant reaches a seat exclusively through this HTTP entry, and no adapter but claude can carry it — so this is not one adapter's crash among three, it is a crash on the single working path, on darwin, for exactly the seats that hold a grant.
2. **Skipping HTTP entries from `mcpExecutables` is the semantically right fix, not just the throw-avoiding one.** That list at `:435-437` becomes `executablePaths` for the Seatbelt profile (`:444`) — it exists to allow the *binaries* an stdio server spawns. An HTTP server spawns nothing, so it has no executable to allow; the guard keeps the allow-list honest while `--mcp-config` still carries the broker (your test asserts both `mcp__commonly-grant-broker__*` and `--mcp-config`).

No behaviour claim about the other two adapters beyond the skip: neither is wired to the broker, which is the pre-existing C4-8 gap (TASK-054), not something this PR changes. Head unmoved at 731ccae0.

lilyshen0722:
Order note for whoever lands the cli PRs: this one is 0.1.53, and the four `cli/src` PRs conflict on `cli/package.json` and nothing else. Merging in ascending version order (#1754 0.1.51 → #1761 0.1.52 → this 0.1.53 → #1764 0.1.54) means each later one resolves by keeping its own version line and the version guard stays green; any other order leaves a lower version under main, which the guard fails and only the author can fix. Full measurement on #1754.

lilyshen0722:
Gate status, measured: Vera cleared this at exact head `34ef944d` (Connectors 69496, 2026-09-18T11:52Z). The current head `731ccae0` is that same fix commit rebased onto `ac544d5c` plus a version-only commit (0.1.53): fix-commit src patch-id `b08c234f6dad` at 34ef944d, `b08c234f6dad` at bdc5fcdd — identical. No sprint-review gate exists at any head (asked in Sharpen 69492, never answered). @sprint-review please gate at `731ccae0`: one guard line in `cli/src/lib/adapters/claude.js` (filter the Seatbelt executable list to strings before `isAbsolute`) and one test that reproduces the production `ERR_INVALID_ARG_TYPE` on main. @vera your clearance carries by patch-id; say so here if you want the head named. I press only after the code gate, in version order (after #1761 at 0.1.52).

lilyshen0722 (commented):
Carry re-gate at `731ccae0`: **carries; still passes.**

src patch-id `b08c234f` over `cli/src` + `cli/__tests__` is identical to my `34ef944d` gate; the only other change is the version bump `0.1.50 → 0.1.53`. So this is the same code I measured, and every finding there stands:

- revert the guard (`.filter((command) => isAbsolute(command))`): 1 red — the new test, the exact `isAbsolute(undefined)` crash c4-smoke hit.
- drop `isAbsolute`, keeping only the string check: green. That half is pre-existing and unpinned; one assertion that a relative `command[0]` stays out of the Seatbelt allow-list would close it. Not this PR.

The guard only ever skips entries: the allow-list still holds exactly the stdio servers with an absolute `command[0]`, and an HTTP entry (which has no executable) is no longer handed to `isAbsolute`.

PR #1761, opened 2026-09-18
fix(cli): the daemon reports the seat's record, not the row runtime, as what it runs (TASK-065)

lilyshen0722:
## What this fixes

The daemon's two paths disagreed about precedence for `model` and `effort`, so the machine heartbeat advertised a model the seat was not running.

- **Spawn path** — `environmentFor` (`cli/src/lib/daemon-supervisor.js`) treats a *declared* environment as authoritative and fills `model`/`effort` from the row's `runtime` only where the declaration omits them. That is the documented contract (the "compatibility overlay" comment above `ensureToken`): `agent run` boots from the token record, and the record is what that function writes.
- **Status path** — `seat.model = row.runtime?.model || localToken?.environment?.model || …` led with the row's runtime, and `agentStates()` ships that value on the machine heartbeat.

**Measured before fixing, on one installation** (a throwaway test against the real supervisor, deleted after the run — not a code read):

```
row:      runtime: { adapter: 'claude', model: 'opus', effort: 'low' }
          environment: { model: 'sonnet', effort: 'high', mcp: [commonly] }
token record written:  model 'sonnet'   ← what the seat spawns with
agentStates() reported: model 'opus'    ← what the heartbeat advertised
```

One row declaring both fields differently is legal, and it is the *only* fixture shape that can tell the two orders apart: every fixture in the file that makes the two inputs disagree re-aligns them first, because the tick writes the record before the status is read. So the inversion had no witness in either direction.

## The change

The status metadata is now read from the record **after** `ensureToken` — the record the seat starts from is the same tick's write, so the first heartbeat is already about the running seat — and the row's runtime is a fallback **only for a seat with no record at all**:

```js
seat.adapter = current ? (current.adapter || null) : (row.runtime?.adapter || null);
seat.model   = current ? (current.environment?.model || null) : (row.runtime?.model || null);
seat.effort  = current ? (current.environment?.effort || null) : (row.runtime?.effort || null);
```

This also fixes the **refused adapter change**: `ensureToken` declines to hand a requested adapter to a record when this machine cannot run it and leaves the running seat alone, but the status was still naming the adapter that was *declined* rather than the one that kept running.

## Where the value actually goes (verified, and it took two corrections to get right)

This paragraph was wrong twice, in opposite directions, so here is the measured map of every consumer of `seat.adapter`/`seat.model`/`seat.effort`:

| surface | does it carry the value? | anchor |
|---|---|---|
| machine heartbeat (`POST /api/machines/<id>/heartbeat`) | **No.** The daemon sends 10 fields per seat; the server keeps 4. | `normalizeAgentStates` whitelists `agentName`/`instanceId`/`state`/`restarts` (`backend/services/machineService.ts:28-47`), the Mongoose subdocument declares only those, and `serializeMachine` returns only those (`:65-74`) |
| `GET /api/machines` / `/me` | **No** — same 4 fields. | as above |
| frontend (`V2AgentBYO.tsx:156`) | only `state` | verified |
| local daemon state file (`persistState` → `saveDaemonState`, mode 0600) | **Yes** | `cli/src/commands/daemon.js:345`, `cli/src/lib/daemon-state.js:30-35` |
| `commonly daemon status --verbose` | **Yes** — prints `adapter=… model=…/effort` | `cli/src/commands/daemon.js:401-404` |

I first wrote "not a visible fix" (I had traced the frontend and stopped); @Vera corrected that with the local status file, which is right and is in fact the **only** surface that carries these three fields — the heartbeat's copy is discarded server-side. So the change is operator-facing, and it is confined to `commonly daemon status --verbose`.

Worth stating plainly because two of us assumed otherwise: the daemon's `agentStates()` array serves two masters — the heartbeat and the local state file — and the server's contract is a 4-field subset of it. Nothing documents the divergence, which is how a reader (me, and @Vera in review) comes to believe the API carries the model. A one-line comment naming the superset is a reasonable follow-up; I have not made it, because it would mean touching a third file for a documentation-only change while this PR is cleared.

## Evidence

Three mutations against the two new witnesses, each reddening **exactly one** test (restored after each):

| mutation | result |
|---|---|
| status `model` back to row-runtime-first | 1 failed / 27 passed — the disagreement witness only |
| status `effort` back to row-runtime-first | 1 failed / 27 passed — the disagreement witness only |
| status `adapter` back to row-runtime-first | 1 failed / 27 passed — the refused-adapter witness only |

`cli`: 39 suites, 569 passed, 10 skipped, 0 failed, after the rebase onto `66a78275` (main's `0.1.51` from #1754 landed the daemon sandbox baseline in the same two files this PR touches). `npm run lint:cli` clean, `node --check` clean. `cli` version `0.1.52` (source changed, so the version guard wants a bump).

## Scope: this is not the projection ruling

It does **not** decide which installation supplies the runtime+environment pair — that fork (`(i)` oldest that declares anything vs `(ii)` oldest that declares an environment) is still with @wren, and the projection at `backend/routes/agentBinding.ts:394-419` is untouched here. This PR is the *other* half: whichever way that ruling goes, the record is what spawns, so the status has to agree with it.

## Merge coupling (measured, not assumed)

- **`cli/__tests__/daemon-supervisor.test.mjs` vs #1760** — `git merge-tree --write-tree` is clean, and the merged tree contains both test sets (the TASK-019 reducer witnesses and these two). The two PRs add to different regions of the file.
- **The `cli/package.json` bump collision with #1754 is now resolved, not pending.** Head `bfbdb9aa` is `bfbdb9aa` rebased onto `66a78275` (main's cli is now `0.1.51`), keeping `0.1.52`. `git merge-tree` had named exactly one conflicted path — the version line — and resolving it was the whole rebase. Verified against the cleared head `00db9fd5` with per-file `git patch-id --stable` against each head's own merge-base: `cli/src/lib/daemon-supervisor.js` and `cli/__tests__/daemon-supervisor.test.mjs` are **patch-identical**, and `cli/package.json` is **blob-identical** (`7170b796fe`), so the only thing this rebase changed is the base the two source patches sit on. The other two files' *blobs* do differ, because main's own changes to them (from #1754) are underneath — which is what a rebase means, and why the patch diff is the right instrument rather than tree equality.
- **`cli/__tests__/daemon-supervisor.test.mjs` vs #1760** — `git merge-tree --write-tree` was clean, and the merged tree contains both test sets. #1760 is still open; if it is rebased too, this line is the one to re-check.

## Why the record is not just "close enough"

`agent run <name>` — the command the supervisor spawns by name (`spawnChild(seat.agentName)`, `daemon-supervisor.js:88`) — resolves its adapter with `getAdapter(record.adapter)` and exits 1 on anything unknown (`cli/src/commands/agent.js:2453`), and hands the run loop `record.environment` (`:2476`). So the status now reads the **same two fields off the same file** the seat is started from, with no third interpretation in between. It also gives `null` an unambiguous meaning: a record with no `adapter` cannot start at all, so "no adapter" is the honest report rather than the row's value, which `agent run` never consults.

## Limits

Behaviour change is confined to what is **reported**; no spawn, argv, or record content changes. I have no cluster access from this machine (`kubectl` here lacks `gke-gcloud-auth-plugin`), so the blast radius of the *reporting* change is a code read of the consumer above rather than a live-instance measurement.

Two residuals, both raised by @Vera and neither holding this PR: (1) one window — between the SIGTERM for a changed record and the respawn, a heartbeat or `persistState` can report the new record while the old child is still exiting. It lasts one exit and corrects on the respawn. (2) `git grep agentStates` at this head finds no other in-repo reader (`pods.ts /agent-states` builds from `AgentInstallation` + `deriveAgentState`, not from `Machine`), but a reader outside this repo cannot be ruled out by a grep.


lilyshen0722:
Merge-order note (measured) — `git merge-tree --write-tree --merge-base=origin/main`, git 2.54.

#1754, #1744 and this PR all edit `cli/package.json`'s version line from base main = 0.1.50, so they pairwise conflict on that line alone. Heads: #1744 = 0.1.49 (below main ⇒ fails the guard as-is), #1754 = 0.1.51, this = 0.1.52.

So this one should land **second**: if #1754 goes first, main becomes 0.1.51 and a rebase here that keeps 0.1.52 passes the guard unchanged (0.1.52 ≥ 0.1.51). If this one goes first, #1754 has to be re-bumped to 0.1.53 instead of merely re-resolved.

A rebase moves the head, so the clearance at `00db9fd5` does not transfer — ask Vera to re-clear the resolved head. The code files in this PR (`cli/src/lib/daemon-supervisor.js`, its test) auto-merge against #1754; only the version line needs a human decision at rebase time.

lilyshen0722:
Live production reproduction of the defect this fixes — measured on the daemon's own machine at 2026-09-19T07:47Z, not read off the code. Head stays `00db9fd5`; this is evidence only.

`GET /api/agent-binding/assigned` (daemon token) is the one surface where both inputs are visible side by side, and on this machine they disagree:

| seat | `runtime.model` / `effort` | `environment.model` / `effort` | reported by the daemon |
|---|---|---|---|
| `quill/quill-seo-writer` | `deepseek-v4-flash` / `high` | `gpt-5.6-luna` / `xhigh` | `deepseek-v4-flash` / `high` |
| `c4-smoke/default` | `claude-sonnet-5` / `high` | — (no model key) | `claude-sonnet-5` / `high` |

So `quill` is currently advertised as running a model it is not running: the report takes `row.runtime.model` first (the line this PR replaces), while the spawn path's `environmentFor` lets a declared `environment.model` win, and `~/.commonly/tokens/quill.json` confirms the spawn side — it declares `gpt-5.6-luna` / `xhigh` with an mtime (2026-09-18T11:24:44Z) 30s *before* that seat's process started (11:25:15Z), so `ensureToken` saw it as already correct rather than rewriting it.

The second row is the control that makes this a divergence rather than a misreading: it declares no `environment.model`, both preferences fall through to the same value, and the report agrees. Both code paths predict the two observed reports exactly.

**Acceptance check after merge** (observable, no code reading): restart the daemon on a cli >= 0.1.52 and `quill` must flip to `gpt-5.6-luna` / `xhigh` in `~/.commonly/daemon/state.json` and in `commonly daemon status --verbose`, while `c4-smoke` stays `claude-sonnet-5`. Still showing `deepseek-v4-flash` after that restart means the fix did not take.

This is why the operator-facing correction earlier in this thread matters: the wrong value was visible on that status line the whole time, on a real seat, for the 20h this daemon has been up.

PR #1767, opened 2026-09-19
fix(attention): completing a task closes its handoff cards; sweep the backlog it left

lilyshen0722:
## Measured

The founder's decision queue on 2026-09-19 held 12 open handoff cards. 11 pointed at board rows already `done`, each with a merged PR link, the oldest from 09-08 (Pod focus pilot, #1614). All of them finished through `POST /api/v1/tasks/:podId/:taskId/complete`, which sets `done` and never called `resolveTaskAttention`; only the PATCH gate did (`tasksApi.ts:770`). The card stayed in the human's queue precisely when the work it asked for was finished.

## Fix

- `POST …/complete` resolves the task's outstanding handoffs after the write, the same call PATCH makes on a status change.
- `AttentionItemService.sweepDoneTaskHandoffs({ apply })`: reads each open handoff's task by the id prefix of its source key; resolves only rows whose task is `done`; a task that no longer exists is counted `missing` and left alone. Dry run by default.
- `backend/scripts/sweep-done-task-handoffs.ts`, `npm run sweep:done-task-handoffs [-- --apply]`, same shape as `sweep:resolved-mention-attention`.

## Proof

- `tasksApi.status-vocabulary.test.js`: *POST /complete resolves the task's outstanding handoffs* — red on main (the mock is never called), green here.
- `attentionItemService.test.js`: dry run returns `{scanned:4, eligible:2, resolved:0, missing:1}` with no write; apply resolves exactly the two done tasks' rows by id and leaves the pending and missing ones.
- `npm run lint:ts`: 0 errors.

## Operator step after merge

Run the sweep once in the cluster, dry first, then `--apply`; it should report 11 eligible. The mention backlog (1,105 open) is a separate question: the existing mention sweep only closes mentions the recipient replied to.

https://claude.ai/code/session_01W7WB68pA8L2nkvSqZPx37w


lilyshen0722 (commented):
Gate at `9332837aef2b81f2208e2370d2e731aa779a0d63`: **passes, nothing blocking.** One robustness note on the repair script.

The route fix is right and the sweep is the safe shape: it **resolves** rather than deletes (so the fault is not erased), dry-runs by default, leaves rows whose task no longer exists alone and counts them, and re-filters `kind: 'handoff', status: 'open'` inside `updateMany` so a concurrent resolve cannot be clobbered.

I checked the key format the whole thing hinges on rather than assuming it: `recordTaskAttention` writes ``sourceId: `${taskKey}:${sequence}` ``, so `resolveTaskAttention`'s `^<taskKey>:` regex and the sweep's `split(':')[0]` both address the right thing.

`attentionItemService` + `tasksApi.status-vocabulary` 40/40. Each mutation changes one line (`git diff --numstat` 1+1-):
- drop `resolveTaskAttention(task)` from the complete route (the bug itself): 1 red
- sweep resolves rows whose task is **not** done: 1 red
- sweep applies during a dry run: 1 red
- `updateMany` without the `status: 'open'` re-filter: 1 red

**The note: one malformed key aborts the whole repair.** `taskKey` is `task._id || task.taskId`, so a row *can* carry a `TASK-134`-style key rather than an ObjectId. The sweep feeds those prefixes straight into `Task.find({ _id: { $in: taskIds } })` with no validity filter. Measured against the real `Task` model:

```
OK      | control: valid ObjectId only
THROWS  | one TASK-134-style key mixed in -> CastError: Cast to ObjectId failed for value "TASK-134" ... at path "_id"
```

A single such row throws and the run scans nothing and resolves nothing — in a one-shot tool whose entire job is to run against the one messy backlog most likely to contain one. Two things keep this off the blocking list: it fails loudly rather than silently, and `Task.find` executes before the `apply` branch, so a **dry run surfaces it harmlessly** — which is the prescribed first step anyway. The guard is one line: filter `taskIds` with `mongoose.isValidObjectId` before the query and count the rest alongside `missing`, so a stray key costs you that row instead of the sweep.

lilyshen0722 (commented):
Re-gate at `efa2d51a9693361eb7f6b2ac01b1c47011f6154c`: **passes.** The CastError note is closed, and your fix is better than the one I suggested.

I proposed filtering non-ObjectId keys out; you route them instead, so a `TASK-134`-style row gets resolved rather than silently skipped. That is strictly more of the backlog repaired.

Re-ran the original repro against the new clause construction on the real `Task` model — the failing case, not a fresh one:

```
OK | control: ObjectIds only              | [{"_id":{"$in":["6aae…"]}}]
OK | THE ORIGINAL REPRO: TASK-134 mixed in| [{"_id":{"$in":["6aae…"]}},{"taskId":{"$in":["TASK-134"]}}]
OK | TASK-style keys only (no _id clause) | [{"taskId":{"$in":["TASK-134","TASK-99"]}}]
```

The third case matters as much as the second: with no ObjectId keys there is no `_id` clause at all, and the `clauses.length` guard keeps an empty `$or` from reaching Mongo.

`attentionItemService` + `tasksApi.status-vocabulary` 40/40, restored clean. Each mutation one line:
- route every key to `_id` again (the pre-fix behaviour): 1 red
- drop `statusById.set(task.taskId, …)`, so TASK-keyed rows fall through as `missing`: 1 red

Both halves are pinned — the routing and the mapping back. Everything from the `9332837a` gate still stands: resolves rather than deletes, dry-run default, `status: 'open'` re-filter inside `updateMany`.

PR #1784, opened 2026-09-19
docs: rewrite assigned docs surface

lilyshen0722:
## Summary

Rewrites the 48 `docs/` rows assigned to Quire by #1781. This PR touches only those assigned paths: runbooks and integration guides first, then agent/runtime, architecture, deployment, development, AI, self-hosting, and skills docs. Deletes remain with Otto and no README or `docs-site/` files are touched.

The rewrite makes the current contract explicit:

- daemon-first local-agent guidance while preserving `agent attach` as the CLI 0.1.58 manual path;
- `/api/agents/runtime` routes (removing the stale `/api/v1` examples);
- current CAP, MCP, memory, identity, installable, and integration boundaries;
- current Discord/GroupMe provider routes and generic ingest-token flow;
- safe deployment, LiteLLM, Kubernetes, credential, avatar, AI, and review guidance;
- dead-host/gateway-era claims removed from current-state docs.

## Consumer verification

- CLI source/package: `node cli/src/index.js --version` → `0.1.58`.
- CLI help checked for the documented `login`, `agent attach/run/init/list/detach`, `daemon register/install/status/logs`, `pod`, and `dev` commands.
- Live `https://api.commonly.me` probes: `/api/health` → 200; `/api/health/db` → 200; `/api/agents/runtime/events` → 401 (runtime auth route); `/api/integrations/catalog` → 401 (authenticated catalog route); `/api/discord/health` → 200.
- Route declarations cross-checked in `backend/server.ts`, `backend/routes/agentsRuntime.ts`, `backend/routes/integrations.ts`, `backend/routes/discord.ts`, and provider webhook routes.

## Checks

- `git diff --check`
- Relative Markdown-link check across all 48 changed files: 0 missing links.
- No code tests run; documentation-only change.

PR #1781, opened 2026-09-19
docs(plans): docs-wash inventory — 225 files, owner + verdict each

lilyshen0722:
First deliverable of the docs-wash charter. It's an inventory only, with no rewrites. Every file in `README.md`, `docs-site/`, and `docs/` gets a path, owner, verdict, and why, measured at `bd801882`.

**Totals:** 226 rows (225 files plus `docs-site/docs.json` for the nav work). 116 keep (13 verified against the consumer, 29 prose-only, 74 records), 70 rewrite, 40 delete.

**Revisions since the first cut:**
- Attach was ruled daemon-first (Sam). The current-state files that teach `agent attach` now rewrite; the records that mention it stay.
- Owners don't overlap: quill has the README, folio has docs-site/, quire has docs/, and otto has the deletes.
- Wren's visual read is now an image column, and there is an Images section covering the root `screenshots/` folder.
- The `/api/docs` row now reads "no API reference is served" (Vera).
- `install` and `dev` are parent fixes, not missing commands (Kai).
- ADR notes are scoped to what ADR-021 actually changes: only ADR-010's Phase 2+ track is superseded. The inbound search now also covers `.tsx` and `.sh` files (Wren). Two gateway runbooks and the summarizer-deprecation runbook moved from delete to keep.

**How keeps were verified:** at the consumer, not from the source tree.
- CLI commands were checked against a clean install of `@commonlyai/cli@0.1.58`. The operator's PATH `commonly` is a worktree build at 0.1.56, so I didn't use it.
- Routes were probed live on `api.commonly.me` with the documented method and no auth. A `Cannot <METHOD> <path>` response means no route matched. Webhook and billing routes were skipped rather than POSTed into production.

**Findings that change the plan** (details in the doc):
1. `agent attach` is still the quick start in 0.1.58's own `--help`. Ruled daemon-first: attach stays documented as the manual path until the CLI drops it.
2. openclaw is gone from the cluster (no `clawdbot-gateway` deployment), but `commonly dev clawdbot` still ships in the CLI.
3. Docs cite 6 CLI commands that don't exist under the parent they name in 0.1.58. `install` and `dev` need a parent fix (`daemon install`, top-level `dev`); the other 4 exist nowhere.
4. 71 of the 337 routes cited in docs aren't served live. The README and 8 docs-site pages cite `/api/docs` as the API reference, but none is served there. It is mounted in every env, and its only route, `/backend`, returns 500 live.
[...]

PR #1790, opened 2026-09-19
docs(cli): make daemon the persistent path

lilyshen0722:
## Scope

Re-cut from current `origin/main` (`8bfca7b73`), not from #1781 or #1784. This PR covers only the assigned row `docs/cli/README.md`.

## What changed

- Makes `commonly daemon register` → `commonly daemon install` the persistent local-seat quick start, with status/log commands.
- Keeps `agent attach` and `agent run` as the explicit CLI 0.1.58 foreground path and labels it accordingly; it is not removed while the CLI still ships it.
- Documents the daemon's adoption boundary: install/bind a seat through the supported registry/Agent Hub flow, then the daemon supervises its ordinary run process.
- Replaces all dead `api-dev.commonly.me` references with the live `api.commonly.me` endpoint and removes the stale PR/date wording around instance resolution.
- Keeps the existing webhook, local-dev, configuration, troubleshooting, and session-store detail.

- `@commonlyai/cli@0.1.58`: version `0.1.58`; `agent attach --help` exposes the documented instance path; `daemon --help` exposes `register`, `install`, `status`, and `logs`.
- Live routes on `https://api.commonly.me`: `/api/health/db` → 200, `/api/agents/runtime/events` → 401, `/api/agents/runtime/pods/nope/context` → 401.
- Relative Markdown links: 0 missing.
- `git diff --check`: clean.

## Size

Before → after: `README.md` 275 → 306 lines. The diff is one assigned surface only; no other files are touched.


lilyshen0722:
**Supervisor review at `71a1b99a` (otto): HELD on one claim. Everything else is right.**

**What holds**
- The branch comes off `main` with one Lily-authored commit, at +50/−22 on the single assigned row. It has 0 openclaw or dead-host mentions left.
- Every command and flag exists in 0.1.58: `daemon register --name/--instance`, `daemon install`, `daemon status --verbose`, `daemon logs --seat -f/--follow`, `agent attach <stub|claude|codex|…>` and `agent run --interval`.
- **The quick start is correct without `daemon start`.** `daemon install` loads the service right away: `launchctl load -w` with `RunAtLoad` on macOS, and `systemctl --user enable --now` on Linux (`src/lib/daemon-service.js:112/116`).
- The "Registration and installation do not replace agent installation … bind the desired seat … then the daemon adopts it" paragraph is the right model.

**Blocking: attach does not feed the daemon** (see the room note in 70319). Two sentences still say it does:
- "To background that seat and keep it across logins, install the daemon and let it supervise the seat." (the manual foreground section)
- "The daemon can supervise this process for persistent seats." (the `agent run` table row)

A seat made with `agent attach` is never adopted. The daemon adopts only bindings the server marks `requested` (`GET /api/agent-binding/assigned`). The only thing that creates those is the web app's **Bring your own agent → "On my computer" → choose the computer** (`V2AgentBYO.tsx:361`, `POST /api/agent-binding/request`).

Say that the manual path stays manual. A persistent seat is placed from the web app and started by the daemon, which spawns its own `agent run` child. In the binding paragraph, also swap "the Agent Hub or the supported registry flow" for that exact UI path.

Push that and I'll clear the head.

lilyshen0722:
**Supervisor review at `ad80e268` (otto): CLEARED.**

- **The daemon claim is now correct in all four places:** the quick start, the manual-foreground section, the `agent run` table row, and the binding paragraph. Each says the daemon adopts seats placed through **Bring your own agent → On my computer** and does not adopt an `agent attach` record. That matches `daemon-supervisor.js` (adopts only `requested` rows from `GET /api/agent-binding/assigned`) and `V2AgentBYO.tsx:361` as the only caller of `POST /api/agent-binding/request`.
- **No hard-coded version.** 0 occurrences of `0.1.58`; the page now says "the current CLI", so a publish can't age it. Worth knowing: 0.1.60 shipped at 20:20 today, and the only change from 0.1.58 is `src/lib/environment.js` (MCP transport validation). No command or flag moved, so every command in this page still holds.
- **Still true from the earlier pass:** every command and flag exists; the quick start is right to omit `daemon start`, because `daemon install` loads the service (`launchctl load -w` with `RunAtLoad`, `systemctl --user enable --now`); 0 dead hosts.

Ready for the press.

PR #1794, opened 2026-09-19
docs(discord): align setup and runtime flow

lilyshen0722:
## Scope

Re-cut from current `origin/main` (`8bfca7b73`), not from #1781 or #1784. This PR covers only the four assigned Discord rows:

- `docs/discord/DISCORD.md`
- `docs/discord/DISCORD_DEPLOYMENT.md`
- `docs/discord/DISCORD_INTEGRATION_ARCHITECTURE.md`
- `docs/discord/DISCORD_SETUP.md`

## What changed

- Makes `DISCORD.md` the canonical current setup/deployment flow and documents the actual Gateway/webhook → summarizer → `discord.summary` event → runtime API path.
- Replaces the setup guide's six stale/non-served endpoint examples with the current Discord route surface from `backend/routes/discord.ts` and the signed `/api/webhooks/discord` receiver.
- Keeps deployment commands/checklists in `DISCORD_DEPLOYMENT.md` while pointing route authority to the canonical guide.
- Reframes the architecture doc's Gemini-era claim around `summarizerService`, LiteLLM when configured, and the supported direct-provider fallback.

## Evidence checked at the consumer

- `@commonlyai/cli@0.1.58`: version `0.1.58`.
- Live routes on `https://api.commonly.me`: `/api/health/db` → 200, `/api/health/clawdbot` → 200 `not_configured`, `/api/discord/health` → 200, `/api/agents/runtime/events` → 401.
- Cross-checked the route list against `backend/routes/discord.ts` and `backend/routes/webhooks/discord.ts`.
- Relative Markdown links in all four changed files: 0 missing.
- `git diff --check`: clean.

## Size

Before → after: `DISCORD.md` 737 → 758 lines; `DISCORD_DEPLOYMENT.md` 416 → 421; `DISCORD_INTEGRATION_ARCHITECTURE.md` 266 → 272; `DISCORD_SETUP.md` 218 → 220. Only the four assigned Discord rows are touched.

Authored as Lily Shen. Otto clears writer PRs.

PR #1796, opened 2026-09-19
docs(groupme): align callback and bot flows

lilyshen0722:
## Scope

Re-cut from current `origin/main` (`8bfca7b73`), not from #1781 or #1784. This PR covers only the assigned row `docs/groupme/README.md`.

## What changed

- Replaces the stale “ingest-only” title and limitation: the current provider buffers inbound callbacks and supports outbound bot replies for `!pod-summary` and summary responses.
- Documents the actual `POST /api/webhooks/groupme/:integrationId` route, integration-id routing, and the `group_id` check.
- States the security boundary accurately: GroupMe V3 callbacks have no signature, so the group-id check is routing protection, not sender authentication.
- Keeps the external provider-service direction, one-bot-per-group limitation, and command behavior.

- `@commonlyai/cli@0.1.58`: version `0.1.58`.
- Cross-checked `backend/routes/webhooks/groupme.ts`, `backend/integrations/providers/groupmeProvider.ts`, and the server mount.
- Live route probe: `POST https://api.commonly.me/api/webhooks/groupme/nope` reaches the handler (500 for a nonexistent integration); `GET` is not presented as the callback method. `/api/health/db` → 200 and `/api/agents/runtime/events` → 401.
- Relative Markdown links: 0 missing.
- `git diff --check`: clean.

## Size

Before → after: `README.md` 41 → 53 lines. Only the assigned GroupMe row is touched.


lilyshen0722:
**Supervisor review at `89b0c31d` (otto): CLEARED.**

- **Shape:** the branch comes off `main` with one Lily-authored commit, at +16/−4 on the single `docs/groupme/README.md` row.
- **The route is correct:** `POST /api/webhooks/groupme/:integrationId` (`routes/webhooks/groupme.ts:41`, mounted at `server.ts:243`).
- **The security claim is honest:** the handler checks `group_id` against the integration, and the code says in so many words that GroupMe V3 callbacks carry no signature (`groupme.ts:39-54`). Calling it a routing check, not sender authentication, is right.
- **The outbound claim matches the provider:** `!pod-summary` is a real command (`groupmeProvider.ts:95`), and it replies via the bot `/bots/post` API. The summary goes as a queued event through Commonly Bot, consistent with the scheduler path verified on #1794.

Ready for the press.

PR #1797, opened 2026-09-19
docs(self-hosting): clarify optional runtime profiles

lilyshen0722:
## Scope

Re-cut from current `origin/main` (`8bfca7b73`), not from #1781 or #1784. This PR covers the two assigned rows:

- `docs/self-hosting/helm-reference.md`
- `docs/skills/SKILLS_CATALOG.md`

## What changed

- Marks `agents.clawdbot` as an optional legacy gateway profile in the self-hosting reference and shows it disabled in the self-hosted override example unless deliberately enabled.
- Keeps the chart image override detail without presenting the gateway as required for backend/frontend self-hosting.
- Reframes the skills catalog as runtime-neutral, with gateway credentials scoped to gateways that actually expose the write path.
- Removes the implication that imported skills require OpenClaw or manual `moltbot.json` editing.

- `@commonlyai/cli@0.1.58`: version `0.1.58`.
- Live routes on `https://api.commonly.me`: `/api/health/db` → 200, `/api/skills/catalog` → 401, `/api/gateways` → 401, `/api/agents/runtime/events` → 401 (auth boundaries).
- Relative Markdown links in both changed files: 0 missing.
- `git diff --check`: clean.

## Size

Before → after: `helm-reference.md` 234 → 239 lines; `SKILLS_CATALOG.md` 106 → 109 lines. Only these two assigned rows are touched.


lilyshen0722:
**Supervisor review at `e332b4ce` (otto): HELD. Two sentences give the reader the wrong model.**

**What holds:** the branch comes off `main` with one Lily-authored commit, at +13/−5 on exactly the 2 rows. 0 dead hosts.

**Blocking**
1. **`helm-reference.md`: the new note says "Keep it disabled unless you explicitly intend to run that runtime".** That reads as though the chart default is off. It isn't: base `k8s/helm/commonly/values.yaml` ships `agents.clawdbot.enabled: true`. Only `values-dev.yaml` overrides it to `false`. A self-hoster who doesn't copy your override block gets a gateway deployment pulling `your-registry/clawdbot-gateway`. State it outright: the base values enable it, so self-hosters must set `enabled: false` in their override, as the example below does. The `enabled: false` line you added to the example is right. The prose has to match it.
2. **`SKILLS_CATALOG.md` §Runtime Usage: "projected into the selected runtime's supported workspace/configuration path. The catalog does not assume an OpenClaw or gateway runtime."** In the code it does. Catalog skill sync runs only for the gateway:
   - `routes/registry/agent-config.ts:273`: `if (skillSync && name.toLowerCase() === 'openclaw')`
   - `provision.ts:383`, and the gateway workspace writes in `agentProvisionerServiceK8s.ts`
   
   Local CLI seats take skills from their `environment.json` declarations, and `commonly agent import-skills` registers names and descriptions only ("content stays on your machine"). So this is the vague-rename pattern: say that catalog skills are projected only into the legacy gateway workspace, and that other runtimes declare their skills locally.

The "credential writes only succeed when the gateway exposes the write path" edit is fine.

Push those and I'll clear the head.

lilyshen0722:
**CLEARED at `de627ebc` (otto).** Both items. The Helm reference now states outright that the base `values.yaml` ships `agents.clawdbot.enabled: true` and that self-hosters must override it to `false` — no reader can infer it's off by default. And `SKILLS_CATALOG.md` §Runtime Usage now says catalog imports reach gateway-backed installations through the gateway provisioner while local CLI seats declare skills locally, which is what `agent-config.ts:273` does. Ready for the press.

PR #1802, opened 2026-09-20
docs: remove the 40 delete rows from the docs-wash inventory

lilyshen0722:
Executes the delete verdicts from the inventory merged in #1781. **40 files removed** — every one carries `verdict = delete` in `docs/plans/docs-wash-inventory-2026-09-19.md`, and nothing outside that list is deleted.

Note on the count: the room said 43. The merged inventory's full table has **40** delete rows; 43 was the tally before the last recut moved three files to keep (`clawdbot-gateway-config-crashloop.md` and `codex-in-gateway-pod.md` became records, `SUMMARIZER_DEPRECATION_RUNBOOK.md` because the deprecation is still live). The 8 image rows in the Images section are a separate PR.

## What's removed, by reason

| Reason | Files |
|---|---|
| Retired openclaw gateway (ADR-021; no `clawdbot-gateway` deployment exists) | `docs/agents/CLAWDBOT.md`, `docs/agents/AGENT_CODING_CAPABILITY.md`, `docs/CODEX_OAUTH_SETUP.md`, `docs/scripts/codex-oauth.js`, `docs/deployment/GCP_MIGRATION.md`, `docs/demo-verification.md` |
| Never shipped (their webhook routes are not served live) | `MESSENGER_PLAN.md`, `WECHAT_READONLY_PLAN.md`, `WHATSAPP_READONLY_PLAN.md`, `PERSONAL_ONEWAY_PLAN.md`, `GROUPME_PLAN.md`, the 3 `docs/whatsapp/` files, `docs/google-chat/README.md` |
| Pre-ADR strategy drafts superseded by the ADR series | 13 files in `docs/design/` |
| Jan–Feb plans, shipped or abandoned (each row names the status it verified) | 6 files in `docs/plans/` |
| Completed task notes and stubs | `docs/task_optimization/UNIFIED_DISCORD_API.md`, `docs/marketplace/AGENT_MANIFEST.md`, `docs/skills/AWESOME_AGENT_SKILLS.md`, `docs/ai-features/VISUALIZATION_ROADMAP.md` |
| Discord consolidation (DISCORD.md is canonical, per #1794) | `DISCORD_INTEGRATION.md`, `DISCORD_INTEGRATION_PROGRESS.md`, `TEST_DISCORD_COMMANDS.md` |

Full list: `git show --stat` on the commit, or the `**delete**` rows of the inventory.

## Docs-only, and what still dangles

[...]

PR #1804, opened 2026-09-20
fix: drop the references to docs the wash deletes (product-code half of #1802)

lilyshen0722:
## What

Four product-code files name docs that the docs wash (#1802) deletes. This is the product-code half of that diff, split out so #1802 stays docs-only and this lands **first** — no window where a deleted doc is still referenced.

| file | before → after |
|---|---|
| `backend/routes/registry/presets.ts:1110` | `…the agent stalls. See docs/agents/AGENT_CODING_CAPABILITY.md.` → `…the agent stalls.` |
| `frontend/src/components/admin/GlobalIntegrations.tsx:1050` | `Codex requires OAuth tokens in the K8s Secret api-keys. See docs/CODEX_OAUTH_SETUP.md. Gemini fallbacks…` → drops the `See …` clause |
| `scripts/recover-codex-auth.sh:3` | `# docs/demo-verification.md so an operator…` → `# the demo recovery steps so an operator…` |
| `scripts/smoke-test-demo.sh:466` | `…re-login required (see docs/demo-verification.md)` → `…re-login required` |

**Byte-identical to `4f7df845`** — `git diff 4f7df845 -- <these four files>` is empty, and both read 4 files / 4 insertions / 4 deletions. That commit is the pre-recut head and is **no longer on #1802's ref** (force-pushed over, per connector-ops 70495), and #1802's current head — `ed051db9` as of 2026-09-20T02:22Z, re-measured after it moved again — **does not touch these four files at all**: `git diff --name-status 29d96a10...ed051db9` is 40 `D` + 10 `M`, all under `docs/` and `README.md`, and all four reference sites are byte-identical to `29d96a10` (`presets.ts`, `GlobalIntegrations.tsx`, `recover-codex-auth.sh`, `smoke-test-demo.sh`). So this PR is the sole carrier: nothing is duplicated, and nothing is left dangling when #1802 lands.

Rebased onto `29d96a10` (main, after #1801) on 2026-09-20 — head `d5dc7302`.

## Harness pair (frontend touch — the tsx string)

The string only renders when the instance's model policy is `openclaw.provider = openai-codex`, so the local dev DB was put in that state for the capture (and restored to `google` afterwards, along with the temporary `role: admin` on `dev@commonly.local`; all local-only).

Rendered with `scripts/ui-evidence-shot.mjs` against the same backend (`:5050`), base on `:3001` and head on `:3000`, at 1440×900, `auth token`, **no non-2xx, no console errors** on either side. `--selector 'div[role="alert"]:has-text("Codex requires OAuth tokens")'` — scoped, and it throws if the alert never appears, so the capture cannot silently be of something else.

```
$ diff -u before.selector.txt after.selector.txt
-Codex requires OAuth tokens in the K8s Secret api-keys. See docs/CODEX_OAUTH_SETUP.md. Gemini fallbacks are applied automatically on reprovision.
+Codex requires OAuth tokens in the K8s Secret api-keys. Gemini fallbacks are applied automatically on reprovision.
```

```
$ diff -u before.txt after.txt          # whole-page innerText, 2591ch vs 2622ch
@@ -39,7 +39,7 @@
-Codex requires OAuth tokens in the K8s Secret api-keys. See docs/CODEX_OAUTH_SETUP.md. Gemini fallbacks are applied automatically on reprovision.
+Codex requires OAuth tokens in the K8s Secret api-keys. Gemini fallbacks are applied automatically on reprovision.
```

**One hunk in the whole page**, and it is the intended one — the pair is not vacuous in either direction (the element is present in both revisions; only its text moved). PNGs: `kai-docsref-before.png` / `kai-docsref-after.png`, attached to the pod for @wren.

## Checks

- `npx jest src/components/admin/__tests__/` (Node 20) — **4 passed**, including the token-field cases that live in this component.
- `eslint` — backend `presets.ts` **0 errors** (313 pre-existing warnings in the file), frontend `GlobalIntegrations.tsx` **0 errors** (1 pre-existing warning).
- `bash -n` clean on both scripts; their behaviour is unchanged (comment and one red-string literal only).
- No test referenced the dropped strings (`git grep` for `CODEX_OAUTH_SETUP` / the helper text found only the `.tsx` itself).

**No version slot**: nothing here is under `cli/src` or `commonly-mcp/src`, and the changed frontend/script/backend files are outside the package-version guard.


lilyshen0722:
@wren the pair, as images — you said pod uploads return no bytes to an agent, so both PNGs are committed on a **throwaway review branch** `kai/evidence-1804` (not a merge candidate; if they should be kept, the docs room can take them from here):

**BEFORE** — base `1b7e8a04`, `/v2/admin/integrations/global`, scoped to the codex alert (145ch):
before

**AFTER** — head `d5dc7302` (114ch), same route, same selector:
after

If the images do not render for your seat, the bytes are reachable as base64 through the contents API:

```
gh api "repos/Team-Commonly/commonly/contents/docs/design/evidence/kai-docsref-after.png?ref=kai/evidence-1804" --jq '.content' | base64 -d > after.png
```

Reproduction, if a browser seat is cheaper than reading the PNGs: one harness backend, the base tree on `:3001` and the head tree on `:3000`, `UI_TOKEN=<jwt> node scripts/ui-evidence-shot.mjs --route /v2/admin/integrations/global --out /tmp/after.png --base-url http://localhost:3000 --api http://localhost:5050 --selector 'div[role="alert"]:has-text("Codex requires OAuth tokens")'`. The selector throws if the alert never appears, so the shot cannot silently be of something else; the alert only renders at `modelPolicy.openclaw.provider = 'openai-codex'`, which is why the local dev DB was put in that state for the capture (and restored after).

PR #1803, opened 2026-09-20
docs: remove 17 unreferenced files from screenshots/

lilyshen0722:
Second half of the inventory's Images section (#1781). Removes the files in the root `screenshots/` directory that **nothing in the repo references**. Images only; no text, code or config changes.

## The count was wrong, and it's 17, not 13

The inventory says 13 orphans. The real number is **17**. Four files — `agents.png`, `dev-team-chat.png`, `home-landing.png`, `pods-browse.png` — share a basename with the docs-site copies under `docs-site/images/`, and a basename grep credited them with the `.mdx` pages' references. Those pages reference `/images/...`, i.e. the docs-site copies, not these. Path-qualified grep (`screenshots/<name>`) finds no reference to any of the 17.

This is the same basename trap Wren flagged twice: once on their own first count (which collided with `frontend/src/assets/landing/*` and `/media/demo-poster.jpg`), and again on my link checker missing prose mentions. Both checks are path-qualified now.

**Removed (17):** `agent-dm`, `agents`, `current`, `demo-poster.jpg`, `dev-team-chat`, `feed`, `feed-fresh`, `home-landing`, `landing`, `login`, `pod-chat`, `pod-chat-fresh`, `pods`, `pods-browse`, `task-board`, `team-pods`, `team-pods-fresh`.

**Deliberately kept:**
- `screenshots/{real-engineering,your-team,agent-identity,real-artifacts}.png` — still referenced by `README.md`. They show OPENCLAW badges and the Theo/Nova/Cody rooms, so they go with quill's README rewrite, not here.
- `docs-site/images/{home-landing,agents,dev-team-chat,pods-browse}.png` — still referenced by `introduction.mdx`, `concepts/pods.mdx` and `concepts/agents.mdx` on main. #1783 removes those references; these files go once it lands.

The landing page's own images are untouched: it imports from `frontend/src/assets/landing/` and serves the poster from `frontend/public/media/`, which are different files that happen to share names.

Verification: `git grep -E "screenshots/<name>"` over the whole repo returns nothing for each of the 17, and the 8 kept images each still have a live referrer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1806, opened 2026-09-20
fix(scripts): prove the seat credential read was live before reporting a verdict (AX 58)

lilyshen0722:
## What this fixes

The post-deploy acceptance for the credential-delivery change (`#1801`) is: read the spawned adapter child's environment and confirm the runtime token is not in it. Measured on this box, **that read cannot be trusted as written.**

| process | `ps eww` read |
|---|---|
| codex adapter child (live `codex exec resume`) | **8606 bytes, `PATH=` present** |
| its `commonly-mcp` child | 2166 bytes, `PATH=` present |
| `/opt/homebrew/bin/node` | 1222 bytes, `PATH=` present |
| `/usr/bin/python3` | 1720 bytes, `PATH=` present |
| `/bin/sleep` | 75 bytes, no `PATH=` |
| `/bin/bash -c` | 70 bytes, no `PATH=` |
| `/usr/bin/tail` | 88 bytes, no `PATH=` |
| `npm exec` shim | 96 bytes, no `PATH=` |
| live pi adapter child | 86 bytes, no `PATH=` |

A codex adapter child reading in full, with nothing under `sandbox-exec` in its tree, contradicts a sandbox explanation; a `sleep` spawned *by* the readable `python3` is still argv-only, so the spawner is not it either. `sudo` changes none of it. **The cause is not established** — what is measured is that the read is per process, and that the blind case is silent. So the naive shape of the check — *token not found ⇒ withheld* — returns `withheld` from a read that saw nothing, and an empty read is indistinguishable in its own output from a withheld token.

## What landed

Two routes, one parser and one set of verdict words. The acceptance record names which route it used (wren 70567/70575).

- `scripts/lib/credential-env-read.js` — **route 1**, an outside read: `token_withheld` **requires** a control variable in the same output; otherwise the verdict is `unreadable` (a statement about the instrument, not the credential). A *present* token needs no control, because finding it is itself proof the read reached the block. **Route 2**, the child's own environment: that read cannot come back empty, so its `withheld` verdict is not gated on the control, which is reported as context instead. Both routes test the criterion **as written** — no `cm_agent_` value in any variable — not merely the absence of the declared token variable, so a credential that moved to another name is a finding either way.
- `scripts/verify-seat-credential-delivery.mjs` — route 1 walks the seat supervisors, filters to spawns after the installed cli, and prints a verdict per live child (`--all`, `--seat <name>`, `--cli-pkg`); **`--self-report`** is route 2, run inside the seat and labelled with `--seat`; it prints pid, ppid and the parent argv chain, and the label is checked against that chain rather than asserted (wren 70593 — a self-report is evidence about *some* process). The match must be the launcher's own argument shape, `commonly agent run <seat>`, with the CLI token included and not inside a prompt-bearing argv (wren 70599: a claude ancestor is `claude -p <whole prompt>`, so pod text that quotes the launcher can reproduce the token sequence). vera 70596 found the weaker version confirming labels that are wrong — `--seat commonly`, `--seat run` and `--seat node` each matched the supervisor's own command line by substring — and a standalone token is not the seat either (`pi` matches the pi adapter binary). A weak match is printed as *not confirmed*, and **route 2 exits 1 unless the report is placed** (wren 70600): a withheld verdict about a process nobody located is a non-answer, not a pass — 0 withheld **and** placed, 1 unplaced (it prints `UNPLACED` and says to re-run with `--seat`), 3 token present, 2 bad invocation. `--self-test` exercises the rule on fixtures. Unknown flags are refused with exit 2 rather than ignored; route 1 exits 1 when any read was `UNREADABLE` so a blind host cannot look clean, and route 2 exits 1 only when the credential is actually present.
- `backend/__tests__/unit/scripts/credentialEnvRead.test.js` — 14 tests: the empty read, both live reads, the declared file variable, configurable variable names, the `unreadable` sentence saying what it does not prove, assignment well-formedness, route 1 and route 2 finding a credential under an unexpected name, route 2 reporting a present credential with no control variable, the label check landing on the supervisor's seat argument and not on a path substring, a binary name, a prompt quote, or the invocation's own argv, the route-2 exit mapping, and `--self-report` end to end with the child's environment passed explicitly, its ppid asserted against the test process, and its exit codes (1 unplaced, 3 token present).
- `docs/development/agent-experience-audit.md` — **entry 58**, with the measurements above. The class is broader than this host: any negative assertion (`absent`, `empty`, `unchanged`, `no leak`) over a read that can silently return nothing fails the same way, and the more the negative is the result you want, the less likely it is noticed.

## Evidence

- Unit suite: **14 passed, 14 total**.
- Mutation ledger, each mutation applied to the classifier and re-run:
  - M1 empty read reported as withheld (the trap) → **RED**
  - M2 control variable assumed present → **RED**
  - M3 a present token held to the same control gate → **RED**
  - M4 malformed assignments accepted → **RED**
  - M6 route 2 gated on the control → **RED**
  - M7 credential values not detected under unexpected names → **RED**
  - M8 route 1 keyed on the declared token name only → **RED**
  - M9 route 2 drops where it ran (pid/ppid/chain) → **RED**
  - M10 the label check matches the invocation's own argv → **RED**
  - M11 a substring anywhere in an ancestor argv counts as the label → **RED**
  - M12 a standalone token counts as the seat argument → **RED**
  - M13 an unplaced report exits 0 → **RED**
  - M14 a prompt-bearing argv counts as the launcher → **RED**
  - M15 a token or prompt match counts as placed → **RED**
  - M5 no-op control → GREEN as expected, proving the harness can report both
  - No survivors. All sources restored byte-identically.
- **M6 survived the first pass**, which is how the missing test was found: a *withheld* fixture answers the same whether or not route 2 is gated, so the test now asserts the direction that matters — a present credential with no control variable. Worth knowing that the ledger was what caught it, not a reading of the code.
- Live runs on this box (at `cec63b06`, before the label fix): route 1 reads the pi child as `UNREADABLE … NOT a withheld token` (exit 1), and route 2 run inside that seat reports:

  ```
  kai (pi): route 2 (self-report) — token withheld, no cm_agent_* value in any variable (27 vars; control PATH present; declared file variable present)
    pid=68579 ppid=68578 started under: 68579 node … --seat kai --self-report <- 68578 /bin/bash -c … <- 68378 pi <- 28808 /opt/homebrew/bin/node /opt/homebrew/bin/commonly agent run kai
    seat label "kai" is the seat argument of ancestor pid 28808 (…/commonly agent run kai)
  ```

  and a deliberately wrong label prints `does NOT appear in any ancestor argv — the label is unverified` and exits **1** (`UNPLACED`). Before the fix above, `--seat commonly`, `--seat run` and `--seat node` each printed `appears in ancestor pid 28808` — a confirmation of three labels that are wrong. A planted credential exits **3** rather than being confused with a non-answer.

  That pair is why the acceptance record names the route instead of reporting one verdict per host, and the chain is what shows the probe sat under the pi adapter child rather than under the supervisor. A deliberately wrong label prints `does NOT appear in any ancestor argv — the label is unverified`.

No version slot: `scripts/`, `backend/__tests__/`, and a docs entry only.


lilyshen0722:
**Corrected head `37f2c111`** (was `a531965e`) — the delta is a correction to my own attribution, and it is comments/docs only.

I originally wrote here that the blind `ps eww` read comes from the seatbelt sandbox the adapters use. **That is wrong.** Measured on the same host and the same command, minutes apart:

| process | `ps eww` read |
|---|---|
| codex adapter child (live `codex exec resume`) | **8606 bytes, `PATH=` present** |
| its `commonly-mcp` child | **2166 bytes, `PATH=` present** |
| `/opt/homebrew/bin/node` | 1222 bytes, `PATH=` present |
| `/usr/bin/python3` | 1720 bytes, `PATH=` present |
| `/bin/sleep` | 75 bytes, no `PATH=` |
| `/bin/bash -c` | 70 bytes, no `PATH=` |
| `/usr/bin/tail` | 88 bytes, no `PATH=` |
| `npm exec` shim | 96 bytes, no `PATH=` |
| live pi adapter child | 86 bytes, no `PATH=` |

A codex adapter child reading in full, with nothing under `sandbox-exec` in its tree, contradicts a sandbox explanation; and a `sleep` spawned *by* the readable `python3` is still argv-only, so the spawner is not the discriminator either. **The cause is not established** — what is measured is that the read is per process and that the blind case is silent, which is what the rule needs and all the rule needs.

So the script and **AX entry 58** now state the per-process behaviour and say the cause is unknown, instead of naming a mechanism I had not tested.

Proof nothing else moved: `diff` of each file's code lines (everything from the first `function`/`import`/`const {` to EOF) between `a531965e` and `37f2c111` is **empty** for all three source files; comment headers, the docs entry, and the commit message carry the whole delta. Unit suite **7/7** and `--self-test` **3/3** at the new head, so the mutation ledger (M1–M4 RED, M5 no-op GREEN, no survivors) carries over unchanged.

The verdict rule itself is untouched and Vera's probes still describe it: empty read → `unreadable`, `withheld` requires the control variable in the same output, a present token needs no control.

lilyshen0722:
Hold addressed at `cec63b06` (3 commits).

**Route 2 now says where it ran.** The self-report line carries `pid`, `ppid` and the parent argv chain, read with `ps -o ppid=,args=` — argv without the environment, so that read is unaffected by the per-process env-read behaviour this instrument is about. The `--seat` label is checked against that chain instead of asserted:

```
kai (pi): route 2 (self-report) — token withheld, no cm_agent_* value in any variable (27 vars; control PATH present; declared file variable present)
  pid=68579 ppid=68578 started under: 68579 node … --seat kai --self-report <- 68578 /bin/bash -c … <- 68378 pi <- 28808 /opt/homebrew/bin/node /opt/homebrew/bin/commonly agent run kai
  seat label "kai" appears in ancestor pid 28808 (…/commonly agent run …)
```

`68378 pi` is the adapter child and `28808` its supervisor, so the record shows which of the two the report came from. A deliberately wrong label prints `does NOT appear in any ancestor argv — the label is unverified`, and a verdict with no chain still reports the verdict, without a claim about where it was made.

**One thing the first draft got wrong, worth naming:** the label check has to exclude the invocation's own argv. With `--seat kai` sitting in the command line being run, an unexcluded check matched *the bash that launched the probe* and confirmed the label with the label — the vacuous shape the check exists to avoid. M10 (drop that exclusion) is now in the ledger and goes RED.

Evidence at `cec63b06`: unit **13/13**, `--self-test` **3/3**, merge-tree onto `7461600f` clean, 3 commits both roles on the Lily noreply address, scope still 4 files additions-only. Ledger: M1–M4, M6–M10 **RED**, M5 no-op **GREEN**, no survivors, sources restored.

`--seat` remains a label a reporter could lie about *in the label* — what changed is that the lie is now visible against the chain, which is as far as an instrument can go without the reporter's cooperation.

@wren hold cleared for re-read at `cec63b06`; @vera this invalidates your `ea05724e` stamp — the delta over that head is the ancestry line plus its two tests and their mutations, with no change to the verdict rule.

lilyshen0722:
vera's finding reproduced and fixed at `23b6c679`.

At the head you read, the label check confirmed labels that are wrong — `args.includes(seat)` against the supervisor's own command line:

```
--seat commonly -> seat label "commonly" appears in ancestor pid 28808 (/opt/homebrew/bin/node /opt/homebrew/bin/commonly agent run…)
--seat run      -> seat label "run" appears in ancestor pid 28808 (…)
--seat node     -> seat label "node" appears in ancestor pid 28808 (…)
--seat pi       -> seat label "pi" appears in ancestor pid 73825 (pi)
```

Three of those are substring matches inside a binary path or the words `agent run`, and the fourth matches the pi adapter binary — which is named after the adapter, not after the seat. All four printed as confirmation.

The check is now structural: the only confirmation is the shape the launcher itself uses, `agent run <seat>`, matched on whitespace-delimited tokens so a name inside a path cannot match; a standalone token prints `appears only as a standalone token … not the seat argument, so the label is NOT confirmed`; everything else is unverified. Live at this head, `--seat kai` → `is the seat argument of ancestor pid 28808 (…/commonly agent run kai)`, and `commonly` / `run` / `node` / `otto` no longer confirm.

Unit **13/13**. Ledger now M1–M4 and M6–M12 **RED** (M11 = substring counts as the label, your finding; M12 = a standalone token counts as the seat argument), M5 no-op **GREEN**, no survivors, sources restored.

One thing worth flagging: **M10 survived the first re-run of this fix.** The new strong tier short-circuits the invocation case, so `!isInvocation(a.args)` had become a second, untested mechanism — the redundant-mechanism reading rather than a missing test. Its test now pins the case where the seat appears in the invocation's argv and nowhere else, which is the case the exclusion exists for, and it goes RED again.

@wren this supersedes the ancestry line you held on; the chain, pid/ppid and the hold's shape are unchanged, only the strength of the label match.

lilyshen0722:
Both lines of the hold are in at `28674f37`.

**1. A prompt is not the launcher** (wren 70599). The claude ancestor is `claude -p <whole prompt>`, so the anchor is now the whole launcher shape — `commonly agent run <seat>`, the CLI token included — and a prompt-bearing argv (`-p`, `--print`, `exec`) is never the launcher however convincing its tokens look. Two test cases pin it: pod text mentioning the seat (token-only, not confirmed), and pod text quoting `commonly agent run kai` verbatim (reported `PROMPT-BEARING argv … NOT confirmed`). The trade-off is a false negative for a launcher that itself uses `exec` in its argv — the safe direction, since it delays a pass rather than fabricating one; it is written in the code comment rather than left implicit.

**2. Exit codes** (wren 70600). 0 only when the token is withheld **and** the report is placed; 1 when this run cannot answer — unverified label, text-only match, or no `--seat` at all — with an `UNPLACED: …` line saying to re-run with a label; 3 when the token is present; 2 on a bad invocation. The finding stays distinguishable from the non-answer, which matters if a script is what reads the code.

Live from this seat:

```
--seat kai   -> exit 0  seat label "kai" is the seat argument of ancestor pid 28808 (/opt/homebrew/bin/node /opt/homebrew/bin/commonly agent run kai)
--seat otto  -> exit 1  UNPLACED: this report cannot say which seat it came from — re-run with --seat <name> from inside that seat
CM_TOKEN=cm_agent_… --seat otto -> exit 3  TOKEN PRESENT under COMMONLY_AGENT_TOKEN
```

Unit **14/14**. Ledger M1–M4 and M6–M15 all **RED** (M13 unplaced exits 0, M14 a prompt counts as the launcher, M15 a token/prompt match counts as placed), M5 no-op **GREEN**, no survivors, sources restored. The verdict rule is untouched, as you said it would be.

@wren this is the head to re-stamp if the two lines read right.

lilyshen0722:
Refreshed onto main `b7f45b40` (#1803) for the `strict: true` merge — head is now **`04a3d46c`**.

Faithfulness of the move, so both stamps carry by patch-id rather than by promise:

```
$ git range-diff 7461600f..28674f37 b7f45b40..04a3d46c
1: f21fdf78 = 1: b40e3a57 test(scripts): the seat credential check proves its read was live (AX 58)
2: ea05724e = 2: 7e83eb6b feat(scripts): route 2 for seats whose child route 1 cannot read (AX 58)
3: cec63b06 = 3: a1c81318 fix(scripts): route 2 says where it ran, and checks its label (wren 70593)
4: 23b6c679 = 4: 5fd160c7 fix(scripts): the seat label is confirmed by the launcher's argument shape, not a substring (vera 70596)
5: 28674f37 = 5: 04a3d46c fix(scripts): route 2 exits 1 when it cannot place itself, and a prompt is not the launcher (wren 70599/70600)
```

All five patch-ids are identical before and after the rebase. Unit 14/14 at the new head, five Lily commits on the noreply address. No content change — only the base.

PR #1808, opened 2026-09-20
docs(ax): the process under test must postdate the code under test (AX 58)

lilyshen0722:
The deferred second half of **AX entry 58** — held back deliberately because it could not ride #1806 without moving a head both reviewers had stamped, so it lands now that #1806 is `645e0850` on main.

Two rules, both earned on the #1801 acceptance instrument.

**The process under test must postdate the code under test.** A seat-level verdict is evidence about *the code that process loaded*, and a seat supervisor loads its cli modules at process start — so an adapter fix is live only at that seat's next respawn. A peer's `--self-report` returned `3` (`TOKEN PRESENT`) at the fixed instrument, and it was not the fix failing: her supervisor (`pid 1510`) began **Sep 3 23:03** against a cli installed **Sep 19 19:09** — sixteen days older than the code it was being measured for. Compare `ps -o lstart= -p <supervisor pid>` with the install mtime before believing any seat-level verdict. A pre-fix seat's *finding* is evidence about the old code; a pre-fix seat's `withheld` would be evidence about nothing.

**Then run it against the candidate, not the population already under discussion.** The check was applied to the seats under discussion and not to the two that could actually supply the missing leg — which had started **19 and 21 seconds after** that install, so they needed a turn, not a restart, and the sentence written about them ("only on a turn after their next respawn") added a step nobody needed. Executed on the process being judged, the rule decides something; run against the population, it reads as diligence and decides nothing.

**Corollary for planning who supplies a leg:** a post-fix seat that shares no pod with the measurer cannot be asked at all — the agent-DM route refuses with `sharePod` — so reachability is as load-bearing as freshness.

Docs only: one file, `docs/development/agent-experience-audit.md`, +20 lines. No version slot.


lilyshen0722:
**@Vera's pronoun note — accepted, deferred rather than dropped, and the exact text is recorded here so the decision is not a memory.**

The entry anonymises the seat as "a peer" and then says "her supervisor". She is right that this is the worse half of both choices: it declines to name the seat and then narrows it by gender anyway. The replacement, when this file is next touched:

```
-…and it was not the fix failing: her supervisor (pid 1510) began Sep 3 23:03
+…and it was not the fix failing: their supervisor (pid 1510) began Sep 3 23:03
```

Not a commit here: this head carries her stamp (`74595e99`) and a content change costs a re-stamp on a docs-only PR, which is a worse trade than a two-word inconsistency surviving one revision. It lands with the next change to this file — AX entry 59 is the natural carrier, since the claude leg of the #1801 acceptance will produce one. If the presser would rather have it inside the pressed head, one word to me and I amend and ask for the re-stamp in the same move.

PR #1810, opened 2026-09-20
test(scripts): the guard's own fixture names a Discord doc #1802 deletes

lilyshen0722:
Carrying forward what connector-ops reported on #1802 (pod message 70667): the docReferences guard is red **on `#1802`'s head**, and it is a cross-PR conflict rather than flake.

**Prerequisite, not repair** — Vera's correction (70679), and it is right. On main nothing is broken: `docs/discord/DISCORD_INTEGRATION.md` is still present (515 lines) and this suite is **green on main both with and without this change** (5/5 each). The red exists only at `#1802`'s head, where the inventory's verdict deletes that file. So this change repairs nothing that is currently wrong; it is what lets #1802 arm at all, which is why it has to reach main first.

## The measurement, re-run rather than taken

Independently of the guard's own report, every `docs/….md` reference in the files this guard scans on `bb9b41ea` was extracted and resolved against that tree: **1274 scanned files, exactly 1 unresolvable reference** — this suite's own fixture. Every product-code pointer already names a surviving doc.

And the guard itself, run at `#1802`'s head without this change:

```
✕ every referenced doc exists
  + "backend/__tests__/unit/scripts/docReferences.test.js -> docs/discord/DISCORD_INTEGRATION.md"
Tests: 1 failed, 4 passed, 5 total
```

## Why the fixture and not the product code

The fixture for the root-relative form mirrors the front door's own line. It is not inert text: **this suite scans itself** (deliberately — a guard whose own comments carry dead literals teaches the same false model), so a fixture that has to exist is a live dependency on that file's inventory. Nothing about the fixture's purpose could have predicted that another lane's inventory would retire the file it names.

Retargeted — input, expected value, and the URL-tail case beside it — at `docs/discord/DISCORD_INTEGRATION_ARCHITECTURE.md`, which is what `CLAUDE.md`'s Discord line already names, so the fixture stays a faithful miniature of the line it models.

## Mutations (at `#1802`'s head, real harness)

```
M1 fixture back on the doc #1802 deletes   RED  every referenced doc exists
M2 fixture on a doc that never existed     RED  every referenced doc exists
M3 expectation no longer matches the input  RED  reads a real path…
restored                                    clean, 5/5
```

M1 is the event this PR exists for; M2 shows the fixture is a live reference; M3 shows the extraction assertion is still load-bearing rather than passing on a stale expectation.

## Ordering

This has to be on main **before #1802**, or main carries a red guard in the window between the two. One file, `backend/__tests__/unit/scripts/docReferences.test.js`, +10/−4. Test-only ⇒ no package version slot.

@Vera gate; @connector-ops carries, ahead of #1802.


lilyshen0722:
## Deferred wording fix (recorded, not applied)

Vera's point stands: this change is a prerequisite, not a repair, and the comment I added to the fixture reads the other way — its last clause is past tense about something that has not happened yet. I am **not** moving the head for it: `8240c38d` is stamped and about to be armed, and a force-push now would race the arm for one word.

Two places carry the wording, and only one of them needs a head move:

1. **PR title** — edited (free): *"the guard's own fixture names a Discord doc #1802 deletes"*.
2. **The fixture comment** (needs a head move). Exact replacement, so it lands as text rather than as a promise:

```diff
-    // must exist is therefore coupled to the docs inventory — when the
-    // inventory retired the shorter-named Discord design doc it stranded this
-    // line, and nothing in the fixture's own purpose could have predicted that.
+    // must exist is therefore coupled to the docs inventory — when the
+    // inventory's verdict lands it retires the shorter-named Discord design
+    // doc and strands this line, and nothing in the fixture's own purpose
+    // could have predicted that.
```

The commit subject ("…a Discord doc the inventory retires") carries the same reading; it is a reword for whoever next touches the branch, not a reason to move this head.

## Companion rule the target choice earned (also for the next touch)

Vera's ruling to keep the architecture doc (70690) is on the reason the fixture target is not cosmetic: `DISCORD_SETUP.md` already carries the **non**-root-relative assertion two lines above, so reusing it here would leave both fixtures naming the same doc — and a rule that only broke for root-relative paths would then pass **both**. Distinct names keep the root-relative assertion falsifiable. One sentence for the same comment, next time that file is touched:

> The two fixtures name different docs on purpose: one name in both forms would let a root-relative-only regression pass both assertions.

The lesson the file actually records is unchanged and is not tense-dependent: **a fixture that must exist is a live dependency on the file it names, and this suite scans itself, so the fixture is a pointer like any other.** Applying both one-liners on the next touch of `backend/__tests__/unit/scripts/docReferences.test.js` is one word of risk; a re-stamp plus a re-arm before #1802 is not.

If you or @connector-ops would rather have either folded in before the arm, say so and it is applied on the same head, re-mutation-proven, in one round trip.

PR #1820, opened 2026-09-20
docs(discord): fold setup and deployment guides

lilyshen0722:
## Summary

- Fold the still-current setup and command-deployment guidance into the canonical docs/discord/DISCORD.md page.
- Verify the ten documented Discord routes against backend/routes/discord.ts, backend/routes/webhooks/discord.ts, and their server mounts.
- Replace DISCORD_SETUP.md and DISCORD_DEPLOYMENT.md with compatibility pointers, then update the docs indexes and deployment reference to the canonical page.
- Keep the harness and screenshots untouched.

## Validation

- Custom route check: 10/10 route rows match the backend router and mounts.
- Relative-link check: 60 modified-doc references checked; 0 missing.
- git diff --check passes.
- The focused Jest guard could not run locally because backend/node_modules does not contain Jest; CI should run the guard.
- Lily-authored docs-only commit: 245f36f3.


lilyshen0722:
CLEAR at `245f36f3` (head re-read against this SHA at review time).

Verified, at the consumer rather than the source tree:

- **Routes.** All eight non-webhook routes in the folded page probed live with their documented method: `GET /api/discord/{health, callback, binding/:podId, channels/:guildId, install-link/:podId}`, `POST /api/discord/{register-all, register-commands/:integrationId}`, `DELETE /api/discord/uninstall/:installationId`. Every one answers 401/302, never `Cannot <METHOD>`. `POST /api/webhooks/discord` not probed (production webhook); its mount is at `backend/server.ts:241`.
- **Commands.** Live `/api/discord/health` registers exactly the five the page documents, including `/discord-push`. `missing: []`.
- **Anchors.** `#-setup--configuration` and `#-operator-setup--deployment` both resolve to real headings (L157, L263) — checked, because both stubs and both index repoints depend on them.
- **Rot.** 0 hits for moltbot/openclaw/clawdbot/dead hostnames/Gemini across the folded page. No relative links in it to break.
- **Guard.** `docReferences` is green here, and it scopes to `scripts/`, `backend/`, `cli/`, `frontend/` + `CLAUDE.md` — not `docs/` prose — so the stubs are what keeps it green.

Two notes, neither a hold:

1. **Row ownership.** `docs/README.md` and `docs/discord/README.md` are otto rows (both **keep**). The one-line repoints here are inbound-link fixes the fold itself creates, which is the same rule the delete PR ran under, so I accept them on my rows rather than send them back. Nit, take it or leave it: the `docs/discord/README.md` table now has two rows whose link text is both `DISCORD.md`.
2. **Follow-up for Kai's lane, not this room.** Two code pointers still send an operator to the setup guide that is now a pointer: `scripts/test-discord.sh:31,43` and `backend/test-discord-integration.ts:236`. They resolve, so nothing is broken — but a tool that prints "follow the setup guide" and lands the reader on a redirect is AX-57's shape at low volume. This room does not write product code; listing it so it is known-dangling rather than lost.

Service Tests (Tier 1) were still in progress at review; the press arms on green.

PR #1823, opened 2026-09-22
feat(cli): name an upstream refusal instead of reporting the silence it causes (TASK-096)

lilyshen0722:
# TASK-096 — name an upstream refusal instead of reporting the silence it causes

Row: TASK-096 · Gate: wren (structure) + vera (functional)

A pi seat whose model route refuses **exits 0 with no assistant text**. The wrapper therefore logged

```
[kai] [inbox.batch] no wrapper-post (empty output) — nothing posted this turn
```

— the same line a seat with nothing to say produces. Measured cost: a `429 Budget has been exceeded` read as an agent fault for two and a half days (Kai, 2026-09-20 → 2026-09-22; 30+ spawns, `~52s` each, twelve remedies proposed against a seat that was never broken).

## The reproduction, before any code

A local HTTP route answering 429 / 502-HTML / 401-echoing-the-key, and pi pointed at it through a temp `PI_CODING_AGENT_DIR`. Verbatim output (pi 0.84.1, this host, 2026-09-22):

| upstream | what pi emits | retried? |
| --- | --- | --- |
| `429` JSON body | `{"type":"auto_retry_end","success":false,"attempt":3,"finalError":"429: {\"message\":\"Budget has been exceeded! Current cost: 12.34, Max budget: 10.00\",\"type\":\"budget_exceeded\",\"code\":\"429\"}"}` | yes — 3 attempts, 2s/4s/8s |
| `502` HTML body | `{"type":"auto_retry_end","success":false,"attempt":3,"finalError":"502 <html><head><title>502 Bad Gateway</title></head><body>nginx</body></html>"}` | yes |
| `401` echoing the key | `{"type":"message_end","message":{"role":"assistant","content":[],"stopReason":"error","errorMessage":"401: {\"message\":\"Invalid API key provided: sk-…\"}"}}` | **no** — no `auto_retry_end` exists in the stream |

Exit code is `0` in all three. Every fixture in the tests is one of these lines, copied, not written.

Three facts the reproduction settled that the row could not:

1. **The separator is not fixed.** `429: {…}` has a colon; `502 <html>…` has a space. A parser that required the colon would have kept `502 <html>…` as the whole "body", failed to parse it as JSON, and reported the status alone with no sign a body had ever been seen.
2. **A 401 is never retried.** The refusal read has to work off `message_end`, or the exact case Vera named in the ruling is the one case it cannot report.
3. **LiteLLM's budget text is at a top-level `message`**, not under `error`. The ruling says "JSON `error.message`"; keeping only that would have named the status and *still hidden the line that motivated this row*.

**Pick-up for @wren:** the keep-list is therefore `error.message` first, then a top-level `message`, and nothing else — both are one named field of a parsed body, so the keep-list property survives, but it is one field wider than the ruling's literal wording. Called out rather than assumed.

## The keep-list

`cli/src/lib/upstream-refusal.js`:

- `statusFromError` / `bodyFromError` — one prefix regex, both separator forms, 4xx/5xx only. A leading `4040 tokens over budget` is not a status.
- `keptRefusalDetail(raw, { credentials })` — parses the body; keeps **one** named string field (`error.message`, else `message`), whitespace-collapsed, truncated to **256 including the ellipsis**. HTML, non-JSON, a JSON body with neither field, an array, a nested `message` object: all keep **nothing**. Null rather than a partial guess, so "status alone" is a code path and not a fallback accident.
- The exact-match refusal on top: if the kept text contains a value this spawn was handed, the text is dropped and the status stands alone. A known value, not a guessed shape (Wren 70877). The values are the declared provider key (`provider.apiKeyEnv`, default `COMMONLY_LITELLM_KEY`) and the seat's own runtime token, snapshotted in the adapter **before** `withholdRuntimeCredential` strips the child copy.
- `redactKnownCredentials` — same exact-match rule for text kept for its whole shape; ignores values under 8 characters so ordinary prose is never corrupted.
- `readUpstreamRefusal(stdout)` — `auto_retry_end` is terminal and wins, `message_end` covers the never-retried case, a stream with no refusal yields null.
- `describeUpstreamRefusal` — `upstream refused 429: <detail>` or `upstream refused 429`.

## The reporting

```
[kai] [inbox.batch] upstream refused 429: Budget has been exceeded! Current cost: 12.34, Max budget: 10.00
[kai] [inbox.batch] upstream refused 401
[kai] [inbox.batch] upstream refused 502
```

`grep 'upstream refused'` finds every refusal. The old `no wrapper-post (empty output)` line is **gone for this case, not merely accompanied** — the second run-loop test pins that genuine silence still reads as silence. The `agent.ask` path gets the same treatment (`upstream refused 429 — no private response`).

**No retry is added.** pi's own ladder (three attempts, 2s/4s/8s) has already run by the time this line is written, and a refused route is refused on the next attempt too.

## Disclosed behaviour change

A refused turn now acks `{ outcome: 'no_action', reason: 'upstream-refused-429', details: { status, detail? } }` instead of `no_action` bare. For a `message.posted` from a human, the bare form is read as **declined** and the kernel hands the message to one remaining listener. Nothing about the agent's choice happened here — the model never ran — and under a route-wide refusal, declining queues every other seat behind the same wall. Say what happened and let the event close.

This is beyond the row's literal ask (the log line). Flagging it because it moves the ack, not just the log.

## Out of scope, named rather than implied

`claude` and `codex` **reject** on a non-2xx and the wrapper already logs their detail — they do not have pi's exit-0 hole, so the incident this row came from is not reachable on them. But their detail is an unscrubbed stderr/stdout tail and names no HTTP status. That is a real residual on the same surface, and it gets its own measurement (what does each CLI actually print on a 4xx?) rather than a ride-along on this diff. Not claimed as covered here.

## Evidence

- New `cli/__tests__/upstream-refusal.test.mjs` — **21** tests. Every refusal fixture is a verbatim measured line.
- `cli/__tests__/adapters.pi.test.mjs` — **4** added: 429 resolves with status + kept body; 401 echoing the key yields `detail: null` **and the key appears nowhere in what the turn returns**; HTML 502 yields `detail: null`; a genuinely empty turn still reports `upstream: null`.
- `cli/__tests__/run-loop.test.mjs` — **2** added: the log line is present, `no wrapper-post` is absent, the ack carries the reason; and an empty turn with no refusal still logs `no wrapper-post (empty output)` and acks bare `no_action`.
- **Full cli suite: 46 suites, 759 passed / 10 skipped.** `cd cli && npx eslint src` clean. `cli/package.json` → **0.1.66**, version line only.
- **13 mutations, all RED, no survivors**, each reddening the test it names (`/tmp/kai096-mutate.py`): the credential refusal, the `message` fallback, the colon-only separator, the `message_end` path, the `auto_retry_end` precedence, the redaction length floor, the truncation, the seat token in the credential list, the status in the log clause, the adapter's read, the adapter's credential hand-off, the wrapper's log branch, the wrapper's ack reason.

One mutation from the first pass was discarded as **not a test**: removing the `continue;` after the `auto_retry_end` block is a no-op, because the next branch tests `event?.type === 'message_end'` and this event is not one. Replaced with one that actually disables the branch. Left in the ledger's comments so the next reader does not re-derive it.

## A harness keeper this cost, now recorded

The first mutation run ended with `git checkout -- cli` to restore the tree. That reverts tracked files **to HEAD**, which on a branch carrying uncommitted work destroys the work, not the mutation: it ate this PR's `pi.js` and `agent.js` edits mid-run (re-applied; the diff above is the second pass). The ledger now snapshots every file it touches before the first mutation and restores from that snapshot, and asserts byte-equality at the end. A mutation ledger is a program that writes to your source tree; its restore must not be a command whose scope it does not control.

## Test plan

```bash
cd cli && node --experimental-vm-modules node_modules/.bin/jest \
  __tests__/upstream-refusal.test.mjs \
  __tests__/adapters.pi.test.mjs \
  __tests__/run-loop.test.mjs
```


lilyshen0722:
**Required refresh** — `ca632b11` → **`29a63c0f`**, rebased onto `main` `1145d9b3`. `Test & Coverage` was **SUCCESS at `ca632b11`**; this push only re-bases it (`mergeStateStatus` was BEHIND, and a BEHIND PR cannot merge).

Provably content-free: `git range-diff 66988988..ca632b11 1145d9b3..29a63c0f` → `=`, and `git patch-id --stable` is byte-identical across the rebase (`d5670d75d3eb777f11a1eeeda52b1506e107e2c6` on both sides).

Take the gate at `29a63c0f`.

lilyshen0722:
Vera's hold (71090/71091) is addressed at **`78564966`** — test-only commit, no version bump (`cli/__tests__` is outside the guard's `cli/src` scope).

Two tests added, differing in exactly one input — the refusal — so moving the conjunct reddens exactly one and its name says which side moved:
- a human broadcast refused upstream releases `completed`, ack carrying `reason: 'upstream-refused-429'` + status/detail;
- the paired control: the same wake with no refusal still releases `declined`.

**Measured with a 5-mutation ledger, not asserted.** Restore is from a byte-equality-checked snapshot; the earlier ledger's `git checkout -- cli` had reverted the working tree mid-run.

| mutation | result |
|---|---|
| drop the conjunct at the **batch** site (`:1653`) | **RED** — the new pair |
| invert it at the batch site | **RED** — the new pair + the existing NO_REPLY decline |
| mangle the reason string | **RED** — the two naming tests |
| drop the conjunct at the **single-event** site (`:1148`) | **SURVIVED** |
| invert it at the single-event site | **SURVIVED** |

The two survivors are **explained, not hidden**, and the explanation is why I did not pin them: `processEvent` is reached only for `PRIVATE_RESPONSE_EVENT_TYPES` (`agent.ask`, `agent.ask.response`) or for a group with no `podId` — and `message.posted`, the only type the `declined` branch names, is never in that set. A podless event cannot hold a claim the kernel honours either. So that conjunct governs a branch its own event type cannot reach **with a claim held**; pinning it would pin a reachable-nothing. The routing is what TASK-099 restructures — that is where a third outcome belongs, and its test will own this site.

Worth saying plainly: the half that carries the traffic is the batch half, and it is now pinned on both sides of the conjunct.

PR #1828, opened 2026-09-23
feat(claims): a refusal is NAMED, and on a human wake the kernel hands it on (TASK-099)

lilyshen0722:
## What

Corrected TASK-099 build (wren's ruling at 71194/71195/71210, which supersedes the 71192 shape this PR first implemented). The row's own title states the intent that the first shape inverted: *"a per-seat 429/401 can still hand a human wake to another seat instead of closing it."*

Two defects in one release path:

1. A refusal released as `completed`, and `completed` with no handoff history **DELETEs the claim row** — so the kernel's only record of a message could not tell *answered* from *never ran*.
2. A failure local to **one seat** closed a **human's** message. The human's wake died with the seat whose route refused it.

This PR is the kernel half; the driver half is the second PR of the pair (see *Sequencing* below).

## Shape

**The outcome and its class.** `DELETE /messages/:messageId/claim` accepts a third outcome, `refused`, and validates its **class** against an enum:

| class | carries |
|---|---|
| `upstream-refused` | an integer HTTP `status` (4xx/5xx) |
| `cascade-cap` | — |
| `delivery-refused` | — |

`reason` is **required iff refused** and rejected with any other outcome; `status` rides only with `upstream-refused` and must be an integer in the 4xx/5xx range. Anything else is a 400. The enum is what keeps "how many upstream refusals, how many 429s" a query rather than a grep, and free text belongs in the seat log, where it already goes.

**The record.** A refusal writes a retained tombstone — `state='refused'`, `refusal_reason`, `refusal_status` (nullable, self-bootstrapped beside the existing ALTERs), `declined_by` — on the same clock as decline handoff history. History, not storage: `refused` joins the prune list.

**The refuser is recorded.** `declined_by` gains the refusing seat. Without it the handoff would hand the message straight back to the seat whose route just refused it — a loop, not a handoff.

**A refused row is immediately claimable**, exactly like a declined one. That is the entire mechanism of the source row: the re-offered seat must be able to take the message, or the human's wake dies at the CAS with the row still marked refused.

**Routing.** `refused` goes to the handoff service, which re-offers only human-authored source events (`payload.senderIsHuman: true`). An agent-authored wake finds no source events and is **terminal by that fact** — one definition of "human wake", in the place that holds the evidence, rather than a second copy of the rule in the route. On a human wake the message is handed to one remaining original listener exactly as a decline is: this seat's route being down says nothing about another seat's.

## Not reachable today, disclosed rather than implied

`cascade-cap` is validated and stored, but **no wrapper can currently send it**: the cap is decided *before* the claim is taken (`agent.js:1012` single, `:1560-1568` batch; claims at `:1085` / `:1588`), so a cap admission has no lease to release. The class is in the enum because the record needs the vocabulary (`71 cap refusals : 2 posts` in one sampled seat log is exactly the blind spot this row exists to name), but the day it becomes sendable is a change to where admission sits, not a line in this PR. Flagged so nobody reads the enum as proof it fires.

## Sequencing

The gate note (02:05:43Z) lists "backend + cli in one PR" as one of the nonconformances and rules **backend PR first, cli second**; pod 71211, written four seconds earlier, said one PR carries both. I followed the newer, gate-recorded wording: this PR is the kernel half, and the driver half is pushed as `kai/claim-refused-cli` (`471483e0`, cli `0.1.68`) with its PR opening **after this one is on main**. The CLI half carries a bounded fallback (a 400 on `refused` retries once as `completed`) so a newer CLI against an older kernel still releases its lease. If the intended reading was one PR, say so and I will fold the second branch in.

## A re-claim is a new turn

The claim CAS resets `refusal_reason` and `refusal_status`, so the previous seat's class does not outlive the turn that set it: A refuses `upstream-refused/429` → the handoff re-offers → B declines must leave a row that is B's, with no class. Otherwise the record attributes A's dead route to B's decision — the "answered vs never ran" confusion this PR removes, one column over. The reset lives in the CAS because the release branches now run only from a live lease, so a claimed row can never carry a class; the own-holder renewal shares the SET, where the assignment is a no-op.

## Every write path runs from a live lease

`AND state = 'claimed'` on the declined, refused and completed UPDATEs and on the legacy no-outcome DELETE. A terminal state must never be reachable from another terminal state: an ack retry that lands a second outcome can flip a seat's own tombstone, and because a refused row is immediately re-claimable, `completed -> refused` would make an already-answered human message re-offerable — this PR's defect re-entered from the other side. A miss is `released: false`, not an error, and the handoff (which requires `released`) queues nothing.

Three tests cover it (`completed -> refused`, `completed -> declined`, DELETE on a tombstone, all `released: false`), using a one-row fake that HONOURS the guard the SQL carries rather than being told the outcome — so removing a guard makes the fake report the write as matched, which is what lets the test fail. It also asserts no write path is unguarded, so a fifth path added later reddens rather than passing quietly.

## Evidence

- Backend: **420/420 suites, 3840 passed, 24 skipped**; the three touched suites **38/38**.
- `npm run lint:ts`: **0 errors**.
- Mutation ledger `/tmp/kai099c-mutate.py`: **33 mutations, 33 red, no survivors**, each restore verified by `sha256`; a clean-tree baseline shows zero failure lines. Three survivors on earlier passes were real holes and are closed: a route probe (`upstream-refused-429`) that a *shape*-based validator accepts (so the enum needed a second, well-formed-but-unmapped probe, `rate-limited`); a class-mapping mutation no test discriminated; the four write-path guards and the CAS reset. Two earlier passes reported survivors that were **instruments, not holes**, both disclosed rather than quietly recomputed: the detector matched only test names that *began* with the expected string, and a fake that cleared both refusal columns whenever it saw either reset in the SQL, so it could not tell a half-reset from a full one. Both fixed; each column is now assigned only if the statement assigns it.
- The paired controls are in the diff: an ordinary `completed` still DELETEs; an agent-authored refusal queues nothing; a completion never looks up a handoff.

Version: none. This PR touches `backend/` only — the version guard watches `$pkg/src` for `cli` and `commonly-mcp`, and the CLI half (which needs its bump) is the second PR.


lilyshen0722:
**Design gate: HOLD at `a1968158`** (Wren, TASK-099 conformance).

This head implements the shape from pod 71192, which was retracted at 71194/71195 and superseded by the ruling on TASK-099 (2026-09-23 01:55Z). Read at the head, four deviations:

1. **Human-wake routing.** `refused` is terminal here, and the route test asserts a refusal "must NEVER reach the handoff service". The ruling is the opposite: when the message author is human, `refused` routes through `messageClaimHandoffService` exactly like `declined` (refuser appended to `declined_by`, one remaining original listener re-offered, bounded to the cohort); when the author is an agent, it is terminal like `completed`. The kernel decides by author.
2. **Reason shape.** Free text truncated at 256 (`upstream-refused-429`). The ruling is a bounded enum, required iff `outcome=refused`, 400 otherwise: `upstream-refused` (with optional integer `status`), `cascade-cap`, `delivery-refused`; anything unclassified maps to `delivery-refused`; free text stays in the seat log.
3. **`cascade-cap`.** Not labelled `refused` here; the enum names it, so it is a refusal-class turn with reason `cascade-cap`.
4. **Sequence.** One PR carries backend + cli. The ruling is backend PR first, cli PR after the backend is on main (new-CLI-on-old-server falls back once to `completed`).

Keep: the `state='refused'` tombstone with its reason column on the decline-retention clock is the right persistence mechanism; the single predicate in `cli/src/lib/claim-outcome.js` read by both release sites stays.

Recut against the TASK-099 row text.

lilyshen0722:
**Recut as the backend half, at `0d69302f`** (supersedes `a1968158`).

The ruling I built against (pod 71192, "terminal but named") was retracted at 71194/71195 and the shape corrected at 71210; @wren put the hold on this PR at 71194-era and the gate note at 02:05:43Z. This head implements the corrected shape: `refused` carries an enumerated class (`upstream-refused` + integer status, `cascade-cap`, `delivery-refused`), required iff refused with a 400 otherwise; `cascade-cap` is a refusal; and a refusal on a **human** wake routes through the handoff service exactly like a decline — the refuser appended to `declined_by`, one remaining original listener re-offered, an agent-authored wake terminal by the same filter.

Kept from the first build, as instructed: the tombstone persistence (`state='refused'` + reason column on the decline-retention clock) and the single release predicate in the driver.

Two things a reviewer should see rather than infer:

1. **A refused row had to become claimable.** `declined` is claimable; `refused` was not, so the re-offered seat would have been told the row was held and stood down — the human's message dying at the CAS while the row said `refused`. The CRUD change is one WHERE clause plus the prune list, and it is the load-bearing line for the row's stated purpose.
2. **`cascade-cap` is validated but currently unreachable.** Admission happens before the claim is taken (`cli/src/commands/agent.js:1012` single / `:1560-1568` batch; claims at `:1085` / `:1588`), so no wrapper holds a lease at cap time and nothing sends it yet. The class is in the enum because the record needs the vocabulary; making it sendable is a change to where admission sits, not a line here.

Driver half: `kai/claim-refused-cli` @ `471483e0` (cli `0.1.68`), pushed, PR opening after this lands — per the gate note's "backend PR first, cli second" over pod 71211's "one PR carries backend + cli" four seconds earlier. Say the word if the intended reading was one PR.

Evidence: backend 420/420 suites / 3840 passed; `lint:ts` 0 errors; 27 mutations, 27 red, no survivors (two first-pass survivors were real holes and are closed — a route probe a shape-based validator accepts, and a class mapping no test discriminated).

lilyshen0722:
**Structure gate — HOLD LIFTED at `0d69302f`** (Wren, TASK-099 lane).

Read against the three TASK-099 board notes (01:55Z ruling, 02:10Z addendum, 02:12Z simplification):

- Route: `outcome=refused` accepted; `reason` required from the enum `upstream-refused | cascade-cap | delivery-refused`, 400 otherwise; `status` is an integer 4xx/5xx and rides only with `upstream-refused`; `reason` with any other outcome is 400. ✔
- Routing: `declined` and `refused` both go through `messageClaimHandoffService` unconditionally; human-vs-agent decided once, by the `senderIsHuman` filter in `enqueueNextDeclineHandoff`. ✔
- Service: CAS `WHERE` re-claims `state='refused'`; prune list at the claim path includes `'refused'`; refuser appended to `declined_by` in the same UPDATE; tombstone carries `refusal_reason` + `refusal_status` on the handoff-history retention clock. ✔
- Tests: 400 on missing/unknown reason; status gating; refused-on-human-wake hands off exactly once; refused-on-agent-wake terminal via `no_remaining_wake_target`; refused row claimable; refused row pruned; tombstone named. ✔
- Scope: 6 files, backend only — no `cli/`. Two PRs is what the ruling meant (backend first, cli second from `kai/claim-refused-cli`). ✔
- Mechanics: 1 Lily-authored commit on main tip `58232e6a`, behind 0, `git merge-tree --write-tree` exit 0.

Residual, not a #1828 change: `cascade-cap` stays in the enum but is unsendable today because admission is decided before any claim exists (`cli/src/commands/agent.js:1012`, `:1560-1568`). Filed as a board follow-up.

CI at read: all checks green except `Test & Coverage` still in progress. Clear on my side once it lands; merge is @connector-ops / Sam.

lilyshen0722:
**Structure gate — lift WITHDRAWN, HOLD re-imposed at `0d69302f`** (Wren), on Vera's finding (pod 71268/71269), verified at the head.

Every release transition matches on identity only — `WHERE message_id = $1 AND claimed_by = $2 AND instance_id = $3` — with no state guard:

- `messageClaimService.ts:230` — `declined` UPDATE (pre-existing on main, same shape)
- `:283` — `refused` UPDATE (new)
- `:312` — `completed` UPDATE
- `:333` — legacy holder-only DELETE

With `refused` now re-claimable, a seat that already released `completed` can flip its own tombstone to `refused` with a second ack (an ack retry is enough), the row becomes claimable, and the handoff re-offers an already-answered human message to another seat. The same path exists today for `completed → declined`; the outcome-less DELETE on a tombstone erases `declined_by` and resets the chain.

**Ruling:** a terminal state is reachable only from a live lease. Add `AND state = 'claimed'` to all four clauses. One line each; no other change to the shape stamped in the previous comment.

**Tests to add:** on one row, release `completed` then `refused` → second returns `released: false`; `completed` then `declined` → `released: false`; outcome-less DELETE on a `declined`/`refused`/`completed` tombstone → `released: false`.

Everything else from the lift comment stands. Re-run on the recut head; @connector-ops please hold.

lilyshen0722:
**Guard item implemented at `63d61255`** (wren 71282 / 71300, PR comment above) — the four write paths now run only from a live lease:

| path | guard |
|---|---|
| declined UPDATE | `AND state = 'claimed'` |
| refused UPDATE | `AND state = 'claimed'` |
| completed UPDATE (tombstone) | `AND state = 'claimed'` |
| legacy no-outcome DELETE | `AND state = 'claimed'` |

Losing that race is not an error: the caller gets `released: false`, the tombstone is untouched, and the handoff — which already returns early unless `result.released` — queues nothing. That last part is what makes the flip benign *at the handoff*, but the guard is the fix: `completed -> refused` would otherwise make an already-answered human message re-claimable and re-offerable, which is this PR's defect re-entered from the other side.

Three tests, the shape you asked for, in `messageClaimService.test.js`:

- `completed -> refused` → `{ released: false }`, row still `completed`
- `completed -> declined` → `{ released: false }`, row still `completed`
- legacy DELETE on a tombstone → `{ released: false }`, row still present

They use a one-row fake that **honours the guard the SQL carries** rather than being told the outcome, so removing a guard makes the fake report the write as matched — that is what lets the test fail. It also asserts no write path is unguarded, so a fifth path added later without the clause reddens rather than passing quietly.

**A disclosure about my own evidence.** The first run of the extended ledger reported these four as `SURVIVED`. That was a **harness** false negative, not a coverage hole: the detector looked for `✕`/`●` followed by the expected string, which only matched test names that *began* with it — the three other verdicts survived only because their names were prefixes. I verified by hand (guard removed → `✕ a terminal row cannot be re-released`), fixed the detector to match the marker and the string on one line, and re-ran: **31 mutations, 31 red, no survivors**; a clean-tree baseline shows zero failure lines. Any mutation ledger run with the old shape under-reports, so it is worth re-checking elsewhere.

Backend: 420/420 suites / 3840 passed; `lint:ts` 0 errors. Rebased/amended single commit, `58232e6a` base, no `cli/`. Re-run structure when you have the head.

lilyshen0722:
**Structure gate — HOLD LIFTED at `63d61255`** (Wren, TASK-099 lane).

The one open item from the previous comment is closed on this head:

- `AND state = 'claimed'` on all four write paths in `messageClaimService.ts` — declined `:239`, refused `:293`, completed `:325`, legacy holder-only DELETE `:345`. A terminal state is reachable only from a live lease.
- Test `a terminal row cannot be re-released` releases `completed` then attempts `refused`, `declined` and the outcome-less DELETE on the same row; each returns `released: false` and the tombstone stays `completed`. The fake honours the guard the SQL carries, and the trailing assertion reddens any future write path added without it.
- Non-test diff against `0d69302f` is the guard lines plus one comment block; everything stamped at `0d69302f` (enum-gated reason, unconditional handoff routing, CAS + prune with `'refused'`, `declined_by` append, tombstone) is unchanged.
- Mechanics: 1 Lily-authored commit on main tip `58232e6a`, behind 0, `git merge-tree --write-tree` exit 0. 6 backend files, no `cli/`.

CI at read: `Test & Coverage`, `E2E Tests`, `Analyze (javascript-typescript)` in progress; everything else green. Clear on my side once they land; merge is @connector-ops / Sam. The cli half (`kai/claim-refused-cli` @ `471483e0`) gets its own gate after this merges.

lilyshen0722:
**Structure gate — lift at `63d61255` withdrawn; one item, ruled into this PR** (Wren), on Vera's finding (pod 71315/71316/71321), verified at the head.

The claim CAS `ON CONFLICT … DO UPDATE SET` (`messageClaimService.ts:140-144`) sets `claimed_by`, `instance_id`, `expires_at`, `state`, `created_at` and leaves `refusal_reason` / `refusal_status` in place. On the PR's headline path — A releases `refused` (`upstream-refused`, 429), the handoff re-offers, B claims and declines — the final row reads `state='declined'`, `declined_by=[A,B]`, still carrying A's class and status. A class outliving its state is the same "answered vs never ran" confusion this PR exists to remove, one column over.

**Ruling:** fix here, not TASK-107 (that row is cli admission, a different thing). A re-claim is a new turn: add `refusal_reason = NULL, refusal_status = NULL` to the CAS SET. With the state guard now on every release branch, those branches only ever run from `'claimed'`, so the CAS is the single place the reset belongs.

**Test:** A `refused` 429 → B `claim()` wins → B `declined`; assert the returned/stored row has no `refusal_reason` / `refusal_status`, and `declined_by` is `[A, B]`.

Everything else stamped at `63d61255` carries. Re-run on the recut head; @connector-ops please hold.

lilyshen0722:
**CAS reset implemented at `8198012d`** (wren 71322, raised by connector-ops 71316).

The `ON CONFLICT … DO UPDATE SET` now carries `refusal_reason = NULL, refusal_status = NULL`, so a re-claim is a new turn. The one place it matters, and the one that is easy to miss: A refuses `upstream-refused/429` → the handoff re-offers → B declines. Without the reset the row reads `state='declined'` **still carrying A's class**, which attributes A's dead route to B's decision — the same "answered vs never ran" confusion this PR exists to remove, one column over. It is the CAS rather than the release branches because the release branches now only run from a live lease, so a claimed row can never carry a class.

The own-holder renewal shares that SET: clearing there is a no-op, since nothing can set a class without a lease and the CAS is what clears it on the way back in.

Test: `a re-claim is a new turn: the previous seat's refusal class does not outlive it` — A refuses 429, asserts the class is on the row, B re-claims, asserts both columns are NULL, B declines, asserts the row is B's with no class and `declined_by=[A, B]`.

**Second instrument disclosure, same family as the first.** This test's first fake cleared *both* columns whenever it saw the reason reset in the SQL, so `M20` (delete only `refusal_status = NULL`) reported `SURVIVED`: the fake could not tell a half-reset from a full one. Fixed — each column is now assigned only if the statement assigns it — and `M19`/`M20` both go red. Ledger now **33 mutations, 33 red, no survivors**, clean-tree baseline zero failure lines. A fake that models a statement's effect in aggregate rather than per column is the same defect as a detector that matches a prefix: it reports on the instrument, not on the code.

Backend: 420/420 suites / 3842 passed; `lint:ts` 0 errors. `8219…`→`8198012d`, single commit, base `58232e6a`, no `cli/`.

lilyshen0722:
Structure gate at `8198012d`: **CLEAR — design hold lifted.**

- Single Lily commit on main tip `58232e6a`, behind 0, `git merge-tree --write-tree origin/main` exit 0, 6 backend files, no `cli/`.
- Non-test delta from `63d61255` is exactly the CAS reset: `refusal_reason = NULL, refusal_status = NULL` in the `ON CONFLICT … DO UPDATE SET`, plus the comment. Nothing else moved in `backend/services` or `backend/routes`.
- Headline-path test reads as specified: A refuses `upstream-refused/429` → class on the row → B re-claims → both columns NULL → B declines → `state='declined'`, no class, `declined_by=[seat-a:default, seat-b:default]`. The fake assigns each column only when the statement assigns it, so a half-reset is observable.
- Column walk from 71324 now closes: every writer of the two columns is the CAS (NULL) or the refused branch, and with the state guard a `'claimed'` row cannot carry a class.
- CI at this head: all 19 check-runs success, including all three Analyze legs.

Press to @connector-ops. The cli half (`kai/claim-refused-cli`, 0.1.68) opens after this lands.

PR #1831, opened 2026-09-23
fix(cli): the codex and claude failure tails go through the refusal reader

lilyshen0722:
Closes TASK-103 — the remainder of #1823/#1827 for the two adapters pi's refusal reader never reached.

## The gap, in the row's corrected words

An earlier version of this row said codex and claude were *silent* about a refused route. **They are not** — vera's 71177 held #1827 on that sentence and the tree agrees: `claude.js` rejects on `code !== 0` and reports stdout with it (a `-p` run writes quota conditions there), `codex.js` rejects on `turnFailedMessage` even at exit 0. The real gap is the opposite one, and it is two things:

1. **the tail is the raw upstream text**, unscrubbed by the keep-list + exact-match rules in `cli/src/lib/upstream-refusal.js`. For codex's `turn.failed` branch that is not just a log-quality problem: the adapter's own comment says the run loop posts that message **as the agent's reply**, so a body echoing a credential this spawn was handed would be posted into a pod.
2. **no HTTP status reaches the circuit breaker**, so `classifySpawnFailure` keyed on regex alone against a provider that had already answered 429. A missing string there is not one class off — it is 180× faster on the first retry, and the only class that does not open the circuit at n=1 (`spawn-retry.js`).

## The change

`scrubAdapterFailure` + `adapterFailure` join `upstream-refusal.js` as the shared reader both adapters now go through, so the two rules cannot drift between three adapters:

- **Non-JSON tails are kept whole and exact-match redacted.** Claude's usage-limit line is prose (`Claude usage limit reached. Your limit will reset at 11:40pm.`) and it is the entire reason stdout is reported at all — the 2026-08-03 outage where 361 consecutive failures carried no reason. Dropping it for being non-JSON would undo that fix, so the keep-list applies only when the tail parses as a body; the whole-shape case is what `redactKnownCredentials` (substitute, not drop) exists for.
- **JSON bodies are reduced to one named field**, and a field that echoed a handed-value is dropped with the body redacted — the same trade #1827 makes.
- **The status is read at the start *or mid-string*.** Claude joins stderr and stdout with ` | `, so when stderr has anything at all the status is not at position 0 — a prefix-only read would report a 429 body as an unclassified runtime failure, which is the misclassification this exists to remove. A standalone 3-digit token is required and it must be 4xx/5xx, so a timestamp, a `1234ms` count, or a logged `200` is not read as a refusal.
- **The status is attached as `error.status`, not only written into the text**, because `classifySpawnFailure` tests the field before the message; the message still names it so a human log agrees with the classifier.
- Message shapes are **preserved** (`claude exited with code 1: …`, `codex turn failed: …`), so nothing downstream that greps them changes.

`cli` **0.1.69**. (0.1.68 is held by the unopened CLI half of #1828, which opens after the backend lands — renumber if you would rather have the strict sequence.)

## Tests

`cli/__tests__/adapter-failure-scrub.test.mjs` (9) — the shared reader unit-level, plus a real local HTTP server answering **429** whose body genuinely echoes the key (asserted, so the redaction test cannot pass vacuously: the body *does* contain it and the detail does not), the mid-tail status shape, the false-positive shapes (`2026-09-23T04:00:00Z … 200 in 1234ms`), the keep-list, the field-vs-text classifier integration, and claude end-to-end through its `_spawnImpl` seam. One addition to `adapters.codex.test.mjs` for the `turn.failed` path (wording survives, credential does not, `.status === 429`).

**Mutation ledger: 10/10 red, clean baseline with zero failure lines, every file restored by sha256.**

## Two disclosures, both about instruments rather than code

1. **The provider-side leg is not measured.** The row asked for "a measured reproduction per adapter against a local server answering 429". The *server* half is real (above). The *CLI* half is not: no `claude` binary is installed on this machine, so the claude leg is measured at the adapter contract — the failure text in, the error object out — not against the provider. `@openai/codex@0.153.4` is installed, so a live codex leg is possible; it is not in this PR because it needs a codex config pointed at the local server and I would rather disclose the gap than ship a fixture dressed as a measurement.
2. **The baseline detector counted a `● Console` header as a failure line**, which made a green baseline look broken on the first run — the instrument reporting on itself, the third instance of this family after #1828's prefix regex and aggregate fake. Fixed to key on exit code + `✕` + the `Tests:` summary. Separately, one mutation's expectation string named a test that does not exist, so a genuine red (`exit=1`) was reported as `SURVIVED`; pinning it to a real test name turned it red. **A ledger's expectation strings are part of the instrument, and a wrong one makes a red look like a survivor.**

## Gates

Full cli suite green on this tree (47 suites, 771 passed, 10 skipped); `npm run lint:cli` from the repo root clean. @wren (lane) — and @vera for the security half, since the redaction rules are hers.


lilyshen0722:
Lane structure read at `75b5016f`: conforms, **HELD** on the redaction half (Vera, pod 71350–71352).

- Single Lily commit on main tip `58232e6a`, behind 0, `git merge-tree --write-tree origin/main` exit 0, 6 files all under `cli/`.
- Confirmed Vera's first item at the head: `scrubAdapterFailure` resolves `detail` as `kept ?? redactKnownCredentials(body, credentials)`, so every null from `keptRefusalDetail` — including the credential-echo drop — emits the whole body. Gate the fallback on "not a JSON object"; a JSON body with no keep-listed field reports status only.
- Second item stands as written: `statusInText` needs a status-shaped context, not any 3-digit 400–599 token.
- Versioning ruling: main never skips a number. Recut this PR to `0.1.68`; the #1828 cli half takes the next free number when it opens, and whichever of the two lands second rebases onto the other's bump.

Re-run on the recut head.

lilyshen0722:
Both holds fixed at `993fff41` (Vera 71350/71351 and 71352 — correct, and the first one was subtler than the comment it contradicted).

**1. The fallback predicate.** You are right that `keptRefusalDetail` returns null for four different reasons and I collapsed all four into "emit the whole body". The fallback is now gated on the shape — `isJsonObject(body) ? '' : redactKnownCredentials(...)` — so a JSON body whose keep-listed field echoed a handed value ships **nothing**, and a body with no keep-listed field reports status only. The non-JSON prose path is untouched, because that is the one case where the text *is* the diagnostic.

**And it was two layers deep.** The first cut of the fix still shipped the body, because a mid-tail read found the status at the ` | ` boundary and then handed the *whole joined string* to the body rules — which no longer looked like a JSON body, so it fell through to "keep it all" again. Finding the status and finding the body are now one function (`readStatusAndBody`), with M16 in the ledger pinning the boundary rather than the status.

**2. `statusInText`.** Also right, and the false positives are the ones you named: `500 tokens`, `"max_tokens": 500`, `retry in 500 seconds` all read as 500, and since `classifySpawnFailure` tests the status field *before* the message, an invented status stands the fleet down on a refusal that never happened. A status is now read only where the tail names it (`status` / `code` / `HTTP`) or where it opens a ` | ` segment followed by a colon — the shape claude's join actually produces for a body that leads with its status. `HTTP 200` and `| 200: ok` are asserted to stay null.

**One more case your predicate exposed, which I had not covered at all:** a JSON **string** tail (`"Too many requests, slow down"`) parses but is not an object, so it must keep its wording like any prose. It does, and M12 pins it — the mutation that made everything-parsed count as a body is now red instead of a no-op.

Ledger 16/16 red on a clean baseline. Two harness notes, since they are the same family as the code defect: M12's first form was a **no-op mutation** (the `JSON.parse` catch fired before the mutated line, so nothing changed — removed/re-aimed rather than reported as a survivor), and one mutation's expectation string named a test that does not exist, which made a genuine `exit=1` report as `SURVIVED`. A ledger's predicates need the same treatment as the code's.

cli suite 47 suites / 776 passed; `lint:cli` clean.

lilyshen0722:
Both residuals fixed at `30276c40`, plus the numbering recut. Your two are correct and measured — replying in order.

**`HTTP/1.1 429 Too Many Requests` → `status: null` (71359).** Right, and the mechanism is exactly as you read it: the named-shape pattern handles `HTTP 429`, but its `\W{0,4}` gap cannot cross the `1` in `/1.1` because `1` is a *word* character — so the two never come within the gap. The most canonical refusal shape there is classified as an unclassified runtime failure, which is the misclassification this row exists to remove. It gets its own pattern (`HTTP/\d(\.\d)?\s+(\d{3})`) rather than a looser gap, because widening the gap starts reading `/v1/models responded 500` as a status — asserted null now, along with `HTTP/1.1 200 OK`. `HTTP/2 503` is covered. A status **line** names the status but its reason phrase is prose, so it marks no body boundary: the whole text stays the body and exact-match protects it, same as `HTTP 503`.

Also worth recording: my first cut of this removed `http` from the named alternation, read the versioned line, and **broke `HTTP 429`** — caught by an existing test, and the fix keeps both shapes. A narrowing that removes the working case is the same defect class as the tightening you found, one level over.

**The array carve-out (71360).** Correct, and your framing is the one that decides it: an array is structured upstream data with fields this code does not know, so it is precisely what rule 1 is for — while a JSON *string* has no fields to reduce and is a whole-shape text, which is why the string case is a real carve-out and the array is not. `isJsonObject` → `isStructuredBody` (object **or** array). Measured with your shape: `[{"error":{"message":"boom","api_key":"…"}}]` now returns `detail: ''` — nothing echoed, and the assertion is written against the secret's absence so it cannot pass vacuously.

**Version recut to `0.1.68`** per Wren 71355 (main never skips a version), and I bumped the #1828 cli half to **`0.1.69`** on its branch now rather than leaving both halves on the same number — it takes the next number when it opens either way, so this only removes a rebase.

Ledger **18/18 red** on a clean baseline: three new mutations aimed at these two fixes (the array exclusion, the structured predicate loosened to "parses", the status-line pattern removed), and **M3/M11 re-anchored** — the rename moved the line they pointed at, and a stale anchor reports as `SURVIVED`, so a ledger needs re-anchoring whenever the line it names moves. That is twice now on this PR that a survivor was the instrument rather than the code; I would rather report the count of instruments fixed than the count of greens.

cli suite 47 suites / 778 passed; `lint:cli` clean. Ready for the structure re-run.

lilyshen0722:
Lane gate at `30276c40`: **CLEAR** on the lane half. Merge still waits on Vera's clearance of the redaction half at this head (her 71350–71352 and 71359–71360 items).

- Three Lily commits on main tip `58232e6a`, behind 0, `git merge-tree --write-tree origin/main` exit 0, 6 files all under `cli/`.
- `cli/package.json` `0.1.67 → 0.1.68`, per the numbering ruling (pod 71355); the #1828 cli half moves to `0.1.69`.
- Source delta from the first stamp `75b5016f` is confined to `cli/src/lib/upstream-refusal.js`: `readStatusAndBody` (status and body boundary found together; only the `| 429:` segment shape moves the boundary), `isStructuredBody` (object or array; a JSON string stays prose), `STATUS_NAMED` / `STATUS_LINE` / `STATUS_COLON` replacing the bare 3-digit scan, and the fallback `kept ?? (isStructuredBody(body) ? '' : redactKnownCredentials(body, credentials))`. The error message omits the `: detail` suffix when detail is empty.
- CI at read: `Test & Coverage` and `Analyze (javascript-typescript)` in flight; every completed check success.

Press to @connector-ops once Vera clears and the in-flight checks land.

PR #1832, opened 2026-09-23
docs(audits): triage the CodeQL js/missing-rate-limiting backlog (TASK-097)

lilyshen0722:
Triage for **TASK-097** — docs-only, one file: `docs/audits/codeql-missing-rate-limiting-triage.md`.

The row asked for two things: reconcile the disagreeing counts, then say which routes need a limiter and which get a documented exemption. This measures the first and frames the second; it does not touch a route.

**Measurement (at `58232e6a`)**

| | |
|---|---|
| open alerts, `js/missing-rate-limiting`, `ref=refs/heads/main` | **349** |
| same query with **no** ref filter | **349** — the ref filter is a no-op; every alert's most-recent instance is already on main |
| distinct `(path, start_line)` sites | **200** (33 files; 3 are test-file express apps) |
| created in 2026-04 / in September | **322 / 3** |
| the guard's own view of the same surface | **232** baselined registrations, 204 limited, **0** fresh violations |

**The finding that matters:** these are **true positives, not indirection.** Walking from each flagged line back to its enclosing registration (`runtimeRouter`/`templatesRouter` included) shows **0 of the 200 sites carry a limiter anywhere in the chain** — while the 7 `phase4RateLimit` routes in `agentsRuntime.ts` are **absent from the alert set entirely**. So CodeQL does recognise our inline limiter shape, and the alert set is the burn-down list the repo already keeps, not a false-positive cloud. 349 is an instance count; the defect count is ~232 registrations, 197 of which the query also sees.

**Instrument note, because a first pass was wrong:** testing only the flagged line reported two `auth.ts` sites as limited and missed every registration under a non-`router` variable name. Its two hits were the *next* route's limiter, one line below. The chain walk is the measurement above.

**Not mine to decide (Wren + Sam):** a limiter on the agent-runtime family changes behaviour for every seat. The doc names four options with the existing budget (IP 3000/60s, token 120/60s), the blast radius, and the parameter the decision needs — the fleet's real per-token steady-state — and recommends per-route on the public surface first, then the mount-level stack once that rate is measured.

**One count I could not reproduce:** the row's "Vera counted 78 repo-wide". None of my shapes produce it; §7 asks for the query rather than guessing, because the answer decides whether the list is 232 or narrower.

No version slot: docs-only.


lilyshen0722:
Recut at `866eb01b` per your two amendments (71374) and the §6 ruling (71373), plus vera's correction (71371).

**What changed**

1. **Status is no longer "open".** §6 now records the ruling: **(D) split** — (B) per-route on the public/anon surface is a builder's row under TASK-097; (A) mount-level on `/api/agents/runtime` is **Sam's own order, and not yet**, with both preconditions named (per-token steady state measured against 120/60s; a runtime 429 reading as a *named* refusal in one seat's tool path).
2. **Callbacks and long-polls are no longer listed as exemptions.** §5 now carries your treatment: **IP tier on, token tier off** — a retry storm from a third party is still an IP. Mechanically that is registering `phase4IpRateLimit` alone, which is worth writing down because the stack is two-tier and the natural call is to register `phase4RateLimit`.
3. **The 78 is reconciled, and it was a page, not a population** (vera 71371): her query was the alerts call without `--paginate` — the first 100 open alerts of all rules, 78 of which happened to be this one; with `--paginate` it is 349. Her "38 in one file" comes from the same page and is unmeasured. §1 and §7 both corrected.
4. **Scope of (B) is measured rather than assumed.** Public/anon = no auth/token/member/scope middleware anywhere in the registration chain, with string literals stripped so a path like `/invite/:token` cannot masquerade as a guard. **9 of the 200 flagged sites qualify; one is the test-file app — so 8 production sites, all `GET`.** Of those only **3 are actionable**:

| route | file | budget |
|---|---|---|
| `GET /x/oauth/callback` | `admin/globalIntegrations.ts` | IP 600/60s |
| `GET /registration-policy` | `auth.ts` | IP 60/60s |
| `GET /verify-email` | `auth.ts` | IP 30/60s |

The other five (`health.ts` ×2, `pg-status.ts`, `docs.ts`, `stats.ts`) are liveness/meta and stay exemption candidates. The remaining **190** flagged sites are authenticated — (A)'s or the burn-down list's, not this row's.

**The 429-body gap, written down with its reason.** The existing limiter bodies in `agentsRuntime.ts` are `{message, code: 'rate_limited'}` — no `status`, no named reason — so the current shape is not enough for #1823's reader and a seat would see raw text. Each new 429 carries `status: 429` plus a named reason, and **the name has to separate a PLATFORM 429 from an UPSTREAM one**, because those are different facts to a seat deciding whether to retry. That also makes this shape precondition (2) of (A) rather than a separate convention.

**One thing I would put in front of the (A) budget before it is set:** the fleet's seats run on **one host** under the launchd supervisor, so they share a single egress IP — an **IP-tier-only** bucket is a *fleet-aggregate* bucket, not a per-seat one. 3000/60s is 50/s against ~30 seats reconnecting long-polls on the order of once per 30s, so the headroom is large; but that is headroom, not the per-token measurement (A) is gated on.

TASK-108 filed for (B) with the route list and a budget each. Structure re-run on `866eb01b` please — the previous head `acb474ae` is still the first commit in the branch.

lilyshen0722:
Head moved `1bcd545f` → `d47ed3f2` (6 commits on `58232e6a`, behind 0). Two folds, both requested:

**1. The probes ruling** (`d47ed3f2`, wren 71392 — carried as the outstanding fold in 71409/71422). §5's class row no longer calls health/status/read-only-meta wholly an exemption candidate; §6 lists `GET /public` with its budget instead of as a candidate. The deciding rule is now written down rather than re-litigated per route: **does an anonymous hit do database work.** Exempt: `health.ts` ×2, `pg-status.ts` ×2, `docs.ts` (a file read) — a limiter on a liveness probe can *be* the outage. Not exempt: `GET /stats/public`, three `countDocuments` per hit uncached, so **IP tier 600/60s** beside the OAuth callback. (B)'s actionable surface is therefore **4 routes**, not 3.

**2. The instrument's recorded shape** (`8f85f41d`, vera 71402/71405/71406, tier recorded per wren 71410). §6 now states the exact line, what is deliberately absent, that it measures the token tier alone, that every count is per backend replica, that it is a lower bound because the limiter answers the crossing request, and that the row names its own removal.

PR #1833, opened 2026-09-23
TASK-109: watermark-log the agent-runtime rate-limit counter (the instrument (A) is gated on)

lilyshen0722:
Closes TASK-109. The instrument that lets `TASK-097 §6 (A)` be decided on a measurement instead of a derivation.

**Why it is needed.** Nothing persists a per-key request rate. `phase4AgentRateLimit` uses `express-rate-limit`'s default in-memory store, so the only place the real count exists is `RateLimit-Remaining` on a live response, per backend process; `tool_calls` cannot stand in (5 rows all-time, one seat, 2026-09-18). The mount-level limiter on `/api/agents/runtime` is gated on the fleet's real per-token peak against the existing 120/60s budget, and the current answer to "what is that peak" is a bound derived from source (12/min idle floor) plus a burst argument — not a reading.

**Shape** (scoped by @vera in 71401 — a watermark log, not an access log):

- A line is emitted only when `used >= RATE_LIMIT_OBSERVE_WATERMARK` (default **60**, i.e. half of 120). An idle fleet emits nothing; the log's volume is proportional to the thing being measured.
- Off unless `RATE_LIMIT_OBSERVE` is truthy, so this ships inert.
- The line carries the limiter's own key plus `used`, `limit`, `remaining`, `resetTime`, the **replica** and a timestamp. Nothing else: no method, path, query, status, headers or body. The key is already `tok:<sha256>` / `hdr:<sha256>` / `ip:<prefix>` from the limiter, so no credential can reach a log.
- Mounted **last** in the `phase4RateLimit` stack (`routes/agentsRuntime.ts`), because each limiter overwrites `req.rateLimit` and last wins — that is the per-token tier, which is the tier `(A)` is about. A presence test pins the position so a future edit cannot silently move it to the IP tier.
- Both the flag read and the sink call are guarded: an instrument that can turn a 200 into a 500 is worse than no instrument.

**Two facts measured against the installed 8.3.2, not taken from the docs:**

1. `req.rateLimit` is `{limit, used, remaining, resetTime, key}` — asserted through a real limiter in the suite.
2. **The middleware never sees the refusal.** The limiter answers the request that crosses the ceiling, so the chain stops there: a 60s window with 50 refused requests still logs at most `used == limit`. The log is a **lower bound** on peaks — it says "the budget was reached", not "this was all the traffic". The suite asserts this (`statuses [200,200,200,429]` → two lines, `used [2,3]`).

**Ledger: 17 mutations, 17 killed** — clean baseline with zero failure lines, and a clean re-run after every restore is verified by `sha256`. Disclosures, because they are the parts a reader cannot see from the diff:

- **The first pass reported five "survivors" that were really ts-jest diagnostics** — the suite failed to run, which is a red for the wrong reason. The harness now separates `BROKEN-RUN` from `SURVIVED`, and those mutations were re-aimed at value changes (`key: 'unknown'`, `remaining: 0`, `at: ''`, `replica: 'unknown'`). A mutation that is a type error proves nothing about the code.
- **The `event` name has no type-valid mutation**: it is a single-member literal type, so any other value is a compile error. The compiler pins the name; the test pins the value, which is what catches a simultaneous type+value rename.
- **The "against the real limiter" tier cannot discriminate `used` from the deprecated `current` getter** — the package exposes both (`'current' in req.rateLimit` is `true`), so the mutation that reads `current` survives *that* test while the unit tests kill it. Field-name coverage therefore rests on the unit tier, and it is stated here rather than implied.
- `Number.isInteger` was removed from the watermark guard: `NaN > 0` is already false, so the clause was provably redundant — dead weight no mutation could kill.

**Verification:** full backend suite green — 421 suites / 3863 tests passed, 24 skipped (`npm test`, exit 0); `npm run lint:ts` 0 errors. Backend only, so no cli/mcp version slot.

**What this does not do:** it does not arm anything and it does not change a budget. `(A)` stays Sam's order, and `(B)` (TASK-108) stays behind TASK-110 — a watermark log measures the key the limiter uses *today*, which is the pre-auth header hash whose caller-controllability TASK-110 exists to fix. A reading taken before that lands describes the current key, not a trustworthy per-IP one.


lilyshen0722:
**Hold at `9ad8d950` resolved — head now `c426d701`.** Both halves of the hold were right, and the fields are now your spec.

**The `key` (71425/71426).** You're right that `tok:${req.agentTokenHash}` is the credential's DB index — the exact `AgentCredential.findOne({ tokenHash })` lookup key — and that the `hdr:` fallback is a full sha256 of the Authorization header. Both are now **prefix + first 12 hex**: `tok:abcdef012345`, `hdr:fedcba987654`. 12 hex is 48 bits against ~30 seats, so correlation survives and the 1:1 join does not. **`ip:` is kept whole** per 71426 — a bucket prefix, not a secret — and a test names that exception so the mutation removing it has an owner.

**The route pattern (71427).** You're right that "no request material at all" doesn't cover `req.route?.path`: it is a static string from our own router, not request material. Measured before implementing — inside route-level middleware `req.route.path` is `/messages/:messageId/claim` while `req.originalUrl` carries the id and the query — and all 18 `phase4RateLimit` registrations are `router.<verb>(path, …)`, so the pattern is always available. A test asserts a request whose `originalUrl` holds an id and a query string leaks neither.

The line, and only:

    { event: 'agent_rate_limit_watermark',
      key: 'tok:abcdef012345', route: '/pods/:podId/messages',
      used: 61, resetInSeconds: 45, replica: 'commonly-backend-...' }

`limit` and `remaining` are absent (71405): one restates the constant 120, the other is arithmetic on the fields that are there. `resetInSeconds` is floored at 0, so clock skew cannot emit a negative window.

**Recorded rather than implied** (71410): the observer measures the **token tier alone** — it is mounted last and last wins in a stack — and the IP tier (3000/60s) is explicitly not observed; a second observer between the two limiters would be a separate row. **Per replica** (71406): the store is in-memory and per-process, so one replica today makes the reading the whole picture and a token spread across pods undercounts. **A lower bound**, read the way you said you'd read it: the limiter answers the crossing request, so the observer never runs for it — a window that refused 200 requests still logs at most `used == limit`. **Removal named** (71402): the middleware and its flag come out in the same PR that closes TASK-109.

**Your two assertions are tests rather than assumptions:** `reports the token tier, not the IP tier, when two limiters are stacked` — a real `[ipTier(max 1000), tokenTier(max 3), observer]` stack, statuses `[200,200,200,429]`, emitted `used` `[2,3]`, every key `tok:testkey`; and `is a no-op — on the wire, not just in the sink — with the flag unset`.

**Ledger: 21 mutations, 21 killed** (clean baseline, clean post-restore; full backend suite `3878 passed / 3902`, `lint:ts` 0 errors). Two disclosures rather than quiet fixes:

- **The harness had no per-mutation timeout, and M5 does not fail the integration block — it hangs it.** With `next()` only on the observing path, nothing answers the supertest requests with the flag off, and the first version of this ledger was killed mid-run and **left the mutation in the tree**. Fixed: 300s per-mutation timeout, restore-on-signal, and every run's output kept. M5 is reported `RED*` — killed, but via the timeout path, with its expected assertion in the partial output.
- **The "did the suite run" predicate was `"0 total" not in summary` — a substring test on a formatted count.** It therefore rejected a green run of **50 tests**, because `"50 total"` contains `"0 total"`, and reported four real kills as BROKEN-RUN. It parses the number now. Neither defect was visible until each mutation's output was saved, which is why the output is kept.

PR #1834, opened 2026-09-23
feat(v2): Commonly's own avatar faces replace DiceBear Big Smile

lilyshen0722:
## What

Sam chose the **Cut** direction on the Avatar kit canvas (2026-09-23), over Mark and Grid, and asked for it everywhere in the app. The faces are drawn in this repo, so the app no longer depends on DiceBear.

Every face in the app goes through `characterAvatarFor`: chat, thread previews, the inspector, Your Team, and the profile picker. That function keeps its signature and data-URI output, so this change is one generator swap with no caller edits.

faces from this branch

## The change

| file | change |
|---|---|
| `v2/utils/avatarKit.ts` (new) | Deterministic generator, paths and fills only, about 1 KB per face, no library, nothing fetched |
| `v2/utils/avatars.ts` | Uses the kit; the 24 picker archetypes are mapped to the kit's seven hair cuts |
| `V2Login.tsx` | Removes the DiceBear CC BY credit, which the licence required only while we used Big Smile |
| `package.json` / lockfile | Drops `@dicebear/core` and `@dicebear/collection` |
| docs | Signal identity rule 4, design-system README, brand direction, and a CLAUDE.md claim about `npm ci` that was already stale on `main` |

**Species still reads at 20px.** People sit on warm grounds in light shirts. Agents sit on cool grounds in ink with a cobalt collar and keep level eyes.

**No migration.** The stored `bigsmile:<seed>` and `bottts:<seed>` values are species tags, not artwork names. A face someone picked keeps its seed and redraws in the new style.

**Out of scope.** Marketplace agents with a published avatar image still show that image, because an uploaded image outranks a generated face.

## Evidence

- Targeted suites: 11 suites and 275 tests pass (avatar character, preset faces, avatars, V2Avatar, initials, avatarUtils, UserProfile, V2Composer, V2ThreadReveal, V2Login, layout invariants).
- `tsc --noEmit` is clean. ESLint on the changed files shows 0 errors and two pre-existing warnings on lines this PR does not touch.
- Mutation checks on the generator each turn a test red: removing the collar, dressing agents light, dressing people in ink, dropping the mustache, and flipping the arc winding.
- The winding test guards a bug caught on the canvas. Arcs drawn counter-clockwise punched a hole under every chin.
- `npm ci --dry-run` succeeds on this branch and on `main`.
- Design canvas, with the three directions and the chosen one in context: https://claude.ai/artifact/7kSSUiQnLabdjnMXGU2Qws (Avatar kit page)

After deploy, the live check is Your Team and a pod thread at 1440 and 390 on commonly.me.


lilyshen0722 (commented):
UX-GATE: PASS @ 469512f7 — design gate on the Cut kit, rendered from this head at 96/40/28/20 on the real grounds at DPR 1 and DPR 2 (24 people + 24 agents, mixed 20px transcript on white and on tint, the 24-cell picker, same-seed both kinds, every cut × every cool ground).

Species at 20px reads, and it reads from the shirt: light shoulders vs ink shoulders is a whole-cell value difference that survives DPR 1. The cobalt collar is 2–3 px at 20/DPR 1 and ~5 px at DPR 2 — it is a 40px+ cue, not the 20px carrier; the doc wording ("ink with a cobalt collar") is right as long as nobody later removes the shirt fill thinking the collar carries it. Level eyes are a 40px+ cue too. Same seed human/agent never collide. Weakest cell is an agent with black hair on slate #475569 (dark blob, one skin dot at 20) — still unmistakably agent, and identity is the 28+ job.

Bar: cobalt appears only as the collar (a mark, same budget as the mono agent name), ink acts untouched, no green, radius 4 frames unchanged, no third-party credit left on login. Picker stays human-only so no cobalt there.

One line to fix before press, not a design call: frontend/design-system/README.md line 178 still ends "Big Smile is CC BY 4.0; retain the visible DiceBear credit on login" while this PR removes that credit from V2Login. Drop the sentence. I re-stamp the moved head quietly.

lilyshen0722 (commented):
**GATE: PASS at `469512f7`** — exact head, measured. One gap, non-blocking.

BASE **275/275** across the 11 suites you name, RESTORED 275/275, tree clean after every mutation. Your count reproduces exactly.

**Your ledger holds — all five re-run here, each with anchor count 1, each red exactly 1:**

| mutation | result |
|---|---|
| remove the agent collar | 1 red |
| dress agents light (`shirt: kind === 'agent' ? INK : shirt` → `shirt`) | 1 red |
| dress people in ink | 1 red |
| drop the mustache layer | 1 red |
| flip arc winding (`0 1 1` → `0 1 0` in `ellipse`) | 1 red |

**The `npm ci` claim checks out, including the part about `main`.** I ran the real thing, not `--dry-run`: exit 0 at this head (1139 packages) and exit 0 at `origin/main` (1172). `main`'s lockfile does carry both `@dicebear` blocks and lists them in root deps, so the CLAUDE.md line this PR deletes was indeed already stale. I had the same stale claim in my own notes and have corrected it.

---

### Gap: the background half of the species signal is unpinned

`M6`: `const AGENT_BG = [1, 2, 6, 7];` → `[0, 3, 4, 5];` (anchor count 1, `avatars.ts:143`). **275/275 still green.**

The PR body says species is carried twice — the background family and the dress. The dress half is pinned twice over (M2 and M3 above). The background half has no test at all, and `avatars.ts:141` calls the two families "disjoint on purpose".

Probe across 200 keys, reading the first `fill` of the rendered SVG:

| | human backgrounds | agent backgrounds | shared |
|---|---|---|---|
| at `469512f7` | `15803d 2f6feb b45309 be123c` | `0e7490 0f766e 3b82a0 475569` | **none** |
| under M6 | same four | same four | **all four** |

So the code is right today and nothing would notice if it stopped being. The failure mode is quiet: a future palette reorder or an index edit silently collapses one of the two species signals, and at 20px the remaining cue is shirt colour.

**Suggested test** (passes at head, fails under M6):

```ts
const bg = (uri: string) => decodeURIComponent(uri.replace('data:image/svg+xml;utf8,', ''))
  .match(/fill="#([0-9a-f]{6})"/i)![1];
it('species background families stay disjoint', () => {
  const keys = Array.from({ length: 200 }, (_, i) => `seed-${i}`);
  const human = new Set(keys.map((k) => bg(characterAvatarFor(k, 'human'))));
  const agent = new Set(keys.map((k) => bg(characterAvatarFor(k, 'agent'))));
  expect([...human].filter((c) => agent.has(c))).toHaveLength(0);
});
```

Not blocking: the dress signal is pinned, so species still reads if the backgrounds drift.

### One note, no action

The "no migration" claim rests on `bigsmile:` / `bottts:` staying species tags. Only one test file mentions either string, and in a comment rather than an assertion. That contract is carried by prose right now. Worth a row if you want it pinned, since it is the thing that makes a future art swap free.

PR #1835, opened 2026-09-23
feat(cli): name a refused turn's class on release, with one predicate for both sites (TASK-099)

lilyshen0722:
Driver half of TASK-099. The backend half landed as #1828 (`184cf329`); this is the part that has to *produce* a named refusal, carry it to the release, and read it back.

**The third outcome.** A finished turn releases its attention claim with one of three outcomes. `refused` is the new one: the lease holder could not deliver — the model never ran (an upstream 429/502/401, named by TASK-096) or the server refused the post. The class is an **enum**, not free text:

| class | carries | when |
|---|---|---|
| `upstream-refused` | `status` (integer) | the upstream answered non-2xx |
| `cascade-cap` | — | the cascade hit its cap |
| `delivery-refused` | — | delivery refused the post |

`status` rides **only** with `upstream-refused`: a cap refusal and a refused post never saw an HTTP status, so attaching a number to them would put a field on the record that cannot mean what a reader assumes. An unclassifiable refusal maps to `delivery-refused` — the class that claims the least — rather than becoming a fourth class; the server's 400 is the backstop, not the first line.

**One predicate, both sites.** `claimReleaseFor(event, turnResult)` was extracted out of two byte-identical copies in the run loop and is now called at **both** release sites (`agent.js:1153` single-event, `:1668` batch). The batch site is the one a `message.posted` turn actually reaches — `:1148` is unreachable for that event type, pinned in the backend half — so a rule applied to only one of them would pass review and do nothing.

**The marker does not leak onto the wire.** `refused` is a *local* release marker. The event ack is a wire payload with its own readers, and what those readers need already arrives as `reason`, so `ackResultFor` drops the marker at the ack boundary (`:1761`, `:1861`) rather than widening the ack with an internal flag. It returns the result unchanged when there is no marker, so the common path never rebuilds the object.

**A CLI that ships first must still release its lease.** `enforcement.release(outcome, { reason, status })` carries the class, and when an older kernel rejects the third outcome with a 400 it retries **once** as `completed` — which is what the old release did. Bounded to a 400: a 500 or a timeout is not turned into a second write, and a second 400 is not retried again. Without that, a CLI running against an earlier server would leave the message claimed until the lease expired.

**Tests.** `claim-outcome.test.mjs` (new) pins the enum, the `status`-only-with-`upstream-refused` rule, the unclassifiable → `delivery-refused` mapping, and the ack-boundary strip; `run-loop.test.mjs` covers each class end-to-end through the batch path.

**Verification**: cli suite **47 suites / 781 passed / 10 skipped, exit 0**; `lint:cli` 0 errors.

**Version `0.1.69`, and it must land after #1831.** Main's cli is `0.1.67` and #1831 carries `0.1.68`; merging this first would make main skip a slot (Wren's numbering ruling, 71355). If the order inverts, the fix is a renumber, not a skip.


lilyshen0722:
Structure read at dfaf8344: CLEAR on content (pod 71584). One fold for the refresh you need anyway (BEHIND 2): `cli/src/commands/agent.js` batch release site carries the line `// Same decision point as the single-event path.` twice in a row. Drop one. That is a content change, so name the new head and I re-read the one-line delta.

lilyshen0722:
Hold response (vera 71537 / 71448-71449) landed as `dfaf8344`.

- `claimReleaseFor` now carries the kernel's bound — `MIN_UPSTREAM_STATUS` 400 / `MAX_UPSTREAM_STATUS` 599, mirroring `messageClaimService:45-46`. An out-of-range status keeps the **class** and drops only the number; forwarding it made the kernel answer 400, and the 400 fallback re-released as `completed`, which is terminal, so the refusal lost its handoff rather than just its field.
- Tests pin 400 and 599 as literals rather than reading them back from the constants they test, with 200/399/600/0/1000 as the out-of-range set.
- Mutation ledger (`/tmp/kai099status-ledger.py`): 6/6 killed — range removed entirely; either bound made exclusive; either bound dropped alone; and the constant loosened with the predicate left intact. That last one is caught only by the literal-pinning test, which is why it is pinned.
- Clean-tree baseline: 0 failure lines, 19/19 passed, exit 0. Full cli suite 48 suites / 801 passed / 10 skipped; `lint:cli` 0 errors; CI 12/12 at this head.

Harness disclosure, since it is part of the instrument: the ledger's baseline line was mis-specified on the first run — I passed a never-matching expectation string, so a green tree reported `SURVIVED` and the script printed `BASELINE PROBLEM`. No mutation verdict was affected (each is evaluated against its own named test), and the re-run with the correct predicate is the one quoted above.

PR #1838, opened 2026-09-23
feat(cli): withhold the grant broker on the run path (TASK-111)

lilyshen0722:
Ruling **B** (wren 71640/71641), and my (A) premise was wrong — correcting it here first, since I proposed it in the pod and wrote it into the row.

**Why (A) has nothing to refuse at.** `toolInstallables.ts:28/84` declares the broker as the GitHub Installable's *catalogue* component: its consumers are `installableCatalogService` (the page) and `grants.ts`'s `resolveBrokerFor` (which names the brokerId). Nothing writes a seat's `environment.mcp[]`. `GRANT_BROKER_URL` is imported by exactly one route, `agentBinding.ts` — the daemon assignment — and that one already refuses. Wren's 71631 named the missing write and I could not find one either.

**What the hole actually is.** The broker reaches an unconfined seat as an entry **hand-authored into that seat's own record**, which `commonly agent run` spawns verbatim. So the withhold goes at the run site, in memory:

- `performRun` derives the seat environment with the daemon's own `withholdGrantBroker` (`grant-broker-guard.js`), passing `instanceUrl` — which is what makes the injected `${COMMONLY_API_URL}/api/mcp/grants/<id>` placeholder identifiable at all.
- Same predicate, same typed code (`grant_broker_unconfined`), one refusal with two emitters; the refusal is logged with its reason.
- **Record untouched**: the stored record keeps its broker so the daemon remains the layer that judges a host it knows. Applied once for the run, because `agent run` reads its record once at boot. An environment with nothing to withhold is passed by identity, which the derive sites rely on.

**Building it found a real defect — the case vera asked to be covered.** `resolvePublicSandboxMode` (`cli/src/lib/sandbox/mode.js`) compared the **raw** trust while `confinementReason`, in the same module family, maps it first via `effectiveTrust`. A `{ trust: 'internal' }` record with **no mode** therefore resolved to no mode at all → `sandbox_mode_unenforceable` → the broker was **withheld from a seat this daemon would have confined**. The mapping is documented in `environment.js` ("`internal` is read as `public`, so such a seat is confined where it can be and refuses to derive where it cannot") — it was applied in the trust check and not in the mode resolution. Fixed here, since this change depends on it.

**A sibling I am reporting rather than fixing in this PR.** `agent.js:514-517` makes the same raw comparison (`environment.sandbox?.trust === 'public'`) on the attach path. Read, not measured: with `internal`, `sandboxMode` falls to `'none'`, and the guard that would refuse `'none'` is gated on the same raw comparison — so the seat neither refuses nor confines. That is a different path with different consequences (fail-open there, fail-closed here), so it wants its own review rather than a rider on this PR. Say the word and I will open it as a one-line row.

**Tests** (seven, at the run path): withheld when no sandbox is declared; kept when confining; kept for a legacy `internal` trust with no mode; withheld for a non-public trust (`sandbox_trust_not_public`) and for an adapter that cannot confine (`adapter_cannot_confine`); the record untouched and an untouched environment passed **by identity**; a foreign origin under the broker path left alone. Fixtures use a unique delivery id per call — a reused one is answered with `duplicate delivery … without batch spawn`, which reads as a broker bug when it is only a fixture collision with the file's earlier tests.

**Verification.** Mutation-proved 5/5 (`/tmp/kai111b-ledger.py`): withhold not applied at the run site; the legacy mapping not applied when resolving the mode; withholding mutating the record instead of copying it; `instanceUrl` not passed (placeholder unresolvable); a foreign origin drawn in by its path. Clean-tree baseline: 0 failure lines, 78/78 passed, exit 0. Full cli suite **47 suites / 785 passed / 10 skipped, exit 0**; `lint:cli` **0 errors**.

**Version `0.1.70`**, and it must land **after #1835** (`0.1.69`): main's cli is `0.1.68`, so merging this first would skip a slot (Wren 71355).


lilyshen0722:
Rebased after the stacked base merged: #1835 (cli `0.1.69`) landed as `64fa5116`, taking the same two files. `5b3c74e0` → **`0d90311c`** onto main `fe98462f`; only conflict was the version line, resolved to `0.1.70`.

Post-rebase content verified, not assumed — the delta against main is exactly the withhold (import, derivation, `environment: seatEnvironment` at the spawn site) plus the `mode.js` trust mapping; #1835's claim-outcome code is untouched. Diff 4 files, +162/−9.

Gate on the rebased head: **cli 48 suites / 808 passed / 10 skipped, exit 0; `lint:cli` 0 errors.**

Disclosed flake: the first full local run failed `adapters.pi.credential-channel.test.mjs` (5s child-probe timeout under parallel load) — passes 27/27 isolated, three times, and the full re-run was green, so it is load-dependent, not this branch. Same class as TASK-112.

lilyshen0722:
Structure read at head `0d90311c` (1 Lily commit, behind 0 of main `fe98462f`, `merge-tree` exit 0, CI 12 pass / 1 skipped): **HELD on one fold**, everything else reads clean.

**The fold.** `effectiveTrust` is now written twice — `cli/src/lib/sandbox/mode.js:37-41` and `cli/src/lib/grant-broker-guard.js:96-100` — beside a third spelling of the same mapping already in `cli/src/lib/environment.js:63-71` (`isLegacySandboxTrust` / `normalizeSandboxTrust`). The defect this PR fixes is the mapping being applied at one site and not another; adding a second private copy reproduces the shape that caused it. Export `effectiveTrust` from `environment.js` beside `LEGACY_SANDBOX_TRUST` (the module that owns the mapping), import it in `mode.js` and the guard, delete the guard's copy. No cycle: `environment.js` imports nothing from `sandbox/` or the guard.

While there, one row in the `grant-broker-guard.test.mjs` table so the `mode.js` change has a direct test rather than only the run-loop one: `['a legacy internal trust with no declared mode', 'claude', { sandbox: { trust: 'internal' } }, null]`. Line 122 covers `internal` *with* `mode: 'workspace'`, which passed before this PR too.

**Reads clean.**
- The withhold sits before the loop, applied once, `log` is the `performRun` param; `seatEnvironment` is the only `environment` handed to `spawn` (one site, `:1220`; no other `environment` use inside `performRun`, which ends at `:1942`). `adapter?.name` is the adapter object's `name` (`claude.js:569`, `codex.js:477`, `pi.js:373`). `withholdGrantBroker(null, …)` returns `null` by identity via `declaresGrantBroker`.
- The stored record is never written on this path (`performRun` has no `saveToken`), so the in-memory identity assertions are the right instrument for "record untouched".
- The five ruled tests are present: absent → `sandbox_absent`; `{public, workspace}` through; non-confining adapter → `adapter_cannot_confine` (exercised with `stub`, same set membership as `pi`); no broker → identity; `{trust: 'internal'}` no mode → passes. Plus foreign-origin left alone and `sandbox_trust_not_public`.
- The `mode.js` change is inert for the claude/codex adapters: both normalize the sandbox before their raw `trust === 'public'` gates (`claude.js:615`, `codex.js:546`). The un-normalized raw comparison is the attach path only (`agent.js:514-518`); filed as its own row, not a rider here.

Re-read on the recut will be the fold delta only.

lilyshen0722:
**Fold delta** (Wren 71662), head `0d90311c` → `bdfed011`, same version slot (`0.1.70`).

- `environment.js` exports `effectiveSandboxTrust` beside `LEGACY_SANDBOX_TRUST` — the scalar form of the mapping, distinct from `normalizeSandboxTrust`, which rewrites a whole sandbox object.
- `mode.js` and `grant-broker-guard.js` import it; **both local copies are deleted** (was `mode.js:36`, `grant-broker-guard.js:95`). The name mirrors the server's export in `grantBrokerConfinement.ts`, so the two sides read as one predicate rather than two that agree today.
- Guard table gains the row you asked for: `internal` + no declared mode → `null`.

Ledger 3/3 killed (`/tmp/kai111fold-ledger.py`):

| mutation | expected victims | result |
|---|---|---|
| M1 the shared export becomes identity | guard row **and** run-site test | KILLED — both red |
| M2 the guard site compares raw trust | guard row | KILLED |
| M3 the mode site compares raw trust | run-site test | KILLED |

M1 is the one that proves the fold: a copy left in either file would have kept that site green. Full cli suite 48/48, 809 passed / 10 skipped; `lint:cli` clean.

lilyshen0722:
**Structure: CLEAR at `bdfed011`** (2 Lily commits onto `fe98462f`; behind 0; `git merge-tree --write-tree origin/main bdfed011` exit 0). Read at main `fe98462f`.

The fold is exactly the hold at `0d90311c` (comment above): `effectiveSandboxTrust` exported once from `cli/src/lib/environment.js` beside `LEGACY_SANDBOX_TRUST`; imported at `sandbox/mode.js:32` and `grant-broker-guard.js:77`; both local `effectiveTrust` copies deleted; `detailFor` reads the same export. Mirroring the server's spelling in `grantBrokerConfinement.ts:101` is the right call — one predicate, one name on both sides. Guard table carries the missing row (`internal` + no declared mode → `null`).

Delta vs `0d90311c` touches only `environment.js`, `mode.js`, `grant-broker-guard.js`, `grant-broker-guard.test.mjs`; the run-path change, spawn site, and the five ruled tests are unchanged from the head Vera cleared.

CI at `bdfed011`: 11 success / 1 skipped, `Test & Coverage` still in progress at the time of this comment — press when it lands green. Slot `0.1.70` unchanged.

PR #1841, opened 2026-09-23
docs(design): evidence widths are 1200 and 390; add the landing demo spec

lilyshen0722:
Two docs changes, both approved by Sam on 2026-09-23:

- **Evidence widths.** `signal-identity.md` and the design-system README said 1440 and 390, while Sam's evidence rule (68655) and most of the evidence folder use 1200 and 390. Both docs now say 1200 and 390. Older plan docs keep their 1440 as dated records.
- **Landing demo spec.** `docs/design/landing-demo/` holds the approved interactive hero and the README frames as board sources, with a guide, so the seat building TASK-147 can read them. The canvas itself is private to Sam's account. The guide requires the build to use the real v2 components with fixture data rather than this markup, and the `characterAvatarFor` faces rather than the canvas image URLs.

Docs only.


lilyshen0722 (commented):
UX-GATE: PASS @ 7de01bce — docs only, gated from the files, nothing rendered. The width change is consistent in both files (signal-identity.md ×3, design-system README:234). Notes that carry into the TASK-147 hero gate, not blocks on this PR:

1. Demo.dc.html draws the active pod row as a cobalt-filled button. The live component (`.v2-pods__item--active`, v2.css:1125) is the tint fill, a 3 px cobalt mark on the left, and an accent-text title. The README says the hero is built from real v2 components, so the build follows the component and I gate the hero against the shell's pods list, not this frame. The cobalt-filled primary option and the cobalt-text "Other…" match main's decision card (`--choice--primary`, `__other`), so those stand.
2. No 390 board exists for the hero or the landing while README.md makes 390 evidence required. The 1312×720 grid (56/212/1fr/272) has no phone answer in the spec; the hero PR has to bring one and that is what I gate at 390.
3. Landing close uses a cobalt-filled Get started where the hero uses white-on-cobalt. Both are front-door blocks under "one system, two volumes"; fine, keep it to the one word.
4. The 1200 evidence width cites Sam's 68655. I could not read that thread from the pod API tonight, so the ruling is taken from the doc as written.
5. Readme.dc.html's green badges are GitHub's own, not product colour. Fine.
6. Shot*.dc.html: no `room`, no green, no shadow beyond the focus halo, nothing under 11 px.

PR #1842, opened 2026-09-23
fix(auth): register stops waiting on the verification email (TASK-142)

lilyshen0722:
TASK-142. `POST /api/auth/register` awaited `sendVerificationEmail` inline, so the first button a stranger presses blocked on SMTP2GO: **6211 ms** measured live on build `58a6f2c2` for a fresh example.com account. The ceiling is worse than the median — `emailService.sendEmail` sets a **30 s** timeout on that axios call, so a slow provider can hold a registration request for half a minute.

## The fix

The send moves to the background with one immediate retry; the response returns 201 as soon as the account exists. `authController.ts` gains `sendVerificationEmailInBackground`, called only when email delivery is configured.

## What this deliberately gives up

The `502 "Email delivery failed"` response was the only signal that the mail never went out. It is gone, and that is a real trade rather than a free win:

- **No test pinned it.** Checked before removing it: no suite asserts a 502 on register, and no frontend code branches on it.
- **It is replaced by two things.** A `console.error` naming the user id and the provider reason after the retry fails, and an in-product recovery path that already existed: an unverified account is shown `V2EmailVerificationBanner`, whose resend button hits `POST /api/auth/resend-verification`, separately rate-limited (it has its own limiter, per the comment on that route).

**Stated limit:** the send is in-process. It is not durable across a pod restart — a restart between the 201 and the send loses that email, and the banner's resend is the recovery. A durable outbox was considered and not built: it needs a model plus a drain in the cron registry, and the recovery path already exists in the UI.

`resendVerification` keeps awaiting its own send on purpose: there the user asked for the mail, so a 502 is feedback they can act on.

## Tests — mutation-proven, not presence-checked

Both new tests fail if the defect is reintroduced, which is the point of them:

| test | mutation | result |
|---|---|---|
| responds 201 without waiting for the verification email | restore the inline `await` | **red** — 30013 ms timeout |
| retries once and logs when the background email fails | cut the retry loop to one attempt | **red** — `Expected number of calls: 2` |

The first hands the handler a send that never settles, so it can only pass if the handler does not wait for it. Verified by making each mutation and reverting it.

- `authController` suite **22/22** (2 new)
- five neighbouring auth/email suites **20/20** (`authController.workspacePgMirror`, `authController.guideIdentity`, `invitationEntitlement`, `auth.api-token`, `emailService`)
- `eslint controllers/authController.ts` — **0 errors** (7 pre-existing `max-len` warnings unchanged)

## Not verified

The acceptance bar on the row is *"register responds under 1 s at p50 on commonly.me"*. This removes the only blocking dependency on the mail provider, but I have no way to measure p50 on the live instance from here — that recapture is a browser/ops measurement after deploy, not something this diff can show.


lilyshen0722 (commented):
**GATE: PASS at `e7710932`** — measured. BASE 22/22, RESTORED 22/22, tree clean after every mutation.

**The two new tests discriminate** (anchor count 1 each):

| mutation | result |
|---|---|
| M1 `sendVerificationEmailInBackground(user)` → `await sendVerificationEmail(user)` | **2 red** — both new tests |
| M2 `attempt <= 2` → `attempt <= 1` | **1 red** — the retry test |
| M3 drop the call entirely | **3 red** |

M1 is worth a word: the "responds 201 without waiting" test fails by **taking 30005 ms**, i.e. it detects the regression by actually hitting `emailService`'s 30 s ceiling. That is a real discriminator, and also the clearest possible statement of the bug you fixed.

**Your three give-up claims check out**, which is the part a test cannot cover, so I verified each at the merge-base rather than taking them:

1. *No test pinned the 502* — `grep 502` across `backend/__tests__` at the base returns nothing for register/email/verify, and nothing at all in `authController.test.js`. True.
2. *No frontend branches on it* — `grep "Email delivery failed"` over `frontend/src` at the base: no hits. True.
3. *`resendVerification` still awaits on purpose* — confirmed, `authController.ts:612`. The asymmetry is right: there the user asked for the mail.

---

### One gap, pre-existing, low severity

`M4`: collapsing `if (hasEmailConfig) { sendVerificationEmailInBackground(user); }` to an unconditional call leaves **22/22 green**. Nothing pins that an instance with no mail configured doesn't attempt a send.

It predates this PR (the same gate sits at `:376` on the base), so this is not a regression and I am not asking you to fix it here. It is worth one line because this PR changes its failure mode: before, a misconfigured instance got a loud 502; now an unconfigured one would quietly burn two attempts and two `console.error`s per registration. The gate is the only thing preventing that, and it is untested.

### Note for the evidence loop, not a change request

The `6211 ms` is a solid before-number. There is no after-number yet, and the register path still awaits `createDefaultWorkspacePod(user._id)` inline (`:412`). If signup is still slow on the next deploy, that is the next thing to time — which is TASK-145's territory, not this PR's.

PR #1843, opened 2026-09-23
test(v2): pin that people and agents draw on disjoint background families

lilyshen0722:
Closes TASK-148. Follow-up to #1834 (`58a6f2c2`), which moved the species signal onto disjoint background families.

## Why

sprint-review measured (71419) that pointing `AGENT_BG` at `HUMAN_BG` in `frontend/src/v2/utils/avatars.ts` left **every** suite green. The dress half of the signal (agents in ink, cobalt collar) is pinned twice; the ground half was pinned nowhere — so a same-seed human and agent could wear the identical wash and nothing failed.

## What

Test-only. `characterAvatarFor` is walked with 200 seeds for each kind, the ground is read back out of the rendered SVG (the first layer `renderFace` paints), and the two ground sets must not intersect. The sweep matters: the family index is `hashString(key) % 4`, so one hand-picked pair exercises one slot and a wrong-but-lucky family would pass.

## Proof

Mutation-proven in both directions, which is the asymmetry the row was filed about:

| mutation | result |
|---|---|
| baseline | 14/14 green |
| `AGENT_BG := HUMAN_BG` | **this test red**, 13 others green |
| `HUMAN_BG := AGENT_BG` | **this test red**, 13 others green |
| restored | 14/14 green |

The failure is the disjointness assertion (`expect(received).toEqual(expected)`), reporting the shared families — not a count or a length.

With the three neighbouring avatar suites (`V2Avatar`, `avatarInitials`, `avatarPresetFaces`): **4 suites, 29/29**. eslint clean on the touched file.

## Limit

It pins *disjointness*, not the 4+4 partition or the specific warm/cool split — asserting exact membership would freeze the palette, which is a design call. It also depends on the ground being the first painted layer; the existing "the traits draw is stable" test is what guards layer order, so a reorder must change both together.


lilyshen0722 (commented):
**GATE: PASS at `92632be5`** — measured. This closes the gap I filed on #1834 (71419). BASE 31/31 across four avatar suites, RESTORED 31/31, tree clean after every mutation. Anchor count 1 on each.

| mutation on `avatars.ts` | expected | result |
|---|---|---|
| M1 `AGENT_BG := HUMAN_BG` | red | **1 red** — the new test |
| M2 `HUMAN_BG := AGENT_BG` | red | **1 red** |
| M3 **partial** overlap, `AGENT_BG = [1, 2, 6, 0]` | red | **1 red** |
| M4 legit reorder, `HUMAN_BG = [3, 0, 5, 4]` (same set) | **green** | **31/31 green** |
| M5 `AGENT_BG = [1, 1, 1, 1]` (disjoint but no variety) | red | **1 red** |

M1 and M2 are the two you ran. The three I added are the ones that decide whether the test is the right shape rather than merely red:

- **M3** is the realistic drift. A future edit does not swap a whole family, it changes one index — and a full-swap-only test would sleep through that. This one catches a single slot.
- **M4 is the important green.** Your stated limit — disjointness is pinned, the 4+4 partition and which family is warm are not — is now measured rather than asserted: reordering a family within itself keeps all 31 green, so a designer can re-rank the palette without fighting the test.
- **M5** shows the two `toBeGreaterThanOrEqual(3)` lines are load-bearing, not decoration. Without them a family could collapse to one colour and still be "disjoint".

The seed sweep is the right call for the reason your comment gives: `hashString(key) % 4` means a single seed exercises one slot, so a hand-picked pair could pass on a wrong family by luck. No gap found — nothing further from me.

PR #1845, opened 2026-09-23
fix(copy): signup names what verification actually gates — the Community pod

lilyshen0722:
Closes TASK-143. Implements ux-lead's ruling on the row.

## Why

Lily's smoke: register answered *"Check your email for verification"*, login then **worked immediately unverified**, and the shell showed a *"Verify your email"* banner. Verification gates exactly one thing — joining the Community pod (`joinCommunityPodBestEffort` is called only from the verify handlers and the auto-verified register path). So the copy now names that gate and implies nothing more: no new gate, no wall, no second step before the workspace, and the first-run hero is untouched.

## Change

| where | before | after |
|---|---|---|
| `auth.verificationBanner.message` | "Verify your email — we sent a link to {{email}}." | "Verify your email to join the Community pod — link sent to {{email}}." |
| `auth.register.success.checkEmailTitle` | "Check your email" | "You're in" |
| `auth.register.success.checkEmailMessage` | "We sent a verification link to your email." | the banner's sentence, with the submitted address |
| `authController.register` 201 message | "User registered successfully. Check your email for verification." | "Registered. Verify your email to join the Community pod." |

zh-CN follows in the same PR (same three keys). The `verifyThenSignIn` hint already said *"You can sign in now; a reminder stays visible until you verify"* — accurate as written, left alone. `ApiDevPage`'s documented `exampleOutput` for this endpoint is updated too, so the sample output is not a string the API no longer returns.

## The coupling this would otherwise have broken

`V2Register.tsx` chose its success screen by **substring-matching the API prose**: `message.toLowerCase().includes('check your email')`. Ruling item (3) alone would therefore have switched ruling item (2)'s screen off — the "You're in" screen would never have rendered, and nothing would have failed. It now matches both phrasings, so a rolling deploy with the old backend still lands correctly. **The coupling is still prose**, which is the limit of this change: a structured flag on the response would remove it, and that is a separate decision, not something to smuggle into a copy PR.

## Proof

| mutation | result |
|---|---|
| baseline | 3/3, banner+AuthContext 9/9, authController 21/21 |
| predicate back to `includes('check your email')` | verify screen test **red** |
| predicate ignores the prose (always pending) | created screen test **red** |
| screen stops passing the address | sentence-with-address assertion **red** |
| API sentence reverted | new backend test **red** |
| restored | green again |

Both screen directions are pinned (`V2Register.test.tsx`): a response naming verification shows the reminder, one saying none is needed shows "Account created". The backend's verification-required branch previously had **no** test — the suite clears `SMTP2GO_*`, so only the auto-verify branch was ever asserted; that assertion is now explicit about which branch it means.

## Limit

The 390/1440 copy rendering is ux-lead's visual gate; this seat has no browser. Adjacent, not touched: the legacy v1 `components/Login.tsx` still refuses an unverified sign-in client-side, so a user who reaches that old form gets the old model — out of this ruling's scope and reported rather than silently changed.


lilyshen0722 (commented):
**GATE: PASS at `073ed7fb`** — measured. One gap, and it is a claim in the code comment that is not true yet.

BASE: frontend **9/9** (`V2Register`, `V2EmailVerificationBanner`, `AuthContext`), backend **21/21**. Restored clean after every mutation, anchor count 1 each.

**Pinned:**

| mutation | result |
|---|---|
| M1 regex keeps only `check your email` (drops the new phrasing) | **1 red** (frontend) |
| M3 regex → `true` (always the verify screen) | **1 red** (frontend) |
| M4 backend message reworded so it matches neither phrase | **1 red** (backend): *names joining the Community pod as the only gate when a mail was sent* |

M4 is the reassuring one. The producer side is guarded, so the cross-repo coupling cannot rot silently from the backend: a copy edit there reds the backend suite even though the frontend mocks its own response.

---

### Gap: the legacy arm is unpinned, and the comment says it is

`M2`: `/check your email|verify your email/i` → `/verify your email/i`. Frontend stays **9/9 green**.

The comment on that line says the pre-2026-09-23 wording is matched *"which is why `V2Register.test.tsx` pins both directions."* It does not: `grep -i "check your email"` over that file returns nothing. Every test mocks the new sentence. What is pinned is the new phrasing and the no-verification negative — not the legacy arm.

That arm is not decoration. It exists for exactly one situation — a rolling deploy where an old backend answers a new frontend — which is the window this PR will actually ship through, and the only arm with no test.

Probe (passes at head, fails under M2):

```js
axios.post.mockResolvedValue({
  data: { message: 'User registered successfully. Check your email for verification.' },
});
// …register…
expect(await screen.findByRole('heading', { name: "You're in" })).toBeInTheDocument();
expect(screen.getByText(/join the Community pod/)).toBeInTheDocument();
```

Under M2 it fails on the heading, i.e. the user is sent to the wrong screen. Not blocking — the code is correct today. Either add the test or soften the comment, but the comment should not claim a guard that is absent; the next person to touch that regex will read it and believe they are covered.

lilyshen0722:
**zh-CN gate: changes required.** This is the eng lead's review under the `zh-cn-ui-localization` skill. On 2026-09-23 Sam delegated the zh-CN read ("Don't block on me please, I think I shed enough direction"), so this table stands in for his pass.

| key | head | should be | reason |
|---|---|---|---|
| verification banner `message` | 请验证你的邮箱以加入 Community 群组 — 链接已发送至 {{email}}。 | 验证邮箱后即可加入 Community Pod，链接已发送至 {{email}}。 | Glossary: **Pod** stays English with half-width spaces, so 群组 ("group") is out. 以…加入 is the purpose-clause calque the skill bans. A Chinese sentence takes ， and not an em dash. |
| `checkEmailTitle` | 你已进入 | 欢迎加入 | 你已进入 is a literal "you have entered". The title is a welcome, and 欢迎加入 is how a Chinese product says "you're in". |
| `checkEmailMessage` | (same as the banner) | 验证邮箱后即可加入 Community Pod，链接已发送至 {{email}}。 | Same fix as the banner, so the two stay identical. |

English copy is fine as written. After the amend, `translationKeys.test.ts` still needs en and zh-CN key parity. The code gates (sprint-review, ux-lead) are separate from this and still owed.

lilyshen0722 (commented):
UX-GATE: PASS @ 073ed7fb — copy only, gated from the diff, no fixture. The strings are the TASK-143 ruling as written: verification gates only the Community-pod join, login stays open, the banner stays a banner (TASK-103). Register title "You're in" says the state the account is actually in now. Banner length at 390: `.v2-verification-banner__text` wraps with `overflow-wrap: anywhere`, `min-height: 40px` is a floor not a cap, and the ≤480 rule (v2.css:4437) stacks Resend under the text at 16 px, so the longer string grows the banner rather than clipping. API + ApiDevPage example strings match.

For Sam's zh-CN read, not blocking: "Community 群组" is the only place the file translates the pod noun; everywhere else it keeps "Pod" untranslated ("Pod 成员", "加入 {{podName}}", "前往 Pod →"). One term either way. "你已进入" for "You're in" is literal and reads as a fragment; his call.

Residual, already reported by sprint-impl: the verify-pending screen is still selected by prose match on the API message; a response flag would remove the coupling. Follow-up row, not this PR.

lilyshen0722:
**Amended at `368f9c17`** — lily-shen's zh-CN read applied, plus the one gap from the code gate.

### zh-CN, per the gate table (verbatim)

| key | was | now |
|---|---|---|
| `auth.verificationBanner.message` | 请验证你的邮箱以加入 Community 群组 — 链接已发送至 {{email}}。 | 验证邮箱后即可加入 Community Pod，链接已发送至 {{email}}。 |
| `auth.register.success.checkEmailMessage` | (same as the banner) | 验证邮箱后即可加入 Community Pod，链接已发送至 {{email}}。 |
| `auth.register.success.checkEmailTitle` | 你已进入 | 欢迎加入 |

English is unchanged. `i18n.test.ts` now pins the three strings verbatim, that the banner and the screen stay one identical sentence, and that neither carries 群组 or an em dash — so a later edit that reintroduces the glossary violation cannot pass silently.

### The gap: the legacy arm of the screen-choice regex was unpinned

The comment claimed `V2Register.test.tsx` pinned both directions; it pinned the new phrasing and the no-verification negative, never the legacy wording — the one arm that exists for the rolling-deploy window this ships through. New case: mocks `User registered successfully. Check your email for verification.`, asserts the reminder screen and the sentence. The comment now names the three cases instead of asserting a guard that wasn't there.

### Mutations — one red each, and each its own

| mutation | result |
|---|---|
| regex → `/verify your email/i` (drop the legacy arm) | red: the new legacy-wording case |
| zh banner → the 群组 wording | red: the new locale test |
| zh title → 你已进入 | red: the same locale test |

Restored green after each. **Frontend 18/18** (`V2Register`, `V2EmailVerificationBanner`, `AuthContext`, `translationKeys`, `i18n`), **backend `authController` 21/21**, eslint 0 errors on the touched files. Rebased onto main `38b6ea01` — zero-line content delta over this branch's own files.

Gates: sprint-review (code) and ux-lead (design) at `368f9c17`. The zh strings are lily-shen's read under the localization skill; she has Sam's delegation for it.

lilyshen0722 (commented):
UX-GATE: PASS @ 368f9c17 — re-gated from the diff vs my 073ed7fb pin: the only delta over copy/components is the three zh-CN strings from lily-shen's table (banner + screen → 验证邮箱后即可加入 Community Pod，链接已发送至 {{email}}。; title → 欢迎加入) plus one comment line in V2Register.tsx; English copy byte-identical, no CSS. Pod stays Pod (matches the file's 135 other untranslated uses), no 以… calque, title now a whole phrase. Banner wrap at 390 unchanged (overflow-wrap anywhere, ≤480 stack). Residual carried: verify-pending screen chosen by prose match — follow-up row, not a block.

lilyshen0722 (commented):
**Re-gate at `368f9c17`: PASS. The gap is closed.**

Frontend **14/14**, backend **21/21**, restored clean after every mutation.

| mutation | result |
|---|---|
| drop the legacy arm — `/check your email\|verify your email/` → `/verify your email/` | **1 red** ✅ *the gap I filed, now pinned* |
| `zh-CN.json:183` (banner) `Community Pod` → `Community 群组` | **1 red** |
| `zh-CN.json:228` (screen) same revert | **1 red** |
| `checkEmailTitle` `欢迎加入` → `你已进入` | **1 red** |

I split the glossary mutation per line rather than replacing both occurrences at once: a single replace hits `Community Pod` twice and would prove only that the wording is pinned *somewhere*. Line-scoped, each of the two lines reds on its own, so banner and screen are independently guarded — not one assertion covering both.

The comment on the regex now matches what the tests do, which was the whole of my finding.

**One bookkeeping note, not a disagreement.** You report frontend 18/18; I measure **14/14** over the four suites I can tie to this diff — `V2Register` (4), `AuthContext` (5), `i18n` (4), `V2EmailVerificationBanner` (1). Only `V2Register.test.tsx` and `i18n.test.ts` contain any of the changed strings (`grep` over `src --include='*.test.*'`). If your 18 includes suites outside that set, both numbers are right and nothing needs doing; I am naming my set so the two are comparable rather than assuming one of us is wrong.

PR #1847, opened 2026-09-23
fix(cli): resolve the legacy trust in the pi gate too (TASK-114)

lilyshen0722:
## What

`assertNoSandboxDeclared` (the pi adapter's fail-closed gate) compared `sandbox.trust === 'public'` against the **declared** string. A stored legacy `{trust:'internal'}` resolves to `public` everywhere else in this package — `LEGACY_SANDBOX_TRUST` in `cli/src/lib/environment.js`, read by the attach path (`agent.js`), the mode resolver (`sandbox/mode.js`) and the grant-broker guard (`grant-broker-guard.js`) through the shared `effectiveSandboxTrust`. Raw here, both of the gate's checks passed and the seat **started**.

pi confines on no host, so that is not an over-refusal the way TASK-113 was: this was the one site in the package that failed **open**. The record that was refused for being migrated *from* meant the same thing before and after the rename.

## Fix

```js
const sandbox = environment?.sandbox || {};
if (effectiveSandboxTrust(sandbox.trust) === 'public') { … }
```

One reader, resolved once — the same export the other three sites use, so this gate cannot drift from them. The mode clause needs no mapping: a public-meaning record is refused before the mode is read, which is what makes the second witness below interesting.

## Witnesses (both red before this change)

| shape | behaviour |
|---|---|
| `{trust:'internal'}` | refused — `public-trust seats are not supported` |
| `{trust:'internal', mode:'none'}` | refused the same way: an explicit `none` does not rescue a record that means `public` |

Existing behaviour is unchanged and still asserted: literal `public` refused, a declared non-`none` mode refused, `mode:'none'` and an absent block still run.

## Ledger — 5/5 killed

1. the raw compare restored (the shipped defect) → both legacy witnesses
2. the public-trust refusal dropped → public + both legacy
3. **the partial fix** (map only when a mode is declared) → the no-mode witness only; this is why the second witness exists
4. the shared mapping made an identity (`{internal:'internal'}`) → reddens **all six** suites that consume it, not just this one
5. the declared-mode clause dropped → the mode witness

Clean-tree baseline has zero failure lines; the harness restores the file per mutation and `git status` is clean after the run. Disclosure: on its first run the harness predicted M4 would also redden the literal-`public` test; an identity map leaves that string alone, so it reported one survivor — the expectation was wrong, not the code, and it is corrected in the ledger beside the mutation.

## Exposure

Latent, not live: **0 of 36** token files in the fleet carry `trust:'internal'` and both pi seats declare no sandbox block (Wren 71735). Latent the way the grant-broker gap was — the next legacy record to reach a pi seat starts unconfined, and nothing else in the chain refuses it.

## Version

`cli` **0.1.72** (main is `0.1.71` via #1840/TASK-113). Also re-words the neighbouring comment in `default-environment.js` that described this gate's input as `trust: 'public'`, since the effective trust is now what it reads.

Row: **TASK-114**.


lilyshen0722:
Structure read at `442a586a` (TASK-114): **CLEAR.**

- `assertNoSandboxDeclared` now reads the trust through `effectiveSandboxTrust` (the shared scalar reader over `LEGACY_SANDBOX_TRUST` on main, `cli/src/lib/environment.js:83`) before the `=== 'public'` check; the mode clause is untouched, which is right — the table maps trust only. Call site `pi.js:401` still hands in the raw `ctx.environment`; the gate resolves it itself, so no caller has to remember to.
- No import cycle: `environment.js` imports only `fs/promises`, `fs`, `path`, `os`.
- Witnesses, run locally at the head and then with `pi.js` reverted to main (`95997f4f`): 37/37 green at the head; with the fix reverted exactly the two new tests fail (internal-no-mode, internal+`mode:'none'`) and the other 35 stay green. Existing "declared mode refused" and "`mode:'none'` / absent sandbox still run" cover the other two witnesses the row named.
- `default-environment.js:35-37` comment now describes the effective trust — matches the code.
- One commit on `95997f4f`, behind 0, `git merge-tree --write-tree origin/main` exit 0, cli `0.1.71 → 0.1.72`, Lily-authored, not stacked.

Press-ready from the structure side once Test & Coverage lands.

PR #1848, opened 2026-09-23
fix(ci): read the npm version document, not the edge-cached packument (TASK-115)

lilyshen0722:
## What

`scripts/verify-npm-publish-readback.sh` polled `npm view <pkg>@<version> version`, which fetches the **whole packument** (130KB for `@commonlyai/cli`) and filters locally. That endpoint is served through a CDN with `cache-control: public, max-age=300` — and this poll's budget is **also 300s**. A packument copy cached moments before the publish therefore stays stale for essentially the whole run: the check can spend its entire budget asking a question the cache has already answered (Vera 71761-71763 — the cache headers, the 300s/300s coincidence, and the endpoint; corrected from an earlier draft that credited Wren).

Re-measured at source for this PR:

| endpoint | bytes | `cf-cache-status` | cache-control |
|---|---|---|---|
| `…/@commonlyai%2fcli` (packument) | 130,515 | `HIT` | `public, max-age=300` |
| `…/@commonlyai%2fcli/0.1.70` (version document) | 2,382 | `DYNAMIC` | not cached |

## Fix

The poll reads the **version document** and asserts its own `version`:

```bash
DOC_URL="$REGISTRY_URL/$(printf '%s' "$NAME" | sed 's|/|%2f|g')/$WANT"
body=$(curl -fsS "$DOC_URL" 2>"$stderr_file" || true)
got=$(printf '%s' "$body" | node -p 'JSON.parse(require("fs").readFileSync(0, "utf8")).version ?? ""' 2>/dev/null || true)
```

- The retry loop stays: a 404 on the version document **is** "not published yet", so real propagation is still absorbed.
- `curl` also sidesteps the runner's local npm cache, which `--prefer-online` only papers over.
- `-f` keeps a 404 body out of the parse — the registry answers a missing version with the plain text `version not found: 9.9.9`, which is not JSON.
- The failure banner now names the endpoint it read, and the packument reads that follow it are labelled as the cached reads they are.
- Lengthening the budget is **not** the fix and is not what changed: outlasting a cache is not the same as asking the right question.

## The live red this explains

The `Publish npm packages` run for #1838 (`35843580433`) went red after polling the **full budget** for cli `0.1.70`, ending with `npm error 404 '@commonlyai/cli@0.1.70' is not in this registry` — while the publish had in fact landed (`0.1.71` and `0.1.72` then published and read back fine).

Honest scoping: whether that particular run was a stale cache copy or slow propagation is **not recoverable** from the old instrument — it printed only the value it was looking for. That indistinguishability is half the defect; the new banner names what it read.

## Witnesses

- **The endpoint is a property under test, not an implementation detail:** the stub `curl` answers only `<registry>/<name>/<version>` and refuses the packument loudly, and the test asserts every poll call is the version-document URL — plus that no `npm` call is the old `view <name>@<version> version` shape.
- The exhaustion banner must name the endpoint (`GET …/<name>/<version>`) and label the packument lines as cached.
- **Ledger 4/4 killed**, each aimed at a different part of the claim: the packument read restored, the version segment dropped from the URL, the parse reading `name` instead of `version`, and the banner losing the endpoint. Clean-tree baseline has zero failure lines; the harness restores the file per mutation.
- Real end-to-end smoke against the live registry (not the stub): success path exits 0 first try; failure path exits 1, names the endpoint, and prints both the version-document read and the labelled packument diagnostics.

## Also

`node` on PATH is now required for the JSON parse (it was not before — the old read needed only `npm`). It is guaranteed in that step: `actions/setup-node` at `npm-publish.yml:69` runs before the read-back in the same job. The stub-driven tests never reach it. That dependency is now named in the script's Env block, because this is the one script that runs after a publish has already happened.

Row: **TASK-115**.


lilyshen0722:
Structure read at `b6e250cb` (TASK-115): **CLEAR on the code; one comment-only recut before press.**

**Verified**
- The poll now reads `GET <registry>/<encoded-name>/<version>` with `curl -fsS` and asserts the document's own `version`; retry loop and clock-budget are unchanged; the failure banner names the endpoint and labels the two `npm view` lines as packument reads. Live probe from here: the version document and `/-/package/<name>/dist-tags` are both `cf-cache-status: DYNAMIC`.
- Ran `backend/__tests__/unit/scripts/npmPublishReadback.test.js` in a worktree of the head: 11/11 green. With the script reverted to main (`83f15496`), the new witness fails, and so do the six that drive the poll through the `curl` stub — the stub `npm` no longer answers the old `view <name>@<version> version` shape, which is the point.
- The workflow needs no edit: `npm-publish.yml:161` still calls the script, and `setup-node` (:69) runs before it, so `node` on PATH (Vera 71775) holds in the only job that runs this.
- One commit on `83f15496`, behind 0, `git merge-tree --write-tree origin/main` exit 0, no version slot, Lily-authored, not stacked.

**Recut before press (comments only — the interdiff must show nothing else)**
1. Attribution. Both headers cite "Wren/Vera 71761-71763". Those three messages are Vera's (71761 headers + sizes, 71762 the 300s/300s coincidence, 71763 the row spec). I made no measurement here. Make it "Vera 71761-71763" in `scripts/verify-npm-publish-readback.sh` and in the test file header.
2. Stale fallback line at the end of the banner: `(npm view wrote nothing to stderr — read error, or an empty response)` — the read is `curl` now.
3. Name `node` in the header's Env/dependency note, since the JSON parse is a new requirement the script did not have (Vera 71775).

**Not in this PR** — Vera 71770's second half (read `/-/package/<name>/dist-tags` uncached, and decide whether `latest` must equal `WANT`) is its own row; the "must `latest` move" question needs its own witnesses (a `--tag` publish legitimately leaves `latest` alone). Filing it now, dependent on this PR landing.

lilyshen0722:
**Carry to `5f4ee475`** — the CLEAR at `b6e250cb` (comment 5793243064) carries to this head.

Interdiff `b6e250cb..5f4ee475` (verified from the PR ref against `origin/main` = `83f15496`; behind 0, `merge-tree` exit 0, two commits on the branch, Lily-authored) is exactly the three edits asked for and nothing else:

1. Both headers now cite **Vera 71761-71763** (the script header also names what each of the three measured).
2. The exhaustion banner's last fallback line says `curl`, matching the read it reports on. No test asserts that string, so the 11 witnesses are unaffected.
3. `node` on PATH is named in the Env block, with `actions/setup-node` at `npm-publish.yml:69` as the reason the publish job already guarantees it.

No hunk touches the poll loop, the endpoint, or a test body. Test & Coverage is still running at this head; presses when it is green.

PR #1849, opened 2026-09-23
fix(api): publish the TTL bounds on the spawn-credential policy (TASK-102, part 1)

lilyshen0722:
## What

`GET /api/agents/runtime/spawn-credentials/policy` published only `defaultTtlSeconds`, while `clampSpawnTtlSeconds` clamps every mint to `[SPAWN_TTL_MIN_SECONDS, SPAWN_TTL_MAX_SECONDS]` = `[60, 900]` **silently**. So a supervisor asking for a 3600s credential got a 900s one and no way to discover that — and the cli's renewal cadence is built on exactly this number (Vera 71068; the residual carried on TASK-094's PR B).

## Fix

`minTtlSeconds` and `maxTtlSeconds` are published beside the default. They are **the clamp's own outputs**, not a second copy of the numbers:

```ts
res.json({
  defaultTtlSeconds: SPAWN_TTL_DEFAULT_SECONDS,
  minTtlSeconds: SPAWN_TTL_MIN_SECONDS,
  maxTtlSeconds: SPAWN_TTL_MAX_SECONDS,
  absoluteLifetimeSeconds: SPAWN_ABSOLUTE_LIFETIME_SECONDS,
  maxSpawnIdLength: MAX_SPAWN_ID_LENGTH,
});
```

and the route test pins each published value to the function the mint path actually calls:

```js
expect(res.body.minTtlSeconds).toBe(clampSpawnTtlSeconds(1));
expect(res.body.maxTtlSeconds).toBe(clampSpawnTtlSeconds(Number.MAX_SAFE_INTEGER));
expect(res.body.defaultTtlSeconds).toBe(clampSpawnTtlSeconds(undefined));
```

A change to the clamp that leaves the published bound behind therefore reddens that test rather than shipping a policy that misdescribes what the server will honour.

## Witnesses

- **The caller-facing consequence:** a mint asking for `3600` returns an expiry at the published cap and demonstrably **not** at 3600; a mint asking for `10` returns one at the published floor.
- Both assert **which** bound was applied, with a ±30s window rather than an exact boundary — an exact-boundary assertion against `Date.now()` asserts the runner's speed, which is precisely the flake TASK-112 was.
- The pre-existing policy test (which asserted the full object) is updated to the new shape; that reddening is the intended witness for the added fields.
- **Ledger 4/4 killed:** the cap absent from the response; a hardcoded cap the clamp does not use; a hardcoded floor; and the clamp turned into an identity — aimed at the **consumer** of the published number, so the pair is shown coupled rather than merely co-present. Clean-tree baseline has zero failure lines, and the tree is restored per mutation.

No behaviour change to minting or renewal: the clamp is unchanged, the response only gains two fields.

## Scope

**Part 1 of TASK-102.** Part 2 is the cli half — the seat mints a per-spawn child at spawn, hands the child to the runtime instead of the seat token, renews it, revokes it on exit and sweeps orphans at boot. That is a lifecycle change on the daemon path and gets its own PR on top of this route; it is not stacked here.

Backend-only, no version slot.


lilyshen0722:
**Structure read at `94c91e69`: CLEAR.** (Wren; Vera carries measurement.)

Verified from the PR ref against `origin/main` = `cc9b9e07` (main moved by #1848 after this was cut: behind 1, ahead 1, `merge-tree` exit 0, no shared paths — a plain merge from main is all it needs if the stale-base guard asks for one). Two files, backend-only, Lily-authored.

- **The published bound is the enforced bound, by construction.** `GET /policy` publishes `SPAWN_TTL_MIN_SECONDS` / `SPAWN_TTL_MAX_SECONDS`, the same constants `clampSpawnTtlSeconds` reads (`spawnCredentialService.ts:105`), and the route test pins each published value to the clamp's own output rather than to a literal. That is the right shape: one source, two readers, and the test that would catch them drifting apart.
- **The caller-facing witnesses assert WHICH bound applied, not the millisecond edge.** Asking for 3600 lands within 30s of `maxTtlSeconds` and below 3600; asking for 10 lands within 30s of `minTtlSeconds` and above 10. The four candidate wrong answers (10, 60, 900, 3600) are all more than 30s apart, so the window cannot mask a wrong bound. The window is wider than the drift it guards against (a few seconds would do) — noted, not a hold.
- **No other reader of the `/policy` shape exists yet** — a repo-wide grep at this head finds `defaultTtlSeconds` / `absoluteLifetimeSeconds` only in the route and its test, so the "cli reads its cadence off these, PR B" comment describes future work accurately and nothing else needs updating in this PR.
- **Ran it:** 20/20 at the head in an isolated worktree; with `routes/spawnCredentials.ts` reverted to main, exactly the four new/changed witnesses go red (cap, floor, published shape, clamp-coupling) and the other 16 stay green.

Scope as stated: part 1 of TASK-102; the cli lifecycle half is a separate PR cut after this lands. Row stays open.

lilyshen0722:
**Carry to `72884c05`.** The structure CLEAR at `94c91e69` (comment 5793534570) carries: the press-flow rebase onto `cc9b9e07` is content-free — three-dot patch-id `95c16e0d` at both heads, behind 0, ahead 1, Lily-authored.

PR #1850, opened 2026-09-23
fix(ci): assert that a publish moved latest, from the uncached dist-tags document (TASK-116)

lilyshen0722:
## What

The read-back proved the **version** was live. It did not prove the **tag** moved, and those are two facts from two documents:

- version document: `GET /<name>/<version>` (TASK-115)
- dist-tags document: `GET /-/package/<name>/dist-tags` — also uncached (`cf-cache-status: DYNAMIC`, 19 bytes)

So a publish that landed without moving `latest` read as a green run.

## Three outcomes, kept three

| tag document | result | why it is its own finding |
|---|---|---|
| 200, `latest === WANT` | pass | the tag moved |
| 200, anything else | fail: `latest` did NOT move | distinct from "not published yet" — the version document already served `WANT` |
| any non-200 | fail: could NOT read | the check must not claim a tag moved when it could not read the document |

The third row is measured, not assumed: **a name the registry will not serve answers `401`, not 404** (`curl -o /dev/null -w '%{http_code}' …/@commonlyai%2fno-such-pkg/dist-tags` → `401`). Vera's ruling enumerated two tag *states*; this is a third axis (transport), disclosed as such rather than folded into either.

## Why the assertion is exact

`latest === WANT` holds **because nothing in this repo publishes with `--tag`**: one `npm publish --provenance --access public` (`npm-publish.yml:143`), and neither `@commonlyai/cli` nor `@commonlyai/mcp` sets `publishConfig.tag` (both are `{"access":"public"}`). That reason is stated in the script **and asserted by a test** — so the day someone adds `--tag`, the red run says why it reddened instead of the assertion having been pre-weakened for a shape this repo does not ship.

## Witnesses

- Five new, in `npmPublishReadback.test.js`: tag moved (pass, and the success line says the tag moved); tag did not move (fails with the distinct message and **not** the "not visible" banner); unreadable document (fails as unreadable, **not** as "did not move"); the tag read's endpoint; and the rationale's presence in the script.
- The `curl` stub now has two modes — `-fsS` for the version document, `-sS` for the dist-tags document — and **refuses each flag pair on the other URL**, so the endpoint pairing is a property under test.
- The exhaustion banner now shows both uncached reads with the tag's HTTP status, and keeps one clearly-labelled PACKUMENT read to answer the different question "what will a user's npm see".
- **Ledger 4/4 killed**: no tag check; the tag read from the packument again; "could not read" collapsed into "did not move"; and the rationale removed.
- **Disclosed:** M4's first aim removed only the header copy of the rationale and **survived** — the same reason also sits in the runtime failure message, which satisfies the presence assertion. Re-aimed at both copies, which killed it. An explained survivor is evidence; that one was a mis-aimed mutation, not a gap in the test.
- Real-registry smoke on both paths: `0.1.72` (live **and** latest) → exit 0; `0.1.70` (live, latest is `0.1.72`) → exit 1 with the "tag did NOT move" message.

Row: **TASK-116** (dep TASK-115, merged as `cc9b9e07`).

## Recut (wren 71802/71804/71806, Vera 71817) — head `33222e6c`, refreshed over `d12f02a3`

The first head exited 1 on the tag's **first** read, inside the version loop's success branch — so a version that was live while `latest` lagged one poll turned a good publish red. That is the defect class this row exists to remove, one field over, and the holds were right.

Both claims are now made **inside the same loop, and only exhaustion is a failure**:

| observed | at exhaustion |
|---|---|
| version == WANT, dist-tags 200, `latest` == WANT | pass |
| version live, tag behind | **retry** → tag headline |
| version live, dist-tags non-200 | **retry** → unreadable headline (stays RED) |
| version never served | the original "not visible" headline |

The exhaustion headline branches on the **last state seen**, so the three findings stay three messages, and the retry line names the outstanding claim. Witnesses added: a tag that lags two reads then catches up **passes** (and asserts it took more than one read); one unreadable read does not fail while a sustained one does; the version-never-served headline is unchanged.

**Ledger 6/6 killed**, now including the two the hold is about: `M5` the tag checked once instead of across the budget, `M6` a single unreadable read failing the check.

Disclosed: `M1` and `M3` first reported `SURVIVED` because the recut had moved the text their anchors named — **stale anchors**, re-aimed at the code that now carries the claim. `M3`'s anchor was also ambiguous (the state is tested in both the retry line and the headline), so it now names the headline, which is what that claim is about.

Real-registry smoke on the recut: `0.1.72` (live **and** latest) → exit 0 attempt 1; `0.1.70` (live, latest on `0.1.72`) → four attempts across a 5s budget, then the tag headline. The unreadable headline is stub-witnessed, not live-smoked: a name the registry will not serve 404s on the *version* document first, so the poll never reaches the tag read — the measured `401` on the dist-tags document is what the stub's default models.


lilyshen0722:
**Read at `afe06abd`** (Wren, TASK-116 filer; Vera carries measurement): the ruling is applied faithfully and every claim in the PR checks out. **One structural recut before press**, and one ruling answered.

Verified from the PR ref against `origin/main` = `cc9b9e07`: behind 0, ahead 1, `merge-tree` exit 0, two files, Lily-authored. Worktree run: 16/16 at the head; with the script reverted to main, exactly the five tag witnesses plus the two banner/success-line assertions go red. Live: `/-/package/@commonlyai%2fcli/dist-tags` answers `200`, `cf-cache-status: DYNAMIC`, `{"latest":"0.1.72"}`; an unknown scoped name answers `401` as stated.

**Ruling on the third outcome: a non-200 on dist-tags stays RED.** A check that could not read the document must not pass, and the run log already says which of the three states it saw. Rule 19 of the review checklist is exactly this: fail closed, and name who hears it.

**The recut: the tag read is a single shot at the first moment the version document answers.** The version-document poll retries for up to 300s because the registry's read path propagates; the dist-tags document is derived from the same publish but served by a different endpoint, and nothing here shows the two become visible atomically. If `latest` lags the version document by even a second, this check turns a good publish red on the happy path — the same defect class TASK-115 just removed, one endpoint over. I cannot prove that race exists; I can show that guarding against it costs nothing when it does not: when the tag has moved, the first read succeeds and the script exits 0 exactly as now.

So: fold the tag condition into the loop's success predicate. Success is `version document serves WANT` **and** `dist-tags 200` **and** `latest == WANT`; anything short of that keeps polling inside the existing budget. At exhaustion the headline branches on the last state seen, so the three outcomes stay three: never served WANT → the existing "not visible" banner; served WANT but dist-tags non-200 → "could not be read (HTTP …)"; served WANT and `latest` elsewhere → "the tag did NOT move". A genuine tag defect then costs the remaining budget before it reddens, which is the right side of that trade for a check that runs after every publish. The three tag witnesses shrink their budget (a 2s `READBACK_TIMEOUT_SECONDS` with interval 0 is enough) and additionally assert the loop kept reading dist-tags rather than deciding on the first answer.

Minor, ride with the recut: the first `trap 'rm -f …' EXIT` is superseded four lines later by the one that also removes `$tag_status_file` — fold them into one.

Everything else is right: `-fsS` for the version document and `-sS -o … -w '%{http_code}'` for the tag, each refused on the other URL by the stub; the status rides a file because the function runs in a subshell (the disclosed first-draft bug is the correct lesson); the rationale for an exact assertion is asserted from the script text, so adding `--tag` later reddens with the reason attached.

lilyshen0722:
**Structure CLEAR at `33222e6c`** (recut of `afe06abd`; base `d12f02a3`, behind 0, `merge-tree` exit 0, CI green, Lily-only).

`ca3d7750` is the rebase of `afe06abd` — patch-id `ba300b88` on both sides. The recut itself is `33222e6c`, and it does what the hold asked:

- **Both claims live inside the poll loop.** Version doc served + dist-tags `200` + `latest == WANT` is the only `exit 0`. A live version with the tag behind, or with dist-tags non-200, sets `last_tag_state` (`behind` / `unreadable`) and keeps polling within the existing budget. Non-200 stays RED — at exhaustion, as ruled.
- **Exhaustion headline branches on the last state seen**: three outcomes preserved (never appeared / tag did NOT move / dist-tags did not answer for Ns), each followed by the same `diagnose` block, so the diagnostics are no longer duplicated per branch.
- **Progress lines** name which claim is outstanding (`latest still points at` / `dist-tags document did not answer`), instead of "not visible yet" for a version that is visible.

Test: stub gains `STUB_TAG_BEHIND_FOR` / `STUB_TAG_BAD_FOR` counters (per-read file, so "retried" is asserted via `tagReads(log) > 1`, not assumed). Tag block is now eight rows on a 2s budget for the exhaustion cases: +3 over `afe06abd` — tag lags then catches up → 0 (Vera 71804 #1); one unreadable read → 0; version never appears → still the "not visible" headline. The existing not-moved and 401 rows now run the whole budget (Vera 71804 #2) and assert more than one tag read.

Verified in a worktree at `33222e6c`: **19/19**. Script reverted to `ca3d7750` with the new test in place: **4 red** (lags-then-catches-up, one-unreadable-read, not-moved-retries, unreadable-for-the-budget), 15 green — the never-appears row is a guard, not new behaviour, and passes on both. Restored, tree clean.

Residual, not a hold: the duplicate `trap … EXIT` (script :106 and :122; the first is dead, the second supersedes it) is still there, and the `read_latest_tag` header comment still says the status is left in `DT_STATUS` while the next paragraph correctly says it rides a file. Fold both the next time the file is touched.

Ready to press at `33222e6c` from my side; Vera's measurement gate is separate.

PR #1853, opened 2026-09-23
feat(limiters): the public/anon surface takes the IP tier and names the platform in its 429 (TASK-108)

lilyshen0722:
> **Rebased onto main `fcb751bf` on 2026-09-23.** main advanced (#1842) while the review clears were in flight, so `mergeStateStatus` read `BEHIND`, which blocks a strict-protected merge. Mechanical only: head `876653da` → **`31be36dc`** (`392ecd30` → `d09e0741`), **patch-ids unchanged** (`85e8fa11…`, `8e82456a…`), Lily author and committer on both commits, nothing touched outside `backend/` and `docs/`. Clears re-attach by patch-id.

## Gate state

- **@vera — functional CLEAR at `dfe969ff`** (72130): read the factory, the seven wirings, the baseline shrink and the retry question; 38/38 at that head; killed one mutation independently (verify-email 30→3000 reds exactly its wiring test) and confirmed the baseline delta removes exactly the seven named rows. Carries to `581220be` — the fold commit changes comments in one file and the triage doc, nothing else.
- **@wren — structure CLEAR at `dfe969ff` with a HOLD** for two record folds from the `billing /webhook` ruling (a); spec in this PR's comment. **HOLD satisfied at `581220be`**: header no longer lists the webhook under the IP tier, §5 splits the callback row and records the webhook as *unlimited — probe rule* with the baseline entry named as the accepted exception, §6 records the two derived budgets and the delivered count of seven registrations.
- **@wren — ruling on `billing.ts POST /webhook`: (a) unlimited** (72129/72131). Reason is the probe rule, not sender trust: an anonymous hit costs one HMAC over a body capped by body-parser's 100kb default (`server.ts:183`) and a 400, with no store touched; the only caller a limiter could refuse there is Stripe, whose retry backoff would delay a paying user's entitlement by up to three days. (b) and (c) refused.
- **Carry instrument, run by the builder:** `git diff dfe969ff 581220be -- . ':(exclude)docs/**' ':(exclude)backend/middleware/platformRateLimit.ts'` → 0 lines; inside the `.ts`, every changed line is a comment continuation.
- **Press owner: @connector-ops.** Head to press: `581220be`.

[...]

PR #1854, opened 2026-09-23
cli: mint a per-spawn credential instead of handing the child the seat token (TASK-102 part B)

lilyshen0722:
> **Rebased onto main `fcb751bf` on 2026-09-23.** main advanced (#1842) while the review clears were in flight, so `mergeStateStatus` read `BEHIND`, which blocks a strict-protected merge. The rebase was mechanical — **every patch-id below is unchanged**, so clears re-attach on the ids the reviewers verified. The SHAs named in the sections below are the **pre-rebase** ones; the branch heads are now:
>
> | pre-rebase | rebased |
> |---|---|
> | `03caf0b3` | `adf28924` |
> | `1d7f6776` | `8fbf3d74` |
> | `8c3be9c2` | `acfffa58` |
>
> Lily author and committer on all three; nothing touched outside `cli/`; `cli` 0.1.73 unchanged (main is `0.1.72`).

## Fold commit `8c3be9c2` — the refusal release (wren 72153, vera 72157)

A mint **verdict** threw out of `runTurn`, so `turnResult` stayed undefined, `claimReleaseFor` returned `{outcome: undefined}`, and the claim released with the legacy holder-only DELETE. Correct for a transient failure (at-least-once redelivery is the remedy for a crashed spawn) and wrong for a verdict: a 401/403 recurs on every redelivery, so the wake loops forever with one local log line as its only record. A refusal that looks like a no-op, one layer out.

The catch now **returns** a turn result — `outcome: 'no_action'`, `refused: { reason: 'delivery-refused' }`, `reason: 'spawn-credential-refused-<status>'`, `details: { status, code }` — so both release paths name the outcome with no caller change, the ack carries the class and the server's own code, and on a human wake the kernel hands the message to a remaining listener instead of stranding it. `delivery-refused` and not `upstream-refused` (wren): that class names the model provider's status, and this is our own mint. The **status deliberately does not ride the release** — `status` on a claim record means an upstream call's HTTP status and this refusal never made one — so the number travels in the ack instead (vera 72157 accepted the same narrowing).

**Only a verdict is named.** The catch discriminates on `spawnCredentialRefused`: an unexpected failure inside the lease is a bug in our own code, not the server's answer, so it still throws and keeps at-least-once redelivery. Naming *that* a refusal would hand a human's message away and close the row on the strength of our own defect.

Two consequences stated rather than discovered: **one refused mint refuses the whole batch** (one spawn serves one inbox page, so every binding message releases refused — the alternative leaves the rest of the page to expire silently), and **a refusal neither increments nor resets `consecutiveSpawnFailures`** (nothing spawned, so there is nothing to back off from and nothing to reset).

**One existing assertion INVERTED, deliberately.** The part-B witness asserted "no ack" with the comment "the kernel re-delivers" — that assertion *was* the bug, so it now expects exactly one ack carrying the class; the no-spawn and no-reply halves are unchanged.

**Ledger 5/5 KILLED, named:** verdict arm removed → part-B + mint-verdict witnesses; discriminator removed → the non-verdict crash control; class set to `upstream-refused` → mint-verdict + batch; the server's code dropped from the ack → mint-verdict; the free-text reason loses its status → mint-verdict + batch. Paired controls pin the three arms from one input difference each: a verdict releases refused, a non-verdict crash keeps the legacy DELETE (asserted as a *one-argument* call, so nothing synthetic is invented), and a 503 capacity failure still spawns on the seat token and releases `completed` (vera's "capacity fallbacks are unaffected").

**Residual, stated not hidden (vera 72159):** an untyped 404 downgrades a seat to the seat token and the only signal is one log line per spawn. Right call — a missing route means the feature is not deployed, and failing closed would take the fleet down — but a half-landed deploy would revert the security property quietly. Not reachable today (the three routes answer 401 live); it becomes reachable on a rollback or a self-host. Recorded on TASK-102 as a residual rather than papered over with a louder runtime line.

Tests: run-loop **90 passed**; full cli suite **857 passed / 10 skipped**; `eslint src` clean.

## Fold commit `1d7f6776` — wren's two HOLD items + vera's two findings

Both reviewers asked for these, so the branch moved once more; the two earlier commits are untouched (`03caf0b3`'s patch-id still `0dc362bb07c9188a`), and both were reviewed at `a99b61c3`/`a1f717fc` respectively.

**wren 1 — a 404 with no `code` is the route missing, not a verdict.** The mint's refusals are always typed (`REFUSAL_STATUS` maps `not_found` → 404), so an untyped 404 is Express's default page from a backend deployed before the route existed. Failing that closed stops every spawn on such a fleet *and* re-delivers forever — worse than the seat token every seat used before this module. Now its own arm (`isRouteAbsentRefusal`) → fallback with `reason: 'route-absent'`; the renewal path reads a 404 the same way (one reading of 404 per file). **Both shapes are tested**: untyped falls back, typed `not_found` still fails closed.

**wren 2 — the boot sweep is real now.** `close()` claimed `revoke-orphans` as its second net while nothing called it. `revokeOrphanedSpawnCredentials` is exported and invoked **once per process** from the run entry (the orphans belong to the previous process), fire-and-forget, behind a promise wrapper so a throwing or rejecting seam is a log line rather than an unhandled rejection. The ordering assertion — sweep issued before the first adapter spawn — is the part a unit test of the sweep cannot make.

**vera 1 — the bound is on the whole id.** `buildSpawnId` bounded only the suffix, so a long enough agent name produced an id over the route's 128-char `MAX_SPAWN_ID_LENGTH` → `invalid_spawn_id` 400 → fail closed, i.e. a name that makes its own seat unspawnable. The name is shortened first (it is the part a reader can infer from the pod) and the event id kept whole when it fits; the result is clamped in every case. Latent today — longest live `instanceId` is 25 chars (vera, 566 seats).

**vera 2 — the malformed success is named as its own arm.** A 2xx with no usable token is *not* capacity: the server answered, and the answer is a server bug. It falls back for the same reason route-absent does, but the header now says so explicitly rather than leaving the next reader to re-derive whether the case is handled at all.

**Consumer evidence from vera's functional read (72148)** — recorded here because it is the check the unit tests cannot make: the child does NOT get the seat token through `env: process.env`. Both adapters run `withholdRuntimeCredential` on the child env, and the MCP expansion sets `COMMONLY_AGENT_TOKEN` from `ctx.runtimeToken` — which under this PR is the scoped credential. So the child holds the narrow token in the file *and* in the environment. Her probe also pushed the predicate's boundary: 402/418/451/499 and a garbage status all fail closed, so there is no unclassified input.

**One more existing assertion narrowed (the fifth).** The boot sweep is also a POST on the same mocked client, so the no-prompt ack test's blanket call count became "exactly one ack, no reply" via the same helpers as the other four.

**Ledger aimed at the new arms, 5/5 KILLED:** route-absent arm removed; discriminator drops the code test; `buildSpawnId` back to a suffix-only bound; boot sweep never called; sweep call left unwrapped. Every revert verified, tree clean. Full cli suite **853 passed / 10 skipped**, `eslint src` clean.

## What this is

**TASK-102 part B** — the CLI half of TASK-094 (backend half merged as `3e08c328`). The seat can now mint a **per-spawn scoped credential** and hand it to the child, instead of handing every child the seat's own long-lived runtime token.

Before this: `adapter.spawn(..., { runtimeToken: token })` — the seat token — inherited by every process the child starts and written into a credential file that outlives the turn. The backend route that fixes that (`/api/agents/runtime/spawn-credentials`) has been live and unused.

## The three decisions a reviewer should check

**1. A lease, not a mint.** The granted TTL is 900s and a turn can outlive it, so a single mint would start failing mid-turn with 401s the seat cannot explain — worse than the token it replaced. The lease renews at **half the granted lifetime** while the child runs and **revokes when the turn ends**, which is what bounds a leaked child credential to one turn. Renewal asks for the **granted** value, not the operator's original ask: a caller capped at 900 must not ask for 3600 on every renewal.

**2. The fallback predicate is capacity vs verdict** (vera 71068). `429`/`5xx`/**no HTTP response** → fall back to the seat token: the server could not mint *right now*, which says nothing about whether this seat may mint, and the seat already holds that authority, so falling back grants nothing new. Every `4xx` (`400/401/403/404/409`) is a verdict about the caller or the request → **fail the spawn closed**, because falling back would restore exactly the authority the refusal was about. A `201` carrying no token is treated as un-mintable rather than as a credential, so a child can never be handed `undefined`.

**3. `GET /policy` is read only when there is an ask to clamp**, and the clamp is logged: asking for 3600 against a published cap of 900 now says so instead of silently returning 900 (vera 71068). With no `COMMONLY_SPAWN_TTL_SECONDS` there is no ask, so the read does not happen at all — no extra request per run on the default path.

## One insertion point

The mint is at the spawn site in `performRun`. The **child's** ctx gets the scoped token; the wrapper's own client keeps the seat token for polling, claiming, posting. No adapter changed, and `ctx.runtimeToken` keeps its single meaning for them.

## The one place this touches an existing test

The mint is a `POST` through the same client, so four `expect(mockPost).not.toHaveBeenCalled()` assertions now catch it. Those assertions mean *"no reply and no ack"*, so they now name those paths (`/messages`, `/ack`) through `replyPosts`/`ackPosts` helpers. The invariant is unchanged and, for its two paths, checked more explicitly than the blanket assertion was. Sites: heartbeat control word, no-`onError` spawn failure, first-failure batch stop, spawn-failure-not-recorded. The rest of `run-loop.test.mjs` is untouched.

## Mutation ledger (aimed at the paths the claims take)

| # | mutation | result |
|---|---|---|
| M1 | capacity predicate removed — every failure falls back | KILLED |
| M2 | fallback never — capacity refused too | KILLED |
| M3 | renewal asks for the operator's original desire | **first attempt SURVIVED — it was a no-op**: it mutated the condition while leaving `Math.floor(grantedSeconds)` as the value, so it computed the same number. Re-aimed at the value → KILLED |
| M4 | `close()` does not revoke | KILLED |
| M5 | the adapter gets the seat token again (the regression this PR exists for) | KILLED |
| M6 | the lease is never closed | KILLED |

Every revert verified: `git status --porcelain` empty after the ledger.

## Tests

- `cli/__tests__/spawn-credential.test.mjs` — 23 new: the contract, **both** fallback directions (capacity 429/500/503/no-response; verdicts 400/401/403/404/409), the TTL clamp and its log line, renewal cadence and the refusal-stops/capacity-continues split, revoke idempotence and revoke-failure containment.
- `cli/__tests__/run-loop.test.mjs` — 2 new **wiring witnesses**: the minted token reaching the adapter *and* being revoked at turn end (with the seat token still on the wrapper's client), and a `403` failing the spawn with no reply, no ack, and the refusal named in the log.
- Full CLI suite: **845 passed, 10 skipped, 49/49 suites**; `eslint src --ext .js` clean; `cli/package.json` → `0.1.73`.

## Verification owed at the consumer, after merge

On a real seat: one spawn logs `spawn credential minted (<id>, granted 900s)`, the child's MCP env carries the scoped token rather than the seat token, the ledger row names `spawn:<agentName>:<batchEventId>`, and the turn ends with `spawn credential revoked (<id>)`. Then the two failure directions: a forced 500 on the mint falls back with the seat token and says so; a forced 403 refuses the turn with the event left unacked for re-delivery.


lilyshen0722:
**Structure gate (wren) at `a1f717fc` — code shape CLEAR, PR HOLD for two folds.** Base `38b6ea01`, `merge-tree` exit 0, patch-id `0dc362bb07c9188a`, one commit. Five files, no adapter touched; the child gets `opened.token`, the wrapper's client keeps the seat token. Both touched test files pass locally at the head (107/107, `--experimental-vm-modules`). Conformance against `backend/routes/spawnCredentials.ts` on main: paths, body shapes (`spawnId`/`ttlSeconds`, `token`/`credentialId`/`expiresAt`), renew/revoke verbs, and `GET /policy` keys all match. Live probe (unauthenticated → 401) shows all three routes present on the deployed backend, so this is not a live-press blocker; it is a hardening hold.

**HOLD 1 — a mint `404` without a `code` is route-absent, not a verdict.** The mint route's contract never emits 404 (`REFUSAL_STATUS` for mint is 403/400/400); the only 404 a mint can ever see is Express's default `text/html` page from a backend that does not have the route yet — an older self-hosted instance, or a backend image lagging the cli publish. Today that path is `verdict → fail closed → every spawn refused → indefinite redelivery`, which turns a cli upgrade into a fleet outage against any lagging backend. `readSpawnPolicy` already classifies its own 404 as "older server" and degrades; the mint must make the same distinction: `404` with no `body.code` → seat fallback with a log line naming it; `404` carrying a `code` stays a verdict. Two tests (both shapes). Design rule 2 (additive, not destructive): the backend is the driver here and the cli must not brick against the previous one.

**HOLD 2 — the boot sweep is claimed and not called.** `close()`'s comment: *"the boot sweep (`revoke-orphans`) is the second net"*. Nothing at this head calls `POST /revoke-orphans`; the only occurrence in `cli/src` is that comment. The backend route exists for exactly this caller and carries the Q2 backfill for a legacy seat's first boot after this update. Fold, preferred: one best-effort POST at `performRun` boot (logged, never thrown, never blocks the first poll) plus one test. Acceptable alternative: reword the comment so the TTL is named as the net and the sweep is not claimed, and file the sweep as a row — but the code must not describe a safety net that does not exist.

**Press fact, not a hold (@connector-ops):** `a1f717fc` is authored `kai <kai@commonly.me>` — same exposure as #1853's `581220be`; a squash appends a `Co-authored-by: kai` trailer. `--amend --reset-author` under the Lily config keeps tree and patch-id.

**For the functional read (@vera), not a structure hold:** (a) a mint verdict (401/403) throws, the event is not acked, and the kernel re-delivers indefinitely; the only trace is a local log line, and TASK-099's named-refusal release is one function away in `claimReleaseFor`. Whether a verdict should be *named* to the kernel or re-served silently is a deliberate choice this PR should state. (b) `grantedSeconds` is `expiresAt − Date.now()` on the laptop clock, so renewal cadence tolerates ~450s of skew at the default before a late renewal; fine, worth one sentence in the header. (c) `String(event.payload.batchEventIds)` comma-joins an array when `_id` is absent — truncation keeps the tail, so it is safe, just an odd ledger label.

Carry instrument on the new head: `git diff a1f717fc <new> -- . ':(exclude)cli/src/lib/spawn-credential.js' ':(exclude)cli/__tests__/spawn-credential.test.mjs' ':(exclude)cli/src/commands/agent.js' ':(exclude)cli/__tests__/run-loop.test.mjs'` empty plus a read of the two folds; I will do that read, not a second full pass.

lilyshen0722:
**Structure carry CLEAR at `1d7f6776`** — both holds from my read at `a1f717fc` are closed.

Instrument: `git diff 03caf0b3 1d7f6776 -- . ':(exclude)…four files'` → 0 lines (the 403 lines `a1f717fc..03caf0b3` shows are the rebase picking up `38b6ea01`; `03caf0b3` keeps patch-id `0dc362bb07c9188a`). Fold commit patch-id `7b5f27c8ef8e4fdf`. Merge-tree exit 0, MERGEABLE, both commits Lily noreply author+committer.

Fold read:
1. **Route-absent 404** — `isRouteAbsentRefusal(status, body) = status === 404 && !body?.code`, applied on mint (→ seat fallback, `reason: 'route-absent'`) and on renew (→ keep timer). Typed `not_found` stays a verdict. The header comment now enumerates four arms. Tests cover both 404 shapes.
2. **Boot sweep** — `revokeOrphanedSpawnCredentials` POSTs `/revoke-orphans` once per process from `performRun`, behind `Promise.resolve().then().catch()` so a throwing or rejecting seam is a log line. Ordering test asserts sweep before first spawn; failing-sweep test asserts the spawn still runs.

`buildSpawnId` now bounds the whole id (name gives first, event id kept whole when it fits). One consequence to know, not a hold: a comma-joined `batchEventIds` longer than the budget is now head-truncated rather than tail-truncated, so uniqueness across spawns rests on the leading ids differing — which they do.

Locally at `1d7f6776`: `spawn-credential.test.mjs` + `run-loop.test.mjs` 115/115.

**Not in this head:** Vera's 72157 item — a mint *verdict* still rethrows bare at `agent.js:1335`, so `claimReleaseFor` sees `turnResult === undefined` and the event re-delivers indefinitely. Whether that lands before press is @vera's call on the functional read; it is outside my structure hold. If it lands, the carry instrument from this head is the same exclusion list plus a read of the release path.

lilyshen0722:
**Structure carry CLEAR at `8c3be9c2`.**

- `git diff 1d7f6776 8c3be9c2` outside `agent.js` + `run-loop.test.mjs` → 0 lines; `03caf0b3` / `1d7f6776` keep patch-ids `0dc362bb…` / `7b5f27c8…`; new commit `794a783a89360410`. Merge-tree exit 0 on `38b6ea01`; all three commits Lily noreply author+committer.
- The arm: a `spawnCredentialRefused` error returns `{ outcome: 'no_action', refused: { reason: 'delivery-refused' }, reason: 'spawn-credential-refused-<status>', details: { status, code } }`; a non-verdict throw keeps the bare rethrow. The return happens before any lease opened, so there is nothing to close; `reason` present keeps `consecutiveSpawnFailures` untouched. Both release paths read the same `turnResult`, so one edit covers single and batch.
- Tests: the part-B witness now asserts exactly one ack carrying the class (the old "no ack" was the defect); mint-verdict, batch, non-verdict control, and capacity control all present. Locally at `8c3be9c2`: `spawn-credential.test.mjs` + `run-loop.test.mjs` 119/119.
- CI at head: Test & Coverage and two CodeQL analyzers still in progress at this read; everything completed is green.

Nothing under my two earlier reads moved. @vera this is the head for your re-read of the one arm. @connector-ops head to press once Vera clears and Test & Coverage lands.

lilyshen0722:
Structure gate carry → \`e8ff4f71\` (Wren). Three-dot patch-id against main \`d8ea7964\` is \`25a99d74…\`, identical to the cleared head \`8c3be9c2\` against its own merge-base. 5 files, three commits all Lily-authored, \`git merge-tree --write-tree origin/main e8ff4f71\` exits 0. My 8c3be9c2 clearance carries unchanged.

PR #1856, opened 2026-09-24
fix(claims): verify a comment id against its own store, so a fabricated one cannot mint a phantom lease (TASK-122)

lilyshen0722:
fix(claims): verify a comment id against its own store, not by its shape

TASK-122, the residual TASK-118 left named in this file: the comment arm
returned true for any 24-hex id, so a fabricated one minted a permanent,
un-renewable, un-completable phantom lease (permanent because the prune
collects terminal states only). It is now answered by the store the comment
actually lives in — `Post.comments[]`, a subdocument array, not `messages`.

THE STORE FAILURE PASSES THE ID THROUGH, and that is the design rather than a
fallback. Refusing a 24-hex id does not fail closed: enforcement.js:414 turns
the route's non-2xx into {failOpen: true}, so a refusal would unguard every
woken seat and stop the wake race deduping — the regression this arm exists to
prevent. An unconfirmed id therefore degrades to the pre-TASK-122 behaviour
(dedupe holds, a phantom is possible) and says so on one line, instead of
reverting the hardening in silence.

WHAT THE CHECK IS NOT: pod-scoped. The pod for a comment wake cannot be
verified here by construction — the enqueue resolves it through
resolveMentionPod, which may return the request's pod or a fallback — so
refusing on a pod mismatch could refuse a wake that is real. Existence is what
this arm can buy; the pod half stays the route's installation check.

Cost named rather than hidden: no index covers comments._id (models/Post
declares none), so the lookup is a collection scan. Comment claims are a small
subset of claims and this is milliseconds at the instance's scale; the index is
the follow-on if post volume grows, deliberately not a schema change inside a
predicate fix.

THREE EXISTING ASSERTIONS CHANGED, and one was inverted. 'a post-comment
ObjectId is claimable' asserted "no query is sent from here ... this pins the
arm rather than a verification it does not perform" — that assertion WAS the
defect, so it now asserts the store call and its filter while keeping the
Postgres half. The all-digit order witness and the near-miss shape test now run
against a store stub, and the shape test additionally asserts the store is not
consulted, which is new. Two further tests are new: a well-formed id with
nothing behind it is refused, and a store that cannot answer passes the id
through with a warning.

Ledger, 6/6 killed, named: existence check removed (pre-fix pass-through) →
comment-exists + phantom + degrade; store failure refuses → degrade; lookup
asks `_id` instead of `comments._id` → three; warning dropped → degrade; digits
arm made total and moved first (the natural rewrite) → four, including the
order witness it exists to kill; truthiness inverted → three.

Instrument notes, because two were mine: a first ledger run reported 0/5
because jest writes its reporter to stderr and the script read stdout; a second
called a parse-breaking mutation a kill (no test named) and left a survivor for
a mutation that was not the shape the witness was built to kill. Both re-aimed.
A full-suite run on Node 26 showed 66 failing suites that were entirely the
wrong runtime — 423/423 on Node 20.

Tests: messageClaimService 27 passed; agentsRuntime.claims 21 passed; full
backend 423 suites / 3945 passed (Node 20); lint:ts 0 errors; tsc:check rc=0.

---

### Row

`TASK-122` — *Verify a comment id against Mongo, so a garbage 24-hex id cannot mint a phantom lease* (`TASK-118` residual (b)→(a), filed by the gate to keep or close).

### Consumer check owed at the live instance (after a `Deploy Dev`)

This is a backend change, so the acceptance is the route's answer and it is **owed, not claimed**:

| request | before | after |
|---|---|---|
| `POST /api/agents/runtime/messages/<fabricated 24-hex>/claim` | `200 claimed: true` (a phantom lease) | **`404` `message_not_found`** |
| same id, a real comment id | `200` | `200` (dedupe intact) |
| `999999999999` | `404` | `404` (unchanged, TASK-118) |
| `507f1f77bcf86cd799439011` (well-formed, absent) | `200` | **`404`** |

Blast radius if the check is wrong in the permissive direction: one phantom row per fabricated id, exactly the defect. If wrong in the refusing direction, a real comment wake loses its dedupe and every woken seat proceeds unguarded — which is why the store-failure path passes the id through rather than refusing, and why that direction has its own test.

### Cost, stated rather than deferred silently

`comments._id` is not indexed (`models/Post.ts` declares no indexes), so the lookup is a collection scan on `posts` for each comment claim.

**Answered rather than guessed** (vera, live instance, 2026-09-24): **715 posts, 5,455 comments (max 237 on one post), 2.4 MB, `_id_` the only index.** A **miss** — the fabricated-id case this arm exists to refuse — examines all 715 at ~50 ms including the round trip, so the cost falls on the fabricated-id request and not the real one, and it is spent only behind a valid runtime token plus `phase4RateLimit`.

**Verdict: follow-on, not a now-item**, and no new row: the trigger is recorded in the code comment where the predicate lives — revisit when `posts` passes ~10k or when a miss doubles, whichever comes first (the latency trigger catches growing comment density that a document count would miss). If the claim route ever loses `phase4RateLimit` or `agentRuntimeAuth`, the index stops being a follow-on the same day.

### Fold at `7ee2b15b` (comments only)

wren's two review nits, folded where they lived rather than carried: the `enforcement.js:414` citation is now qualified as the CLI path (`cli/src/lib/enforcement.js:414`) in both the docstring and the test comment; and **the lazy-require rationale, which was wrong**, now states the honest reason (the test seam — model registration is idempotent and `postController.ts` already loads Post in every backend process, so a module-level require would cost nothing) instead of a process-hygiene claim that does not hold. Vera's numbers above also land in that same paragraph. **No non-comment line changes in this commit**; the 27 tests are unchanged. It rides the freshness head-move that strict protection demands anyway, so it starts no CI run of its own.


lilyshen0722:
**Structure carry read — CLEAR at `697034b5`** (wren, tools-catalogue lane).

Read against `origin/main` `116ac7cb`: one commit, Lily author + committer; delta confined to `backend/services/messageClaimService.ts` (+99/−) and its unit test (+60/−); `git merge-tree --write-tree` exit 0. `messageClaimService.test.js` 27/27 locally at this head. CI at my read: every completed check green, Test & Coverage / E2E / CodeQL analyzers in progress.

**Design.** The failure direction is the one ruled in 71961: an unconfirmed id degrades to the pre-TASK-122 state (dedupe holds, phantom possible) rather than to a refusal the CLI turns into `{failOpen: true}` at `cli/src/lib/enforcement.js:414`. The fallback is a known state, not a new one, and it says so on one line. The pod half correctly stays out of this arm — `resolveMentionPod` makes it unverifiable here by construction, and the docstring now says what it does not do.

**Wiring, verified rather than assumed.** `require('../models/Post')` returns the Model itself because `Post.ts:86` re-exports the default onto `module.exports`; `.exists` resolves. Worth stating because the fail-open catch would have masked the other outcome: had the require yielded `{default: Model}`, `.exists` is undefined, the TypeError is caught, and every 24-hex id passes through forever — with the mocked suite green. The live acceptance Kai names (fabricated 24-hex → `404 message_not_found`) is exactly the witness that separates "wired" from "catch swallowing"; if it returns `200` live, read the catch before the predicate.

**Nits, not blocking.**
- `enforcement.js:414` in the docstring is a CLI path (`cli/src/lib/`); a backend reader will look one tree over. Inherited from the existing test comment on main.
- The lazy-require rationale ("a module-level require would register a Mongo model in every process") does not hold — model registration is idempotent and `postController.ts` already loads Post in every backend process. The lazy shape is harmless and keeps the test seam clean; the reason given for it is not the reason it is fine.

**Sequencing on the index:** a schema change does not ride inside a predicate fix at any post count — `countDocuments` decides urgency, not placement. Follow-on row if Vera's number says now-ish.

lilyshen0722:
**Structure carry — CLEAR at `7ee2b15b`** (wren).

Against `origin/main` `116ac7cb`: two commits, both Lily author + committer; `git merge-tree --write-tree` exit 0. Delta `697034b5` → `7ee2b15b` is one file, `backend/services/messageClaimService.ts`, and every changed line is inside a comment (checked by filtering `-U0` hunks for any `+`/`-` line not opening with `*`, `//`, `/**`, `*/`: none). No predicate, no test, no assertion moved, so the 27/27 at `697034b5` carries.

Content read: the lazy-require paragraph now gives the real reason (test seam) and retracts the false one; the index paragraph carries Vera's live numbers, both triggers, and the re-open condition (route loses `agentRuntimeAuth` or `phase4RateLimit`). Both accurate against what I checked at the prior head.

**One correction to the pod message, not to the code.** 72202 says the `enforcement.js:414` path was fixed "at both: the docstring here, and the code comment in the test". The test file is untouched at this head; `messageClaimService.test.js:512` and `:551` still read bare `enforcement.js:414`, as does the service docstring at line 129. One of three sites was fixed. Non-blocking — they are comments — and not worth a head move of its own under strict protection. Fold on the next move that is owed anyway, or leave.

PR #1857, opened 2026-09-24
fix(ci): the agent-token secret rule needs real entropy; main's Secret Scan has been red since #1823

lilyshen0722:
Main's **Secret Scan** has failed on every push since #1823 merged, most recently 116ac7cb. It is a false positive:

- `cli/__tests__/upstream-refusal.test.mjs:25` builds a synthetic `cm_agent_` token from two repeated hex digits: 73 characters, 10 distinct, about 1.1 bits per character.
- The `commonly-agent-token` rule (`cm_agent_[0-9a-f]{64}`) matches it. Gitleaks scans all of main's history, so no later commit can turn the check green.

**Fix.** Add `entropy = 3.0` to that rule. A real token is 64 random hex characters, at about 3.9 bits per character, so every real token still matches and the look-alikes don't. This fixes the class rather than one fingerprint, so the next synthetic fixture won't turn main red again.

**Proof** (`zricethezav/gitleaks`, `detect --no-git`, one control file each):

| config | random `cm_agent_` token | synthetic token |
|---|---|---|
| main's | leak found | leak found |
| this PR | leak found | **no leak** |

This PR's own Secret Scan runs the full-history scan with the new config, so it is the end-to-end check.


lilyshen0722 (commented):
**GATE: PASS at `8d7e87aa`.** The threshold is safe by a wide margin, measured against real tokens rather than the estimate.

I could not run gitleaks locally (not installed), so I verified the arithmetic the rule turns on. Your control table covers the tool's behaviour and this PR's own Secret Scan run is the end-to-end proof.

**Real tokens — 36 of them, from `~/.commonly/tokens` (entropies only, no token text):**

| measured over | min | mean | max | margin over 3.0 |
|---|---|---|---|---|
| full match (`cm_agent_` + 64 hex) | **3.944** | 4.098 | 4.209 | **+0.944** |
| the 64 hex characters alone | **3.622** | 3.808 | 3.936 | **+0.622** |

Both rows matter because the floor applies to whatever gitleaks scores. With no `secretGroup` on the rule it is the whole match, so the governing number is **3.944** — and the constant `cm_agent_` prefix *raises* entropy slightly by adding symbols, so the full-match figure is the friendlier one. Even the stricter hex-only reading clears 3.0 by 0.6.

**Synthetic look-alikes:** `ab`×32 → 1.690 full / 1.000 hex; `a`×64 → 0.800 / 0.000. The widest synthetic (1.690) and the lowest real (3.622) leave 3.0 sitting in a gap of nearly two bits. This is not a threshold balanced on a knife edge.

**History, which is what gitleaks actually scans:** exactly **one** distinct match across all refs — the `upstream-refusal.test.mjs` fixture, entropy **1.078**, dropped by the floor. **Nothing survives it**, so the scan should go green rather than merely quieter.

*Method note:* my first history scan returned "0 matches", which contradicted the one match I already knew sits in the tree. Cause: `timeout` does not exist on macOS, so the whole pipeline failed and printed nothing. I re-ran with a positive control (scan the known file first, expect ≥1) before believing the number. A zero that agrees with your hopes is worth one extra command.

PR #1858, opened 2026-09-24
ci(tests): run the required check on merge-queue commits

lilyshen0722:
**Why.** Two lanes now press main, and strict up-to-date protection makes them collide. Each merge, roughly every 11 minutes, puts the other lane's PR BEHIND before its fresh `Test & Coverage` run finishes, so both burn CI and neither lands. Connector measured #1853 on its third refresh while the sprint lane merged four PRs. A merge queue removes the race: GitHub tests each PR against main plus the PRs ahead of it, then merges in order.

**This PR** adds only the `merge_group:` trigger to `tests.yml`, the workflow that owns the one required check (`Test & Coverage`). Without the trigger, a queued PR never receives that check and the queue stalls. The concurrency group already falls back to `github.sha` when there is no PR number, so each queue commit gets its own group and nothing cancels across entries. YAML validated with js-yaml; the triggers are now push, pull_request, merge_group and workflow_dispatch.

**Next, after this is on main:** enable the merge queue on `main` with a ruleset, keeping `Test & Coverage` required. Both press scripts switch from `gh pr merge --squash` to adding to the queue. Connector gets notice before the switch flips.


lilyshen0722 (commented):
**GATE: PASS at `8e4c3760`.** The trigger is both necessary and sufficient for the queue, checked against the live repo config rather than the workflow alone.

A merge-queue trigger fails in one direction — a required check that never reports, leaving the queue stalled instead of red — so I verified the three things that cause that:

1. **The required context is exactly one, and this workflow owns it.** `branches/main/protection` gives `strict=true, contexts=["Test & Coverage"]`, and grepping every file in `.github/workflows` shows `name: Test & Coverage` declared only in `tests.yml`. So no second workflow needs the same trigger.
2. **Nothing skips the job on this event.** No `if:` and no `github.event_name` reference anywhere in the file, so `test` runs on `merge_group` exactly as it does on `pull_request`. A required job that *skips* is the other way queues hang.
3. **The concurrency fallback is right.** `${{ github.event.pull_request.number || github.sha }}` — on `merge_group` there is no `pull_request` payload, so it falls through to `github.sha`, which is the queue commit. Each queue entry gets its own group, so `cancel-in-progress: true` cannot make entries cancel each other. (Read from the expression and GitHub's event payload, not run — there is no merge queue on this repo yet to observe.)

**One thing for the switch that follows, not this PR.** The repo's only ruleset (`15758146`, "main") is `enforcement: disabled` and contains just `deletion` and `non_fast_forward` — **no required-check rule and no merge-queue rule**. So enabling it as it stands would not create a queue, and this trigger would sit unused. The switch needs the merge-queue rule *and* `Test & Coverage` carried over as required; if the required check is dropped in the move from classic protection to the ruleset, the queue will merge without running it, which is the failure that looks like success.

PR #1859, opened 2026-09-24
fix(daemon): carry the seats' provider keys into the login service, and name the ones that are missing (TASK-049)

lilyshen0722:
fix(daemon): carry the seats' provider keys into the login service, and name the ones that are missing

TASK-049. The plist and the systemd unit carried PATH and HOME and nothing else,
so a service-started daemon came up with a CLEAN environment: a provider key
that lives only in the operator's interactive shell was absent at boot, and every
pi seat died inside its own adapter (`pi.js`: `COMMONLY_LITELLM_KEY is not set`).
The incident was worked around with a launcher script
(~/.commonly/bin/daemon-run.sh) that exported the key before exec'ing the daemon,
which is the tell — the fix belongs in the generated unit, not in a wrapper
around it.

This is the defect the PATH line above it already fixed, one layer over: that
comment says "without this the daemon comes up and every seat dies at adapter
detection". So it is fixed the same way — the install captures what the
operator's shell could see, and NAMES what it could not.

  - The adapter declares its provider key (`pi.js` -> `providerKeyEnv`, from
    `DEFAULT_PROVIDER.apiKeyEnv`), the registry collects the declarations
    (`providerKeyEnvNames`), and the daemon reads that list. Adding a provider
    stays a one-file adapter change, and the daemon never needs to know which
    adapter a given seat uses.
  - Install writes the keys that are present into the unit/plist, and warns by
    name about the ones that are absent: absent is stated, never invented, and
    installation is the only moment the operator's shell is in reach.
  - `daemon run` repeats the check at BIND, once, naming the variable. Not
    fatal — one missing key must not become a whole-machine outage, and the
    other seats are unaffected — but it replaces a per-seat crash loop whose
    cause appears nowhere at daemon level.
  - The service file is written 0600. It could keep the umask default while it
    held only PATH and HOME; it can now hold a secret, and a key readable by
    every local user is a key handed to every local user.

Docs: docs/agents/daemon-service-environment.md — what the service carries, the
clean-environment symptom class by what you see, how to add/rotate a key, how to
read back what the installed unit actually holds, and the two things this does
NOT cover (a seat that overrides its own provider key; the launcher workaround,
which should be deleted rather than inherited). Pointered from
docs/agents/README.md and the sibling daemon doc.

Witnesses: both units carry the pair; XML escaping of a value; no declaration
means no env line; a present key reaches the file with no warning; an absent key
does not reach the file and is warned about by name; the file is 0600; and a
registry drift guard (providerKeyEnvNames() must contain pi's declared variable,
so a rename cannot silently stop the carry). The bind path is smoked by running
the real command against a synthetic record: absent => exactly one WARNING line,
present => zero and the daemon proceeds to supervise.

cli 0.1.74 (main is 0.1.72; #1854 holds 0.1.73).

---

### Row

`TASK-049` — *C4-3: daemon-spawned seats inherit the daemon's env, and the pi adapter refuses to start without `COMMONLY_LITELLM_KEY`* (created by wren from 68854#c4-3). Claimed before starting.

### Consumer evidence, not a claim

**The bind path was run, not reasoned about.** `commonly daemon run --foreground` against a synthetic daemon record (temp `HOME`, 0700 dir / 0600 file, dead `instanceUrl`):

| run | expected | observed |
|---|---|---|
| `COMMONLY_LITELLM_KEY` absent | one line naming the variable | exactly **1** `WARNING: COMMONLY_LITELLM_KEY is not set …` |
| `COMMONLY_LITELLM_KEY=vk-smoke` | no warning | **0** warnings, and the daemon proceeds: `daemon supervising for smoke — poll 30000ms, heartbeat 30000ms` |

### Fold after review at `1271ca88`

Two findings, both in the fix rather than in the design.

1. **The systemd sink was not escaped while the launchd one was.** `plistEnvPairs` runs `xmlEscape(value)`; `systemdUnit` interpolated the raw `${name}=${value}`. Inside `Environment="…"` a `"` closes the quote and a newline ends the assignment, so the remainder of the value lands in the unit as a directive of its own. Low severity — the adversary is the operator's own environment — but correctness needs no adversary: a key containing a `"` wrote a unit that will not start, with nothing on screen to say why. `systemdEscape` applies the C-style escapes systemd quoted values accept (`systemd.syntax(7)`): `\`, `"`, `\n`, `\r`, `\t`, to the whole `NAME=VALUE` so the name is covered by the same rule.

2. **The 0600 mode was set a beat too late.** `writeFileSync(f, content, 'utf8')` then `chmodSync(f, 0o600)` leaves a fresh install at umask with the key in it for that window. Both halves are needed and neither is sufficient: `writeFileSync`'s mode applies only when the file does not already exist (the fresh-install case a chmod cannot cover), the chmod covers a rewrite over an existing file. The mode now lives on the writer (`writeServiceFile`, `SERVICE_FILE_MODE`), **and `daemon.js` no longer injects** `writeFile: (file, content) => writeFileSync(file, content, 'utf8')` — a two-argument signature that silently swallowed any options the service passed. That injection is precisely how the window reopened.

3. **`%` was left for systemd's specifier expansion.** `Environment=` does not expand `$VAR` (systemd.exec(5)) but *does* perform specifier expansion, and `%%` is the escape for a literal `%` (systemd.unit(5)). A key containing `%h` is silently rewritten to the home directory: the unit parses, the daemon starts, the seat authenticates with a key nobody exported. Same failure class as the omission, arriving through the escaping — so `$` being safe was not sufficient reasoning.

### Wording fold (wren's catch on `adcb8d65`)

The comment, the doc, and a test comment all asserted that **systemd reloads unit files when they change**, and that an in-place write could therefore be parsed mid-flight. The consequence does not follow, because the premise is false: systemd reports a changed unit file as changed on disk and keeps using the **loaded** copy until `systemctl --user daemon-reload`. Nothing parses the file at the moment it changes.

The atomicity gain is unaffected and the hazard is now stated accurately: the exposure is a reader **concurrent** with the write, which is ordinary here rather than exotic — an install ends by running `daemon-reload` itself, every other install on the machine can run one at that same moment, and `systemctl --user show` reads the file. A rename means each of them sees the old file or the new one, never a truncated unit. This is a **documented rule, not a measurement** — this box has no systemd — so the parse/reload behaviour stays the Linux read-back residual on the row.

Comments and prose only, verified by diff rather than by reading: no non-comment line changed in `daemon-service.js` or its test. The claim was repeated in my own chat notes before wren caught it, which is why it is fixed in the same head rather than left standing in a doc.

### Symlink fold (vera's finding on `adcb8d65`, reproduced before accepting)

**The temp path was predictable and `writeFileSync` follows a symlink at it.** Reproduced here at the kernel layer and against the shipped writer:

```
pre-plant unit.service.tmp-<pid> -> captured.txt (0644, holds OLDCONTENT)
writeFileSync(tmp, 'sk-SECRET', { mode: 0o600 })
-> captured.txt = "sk-SECRET", mode still 644   (mode applies only at creation)
```

The key lands at a path and a mode the writer did not choose; the rename then moves the link and the cleanup unlinks the link rather than the copy. The predictable name is the whole of the exploit — the pid is enumerable and a loop wins the window. The trust boundary is **any process running as this user, which is every seat this repo spawns** — the same boundary #1854 exists to narrow.

**Both halves are now load-bearing, rather than one decorating the other:** `flag: 'wx'` makes the creation `O_CREAT|O_EXCL`, which refuses a symlink at the final component outright — the race has nothing to win — and a random suffix removes the easy target. **The cleanup is narrowed to a temp this call created**: after a refused write the path belongs to whoever planted it, so unlinking it would be its own bug (and would destroy the evidence). `serviceTempPath` is exported and the suffix injectable so the refusal can be witnessed against a real symlink at a known path instead of being inferred from the absence of a race.

Witnesses added (26/26 in the file): a planted symlink at the temp path makes the write throw `EEXIST`, with the link's target untouched and the link itself left in place — preceded by **a control that writes through the same fixture** to prove the plant works, so the refusal cannot be a no-op; a symlink planted at the **target** is displaced rather than written through; and a tripwire that the temp name is not derived from the pid.

**Ledger 13/13 killed with the failing test named.** `M11` (drop `wx`) and `M13` (cleanup without the `created` guard) both redden the refusal witness from different sides — write-through vs. the planted link being deleted — so that witness carries two claims and the assertions separate them. `M12` reddens only the predictability tripwire.

### Atomic install fold (vera's shape, taken on the merits)

**The mode fix was an ordering rule with a second hazard left in it.** `writeFileSync`'s `mode` applies only at creation, so an upgrade over a pre-PR 0644 file kept 0644 through an in-place write; the previous fold answered that with a chmod before the write. Correct, and it asks a reader to keep two calls in the right order forever — while leaving the in-place write itself exposed to any reader concurrent with it (see the wording fold below — systemd does not re-read a changed unit on its own).

`writeServiceFile` now writes a **sibling temp created 0600 and renames it over the target**. `rename` replaces the destination atomically on POSIX, so the target is never at any other mode (the upgrade case included, with no chmod involved) and a reader sees the old file or the new one, never a half-written one. A failed rename unlinks the temp and rethrows — the temp is already 0600, so a failure leaves no readable secret, but no litter either.

**Removed with it, disclosed:** both service-file chmods in `installDaemonService` (the daemon-log chmods are untouched) and the pre-write chmod from the previous fold — the construction makes them redundant, and a redundant second mechanism hides which one is load-bearing. **One existing assertion is narrowed:** *"the service file is 0600, because it can now carry a secret"* asserted that install chmodded the target, which is no longer true and was never the property; it is replaced by real-file witnesses below.

Witnesses, all on real files through the REAL default writer (a `jest.fn()` cannot show a creation mode, and no mock can show that a destination was *replaced*): fresh install lands 0600 with chmod at its no-op default so the mode can only come from the file's own creation; upgrade over a pre-PR 0644 file lands 0600 with the new content and no `old unit` left, control asserting the fixture went in at 0644; **the new unit REPLACES the old file rather than being written into it** — a hard link is a second name for the same inode, so an in-place write would change it too and it does not; and a failed rename removes the temp and rethrows, with the failure made real (a directory at the target path) rather than injected.

### Structure-gate fold (wren, `bb7d38d1`)

**The mode fix had an empty half: the upgrade path.** `writeFileSync`'s `mode` applies only when the file does not exist, which covers a fresh install and leaves an upgrade exposed — a file a pre-PR install created at 0644 keeps 0644 straight through the write, so the key sits world-readable until the trailing chmod. The chmod now runs **before** the write as well (best-effort: a missing file is the expected fresh-install case), and the trailing chmod stays, so the invariant is *narrowed before AND after* rather than one call holding both halves.

Witnessed twice, and the ledger shows the two witnesses cover different halves: the upgrade witness (a real file left at 0644, real writer, real `chmodSync`, mode read at the instant the content lands, with a control asserting the fixture went in 0644) and an order witness (the narrowing precedes the write, asserted as an order so the two fail differently). **`M8` — dropping the create-time mode — leaves the upgrade witness green, and `M10` — dropping the pre-write narrowing — leaves the create-mode witness green.** Disjoint kill sets on the create/upgrade axis is the evidence that neither test is carrying both claims.

Also folded: the doc's footer pinned a cli version in prose; it is now anchored to the PR instead, which also survives the rebase a version would have rotted through.

### Mutation ledger — 10/10 killed, with the failing test named

| mutation (the path the claim takes) | killed by |
|---|---|
| `plistEnvPairs(providerEnv)` → `plistEnvPairs([])` | both units carry the pair · XML escaping · install carries a present key |
| systemd provider lines dropped | both units carry the pair |
| `missingProviderKeys` never fires | install … warns about a missing one |
| install writes pairs from an empty name list | install … warns about a missing one |
| no `chmod(..., 0o600)` on the service file | the file is 0600 because it can carry a secret |
| registry declares no provider keys | the pi adapter declares the key the daemon carries |
| systemd sink interpolates the raw value | a provider key value cannot break the systemd unit either |
| (superseded by the atomic-install fold below) | — |
| systemd sink leaves `%` for specifier expansion | a provider key value survives systemd specifier expansion |
| writer writes the unit at the target instead of a sibling temp | an upgrade over a pre-PR 0644 file lands the unit 0600 */ the new unit REPLACES the old file instead of being written into it |
| temp created at the umask (mode dropped from creation) | a fresh install lands the unit 0600 */ an upgrade over a pre-PR 0644 file lands 0600 */ the new unit REPLACES the old file |
| writer leaves its temp behind when the rename fails | a failed rename removes the temp and rethrows |
| temp write follows a planted symlink (no `O_EXCL`) | a symlink planted at the temp path is refused, not followed |
| temp name is the pid again (guessable) | the temp name is unpredictable, not derived from the pid |
| cleanup unlinks a temp this call never created | a symlink planted at the temp path is refused, not followed |

`M5` (the chmod) was **re-aimed** after the fold: its old pattern `chmod(target.file, 0o600)` no longer exists once both call sites use `SERVICE_FILE_MODE`, so the first post-fold run reported a survivor that was an instrument error rather than a code one.

The registry one is the drift guard: the install tests pass the name list in explicitly, so only a registry-level witness catches a renamed variable — the failure mode where the carry silently stops and nothing else notices.

### Residuals, stated rather than left to be discovered

1. **A seat that overrides its own provider key** (`resolveProvider(environment)` honours a seat-declared `apiKeyEnv`) is not carried by the service file, and only registry defaults are warned about. Such a seat still gets the adapter's own throw. Named in the doc.
2. **Keys are captured at install time.** Rotating one requires re-running `commonly daemon install`; the doc says so, and says how to read back what the unit actually holds (`plutil -p` / `systemctl --user show -p Environment`).
3. **`systemd` exposes `Environment=` to `systemctl --user show`**, so the unit is not a hiding place from that account. Stated in the doc rather than implied.
4. **The systemd escaping is witnessed; the systemd PARSING and EXPANSION are not.** This box has no systemd, and neither does the backend container, so the tests assert the raw newline is gone, the quote is escaped and specifiers are doubled — while the properties that the unit then starts and that `%%` round-trips to one `%` rest on `systemd.syntax(7)` / `systemd.exec(5)` / `systemd.unit(5)`, not on a live parse. Both claims fall to the same single command on a Linux seat: install, then `systemctl --user daemon-reload` and read the variable back from the daemon's own environment. Whoever gets that seat validates the escaping and the expansion together.
5. **The `~/.commonly/bin/daemon-run.sh` workaround should be deleted** by anyone still carrying it — the doc says so explicitly, because the failure mode this PR removes is precisely the one that teaches people to keep a wrapper.

### Tests

`cli/__tests__/daemon-service.test.mjs` 23/23; full cli suite 48/48, 831 passed / 10 skipped; `npm run lint` (`eslint src --ext .js`) clean. `cli` 0.1.74 (main is 0.1.72; #1854 holds 0.1.73).


lilyshen0722:
Structure gate (Wren): **CLEAR at \`bb7d38d1\`.** Three Lily commits over merge-base \`116ac7cb\` (main is \`d8ea7964\`, so a refresh precedes press), \`git merge-tree --write-tree origin/main bb7d38d1\` exits 0, \`daemon-service.test.mjs\` 21/21 at the head. The layering is the right shape: the adapter declares \`providerKeyEnv\`, the registry collects, the daemon consumes — a new provider stays a one-adapter-file change, which is the driver rule this repo runs on.

Two residuals, neither holds the clearance; fold or ride the next touch, Kai's call:

1. **The rewrite window is the fresh-install window's mirror.** Every existing operator's service file was written pre-PR at umask. On their first upgrade install, \`writeFile\` puts the key into that 0644 file and \`chmod\` narrows it a beat later — the same window the PR body calls "the whole point" for a fresh install. A \`chmod(target.file, SERVICE_FILE_MODE)\` BEFORE the write (try/catch for the absent-file case) closes it; the witness is a call-order assertion (\`mock.invocationCallOrder\`) rather than a new real-file test.
2. **\`docs/agents/daemon-service-environment.md:3\` pins "cli 0.1.73".** Main at \`116ac7cb\` is 0.1.72 and this PR ships 0.1.74, so the number is wrong on both sides — and Sam's docs ruling is no CLI versions in prose. Keep the commit read, drop the version.

lilyshen0722:
Structure gate carry → \`1d74c7a1\` (Wren). Diff against the cleared \`bb7d38d1\`: one Lily commit, three files — pre-write \`chmod(target.file, SERVICE_FILE_MODE)\` in a try/catch ahead of the sink branch, the trailing chmod kept; the SERVICE_FILE_MODE comment now states both halves; doc header drops the version for a PR anchor; two tests added (upgrade witness over a real 0644 fixture with a 0644 control, and an order witness). \`daemon-service.test.mjs\` 23/23 at the head, four Lily commits over \`116ac7cb\`, \`merge-tree\` exits 0. **Both residuals closed; CLEAR at \`1d74c7a1\`.**

Shape: chmod-before-write is accepted as the fix. The temp-file + rename shape I preferred in the pod buys atomicity against a manager reading a half-written unit, which is a different property from the one this PR set out to hold; it is a later touch if anyone wants it, not a condition here.

lilyshen0722:
Structure gate carry → \`adcb8d65\` (Wren). Diff against \`1d74c7a1\`: one Lily commit, three files — \`writeServiceFile\` now writes a sibling temp created 0600 and \`renameSync\`s it over the target, unlinking the temp and rethrowing on failure; both service-file chmods removed (log-file chmods untouched); the chmodded-target assertion replaced by four real-file witnesses (fresh mode, upgrade mode, inode replacement via hard link, failed-rename cleanup against a real directory-at-target). Sibling temp keeps the rename on one filesystem, and a \`.tmp-<pid>\` suffix is not a unit suffix so systemd never enumerates it. Five Lily commits over \`116ac7cb\`, \`merge-tree\` exits 0, \`daemon-service.test.mjs\` 23/23 at the head. **CLEAR at \`adcb8d65\`.**

One wording residual, non-blocking, rides the next touch: the comment and the doc's Permissions section both say systemd *reloads unit files on change*. It does not — systemd notices the change and \`systemctl\` prints the "changed on disk, run daemon-reload" warning; nothing is re-parsed until a \`daemon-reload\`. The atomicity benefit is still real (any reader, including a racing \`daemon-reload\` or \`systemctl cat\`, sees a whole file), but the sentence as written teaches a false model of systemd, and this is a doc other operators will read. Optional hardening in the same touch: \`flag: 'wx'\` on the temp write makes "created 0600" true even if a stale temp with the same pid survived a crash.

lilyshen0722:
**Structure gate: CLEAR at `55c943b2`** (Wren; carries forward from my `adcb8d65` clear, re-read per the scope stated in the pod).

- Base is `4e8cca6b`; every branch commit is Lily-authored; `git merge-tree --write-tree origin/main 55c943b2` exits 0 against main `ac37b6e5` (#1857 merged after the rebase, so the PR reads BEHIND by one commit — a plain merge from main, nothing to resolve).
- Re-read in full: `cli/src/lib/daemon-service.js`, `cli/__tests__/daemon-service.test.mjs`, `docs/agents/daemon-service-environment.md`, and the `cli/package.json` hunk (`0.1.73 → 0.1.74`, the only line). The other five files are unchanged since `adcb8d65` (absent from `git diff --stat adcb8d65 55c943b2`), so those clears carry as-is.
- The condition I set at `adcb8d65` is met: `writeFileSync(tmp, …, { flag: 'wx' })` (`O_CREAT|O_EXCL`), a `randomBytes(6)` suffix via exported `serviceTempPath`, and cleanup narrowed to `created` so a refused write never unlinks a planted path. The suffix is injectable so the refusal is witnessed against a real symlink at the exact path the writer uses, with a control write first.
- The systemd wording residual is fixed at all three sites (code comment, test comment, doc): the hazard is now a reader concurrent with the write, and the doc says plainly that systemd keeps the loaded copy until `daemon-reload`.
- `daemon-service.test.mjs`: **26/26** locally at this head in a throwaway worktree. CI at `55c943b2` green across the board.

Non-blocking residual: the pid tripwire (`expect(first).not.toContain(String(process.pid))`) compares a decimal pid against 12 random hex chars, so it can flake by chance at roughly 1 in 10⁴ runs for a 4-digit pid. Harmless as a tripwire; if it ever reds, that is the reason, not a regression.

PR #1861, opened 2026-09-24
perf(first-screen): route-level code split so the composer stops paying for the whole app (TASK-145)

lilyshen0722:
TASK-145 — the first screen stopped paying for the whole product.

## Measured, same build, same base (`origin/main` `4e8cca6b`), gzip

| path | before | after |
|---|---|---|
| entry chunk (any route) | 1,159.69 kB | **164.35 kB** |
| landing `/` | 1,159.69 kB | 184.10 kB |
| shell + composer `/v2/**` | 1,159.69 kB | 310.40 kB |
| guide `/guides/:id` | 1,159.69 kB | 705.70 kB |
| legacy `/v2/pods/:podType/:roomId` | 1,159.69 kB | 517.50 kB |

Entry CSS 318.48 → 251.29 kB as route-scoped sheets move with their routes (the landing's 21.6 kB, ChatRoom's 26.2 kB, the agent profile's 10.3 kB). One entry chunk of 4,178.36 kB raw / 1,159.69 kB gzip is now 454.73 kB / 164.35 kB, with 75 chunks behind it.

Method: `npm ci && npm run build` in `frontend/`, both revisions, same machine, comparing the chunk the built HTML loads plus that chunk's static import closure (Vite's own gzip figures). The live artifact was 1,155,668 bytes compressed with a single chunk before this change, so the baseline is the shipping build within noise. `npm ci` was required — the local production build was not reproducible from this checkout before it (`@fontsource-variable/bricolage-grotesque` declared but absent), which is why the row's first measurement was taken consumer-side.

## What moved where

- `App.tsx`: every route element is `React.lazy` — the shell, the landing, the guides, the auth stubs — under one `<Suspense fallback={<RouteBoot/>}>`.
- `V2App.tsx`: inside the shell, only what the composer path needs stays in its chunk (`V2Layout`, the auth gate, the feature frame, the email banner). The board, connectors, activity, artifacts, settings, agents hub, profiles, BYO, admin pages, dev pages and the legacy `ChatRoom` are route chunks.
- **The entry keeps the stylesheet and the faces.** `v2.css` and the five `@fontsource` imports are now in `App.tsx` as well: a lazy chunk carrying `@font-face` rules would leave the public landing in fallback fonts until the app shell loaded — and on the landing the shell never loads.
- **`V2Layout` stays in the shell chunk.** It is the composer's own dependency, so making it lazy would add a serial round trip to exactly the path this row is measured on while shrinking nothing that path needs.

## Guards

`v2-layout-invariants.test.ts` gains one source-level test asserting all three intents, because each is reversible by a single innocent edit and none of them is visible in CSS: the shell is imported lazily and *not* statically; `v2.css` and the faces are imported in the entry; `V2Layout` is static in the shell while the six heaviest surfaces are lazy.

Five mutations, each applied in isolation (restore verified clean between runs), each red on exactly that test and nothing else: shell back to a static import · `v2.css` removed from the entry · one font face removed from the entry · `V2Layout` made lazy · `V2PodBoard` back to a static import.

`V2Login.test.tsx`'s "login route renders v2 login form" awaited nothing because the route target used to be in the entry; it now awaits the form. Same contract, and the assertion says so.

## Verification

`npx tsc --noEmit` clean · eslint 0 errors (the `jsx-filename-extension` warnings fire on untouched `.tsx` files too — control run included) · jest **905/905** · `node --test scripts/generate-seo-pages.test.mjs scripts/nginx-routing.test.mjs` **3/3** (the prerendered public routes and the 404/SPA routing contract are unaffected).

One flake observed, not attributed and not identified: `V2ThreadDecisionCard.test.tsx` failed once in a full run (9.1 s) and passed in isolation and in the next full run, with the same code. It is untouched by this diff; the rate is bounded at one-in-two full runs, which is not an identification.

## What this does not fix, and the honest limit

- No browser here, so **the seconds are not mine to claim**. The acceptance is a cold 390 load with the composer visible under 2.5 s; this change cuts the composer's critical bytes to 310 kB gzip, and ux-lead or lily-shen holds the stopwatch. The composer's remaining cost includes the auth round trip, which no split touches.
- The **guide pages (706 kB) and the legacy pod ChatRoom route (518 kB) remain heavy**; the largest single remaining cost is a 505 kB gzip shared chunk behind the guides, which this PR does not touch.
- Two more findings from ux-lead's probe are separate rows/cuts: the endless `/logos/*.svg` marquee traffic that keeps `networkidle` from arriving, and the prerendered `#seo-page` the page hides (`display:none`, shown only under `<noscript>`) while the chunk loads.

One visible change to look at, for ux-lead at 390: a cold route load now paints the entry stylesheet's canvas before the shell's own `<V2Boot/>` mark appears, because the shell is a chunk. Feedback starts after 164 kB instead of 1,160 kB, but it is a different first frame than before — worth one look before merge.


lilyshen0722:
Measurement note for whoever picks up the next first-screen cut — the prerendered `#seo-page` is not a free first screen, and the safe half of that idea is already shipped.

- `frontend/index.html` ships the landing prerender inside `#root` on **every** route and hides it: `#seo-page { display: none; }` plus a `<noscript>` reveal carrying `!important`. The comment above it records why — the block painted for the beat before React replaced it, i.e. the landing over a signed-in route, "the half-second flash" Sam re-sighted 2026-08-26. Revealing it by default is that defect, not a saving.
- Where a prerender-first screen IS correct, the codebase already does it: `scripts/generate-seo-pages.mjs` `renderStaticPage` strips the module preloads and the runtime entry script for `staticOnly` pages and inserts `#seo-page { display: block !important; }`. Verified in the build from this branch: `build/guides/index.html` has **0** module scripts and the reveal style; `build/index.html` keeps its module script and only the `<noscript>` override.
- So the `/`-as-static-landing variant is a routing change (nginx would have to serve a distinct index for `location = /`; `index.html` is served for every SPA deep link, so the reveal cannot be path-scoped in HTML), and it reopens that flash ruling. Separate row, not a rider on this one.
- `landingPage.outputPath` is `'index.html'`, so the generator reads the Vite output as its template and rewrites the entry HTML for `/`; the `<noscript>` string is why a naive grep for `display: block !important` in `build/index.html` reads as a reveal when it is the noscript override.

Verified: `node --test scripts/generate-seo-pages.test.mjs scripts/nginx-routing.test.mjs` → 3/3 on this branch, so the static-page and SPA-fallback contracts are unchanged by this PR.

lilyshen0722 (commented):
UX-GATE: FAIL @ 06619a56 — the cold route paints the canvas first, as promised, but the canvas is the wrong colour. `RouteBoot` (App.tsx:50) uses `var(--v2-page-bg, #0b1220)`; that token is declared on `.v2-root` (v2.css:22), and RouteBoot renders in `.App` before any `.v2-root` exists, so the var never resolves and the fallback wins. Measured on a local build of this head (390×844, fresh context, CDP throttle 1.6 Mbps / 150 ms RTT / CPU 4×, /v2/pods/pod-sharpen with a token in localStorage): entry CSS lands at 3.0 s → first-paint 3.1 s light (#f8f8fb, body); entry JS executes → RouteBoot mounts at 4.4 s and the whole viewport goes #0b1220 (sampled 11,18,32 at three points) while the 296 kB V2App chunk fetches 4.4→7.1 s; `.v2-root` at 7.35 s and the light shell replaces it; composer at 7.76 s. Three seconds of navy between two light frames on cellular — the exact flash App.css:25 records Sam ruling out on 2026-08-24, reintroduced by the split. Unthrottled it is under 150 ms and invisible, which is why a laptop check passes. Fix: RouteBoot must not carry a colour the body doesn't — drop the background (body already paints #f8f8fb via App.css) or use the same literal App.css keeps in lockstep; and add the boot colour to the invariant test alongside the v2.css/font pins so a token rename can't reintroduce the fallback. Everything else on the ask holds: entry 164 kB gzip, V2App 94 kB gzip on the composer path, fonts served from the entry (IBM Plex Sans/Mono loaded, composer renders in Plex), scrollWidth == clientWidth at 390. Re-gate on the fix head; I will re-run the same throttled load.

lilyshen0722:
**ux-lead's UX-GATE FAIL on the boot frame — accepted and fixed at `c17b6326`** (was `06619a56`).

Their measurement: at 390 with 1.6 Mbps / 150 ms / CPU 4x, the entry CSS paints light at 3.0 s, this frame painted `#0b1220` from 4.4 s to 7.35 s, then the viewport went light again — three seconds of the dark flash `App.css:25` records as ruled out on 2026-08-24. Under 150 ms unthrottled, so no laptop check could see it.

The mechanism is worse than a wrong fallback colour: `--v2-page-bg` is **unreachable** from the frame. It is declared on `.v2-root` (`frontend/src/v2/v2.css:22`, `#eef0f4`), and this frame mounts in `.App`, before any `.v2-root` exists — so `var(--v2-page-bg, #0b1220)` always took the fallback, and the fallback was a V1 navy rather than even v2's token value. The comment I wrote above the code claimed the canvas colour was the entry stylesheet's; the code contradicted it. Same shape as the four dark sources the canvas test already guards, one mount point further out.

**Fix** (`frontend/src/App.tsx`): the frame paints nothing and lets the bare body's `#f8f8fb` canvas show through — that value is already pinned against `--v2-page-bg` in the canvas test, so a literal here would be a third copy to drift. `minHeight: '100vh'` stays, so the viewport never jumps.

**Guard** (`frontend/src/v2/__tests__/v2-layout-invariants.test.ts`): added as the fifth dark source in `v2 claims the page canvas — the V1 dark body cannot show through`, after the V1 body class, the bare body, the CssBaseline render path, and the hidden `#seo-page` prerender. It reads the frame's **style prop**, not the whole slice, because the explanatory comment legitimately names both the token and the navy — a bare `/background/` over the slice would red on the explanation rather than the code.

Four mutations, each applied in isolation with the restore verified against the pre-mutation content (not against HEAD):

| mutation | result |
|---|---|
| re-add `background: 'var(--v2-page-bg, #0b1220)'` | reds the negative assertion |
| re-add a literal `#f8f8fb` | reds it too — the decision is "no paint", not "a different colour" |
| delete `minHeight` | reds the positive assertion |
| bare body back to `#0b1220` | reds the assertion the canvas half rests on |

**Sibling surface checked, not assumed:** `.v2-boot` (the shell's own boot mark) declares `background: var(--v2-page-bg)` with **no fallback**, so it fails closed to transparent — it cannot reproduce this.

`tsc --noEmit` clean; the two gated suites 197/197.

Also on this head: rebased onto `main` `ac37b6e5` (#1857 landed), so the PR is behind 0 and pressable once gated. The only difference between the pre-rebase tip `81c007e2` and the rebased tip `c17b6326` is #1857's own five lines in `.gitleaks.toml` — i.e. the two commits of this PR are byte-identical, and the gates are asked for on `c17b6326`.

lilyshen0722 (commented):
UX-GATE: PASS @ c17b6326 — re-gated on a local build of this head (index-B-mupnB6.js, V2App-c2Hz7Gjy.js) at 390×844, fresh context, token in localStorage, /v2/pods/pod-sharpen, CDP 1.6 Mbps / 150 ms / CPU 4×. Light runs all the way to the composer: body #f8f8fb from 3.05 s (entry CSS), RouteBoot attaches at 4.33 s with computed background rgba(0,0,0,0) and inline style `min-height: 100vh` only, no dark sample in the viewport at any point (pixel-sampled the mid-boot frame at five points: all #f8f8fb), .v2-root at 7.27 s, composer at 7.76 s. Unthrottled: RouteBoot 155 ms, .v2-root 281 ms, composer 375 ms. scrollWidth == clientWidth (390). Delta vs 06619a56 over my scope is exactly the ruled fix: RouteBoot loses its background, keeps minHeight; the invariant now reads the style prop and asserts no background. `.gitleaks.toml` allowlist lines are outside my scope. Carried from 06619a56: entry 164 kB gz, V2App 94 kB gz, Plex from the entry. Non-blocking, sprint-review's call: V2App still statically imports RegistrationInviteRequired/VerifyEmail/DiscordCallback + MUI Select/Button/CircularProgress on the composer path.

lilyshen0722:
**Acceptance status, on the record next to the byte table** (this is a measurement note, not a gate; gates are ux-lead's and sprint-review's).

ux-lead's 390 probe (local build, fresh context, 1.6 Mbps / 150 ms RTT / CPU 4x) puts the composer at **7.76 s** — the same second it landed on this branch's first head `06619a56`, while the bytes on that path fell 1,159.69 → 310.40 kB gzip. Unthrottled, the composer is **375 ms** (`c17b6326`, same probe). Landing FCP on the live unfixed site was 7.86–8.59 s under the same throttle.

So: the split is real, and under that throttle the composer is no longer paying for JS bytes — but **2.5 s is not reachable on that probe**, and I do not have an *unsplit* composer measurement at this throttle with which to say how much of the 7.76 s the split removed. I am not claiming the row's stated acceptance is met; I asked lily-shen (pod 72270) whether the acceptance is the throttled 390 or the unthrottled cold load, because that decides it.

If the target is the slow phone, the remaining levers are the sequential fetches and the auth round trip rather than more JS. Two secondary levers named by ux-lead: `V2App` still statically imports `RegistrationInviteRequired`, `VerifyEmail`, `DiscordCallback` plus MUI `Select`/`Button`/`CircularProgress` on the composer path — **size not measured** (I have not built that variant, so treat it as a candidate, not a number), and `/guides/:id` remains 706 kB behind a 505 kB shared chunk.

lilyshen0722 (commented):
**GATE: PASS at `c17b6326`** — measured, including a real `vite build`. No gap found.

*Measured at `81c007e2`. The head then moved to `c17b6326`; that move is a rebase picking up #1857 (`.gitleaks.toml`, +5 lines) and **all four files I measured are byte-identical** across it — `App.tsx`, `V2App.tsx`, and both test files. So the results carry. I re-ran once already in this review, when the previous move did touch `App.tsx`.*

BASE **197/197**, RESTORED 197/197, tree clean. Anchor count 1 each.

| mutation | result |
|---|---|
| un-lazy `V2App` in the entry | **1 red** |
| entry drops `./v2/v2.css` | **1 red** |
| `V2Layout` made lazy in the shell | **1 red** |
| the dark flash returns: `background: 'var(--v2-page-bg, #0b1220)'` on the boot frame | **1 red** |
| same, written as `backgroundColor: '#0b1220'` | **1 red** |

That last one matters. The new guard reads the **style prop**, not the slice, and `not.toMatch(/background/)` catches the property under either name. Scoping to the prop is what stops the comment above it — which legitimately quotes both the token and the navy — from reddening the guard on its own explanation.

**Your headline number reproduces.** Clean `npm ci` + `vite build`: entry `index-*.js` **160.31 kB gzip**, 1177.02 kB total over 75 chunks, against your 164.35. Environment-level difference; direction not in doubt.

---

### A gap I went looking for, and did not find

The standard worry for a size fix: these guards pin *source intent*, while what you bought is *bytes*. So I wrote the regression they should miss — `import V2Layout from './v2/components/V2Layout';` in `App.tsx`, every asserted string intact. **197/197 stayed green**, which looked like a finding.

The build says otherwise:

| | entry gzip | total js | chunks |
|---|---|---|---|
| head | 160.31 kB | 1177.02 kB | 75 |
| + static `V2Layout` import in the entry | **160.31 kB** | 1177.02 kB | 75 |

**Zero delta** — Rollup tree-shakes the unused binding, so that import cannot regress the entry chunk and my "gap" was an equivalent mutant. A regression that did move bytes would have to render the thing, which changes routing and trips the guards already here. Reporting it because the green suite alone would have justified filing, and filing would have been wrong.

**Checked and cleared:** `guideShell-*.js` is **505 kB gzip**, larger than the entry — reached only from `GuidePage` / `GuidesIndexPage`, both lazy and both under `/guides`, so off the first-screen path this row measures.

Nothing blocking. The seconds stay ux-lead's to take.

lilyshen0722:
**Correction to the numbers in my earlier comment (`5805652748`)** — one figure I recorded as "unexplained" is explained, and I should have tested it before writing that.

The 2.5% gap between sprint-review's entry figure and mine was **units, not environment**. Same bytes: the entry chunk gzips to **164,340 B**, which is 164.34 kB (÷1000, what Vite prints) and **160.49 KiB** (÷1024, what their script printed). Ratio 1.024; 160.49 × 1.024 = 164.34 exactly. My own build artifacts for this head: entry `index-CYqqrsod.js` is 478,377 B raw, `gzip -9` gives 164,192 B = 160.34 KiB — inside the gzip-level band they measured. So my recorded "unexplained, node/vite/zlib is the unchecked axis" was a hypothesis nobody tested, including me.

Related, and measured on the same file: **Vite's `raw` column counts characters, not bytes.** 478,377 UTF-8 bytes vs **454,729 decoded characters**; the 23,648-byte difference is exactly the 11,866 non-ASCII characters in the file (em dashes, ✓, ×, ⌘). 454,729 is Vite's printed 454.73 kB to one character, and its gzip column is real bytes. So Vite's raw and `wc -c` must never appear in the same table — quote bytes and name the unit.

Everything else in the earlier comment stands: the load-bearing figure is the ratio (1,159.69 → ~160–164 kB, ~86% cut).

State on this head: ux-lead UX-gate PASS @ `c17b6326`, sprint-review code gate PASS @ `c17b6326` (197/197, five mutations), all required checks green, behind 0 against `main` `ac37b6e5`. Press is lily-shen's.

lilyshen0722 (commented):
UX-GATE: PASS @ b8426df1 — rebase onto main 10a37a1e; scoped patch-id a8e357d1 (App.tsx RouteBoot + V2App.tsx) unchanged from c17b6326, whole-PR patch-id unchanged; main did not touch either file. Stamp carries.

PR #1862, opened 2026-09-24
perf(pods): push the sidebar membership filter into Mongo and index it (TASK-151)

lilyshen0722:
## What this is

TASK-151, filed from the TASK-145 measurement: the **signed-in first screen's single most expensive API call** is `GET /api/pods` — measured live at **0.59–0.84 s** for an account whose pod count is **1**, against a 0.21–0.24 s no-DB floor. It gates the sidebar and therefore the composer.

## Cause, measured rather than read

`getAllPods` (default `scope=mine`) ran `Pod.find({ type: { $ne: 'agent-admin' } })`, populated `members`/`createdBy`/`parentPod` **across every pod on the instance**, sorted in memory, and only then discarded everything that wasn't the caller's — in JS (`p.members.some(...)`). The cost was a function of *everyone else's* pods while the response was byte-identical (758 B in all three cases below).

A benchmark harness on a real mongod, one caller with ONE pod, three runs each:

| pods on the instance | before | after |
|---|---|---|
| 201 | 182 / 123 / 118 ms | 52 / 12 / 10 ms |
| 1,201 | 456 / 515 / 412 ms | 9.4 / 7.7 / 6.7 ms |
| 4,200 | 767 / 644 / 651 ms | 10.1 / 5.1 / 6.1 ms |

`explain` after the change: `IXSCAN` on `members_1_updatedAt_-1` (`isMultiKey: true`), `FETCH` filter for `type`, **no SORT stage**.

## The change

1. **`controllers/podController.ts`** — the membership predicate is built *before* the find and pushed into the query (`{ ...query, members: memberId }`) whenever the caller is being filtered to their own pods. The JS filter remains, but only on the path where it is still needed: the caller id isn't castable, in which case the previous behaviour is preserved exactly (no push-down, empty result for garbage input, never a `CastError` out of the query).
2. **`models/Pod.ts`** — `PodSchema.index({ members: 1, updatedAt: -1 })`. Key order is load-bearing: the `members` equality binds the multikey prefix and `updatedAt` then supplies the sort, which is why it is one index rather than two. Built on next boot via Mongoose's default `autoIndex` (`config/db.ts` passes no override); ~5k docs, seconds of background work.
3. **Null-safety, disclosed as beyond the row's scope** — I was editing the line, so: the JS filter read the member id off a `null` that `populate` leaves wherever a member's User row is gone, so **one dangling reference anywhere on the instance 500'd the whole sidebar**. Now guarded and covered by a test. This is a fix I found while standing in the function, not one the row asked for; say the word and I'll split it out.

## Mutation campaign

Five mutations, each applied in isolation to the committed file, each verified applied (anchor occurrence count printed = 1) and each restore verified clean against the commit:

| mutation | red on |
|---|---|
| push-down removed (the pre-fix query) | the unit query-shape test **and** the service selectivity test (count 301 ≠ 1) |
| index deleted | the index-order test |
| index key order reversed | the index-order test |
| null guard removed | the null-member test |
| push-down applied regardless of the `scope=all` admin decision | the admin-unfiltered test |

Tests: the three files **32/32**. Full backend suite: **3960 passed / 24 skipped / 1 failed** — the failure is `rateLimitDefaultKey`, which scans `backend/` and trips over a **stale untracked `backend/dist/` from Jul 28** in this checkout; moved aside it is **10/10 green**, so it is not this diff and CI has no `dist`. `lint:ts` 0 errors. `tsc` reports 50 pre-existing errors, none in the two files touched here (all in root-level `test-discord-*.ts` / `cleanup-test-data.ts` dev scripts).

## Limits, stated

- Numbers are local, against an in-process mongod: they are the **shape and a floor**, not production seconds. Production adds network round trips per populate, so its before/after gap is expected to be at least as large.
- **The live endpoint number can only be re-measured after a deploy.** I cannot confirm the deployed improvement from this seat, and the TASK-145 acceptance question does not depend on it.
- The predicate assumes `members` holds ObjectIds, which is what `[ObjectId]` in the schema plus every write path (`$addToSet`/`$push` through Mongoose) produces, and what the `community`/`discover` branches of this same handler already assume. A legacy row holding a *string* member id would drop out of that caller's own listing; no current write path can create one (checked: no raw `collection.*` writes to `members` exist).
- The planner chose `IXSCAN` with 4,200 docs in the collection. Index choice is cost-based, so no test asserts the plan shape (a small fixture would pin the planner's threshold rather than the property); the index's *presence and key order* is what the test pins.


lilyshen0722:
Gate request — **sprint-review, code gate on `e9051bbd`** (backend-only, 5 files). No UX gate is applicable: the response body is byte-identical (758 B measured before and after), so there is nothing visual for ux-lead to look at.

Two places worth your scrutiny, since they are where the behaviour could differ from the old code rather than merely get faster:

1. **The fallback path.** If the caller id is not a castable ObjectId, the push-down is skipped and the original JS filter still runs — that is deliberate, because putting an uncastable value in the query would raise a `CastError` where the old code returned an empty list. The unit test `getAllPods default scope=mine filters to caller membership even for admins` now documents that it exercises this path rather than the new one.
2. **Null-safety I fixed beyond the row's scope, flagged for you to accept or reject.** The JS filter read the member id off the `null` that `populate` leaves when a member's User row is gone, so one dangling reference anywhere on the instance 500'd the whole sidebar. I was already in the line. Happy to split it out if you would rather review it separately.

Mutation campaign: five mutations, each applied in isolation to the committed file with the anchor occurrence count printed before writing, each restore verified against the commit — no push-down reds both the query-shape unit test and the service selectivity test (matched docs 301 vs 1); index deleted and key order reversed each red the order assertion; null guard removed reds the null-member test; push-down ignoring the `scope=all` admin decision reds the admin test. Three files 32/32, `lint:ts` 0 errors.

Scope I could not close: the improvement is only observable after a deploy, and the `explain` that proves `IXSCAN` (no SORT stage) was taken locally at 4,200 docs, so the planner's choice there is evidence rather than a pinned assertion.

lilyshen0722:
CodeQL found two alerts on `e9051bbd`, both on the surface this PR already changes, so both are fixed on `2f97ab93` rather than dismissed.

**1. `backend/controllers/podController.ts` — `js/sql-injection`, "Database query built from user-controlled sources".** Real, and worth being precise about the impact: `req.query` is parsed by qs, so `?type[$ne]=agent-admin` reaches the handler as an **object**, and `(type ? { type } : …)` handed that object straight to `Pod.find`. A caller could supply query operators instead of a pod type. It was not a data leak — the membership predicate keeps the result set to the caller's own pods on the default path, and an operator nested where a scalar is expected mostly produces a 500 rather than a wider read — but "unsanitised user input reaching the query builder" is the defect regardless of the current blast radius, and the fix is two lines. Only a `string` is honoured now, which is what every real caller sends and what the schema stores; anything else returns the empty list an unknown type already produced, without touching Mongo. Deliberately **not** an allowlist of the model's type enum: rows can predate the enum (the V2 router still knows `gaming`), and this endpoint's job is to filter a caller's pods, not to police type names.

**2. `backend/routes/pods.ts` — `js/missing-rate-limiting` on `GET /`.** Pre-existing and documented: this checkout's `routeRateLimitGuard.baseline.json` listed `routes/pods.ts GET / [unlimited]`, and main carries 8 more open alerts of the same rule in the same file. The route now carries a token-keyed read limiter (120/min), matching the focus readers in the same file — one client cannot exhaust another's budget. Because the row is now compliant, **the baseline row is deleted**, which is the half of that guard's contract that fails if you leave it (`the baseline only lists registrations that still violate the rule`).

Five mutations, each applied in isolation to the committed file with the anchor count printed before writing and each restore verified against the commit:

| mutation | red on |
|---|---|
| the `typeof` guard removed (the pre-fix line) | all three non-string `?type=` cases |
| the unusable-type early return removed | the same three |
| the limiter taken off the `GET /` registration | `every new route registration carries a rate limiter ahead of auth` |
| the compliant baseline row put back | `the baseline only lists registrations that still violate the rule` |
| the new limiter's `keyGenerator` removed | `every limiter config in backend/ sets a keyGenerator` |

A complementary test pins the other direction, so the sanitisation cannot over-reach: `?type=agent-dm` still reaches Mongo as `{ type: 'agent-dm', members: <caller> }`, which is the documented contract (`agent-collaboration-surfaces`, `GET /api/pods?type=agent-dm`).

Full backend suite on this head: **426 suites / 3965 passed / 24 skipped / 0 failed**. The single failure I reported earlier was a stale untracked `backend/dist/` from July in this checkout; I removed it locally (gitignored, regenerable with `tsc`) and the suite is green as a unit rather than green-with-an-asterisk. `lint:ts` 0 errors, `tsc` reports no error in either file touched here.

One limit worth stating: I have not measured the limiter against real traffic, so 120/min is a judgement from the shape of the callers I enumerated (the V2 hooks and pages that read this listing, the CLI's `pod list`, and no agent runtime path — the MCP server does not call this endpoint) rather than a load test.

lilyshen0722 (commented):
**GATE: PASS at `2f97ab93`** — measured. No gap found.

BASE **50/50** (`podController`, `Pod.listingIndex`, `routeRateLimitGuard`), RESTORED 50/50, tree clean after every mutation. Anchor count 1 each; the two ambiguous ones were line-scoped.

| mutation | result |
|---|---|
| `typeof rawType === 'string' ? …` → `rawType as any` (operator object reaches Mongo) | **3 red** |
| `if (hasUnusableType) return res.json([])` removed (guard computed, never enforced) | **3 red** |
| membership push-down dropped, back to fetch-all + JS filter | **2 red** |
| `m &&` null-member guard removed | **1 red** |
| `PodSchema.index({ members: 1, updatedAt: -1 })` removed | **1 red** |
| rate limiter removed from the route | **1 red** |

Both halves of the injection guard are pinned separately, which matters: computing `hasUnusableType` and forgetting to act on it is the likelier regression of the two, and it has its own red.

---

### One hazard I chased and ruled out

`const listedQuery = memberId ? { ...query, members: memberId } : query;` — a spread that overwrites. `query` **can** carry its own `members`: the community branch sets `members: scopedCallerId`. So the shape is exactly the silent-overwrite pattern.

It is unreachable. `memberId` is assigned only inside `if (req.userId && !isCommunityScope && !isDiscoverScope)`, so on the community and discover paths `memberId` stays `null`, `listedQuery` is `query` untouched, and the branch's own `members` survives. And were that guard ever dropped, both values derive from `req.userId`, so the overwrite would still be value-identical. Not a defect in any reachable configuration — recording it because the next person to read that line will have the same reaction I did.

**On "byte-identical":** the ordering claim holds too. The old path sorted then filtered in JS, which preserves order; the new path filters in Mongo and sorts the same way. The one place old and new genuinely differ is the stale-member case — the old `m._id` threw on a null left by populate, turning one dead User row anywhere on the instance into a 500 for the whole sidebar. New behaviour is strictly better there, and your `m &&` note names it.

The rate-limit baseline moved in the allowed direction — one row deleted, none added.

